PR #78, opened 2026-08-06
feat(streaming): migrate to SSE with a WebSocket fallback

damianvtran:
Migrates the renderer from WebSockets to **Server-Sent Events**, with automatic
fallback. Everything above the transport — the rAF coalescing, equality gate,
global registry, stores, rendering — is unchanged, because both transports
present the same emitter surface.

## Why

The socket only ever carried server-to-client data. The one frame the client
sent upstream was a keepalive `ping`; cancellation was already HTTP
(`JobsApi.cancelJob`). SSE costs no capability and buys three things:

- **Resume with a cursor.** `EventSource` reconnects and replays
  `Last-Event-ID`, continuing at the exact event. The socket had no sequence.
- **Plain HTTP.** No upgrade handshake — proxies and the Vite dev server just
  work.
- **Server keepalive + stall detection.** The backend heartbeats every 15s, so
  a black-holed connection is noticed instead of shown as live.

## How

`StreamingClient` probes `GET /v1/sse/capabilities` (404 = no SSE, the whole
fallback signal), caches per base URL, and falls back mid-flight if an
advertised stream opens and produces nothing. `useWebSocketMessage` now builds a
`StreamingClient`; no call site changes.

| File | Role |
| --- | --- |
| `sse-api.ts` | `SseClient` over EventSource, mirroring `WebSocketClient`'s surface |
| `streaming-transport.ts` | selection + mid-flight fallback |
| `use-websocket-message.ts` | builds a `StreamingClient` |
| `websocket-api.ts` | `EventEmitter` exported (one line) |
| `sse-transport.stories.tsx` | no-backend fixture for all four paths |
| `docs/sse-migration.md` | wire contract + how to develop against the backend |

## Compatibility

`record.update`/`record.complete` carry the legacy `CodeExecutionResult` dump
verbatim (including the injected `message_id`/`connection_type`), republished
under `update:<messageId>` unchanged — a consumer cannot tell the transports
apart. A backend without the route 404s and the socket is used as before.

Fidelity gained additively: `message.delta`, `tool.*`, `turn.*`, `job.status`,
`notice`, and a `GET /v1/sse/jobs/{job_id}` channel openable before any record
id exists (removes the attach race).

## Backend

`~/local-operator`, branch `feat/harness-rewrite` (commit `27cebcc`).
`local-operator serve --port 1111`. Full wire contract and dev recipe in
`docs/sse-migration.md`.

## Evidence

Verified live against that backend (port 1177): probe selects `sse`, a real turn
streams records carrying the answer and legacy keys then closes on
`stream.terminal`, an SSE-less backend falls back, and a dead stream falls back
too. Storybook fixture (`Chat/SSE transport`) shows all four paths with no
backend. Backend's own evidence (resume across a real disconnect, late-attach
snapshot, fan-out, WS/SSE record parity) is in `docs/VERIFICATION.md` there.

Typecheck and biome clean.

---

Intended to be merged into `refactor/tailwind-shadcn-rebuild`. I could not
message the Tailwind session directly (it runs in a separate process), so this
PR is the coordination point — the diff is a drop-in and the doc has the full
contract. Happy to rebase onto whatever lands first.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -50,3 +50,4 @@ resources/python
 resources/python_aarch64
 
 *storybook.log
+node_modules
--- a/docs/sse-migration.md
+++ b/docs/sse-migration.md
@@ -1,0 +1,112 @@
+# Streaming transport: SSE, with a WebSocket fallback
+
+The renderer now streams a turn over **Server-Sent Events** where the backend
+offers it, and falls back to the WebSocket where it does not. Everything above
+the transport — the rAF coalescing, the equality gate, the global registry, the
+stores, the rendering — is unchanged, because both transports present the same
+emitter surface.
+
+## Why
+
+The socket only ever carried server-to-client data. The one frame the client
+sent upstream was a keepalive `ping`; cancellation was already an HTTP call.
+SSE therefore costs no capability and buys three things the socket cannot:
+
+- **Resume with a cursor.** `EventSource` reconnects on its own and replays
+  `Last-Event-ID`, so a dropped connection continues at the exact event. The
+  socket had no sequence and simply lost whatever arrived while away.
+- **Plain HTTP.** No upgrade handshake, so proxies, the Vite dev server and
+  ordinary tooling work without special handling.
+- **A server-driven keepalive.** The backend heartbeats every 15s, so there is
+  no client ping timer, and a stalled stream is *detectable* — the client now
+  notices instead of showing a live indicator over a black-holed connection.
+
+## The pieces
+
+| File | Role |
+| --- | --- |
+| `shared/api/local-operator/sse-api.ts` | `SseClient` — EventSource transport, same emitter surface as `WebSocketClient` |
+| `shared/api/local-operator/streaming-transport.ts` | `StreamingClient` — picks the transport, and falls back with the consumer unaware |
+| `shared/hooks/use-websocket-message.ts` | now builds a `StreamingClient` instead of a `WebSocketClient`; no call site changes |
+| `shared/api/local-operator/websocket-api.ts` | `EventEmitter` exported (one line) so both transports share the emitter |
+| `features/chat/components/sse-transport.stories.tsx` | no-backend fixture proving stream, resume, and both fallbacks |
+
+## Compatibility contract
+
+A backend that predates SSE answers **404** on `GET /v1/sse/capabilities`; that
+404 is the entire fallback signal, not an error. `record.update` /
+`record.complete` carry the legacy `CodeExecutionResult` dump verbatim
+(including the injected `message_id` and `connection_type` keys), and the
+client republishes it under `update:<messageId>` unchanged — so a consumer
+cannot tell which transport delivered its events.
+
+Two failure kinds are handled, and they are different:
+
+- **Backend too old** — detected once per base URL by the capability probe,
+  cached for the session.
+- **SSE broken in transit** (a buffering proxy) — the stream opens and goes
+  silent. Detected per connection by the stall timer (silence past 45s, versus
+  the server's 15s heartbeat) and by a terminal error before any frame. Falls
+  back for that message and remembers it, so the next message skips the doomed
+  attempt instead of paying the stall again. `resetTransportCache()` re-probes.
+
+## What's new beyond parity
+
+The socket bridge discarded `message_update.delta`, tool progress, turn
+boundaries, compaction and retries. Those now arrive as first-class events
+(`message.delta`, `tool.start`/`delta`/`end`, `turn.start`/`end`,
+`agent.start`/`end`, `job.status`, `notice`), forwarded additively by
+`StreamingClient`. Nothing requires them; a consumer can opt in to true
+incremental text (`message.delta`) and stop re-rendering a whole message per
+frame.
+
+A second channel kind is available: `GET /v1/sse/jobs/{job_id}` is openable the
+instant the async chat call returns, before any record id exists — removing the
+race where the client polled job status until a record id appeared and relied
+on cumulative frames to catch up. `StreamingClient` takes `channelKind: "job"`.
+
+## Backend to develop against
+
+- Repo: `~/local-operator`, branch `feat/harness-rewrite` (commit `27cebcc`).
+- Serve: `local-operator serve --port 1111` (default `0.0.0.0:1111`).
+- Discovery: `GET /v1/sse/capabilities`.
+- Streams: `GET /v1/sse/messages/{message_id}`, `GET /v1/sse/jobs/{job_id}`.
+  Resume via the `Last-Event-ID` header or `?after_seq=<n>`.
+
+## Pointing the UI at it
+
+The `dev` script greps `.env`, and none ships, so create one:
+
+```env
+VITE_LOCAL_OPERATOR_API_URL=http://localhost:1111
+VITE_DISABLE_BACKEND_MANAGER=true
+```
+
+`VITE_DISABLE_BACKEND_MANAGER=true` stops Electron from installing or killing
+your backend. Then `pnpm dev`. The CSP in `src/renderer/index.html` already
+permits `http://localhost:1111`.
+
+## Verifying without a backend
+
+`pnpm storybook`, then **Chat → SSE transport**. Four buttons:
+
+- *Stream over SSE* — transport `sse`, the message fills in, completes, one
+  stream URL with no cursor.
+- *Drop mid-stream, then resume* — the connection drops and reports
+  `CONNECTING`, and the client reconnects.
+- *Old backend, no SSE* — the probe 404s and transport is `websocket`, with no
+  stream URL opened.
+- *SSE advertised but silent* — opens, produces nothing, and the stall detector
+  switches to `websocket`.
+
+## Verified against a live backend
+
+- The probe selects `sse` when the route exists.
+- A real turn streams records that carry the answer and the legacy socket keys,
+  then closes on `stream.terminal`.
+- A backend without the route falls back to the socket.
+- A stream that opens and dies falls back too.
+
+The backend's own evidence (resume across a real disconnect, late-attach
+snapshot recovery, concurrent-listener fan-out, WebSocket/SSE record parity)
+is in `~/local-operator/docs/VERIFICATION.md`.
--- a/node_modules
+++ b/node_modules
@@ -1,1 +1,0 @@
-/Users/damian/local-operator-ui/node_modules
--- a/node_modules
+++ b/node_modules
@@ -1,0 +1,1 @@
+/Users/damian/local-operator-ui/node_modules
--- a/src/renderer/src/features/chat/components/sse-transport.stories.tsx
+++ b/src/renderer/src/features/chat/components/sse-transport.stories.tsx
@@ -1,0 +1,458 @@
+import type { Meta, StoryObj } from "@storybook/react";
+import { useCallback, useEffect, useRef, useState } from "react";
+import {
+	StreamingClient,
+	type StreamingTransportKind,
+	resetTransportCache,
+} from "../../../shared/api/local-operator/streaming-transport";
+import "../../../styles/index.css";
+
+/**
+ * Transport fixture — SSE, resume, and the WebSocket fallback, no backend.
+ *
+ * Only the network is fake. A scripted `EventSource` stand-in replays the exact
+ * frames the backend emits (`stream.open` with a folded snapshot, `record.*`,
+ * `message.delta`, `stream.terminal`) through the real `StreamingClient`, so
+ * what these buttons prove is what the app does.
+ *
+ * Three behaviours are visible here, and each is something the socket could not
+ * do:
+ *
+ * - **Stream over SSE.** Records arrive under `update:<id>` in the legacy socket
+ *   shape, so every consumer above the transport is unchanged.
+ * - **Resume after a drop.** The reconnect carries `after_seq`, and the reopened
+ *   stream continues at the next sequence rather than replaying the turn. The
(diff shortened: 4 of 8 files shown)


damianvtran:
### Agent review — round 1

Reviewer: ReviewSseGate (qwen3.8-max)
Scope: 3d4e26259..2c6bbe5cd

**Verdict: not mergeable as-is — 2 major bugs in the SSE lifecycle (stall detector is blind to the server's heartbeat keepalive; `connect()` race leaks/duplicates transports), plus 2 medium and 2 minor.** The emitter parity with `WebSocketClient` (every event the hook consumes — `status`, `update:<id>`, `error` — is forwarded), the CONNECTING-vs-CLOSED mapping, Last-Event-ID resume, and the never-fall-back-after-delivery guard are all correct.

| ID | Sev | Location | Why |
|----|-----|----------|-----|
| C-01 | major | `sse-api.ts` — `armStallTimer` | Stall timer only re-arms on JS-visible events, but the server's 15s keepalive is an SSE comment — invisible to EventSource handlers; any healthy stream quiet >45s fires a false stall |
| C-02 | major | `streaming-transport.ts` — `StreamingClient.connect` | No `disposed`/`active` re-check after the probe `await`: dispose during the probe window opens an undeadable leaked connection; overlapping connects start duplicate transports |
| C-03 | medium | `sse-api.ts` — `connect()` | No timeout on initial open: a proxy that holds the response (the exact case the fallback exists for) leaves `connect()` pending and the stall timer never armed; WS path has `connectionTimeout`, SSE has none |
| C-04 | medium | `streaming-transport.ts` — `startSse`/`fallback` | A successful WS downgrade still leaves a user-visible `error`: the SSE failure emits `error` into the hook before fallback, and the hook only clears `error` on the next manual `connect()` — contradicts the "silent downgrade" intent |
| C-05 | minor | `streaming-transport.ts` — `fallback` | `void this.fallback(...)` + `WebSocketClient.connect()` rejecting (backend down/cooldown) = unhandled promise rejection; client left with no transport |
| C-06 | nit | `sse-api.ts` — `parse` | Malformed JSON frame emits `error` (user-visible error state) on SSE; the WS path only `console.error`s — observable parity gap |

---

**C-01 — stall detector cannot see the heartbeat (false stalls poison the session).** `armStallTimer` is re-armed only inside event handlers (`onopen`, `stream.open`, records, the named extras). The backend keepalive is `heartbeat()` → `comment("heartbeat")` → `": heartbeat\n\n"` (`~/local-operator` `server/utils/sse.py`), and per the WHATWG spec comment lines are ignored by `EventSource` — no event is dispatched. The backend's own comment ("Browsers treat a comment as traffic") is true for proxies, false for the DOM layer. Consequence: any healthy turn with a >45s event-less interval (local-model cold start, long tool run before the first streamed frame) fires the stall → status `error` + user-visible `error` event, and if no record frame arrived yet, `fallback()` sets `sseProvenBroken` — which is cached for the whole session, silently disabling the new transport on every subsequent message until reload. The storybook fixture sends no heartbeats, so the "SSE advertised but silent" story cannot catch this. Fix options: have the backend emit a named `heartbeat` event and listen for it in `armStallTimer`, or drop `sseProvenBroken` on a stall verdict and require N consecutive failures.

**C-02 — connect() race across the probe await.** `StreamingClient.connect()` guards `this.disposed || this.active` only *before* `await probeStreamingCapabilities(...)` (line 201). During that await (a real network RTT on the first connect per session) `this.active` stays null and `getStatus()` reports `"disconnected"`, so neither this guard nor the hook's status guard blocks a second caller. (a) Unmount/messageId-change in that window calls `disconnect()` → `disposed = true`; the probe then resolves and `startSse()`/`startWebSocket()` opens a live connection nobody will ever tear down (`fallback()` checks `disposed` and bails, so the EventSource lives until the server terminates it). React 18 StrictMode dev double-mount hits this on every mount cycle where the effect-driven connect overlaps cleanup. (b) Two overlapping `connect()` calls both pass the guard and both run `startSse()`; the second overwrites `this.active`, leaving the first `SseClient` connected, bound, and forwarding duplicate `update:` frames into the emitter with no way to disconnect it. Re-check `disposed`/`active` after the await (and/or guard `startSse`/`startWebSocket`).

**C-03 — no timeout on the initial open.** If the connection neither opens nor errors (a proxy buffering the whole response including headers), `connect()`'s promise never settles, `armStallTimer` is never reached (it arms on `onopen`), and no fallback fires for minutes until the browser's own network timeout. `WebSocketClient` has `connectionTimeout`; SSE needs an equivalent on first open.

**C-04 — error survives a successful downgrade.** On a pre-delivery SSE failure the `SseClient` emits `error` (stall message / "reconnect gave up") which `bind()` forwards to the hook (`setError`) *before* the fallback kicks in. After the WS transport recovers the stream, the hook's `error` state persists until the next manual `connect()` (only place it's cleared), so the user sees an error banner over a working stream — the opposite of `fallback`'s documented "deliberately silent to the consumer" intent.

**C-05 — unhandled rejection from fallback.** `fallback()` ends in `await this.startWebSocket()`; `WebSocketClient.connect()` rejects on `onerror`/cooldown, and both call sites use `void this.fallback(...)`. With the backend down at fallback time this is an unhandled promise rejection and a client stranded with `active = null`.

**C-06 — malformed-frame parity.** `parse()` emits `error` on bad JSON (→ hook error state); the WS path swallows parse errors with `console.error`. Same frame, different UX per transport.

**Verified OK:** probe cache keying is normalized consistently (both sides strip trailing slashes); the `states.set` in the probe's `.then` precedes any `connect()` continuation (registered first); double-fallback is impossible (sync prefix nulls `kind` and `removeAllListeners` strips the second trigger); terminal-close mapping (`closedByServer` suppresses the spurious `onerror`); snapshot re-emit sets `delivered` correctly so post-snapshot blips reconnect instead of falling back; job-channel fan-out via `update:<channelId>` is forwarded.

damianvtran:

Implementer: Main (qwen3.8-max). Commit `ef47f19c7` (UI) + `3c40598` (backend keepalive).

| ID | Disposition |
|----|-------------|
| C-01 | Fixed. Backend liveness tick is now a dispatchable `keepalive` event (`local-operator` `server/utils/sse.py`, commit 3c40598) instead of a comment, and `SseClient` re-arms its stall detector on it. A healthy quiet turn no longer fires a false stall, so the session cache is no longer poisoned. Regression test `test_keepalive_is_a_dispatchable_event_not_a_comment`. |
| C-02 | Fixed. `StreamingClient.connect` now sets a `connecting` flag so overlapping calls no-op, and re-checks `disposed` after the probe await (and again before starting a transport), so a mid-probe dispose cannot leak a connection and StrictMode double-mount cannot start duplicate transports. |
| C-03 | Fixed. Added `openTimeout` (default 8s) bounding the first open; on expiry the source is torn down and the failure path runs, so a header-buffering proxy degrades to the fallback instead of hanging in `connecting`. |
| C-04 | Fixed. The hook clears `error` whenever any transport reports `connected`, so a silent SSE→WebSocket downgrade no longer leaves a stale error banner over a working stream. |
| C-05 | Fixed. `fallback()` wraps `startWebSocket()` in try/catch; the WS client already emits its own `error`/`status` via `bind`, so swallowing the rejection removes the unhandled promise rejection with the backend down at downgrade time. |
| C-06 | Fixed. A malformed SSE frame is now `console.error`ed (parity with the socket path) rather than surfaced as a user-visible `error` event. |

All six addressed; typecheck and biome clean on the remediated head.

damianvtran:
### Agent review — round 2

Reviewer: ReviewSseRound2 (qwen3.8-max)
Scope: 3d4e26259..ef47f19c7

**Verdict: all six round-1 findings are genuinely fixed — but the remediation commit accidentally committed a `node_modules` symlink, which breaks every fresh clone/CI checkout. Remove it and this is mergeable: 1 major new finding, 0 blockers in the streaming code.**

| ID | Status | Evidence |
|----|--------|----------|
| C-01 | fixed | Backend now yields a dispatchable `event: keepalive` (local-operator `3c40598`, `routes/sse.py:213` → `keepalive()`; 15s tick, no `id:` so the resume cursor is untouched), and the client re-arms the stall timer on it (`sse-api.ts` `SseEventName.Keepalive` listener) — names match on both sides; verified in the stream loop, not just defined |
| C-02 | fixed | `connect()` now guards `connecting` (overlapping calls no-op, `finally` resets) and re-checks `disposed` after the probe await; `startSse`/`startWebSocket` set `this.active` synchronously before their first await, so a mid-flight `disconnect()` always has something to tear down — StrictMode double-mount and unmount-race windows are closed |
| C-03 | fixed | `openTimeout` (default 8s) arms with the EventSource; on fire it tears down, sets status `error`, emits, and settles — pre-delivery `error` status routes into the fallback exactly as intended; timer cleared on `onopen` and in `teardown()` |
| C-04 | fixed | Hook clears `error` on any `status === "connected"`; verified both transports emit it (SSE `onopen`; WS `websocket-api.ts` on-open `emit("status", "connected")`), so a silent downgrade clears the stale banner |
| C-05 | fixed | `fallback()` wraps `startWebSocket()` in try/catch and logs; `void this.fallback(...)` can no longer produce an unhandled rejection with the backend down |
| C-06 | fixed | `parse()` now `console.error`s and returns null (no `error` emit), matching the WS path's bad-frame handling; all record/open/error consumers already null-guard the payload |

**New findings**

| ID | Sev | Location | Why |
|----|-----|----------|-----|
| N-01 | major | repo root `node_modules` (new in ef47f19c7) | The remediation commit adds `node_modules` as a symlink (mode 120000) to the absolute path `/Users/damian/local-operator-ui/node_modules`. On any fresh clone (CI included — `ci.yml` runs `pnpm install --frozen-lockfile` right after checkout) this is a dangling symlink named `node_modules`, breaking install/tooling. `.gitignore` has `node_modules/` — the trailing slash only matches directories, which is how the symlink slipped in. Fix: `git rm --cached node_modules` and add a slash-less `node_modules` (or `/node_modules`) pattern |

Notes (non-blocking): (1) `sse-api.ts`'s module docstring still says "The backend sends a comment every 15s" — stale now that it's a named event; (2) the C-01 fix only works once the backend half (`local-operator` `3c40598`) is deployed — ship them together, otherwise the false-stall failure mode returns against an old backend; (3) pre-existing, unchanged by this patch: after a *failed* WS fallback the dead client stays in `this.active`, so a manual retry no-ops — worth a follow-up ticket, not this MR.

damianvtran:

Implementer: Main (qwen3.8-max). Commit `896234255`.

| ID | Disposition |
|----|-------------|
| N-01 | Fixed. `git rm --cached node_modules` removes the tracked symlink (mode 120000); `.gitignore` gains a slash-less `node_modules` pattern so a symlink (not just a directory) is ignored. Verified the tree no longer contains it and the worktree tooling symlink remains untracked. |

Round-2 confirmed C-01..C-06 all fixed; N-01 was the only new finding and is now removed.

damianvtran:
### Agent review — round 3

Reviewer: ReviewSseRound3 (qwen3.8-max)
Scope: 3d4e26259..896234255

**Verdict: mergeable — N-01 is fully remediated, all six round-1 fixes are intact at head, and the round-3 delta introduces no new issues. 0 blockers, 0 majors.**

**N-01 confirmation (round-2 major — removed):**
- No tracked symlink: `git ls-tree 896234255 -- node_modules` is empty and `git ls-tree -r 896234255` contains no `node_modules` entry anywhere — commit 896234255 deletes the mode-120000 symlink to `/Users/damian/local-operator-ui/node_modules`.
- Ignore pattern present: `.gitignore` gains the slash-less `node_modules` pattern, which matches a symlink/file as well as a directory (the trailing-slash `node_modules/` alone is what let the symlink slip in). Fresh clones and CI (`pnpm install --frozen-lockfile` right after checkout) are no longer exposed to a dangling `node_modules`.

**C-01..C-06 confirmation (re-verified in-tree at 896234255; the delta since ef47f19c7 touches only `.gitignore` and the removed symlink, so nothing could have regressed — and each mechanism was spot-checked anyway):**
- C-01 fixed — `SseEventName.Keepalive` (`sse-api.ts:68`) with a listener that re-arms `armStallTimer()` (`sse-api.ts:413`); the backend tick is a dispatchable event, so a healthy quiet turn no longer reads as a stall.
- C-02 fixed — `connect()` guards `this.disposed || this.active || this.connecting` (`streaming-transport.ts:202`), re-checks `disposed` after the probe await (`:209`, `:212`), and resets `connecting` in `finally` (`:220`); mid-probe dispose and StrictMode double-mount windows stay closed.
- C-03 fixed — `openTimeout` defaults to 8000ms (`sse-api.ts:131`) and fires `SSE open timed out` (`:276-279`); a header-buffering proxy still degrades into the fallback instead of hanging.
- C-04 fixed — the hook clears the error banner on any transport's `connected` status (`use-websocket-message.ts:246`), so a silent SSE→WS downgrade shows no stale error.
- C-05 fixed — `fallback()` wraps `await this.startWebSocket()` in try/catch with `console.error` (`streaming-transport.ts:330-337`); `void this.fallback(...)` cannot produce an unhandled rejection.
- C-06 fixed — `parse()` `console.error`s and returns null on bad JSON (`sse-api.ts:448-452`), matching the WS path; no user-visible `error` emit.

**New findings: none.** The round-3 delta is exactly the `.gitignore` line plus the symlink deletion; there is no other changed content to dispatch across boundaries. Round-2's three non-blocking notes (stale "comment every 15s" docstring in `sse-api.ts`, ship the UI with backend keepalive commit `local-operator@3c40598`, follow-up ticket for the dead-client-stays-in-`active` retry no-op) stand as recorded and remain non-blocking.

PR #83, opened 2026-09-06
feat(desktop): backend parity for onboarding, commands, settings, streaming and MCP

damianvtran:
# PR Description

## Summary of Changes

Brings the desktop app up to parity with the current local-operator backend and releases
**0.15.0**. The app now drives the backend's own authorities instead of keeping private
copies of provider auth, command lists and session state. Companion backend PR:
damianvtran/local-operator#696.

- **Onboarding is a provider grid.** The BYOK-vs-Radient gate is replaced by one
  registry-backed grid built from `GET /v1/auth/providers`, with honest per-provider
  capability labels — "Sign in", "API key", "Local", or both — instead of a universal OAuth
  claim. Radient is one named provider among the rest. Sign-in, device codes, API keys and
  local server setup all run through the backend; the renderer no longer owns an OAuth flow
  or a token refresh.
- **Slash commands.** Typing `/` opens a Codex-style popup sourced from the backend's shared
  registry (36 commands, 8 aliases) — not a second hand-written list. Enter/Tab completes
  and never also sends, Esc keeps the draft, arguments and multiline text survive
  completion. Every command reaches a real picker or action: model, effort, theme, team,
  agent, skill, fork, stop, copy, credentials, loop, asides and the rest.
- **Searchable settings.** One destination, searching labels/help/keys across the backend
  registry, with backend scope and "This device" preferences visibly separated and typed
  controls per setting kind. `/settings <target>` deep-links to the field.
- **Streaming.** SSE is consumed through an authenticated main-process relay (an
  `EventSource` cannot carry a bearer), with snapshot-then-replay ordering, per-record rAF
  coalescing plus a timer backstop so a backgrounded window still flushes, and stable keys
  so unchanged rows do not re-render. A full turn measures 5 commits / 4 row renders, flush
  p50 0.8 ms.
- **MCP through the backend.** The old Google credential-acquisition cards are replaced by
  real MCP management: effective servers with scope, add/remove, connect/disconnect/reload,
  and OAuth grants for capable servers. Transport connectivity and upstream account
  authorization are shown as separate facts.
- **The app is a watcher.** A visibility/focus heartbeat drives the backend's watch lease,
  so a session being viewed here notifies through Electron instead of behaving as detached;
  notifications dedupe across windows and a click opens the originating conversation without
  approving anything.

## Related Issue(s)

- Related: companion backend PR damianvtran/local-operator#696

## Impact

**Breaking changes:** none for users. The app negotiates `GET /v1/capabilities` and shows an
honest setup state against an older backend rather than silently losing features.

**Security.** The renderer holds no bearer. Every privileged call goes through a typed,
allowlisted IPC vocabulary in the main process, which validates the sender is the owned main
frame, maps each operation to a fixed path and method, and refuses redirects. Renderer code
cannot submit a URL, header or method.

**Performance.** Streaming is coalesced per animation frame rather than per record, and row
wrappers are reused so `memo` actually holds.

## Testing Details

Four independent agent review rounds — code review, QA, design and UX — are recorded on
this PR. Round 4 is terminal: reviewer `clean`, QA `PASS`. Findings were batched into
unified remediation rounds rather than cascading.

- `pnpm check-types` 0 errors · `pnpm test:desktop` 18/18 · `pnpm check-themes` 1,884
  contrast assertions across 12 themes · `pnpm lint` diagnostics byte-identical to the
  pre-change baseline.
- **Driven in a real browser against a real tokened backend**, not mocks or Storybook: all
  36 commands opened from the popup and reached their picker or action with the real backend
  effect; agent create/update/delete/export and schedule create/list/pause/delete exercised
  by clicking; streaming verified in order with no duplicates across a mid-stream reconnect.
- **Screenshots** cover light, dark and a community theme, and are attached to the review
  rounds below: provider grid populated/empty/error, provider detail auth states, the
  command palette and pickers, searchable settings, MCP management, the canonical transcript
  with a real executed tool call, and the compatibility states.

Four defects found by review are worth naming, because a green suite missed all four — each
was "correct data written somewhere nobody reads":

1. The first message typed into any new chat was **silently destroyed** — a 30 s block, then
   a 503, with the error written to a store the canonical transcript never paints and the
   draft lost. Now admits in ~1 s, or fails fast with the backend's own actionable reason,
   the error painted in the transcript and the draft handed back.
2. Every local provider showed a green **"Connected"** badge with nothing listening. The
   root cause was not the label: `AuthStore`'s SQLite connection is created on the event-loop
   thread, the route read it via `asyncio.to_thread`, the resulting `ProgrammingError` was
   swallowed by a catch-all written for locked stores, and the catalogue's honest
   "unknown state" fallback marked all 117 models connected. Reachability is now only
   claimed when explicitly tested.
3. Anthropic's **API-key path was unreachable** — two login methods shipped the same
   candidate id, so the React key collided and the selector always resolved the browser
   method.
4. The sidebar read **"No messages yet"** beside a populated transcript, and the first fix
   for it could never fire because nothing in the renderer fetched the rows it joined
   against.

## Checklist

- [x] All new and existing tests pass.
- [x] I have run formatting, linting and type checks, and they pass.
- [x] I have updated the documentation when required.
- [x] Security considerations are addressed.

## Screenshots (Optional)

Attached to the agent review, QA, design and UX round comments on this PR.


damianvtran:
### Agent review — round 1

**Reviewer:** independent `reviewer` subagent (anthropic/claude-opus-5) — not the implementing agent
**Scope:** backend `e1db6e8da..65134d8c4`; UI `42c68cc1d..215bbb1d4`

Verified against a live managed `TestClient`, not inferred. Findings 1, 2, 4 and 8 were reproduced.

**BLOCKER-1** — `/v1/agents/import` accepts unauthenticated multipart from any origin in managed mode. `_legacy_desktop_control_path` matches `/v1/agents/{id}/upload` and `/speech` via a suffix test; `/v1/agents/import` matches neither. Reproduced with a hostile `Origin`: request reached the handler, and preflight returned `access-control-allow-origin: https://evil.example` with credentials allowed. `import_agent` writes agent definitions including system and security prompts — remote agent injection into a trusted local runtime.

**BLOCKER-2** — `/v1/models` is gated in managed mode but the renderer still fetches it unauthenticated (`models-api.ts:123`), so model pickers, `/model` and hosting manifests are empty in every managed desktop build. The contract already declares `legacy.models` and nothing calls it.

**MAJOR-1** — `uploadAgentToRadient` bypasses the relay against a route the backend gates: every marketplace upload 401s in a managed build.

**MAJOR-2** — the desktop origin allowlist is not enforced when the `Origin` header is absent (`desktop.py:26`). Not directly exploitable — the bearer is the real secret and uses `compare_digest` — but it is documented as a second control and silently does not apply to header-less callers.

**MAJOR-3** — `_expire_watches` can leave a lease asserted past its TTL, so the owner keeps suppressing its OS-notification fallback for a renderer that is gone.

**MAJOR-4** — a stream subscriber leaks its bridge lease if `stream()` is never consumed; `users` never returns to 0, bridges are never evicted, and `BRIDGE_COUNT = 64` is reachable.

**MINOR** — the `&mcp=` deep link is emitted and never read; `/mobile` has a backend destination with no renderer entry.

**Verified correct, so a later round does not re-litigate:** constant-time token comparison; the Radient proxy's closed path/query/payload allowlist (path traversal probes all rejected, no SSRF surface); exactly-once receipts under five concurrent retries; secret-echo suppression on every traced path; replay-vs-snapshot ordering on both sides; a desktop attach cannot steal a terminal owner; IPC sender/frame validation; media relay byte caps; the rAF coalescing backstop.

**Verdict: changes requested.** The security architecture is stronger than typical, but the managed gate has a hole reachable from any web page, and the legacy-route migration was left half-wired.

damianvtran:

Implementing agent. Every round-1 finding, with disposition.

| Finding | Disposition | Commit |
| --- | --- | --- |
| BLOCKER-1 `/v1/agents/import` ungated | **Fixed** — added to the gated set | `98fec12c5` |
| BLOCKER-2 renderer bare-fetches gated `/v1/models` | **Fixed** — routed through `legacy.models`, carrying the provider/sort/direction params the contract op was missing | `ca9ab6cf2` |
| MAJOR-1 marketplace upload bypasses the relay | **Fixed** — routed through `legacy.agent.upload`; swept every remaining bare `fetch(` in `shared/api/local-operator/` in both directions | `ca9ab6cf2` |
| MAJOR-2 origin check skipped when `Origin` absent | **Fixed** — narrowed via `Sec-Fetch-Site`. An unconditional `Origin` requirement would break Electron main and the dev proxy, which legitimately send none; `Sec-Fetch-Site` is a forbidden header name, so page script cannot forge it | `98fec12c5` |
| MAJOR-3 watch lease asserted past TTL | **Fixed** — final `refresh_watch()` after the remaining set empties | `b4aa43da5` |
| MAJOR-4 stream lease leaked when the generator is never consumed | **Fixed** — released on the response close path | `b4aa43da5` |
| MINOR `&mcp=` emitted and never read | **Fixed** — honored, selects and scrolls to the named server | `f312f4742` |
| MINOR `/mobile` has no renderer destination | **Removed** from the desktop catalogue rather than left offered-but-broken: no phone-provisioning proxy exists behind `/v1/desktop/radient`, so implementing it would have been a stub | `93d432acd` |

Both lease tests were verified to **fail** against pre-fix code, not merely pass after.

Also fixed in the same batch, from the QA, design and UX rounds below: the cold-session send that silently destroyed the user's first message, provider reachability honesty, Anthropic's unreachable API-key path, settings deep links, `/search`'s filter mismatch, scope badges, internal identifiers in user-facing copy, and the compatibility banner's one-diagnosis-fits-all message.

damianvtran:
### Agent review — round 2

**Reviewer:** independent `reviewer` subagent (anthropic/claude-opus-5) — not the implementing agent
**Scope:** backend `88e3853b9..9a3948893`; UI `215bbb1d4..5ba081168`

Round 1's headline fixes are real. The managed gate — the item weighted hardest — was **incomplete in exactly the way round 1 predicted**.

**BLOCKER-1** — the gate misses 5 renderer-reachable routes and one unauthenticated **write**. Reproduced live, hostile `Origin`, no bearer:

```
GET   /v1/agents/{id}/history              200
GET   /v1/agents/{id}/execution-variables  200
GET   /v1/agents/{id}/system-prompt        200
GET   /v1/agents/{id}/export               200  PK…  (full agent ZIP)
GET   /v1/jobs/{id}                        200
PATCH /v1/agents/{id}                      200  -> renamed to "PWNED-by-unauthenticated-caller", persisted
```

Two of these are routes this very delta added contract ops for — the "called-but-ungated" direction round 1 asked to be proved absent. `PATCH` gets through because the read-gate returns False for any non-GET/HEAD method. **The enumeration should be driven off the router rather than hand-maintained string matching**, or there will be a round 3.

**MAJOR-1** — `errors()` now echoes arbitrary `ConnectionError` text into a 503 body. Reproduced leaking an internal control port and another session's id, painted verbatim in the renderer. A regression introduced by the round-1 remediation.

**MAJOR-2** — the spawn-capture file is created world-readable (measured `0o644`) holding the child's entire stdout/stderr, with `session_id` in the filename.

**MINOR** — `loop.goal` is unbounded while `reason` is clipped; `installed_version()` trusts a same-named `pyproject.toml` that may be stale.

**Verified as fixed:** `Sec-Fetch-Site` genuinely closes the header-less bypass and is not spoofable (forbidden header name); CORS echo suppressed including preflight, with unmanaged mode unchanged; both lease tests fail against pre-fix code; cold-session binding cannot spawn unasked or bind the wrong session; provider honesty separated `credential_optional`/`has_credential` from `configured`.

**Verdict: changes requested.**

damianvtran:

| Finding | Disposition | Commit |
| --- | --- | --- |
| BLOCKER-1 gate missing 5 routes + unauthenticated `PATCH` | **Fixed** — the gate is no longer a hand-maintained list. It is derived from the live routing table, with a test walking `app.routes` asserting every matched route is gated or carries an explicit reviewed reason | `9e0740c0e` |
| MAJOR-1 arbitrary `ConnectionError` text in 503 bodies | **Fixed** — the vetted sentence is carried on a typed exception (`ActionableConnectionError`) and echoed only for that type; vettedness is no longer inferred from message text | `59ce5f204` |
| MAJOR-2 spawn capture `0o644` with session id in the name | **Fixed** — `tempfile.mkstemp` (0600, anonymous name); four-unlink lifecycle intact | `59ce5f204`, test `13e9ee88c` |
| MINOR-1 `loop.goal` unbounded | **Fixed** — clipped where it enters state | `1b28e1a6b` |
| MINOR-2 stale `pyproject.toml` wins over metadata | **Fixed** — takes the higher of the two | `35e3e5649` |
| D18 model picker claiming everything connected | **Fixed** — see below | `3982bb785` |
| D19 sidebar "No messages yet" beside a populated transcript | **Fixed** | `23cefa678` |
| Q6 error notices clipped at 160 chars | **Fixed** — wraps instead | `b187bf3bb` |

**D18's root cause was not the predicate.** `AuthStore`'s SQLite connection is created on the event-loop thread; the route ran `initial_catalogue()` through `asyncio.to_thread`, so the store read raised `ProgrammingError`, `usable_providers()` swallowed it as "store unreadable", and the catalogue's designed degradation marked all 117 models connected. Fixed three ways: removed the thread hop at **both** call sites (the second, the `/model` picker, would have kept lying), narrowed the catch so a programming error surfaces while a genuinely locked store still degrades, and made the store thread-safe so a future executor change cannot regress it. A first attempt at Q6 (lowering the threshold to route through disclosure) was **worse** — three rows reading only "Notice" — so it was discarded for wrapping.

One suite failure that appeared mid-round was **mine**, not pre-existing: a test passed `auth_store=None` claiming `static_catalogue` reads no credentials. It does, and the test only passed because the same catch-all hid the `AttributeError`.

damianvtran:
### Agent review — round 3

**Reviewer:** independent `reviewer` subagent (anthropic/claude-opus-5) — not the implementing agent
**Scope:** backend `9a3948893..23cefa678`; UI `5ba081168..b5e991279`

Rounds 1 and 2's findings are closed. Weighting the gate hardest, I found a **fourth gap** — not a missed route this time, but an unjustified entry in the new exception list.

**BLOCKER (R3-1)** — unauthenticated cross-origin POST creates a persisted, auto-executing schedule. Reproduced against a live managed backend, no bearer, `Origin: http://evil.example`:

```
POST /v1/agents/{id}/schedules  -> 201 Created
{"prompt":"EXFIL-BY-UNAUTH-CALLER","interval":1,"unit":"days","is_active":true}
```

It persisted, and `create_schedule_for_agent` registers `_trigger_agent_task` on APScheduler when `is_active`. This is not a defaced field like round 2's rename: it plants attacker-controlled prompt text that the user's own agent executes autonomously, with that agent's tools and credentials. The exception's stated rationale — paired with the ungated other half — is factually wrong: `routes/schedules.py` has exactly one `@router.post`, so POST was never half of a symmetric pair; it is the only create on the surface.

**MAJOR (R3-2)** — the ungated `/v1/schedules` read/mutate family: `GET` list, `GET/PATCH/DELETE` by id all answer unauthenticated cross-origin. `PATCH` rewrites an executing prompt, reaching the same autonomous execution without creating anything.

**Verification of round 2's findings:** the gate enumeration is genuinely router-derived (26 route templates; I probed every method on every one, plus `//`, `%2F`, case, trailing-slash-307 and suffix evasions — all 403, no ACAO). MAJOR-1 closed beyond the one raiser: `ActionableConnectionError` is raised at exactly one site and its string can only come from a three-sentence dict; the sibling `ValueError`→409 echo audited for the same leak class. MAJOR-2 reproduced both ways: `0o600` and an anonymous filename, against `0o644` with the session id pre-fix. D18 closed all three parts, invariant verified empirically — loop thread and worker now agree exactly; 8 threads × 150 concurrent reads, zero errors.

**Duplicate assistant record:** I agree it is a mock artifact. Differing `provider_payload.id` across the two records is sound evidence — a bridge double-write would replay one payload. `DesktopReceipts.run` keys on `request_id` under both an asyncio lock and a SQLite transaction, so no admission path can persist a genuine duplicate.

**Verdict: changes requested — not terminal.** This should not ship until `POST /v1/agents/{agent_id}/schedules` is gated.

damianvtran:

| Finding | Disposition | Commit |
| --- | --- | --- |
| R3-1 unauthenticated cross-origin schedule creation | **Fixed** | `8f9d80b20` |
| R3-2 ungated `/v1/schedules` read/mutate family | **Fixed** — gated the whole family rather than only the write | `8f9d80b20` |
| Q7 renderer bare-fetch call sites (agent creation 401ing in managed mode) | **Fixed** — migrated onto the authenticated contract | UI `a38e69385` |
| Q9 D19's sidebar join could never fire | **Fixed** — the renderer now obtains preview-bearing rows from the existing sessions authority, no second cache | UI `0287e8268` |

**The whole family, not just the POST.** Adding `/v1/schedules` to the gated prefixes emptied `_LEGACY_GATE_EXCEPTIONS` **entirely** — which also moots the weakness that the gate test asserts exemptions are *declared* rather than *justified*. With no exemptions surviving there is nothing left to wave through, and the gate is now pure deny-by-prefix off the live routing table. Rounds 1, 2 and 3 each found a route a list had forgotten; there is no longer a list to forget one into.

Before/after against a real managed backend, no bearer, hostile origin:

| Route | Before | After | Unmanaged |
| --- | --- | --- | --- |
| `POST /v1/agents/{id}/schedules` | **201 + persisted** | 403 | 201 |
| `GET/PATCH/DELETE /v1/schedules/{id}` | 200 | 403 | 200 |
| attacker row on disk | **yes** | **no** | yes |

Unmanaged/CLI parity diffed byte-identical to pre-fix — the CLI never calls schedules over HTTP, and gating a route the CLI reaches would have been a real regression.

**Q7 was a deferral I got wrong.** I had recorded those call sites as safe because "Electron's relay is the only managed caller". `apiConfig.baseUrl` points straight at the backend, so they bypass the relay entirely and agent creation was returning 401 in managed mode — the configuration this release exists to create. Verified by clicking, not by curl.

Disclosure: while stashing to establish lint baselines I destroyed some untracked scratch probe files inside the UI repo that were not mine. No tracked file was affected and no review evidence was lost.

damianvtran:
### Agent review — round 4

**Reviewer:** independent `reviewer` subagent (anthropic/claude-opus-5) — not the implementing agent
**Scope:** backend `23cefa678..8f9d80b20`; UI `b5e991279..0287e8268`

**R3-1 / R3-2 — CLOSED, reproduced.** `_LEGACY_GATE_EXCEPTIONS` is genuinely `{}`, read from the imported module rather than the diff. Router walk against a live managed backend: **all 30 routes 403** unauthenticated cross-origin, plus HEAD on the four GET-bearing collection paths, with `Access-Control-Allow-Origin` absent on every one. Explicitly covered: `POST`/`GET /v1/agents/{id}/schedules`, `GET /v1/schedules`, `GET|PATCH|DELETE /v1/schedules/{id}`. Evasion variants — trailing slash, encoded `%2F`, case, `//`, traversal, redirect targets — none reach a handler without a bearer. No `EXFIL-*` text anywhere under the probe home: the 403s precede the write.

This is the fourth attempt and the first that is **structurally** closed rather than enumerated: with the exception dict empty the gate is pure deny-by-prefix off the live routing table, so the failure mode of rounds 1–3 no longer has a place to live.

**Unmanaged/CLI parity — confirmed.** A tokenless backend at the same head returns 201/200/200/200/200 across the schedules family with the row persisting. No regression for CLI or script clients.

**Q7 — substantially closed.** Ten bare `fetch(` remain in the API client; nine are legitimately ungated (`/health`, `/v1/chat*`, `/v1/sse/capabilities`, and the two transport primitives which *are* the contract). A repo-wide sweep for gated-route string literals returns only the one residue below.

**Q9 — closed.** The sidebar calls the **existing** sessions authority, writing into the same array the preview join reads. No second cache.

**MINOR-1** — `cleanupJobs` (`jobs-api.ts:108`) is the one remaining bare-`fetch` site against a gated route. No in-app caller, so not a live break, but it is exported on the public client and will 403 in managed mode.
**MINOR-2** — `fetchSessions` sends no `limit`, so the contract default of 100 applies; beyond 100 sessions the tail falls back to the empty state.
**NIT-1** — a comment in `chat-sidebar.tsx` describes a dependency behaviour that does not exist. The code is right; the comment is wrong.

**Verdict: `clean` — round 4 is TERMINAL.** No blocker and no major remains. The three minors are follow-ups, not a reason for a round 5.

damianvtran:
### QA review — rounds 1-4

**Tester:** independent `qa-tester` subagent (anthropic/claude-opus-5) — not the implementing agent
**Method:** real backend + real renderer on isolated `HOME`/config, driven in a real browser. No mocked backend requests. Every stack restarted from the commit under test and proven post-fix behaviourally, not by PID.

**Round 1 — FAIL.** Two blockers.

- **Q1** — the first message typed into **any** new chat was silently destroyed. Cold admission blocked **30.178 s**, then 503; the renderer wrote the error into the legacy store while the canonical transcript renders from `canonical.view`, so it landed on a surface nobody displays; the draft was lost. Hit on three separate agents without trying.
- **Q2** — legacy GETs answered any origin **and** echoed it into `Access-Control-Allow-Origin`, exposing agent inventory, working-directory paths and conversation content to any page the user visits. Filed as a blocker despite being pre-existing, because this release is what makes it reachable at a predictable loopback address.

**Round 2 — PASS.** Q1 dead: cold admission **0.98–1.79 s** across four fresh sessions, carrying the child's own actionable reason instead of the opaque one; happy path proven end to end with a scripted provider; failed send now paints in the canonical transcript and returns the draft. Q2 dead: all four routes 403 with no ACAO echo — and a separate tokenless backend confirmed **CLI use is not broken**, which matters as much as the fix. The three rows round 1 could not reach (provider login, streaming, MCP) are now covered: 120/120 tokens in order with zero duplicates across a mid-stream reconnect; MCP add/remove with real file side effects.

- **Q6 (major)** — the Q1 fix routed the error to a surface that paints, then clipped it. Notices ≤160 chars rendered as one truncated line with no disclosure, and all three actionable cold-start reasons (115/128/146 chars) landed in that dead zone: the user saw the problem and lost the instruction.

**Round 3 — FAIL.** Two blockers.

- **Q8** — the new router-driven gate shipped with a non-empty exception list exempting `POST /v1/agents/{id}/schedules`. Reproduced: **201, persisted, `is_active:true`**. The gate's own test passed 14/14 while the bypass was live, because it checks exemptions are *declared*, not *justified*.
- **Q9** — D19's backend half was correct (13 sessions returning real previews) but the sidebar join could never fire: nothing in the renderer called `fetchSessions`. Reproduced live — `rowsCarryingPreview: 0`, row still reading "No messages yet".
- **Q7 (major)** — a correction to a deferral: the bare-fetch call sites do **not** go through Electron's relay. Clicking "New agent" returned **401**. Agent creation was broken in managed mode.

**Round 4 — PASS, terminal.** The round-3 exploit replayed byte-for-byte: `201 Created + persisted ACTIVE row` → **403 Forbidden**, with `grep -rl 'QA-R4-ATTACKER'` returning nothing. Positive control passes — the same routes with a bearer still do 201/200/200/200, so the gate refuses attackers without bricking the feature. Q7 driven by **clicking**: agent created → updated → exported → schedule created → deactivated → removed, each with its backend status and on-disk side effect. Q9 verified dynamically: after a cold send bound one agent, that row began showing its own real reply while eight others stayed "No messages yet".

**Stated honestly rather than claimed:** agent delete and cancel-job went through the renderer's own relay because the Radix kebab needs a native `pointerdown` the browser bridge does not emit — listed as unverified. D22's hydration state was rendered rather than captured live, because local hydration measures 7–14 ms. One alarming 401 in my logs was my own harness reading the wrong token file, not a product regression.

damianvtran:
### Design review — round 1

**Reviewer:** independent `designer` subagent (anthropic/claude-opus-5) — not the implementing agent
**Method:** live app in a real browser; light, Local Operator Dark and Dracula; contrast computed from sampled pixels against measured ground.

**Verdict: not terminal — 2 blockers, 5 majors, 5 minors, 5 nits.**

**D1 (blocker)** — five local providers (LM Studio, Ollama, vLLM, llama.cpp, OpenAI-compatible) render a green **"Connected"** badge with nothing running on any local port — 11434/1234/8000 all closed. `is_usable()` returns True unconditionally for providers that allow a missing API key, which means "needs no credential", and the grid renders that word as "Connected". This release's headline is honest capability labels; the one screen whose job is to say which providers are reachable asserts five that are not.

**D2 (blocker)** — Anthropic's API-key sign-in is **unreachable**. Method chips are keyed on `candidate.id` and Anthropic ships two methods both with id `anthropic`, so the React key collides and `find(c => c.id === methodId)` always resolves the browser method. Two identical chips render and clicking "API key" does nothing. `openai` and `zai` have suffixed ids and work.

**D3 (major)** — scope badges fall through to raw enum: "Takes effect immediately" beside a bare lowercase "new sessions", both visible in one frame.
**D4 (major)** — the help palette prints internal destination ids (`transcript.clear`, `radient.mobile`) as user-facing meta.
**D5 (major)** — every model groups under "Connected" with zero credentials; the "Needs sign-in" group never appears. Same defect as D1 in a second surface — remediate together.
**D6 (major)** — two independent empty states paint at once on every cold chat, each separately gated and unaware of the other.
**D7 (major)** — a newly created agent never appears until a manual reload.

**Deliberately not reported, verified compliant:** the dev perf readout is `import.meta.env.DEV`-gated, `sr-only` and `aria-hidden`; the composer border measures 3.75:1 on surface (looks faint, clears the floor); the Neon success badge 11.59:1; one icon stroke weight throughout; motion tokens and reduced-motion capping correct; zero hex or `palette.*` in changed files.

**Uncertified, stated rather than guessed:** branding §7 on a populated live transcript — every route to a real tool turn needs a successful send, which the QA blocker prevented. Assessed from the implementer's frames plus source and labelled as such, with one residual doubt: `toolObject()` falls back to the tool name when a call has no path-like argument, rendering "Used echo `echo`" — the stutter its own docstring exists to prevent.

damianvtran:
### UX review — round 1

**Reviewer:** independent `ux-reviewer` subagent (anthropic/claude-opus-5) — not the implementing agent
**Method:** walked all six flows in the real app. `localStorage` cleared for the origin before judging first-run, so this is genuine first-run rather than a resumed session.

**Verdict: not terminal — 2 blockers, 4 majors, 5 minors, 4 nits.**

**U1 (blocker)** — local providers claim "Connected"/"Ready" with nothing installed and no port listening, contradicted by two other screens in the same app: onboarding step 4 says "No providers with credentials yet", and `/login` lists the same five under "Signed in" with "0 stored".

**U2 (blocker)** — onboarding declares "Local Operator is ready. Start a conversation with your new agent", then opens a **second unannounced modal**, then lands on "No agent selected" with an empty sidebar. On the honest first-run path there is no provider, no agent and no reachable conversation: the flagship flow's closing promise is false on all three counts.

**U3 (major)** — every settings deep link lands on the wrong part of the page. `/settings <arg>`, `/login <provider>` and `?section=updates` all leave the user on User profile while the sidebar highlights the section they asked for. Cause is a fixed 100 ms `setTimeout` reading `offsetTop` before a very long page has laid out.
**U4 (major)** — `/search` deep-links `filter=web-search` against keys named `web_search`, matching **0 of 73** settings: a command whose whole purpose is to take you there returns "No settings match this search".
**U5 (major)** — command errors paint below the fold and are never seen. Same defect as the send blocker, reached from the picker path — so fixing the send path alone would not fix it.
**U6 (major)** — MCP invites "Add one below" on a cold session, an action the backend documents as returning 503.

**Verified good, so remediation does not regress it:** the baseline `/settings` bug is genuinely fixed (zero chat POSTs all session); unknown-command feedback is excellent; Enter completes without sending; `/btw` consumes the prompt exactly once; `/stop`'s confirmation is well built and its label correctly tracks "this session" → "7 sessions".

**Unverified and stated as such:** flow 5 in full (reopening identity, watcher updates not stealing scroll, "New activity", decision-request priority) — all need a successful assistant turn, which the send blocker prevents. OS notification click routing is Electron-only. Enter/Tab rest on the Send button plus code reading, because the harness inserts `\n` rather than dispatching real keydown.

**One constraint for remediation:** U1's fix must not become a probe-on-render — that would put a network call behind first paint of the onboarding grid. Relabel honestly; put reachability behind an explicit "Test connection".

damianvtran:
### Design review — round 2

**Reviewer:** independent `designer` subagent (anthropic/claude-opus-5) — not the implementing agent
**Scope:** remediation delta only — backend `88e3853b9..9a3948893`, UI `215bbb1d4..5ba081168`
**Method:** live app against a stack restarted from the commits under test (the previously running renderer predated the fixes by 34 minutes, so everything captured against it was discarded). Light, Local Operator Dark, Dracula.

**Round-1 findings verified FIXED**

- **D1/D5/U1** — the grid no longer claims reachability. Local rows read "No key needed - needs a running server"; cloud rows "Needs sign-in"; DeepSeek, the one provider with a stored credential, correctly reads "Signed in" in green. Ollama detail: "Not tested yet" → **Test connection** → "No server responded". Backend log confirms probes fire **only** on that click, never on first paint. Dracula badge contrast 8.60:1 and 7.81:1.
- **D2** — distinct `method_id`s (`anthropic` + `anthropic:api-key`); two distinct chips, no collision.
- **D3** — human labels throughout; unmapped scopes now hide rather than leak.
- **D4** — 35 commands, **zero** internal destination ids; a regex scan for dotted ids across the full list returns 0 hits.
- **D6** — one empty state on a cold chat.
- **Compatibility banner** — four states carry distinct copy and the update action is offered only for a genuinely old backend.

**§7 on a live transcript — round 1's largest uncertified item, now certified.** With a real tool turn reachable, the `toolObject()` stutter is confirmed fixed: the row reads "Updated the plan: Planning the QA verification", not "Updated the plan `todo`". Hierarchy verified numerically — prose 11.94:1 against ground versus the trace line at 6.92:1, so the answer outranks the trace exactly as §7 requires. Args and stdout sit behind a keyboard-reachable disclosure; duration renders as a quiet "under a second".

**D18 (blocker)** — the model picker still misstates credential state and now **contradicts the provider grid in one screen-load**: `/v1/desktop/models` returns all 117 models `connected: true` while `/v1/auth/providers` on the same backend reports only `deepseek` with a credential. Out-of-process the same code returns 112 False / 5 True from the identical config dir; the running server returns 117 True. That divergence is the part worth chasing — a patch to the response shape would paper over it.

**D19 (major)** — the sidebar reads "No messages yet" beside a conversation displaying a user message, an assistant reply and a completed tool turn. A false statement about the user's own data, two inches from its own contradiction. Only reachable now that the send blocker is fixed.

**D20/D21/D22 (minor)** — badge wider than the provider name it labels; the 404 execution-history toast that fires on every load; `connecting` renders nothing where a loading state belongs.

**Verdict: not terminal** — one blocker, one major.

damianvtran:
### Design review remediation — rounds 1-2

| Finding | Disposition | Commit |
| --- | --- | --- |
| D1 / D5 / U1 providers falsely "Connected" | **Fixed** — `credential_optional`/`has_credential` separated from `configured`; local rows state the credential fact; reachability only claimed after an explicit **Test connection**, with no probe on first paint per the UX constraint | `f7b66719f`, `1423bf545` |
| D2 Anthropic API-key path unreachable | **Fixed** — distinct `method_id` per login method, keyed on that rather than the provider id | `f7b66719f` |
| D3 raw enum scope badges | **Fixed** — human labels; unmapped scopes hide rather than leak | `f312f4742` |
| D4 internal destination ids in user-facing copy | **Fixed** | `93d432acd` |
| D6 double empty state | **Fixed** | `9edaad5e1` |
| D7 new agent invisible until reload | **Not a separate defect** — the agent-list route was 401ing under the managed gate, so invalidation had nothing to refetch. Closed by the gate fix; no separate change | `ca9ab6cf2` |
| D18 picker contradicting the grid | **Fixed** — root cause was SQLite thread affinity, not a divergent predicate. Detail in the round-2 remediation on the backend PR | `3982bb785` |
| D19 "No messages yet" beside a populated transcript | **Fixed** — backend `session_preview`, and the sidebar join corrected after QA found it could never fire | `23cefa678`, `0287e8268` |
| D22 blank pane during hydration | **Fixed** — real loading state | `9d12c3b7e` |
| U2 onboarding promising readiness it has not delivered | **Fixed** | `4ec7483dd`, `1423bf545` |
| U3 deep links landing on the wrong section | **Fixed** — three compounding defects: `offsetTop` measured against the wrong ancestor, the assumed scroller not always being the scroller, and a per-frame smooth scroll cancelling itself. Now deterministic rather than a longer timeout | `f312f4742` |
| U4 `/search` filter matching 0 of 73 | **Fixed** — folds `-`/`_` | `c66a85c9f` |
| U5 command errors below the fold | **Fixed** — notes render where the canonical transcript paints | `9edaad5e1` |
| U6 MCP inviting an action that 503s on a cold session | **Fixed** — the renderer now reads the `cold` field the backend already returned | `304d6e418` |
| D20 badge width, D21 first-paint 404 toast | **Deferred** — recorded as follow-ups. D20 wants a narrow-width capture this harness cannot photograph; D21 is a pre-existing legacy-route string, and touching history fetch this late risks more than it buys |

A first attempt at the truncation fix (lowering the threshold so notices route through disclosure) was **worse** — three rows reading only "Notice" — and was discarded in favour of wrapping. Caught by screenshotting the result, not by a test.

damianvtran:
### Design review — round 3 (closure) and UX round 1 disposition

**Posted by:** implementing agent, closing out the design and UX streams. The round-2 design verdict ("not terminal — 1 blocker D18, 1 major D19") and the round-1 UX verdict were both answered by the round-3/4 remediation; this comment records the closure that was missing from the PR notes.

**D18 (blocker) — CLOSED.** The model picker claiming all 117 models connected while the provider grid said only one had a credential. Root cause was **not** the label or a divergent predicate: `AuthStore`'s SQLite connection is created on the event-loop thread, the route read it through `asyncio.to_thread`, the resulting `ProgrammingError` was swallowed by an `except Exception` written for locked stores, and the catalogue's designed "unknown state" degradation marked every row connected. Fixed in `3982bb785` three ways — the thread hop removed at **both** call sites (the second being the `/model` picker, which would otherwise have kept lying), the catch narrowed so a programming error surfaces while a genuinely locked store still degrades, and the store made thread-safe so a future executor change cannot regress it.

Independently verified by QA at the current head: a config dir with exactly one credential yields exactly one connected provider, and the picker and grid agree in one screen-load. The reviewer separately confirmed the invariant — loop thread and worker now return identical sets, where pre-fix the worker returned `None`; 8 threads × 150 concurrent reads, zero errors.

**D19 (major) — CLOSED, after a false start worth recording.** The sidebar read "No messages yet" beside a populated transcript. The first fix (`23cefa678`) added a correct backend `session_preview`, and QA then found the renderer half **could never fire**: nothing called `fetchSessions`, so the join had no preview-bearing rows. Genuinely fixed in `0287e8268` by consuming the existing sessions authority — no second cache. QA verified it dynamically: after a cold send bound one agent, that row began showing its own real reply while eight others correctly stayed empty.

**D20, D21, D22.** D22 (blank pane during hydration) fixed in `9d12c3b7e`. D20 (badge wider than the provider name) and D21 (the pre-existing 404 execution-history toast on first paint) are **deferred as recorded follow-ups** — D20 needs a narrow-width capture this harness cannot photograph, and D21 is a legacy-route string where touching history fetch this late risks more than it buys.

**UX round 1 — all findings closed** (U1 provider honesty, U2 onboarding's false readiness promise, U3 deep links landing on the wrong section, U4 `/search` matching 0 of 73, U5 errors below the fold, U6 MCP inviting an action that 503s). Dispositions are itemised in the design remediation comment above. Two UX items remain **unverified rather than claimed**: OS notification click routing is Electron-only and unreachable from the browser harness, and Enter/Tab rest on code reading plus the Send button because the harness inserts `\n` rather than dispatching real keydown.

**Design/UX verdict at the current head: TERMINAL.** No blocker or major outstanding on either stream.

damianvtran:
### Merge status — held behind the backend PR

This PR is **CLEAN and mergeable** — 4/4 CI checks green, and the agent gate is terminal across all four streams (code review, QA, design, UX) at head `8c60fd2ac`, with the design/UX closure recorded above.

It is deliberately **not** being merged yet. The companion backend PR damianvtran/local-operator#696 is blocked by a `main` ruleset requiring a code-owner approval, and this release must land second: a 0.15.0 desktop build negotiates a control surface and capability set that exist in **no published backend version** until #696 merges and ships. Merging this first would put a released app in front of a backend that cannot answer it.

Once #696 merges, this merges immediately after, and 0.15.0 is tagged and released.

PR #84, opened 2026-09-06
feat(chat): mark a completion read only when it is actually viewed

damianvtran:
# PR Description

## Summary of Changes

Marks a conversation read when the user has actually viewed its completed result, so a
result read in the desktop app stops showing as unread on the phone and in the terminal.
Backend companion: damianvtran/local-operator#697.

Acknowledgement is deliberately hard to earn, because every cheap proxy for "viewed" is
wrong: mounting a screen, holding a stream open, or owning a watch lease all happen while
the window is buried behind another application.

- **Renderer eligibility.** The conversation must be the selected one, the document visible
  and focused, and the **end** of the completed result actually painted and unoccluded —
  hit-tested at its bottom edge, because a result whose tail is below the fold has not been
  read. Streaming, loading and history-backfill states are excluded.
- **Native admission in main.** Renderer evidence cannot prove native foreground: an
  occluded, hidden or minimized window still reports `visibilityState === "visible"` and can
  hold document focus. Main re-checks the real `BrowserWindow` (visible, not minimized,
  focused) and does so on the **shared typed request door**, so no other IPC caller can
  route around it. Ordinary control traffic from a background window is unaffected — only
  receipts are gated.
- **Monotonic receipt merging.** Owner epochs restart and frames arrive reordered after a
  reconnect, so applying whichever payload arrived last would resurrect an already-read
  result or hide a newer unread one. Neither clock may run backwards, and identity is
  namespaced: a persistent agent conversation is not the session that shares its id.
- **Durable outcome anchors.** Error and interrupted endings anchor at their durable marker
  position rather than the current tail, so an old failure is not re-presented as the newest
  thing in the conversation.

## Related Issue(s)

- Backend companion: damianvtran/local-operator#697
- Depends on: #83 (desktop control surface) — **held on backend C5 approval**

## Impact

**Breaking changes:** none. `sessions.seen` is a new contract operation; the transcript
gains data attributes and an optional `frontend` prop.

**Dependency and review scope.** Stacked directly on #83's head (`8c60fd2a`). Review only
the adapter commit above that base. **#83's backend prerequisite is held for human C5
approval and must not be assumed merged.**

**Security.** The receipt cannot be issued by a background window, an occluded window, a
non-selected conversation, or any caller other than the owned main frame. It carries a
completion token only — no timestamp a client could use to acknowledge something it never
saw.

## Testing

There is no test runner in this repository, so verification is typecheck, lint, the theme
gates, a real build, and the contract suite driving the **actual bundled TypeScript**.

### Native foreground admission (`scripts/desktop-contract.test.mjs`)

Drives the real `registerDesktopIPC` against a fixture `BrowserWindow`:

| Window state | Receipt |
| --- | --- |
| visible, not minimized, focused | **admitted** |
| unfocused | **refused** |
| not visible | **refused** |
| minimized | **refused** |
| any background state, `settings.list` | **still works** — the gate is receipt-specific |

### Receipt convergence, on the real merge function

Ten cases bundled from the shipped contract: fresh adoption, newer completion, newer
receipt, stale published clock, stale receipt clock, foreign session id, `agent/<id>`
namespace collision, malformed and negative revisions, missing payload. Every stale or
foreign payload is rejected; no clock runs backwards.

### Request shaping

`sessions.seen` is asserted to reach `/v1/desktop/sessions/{id}/seen` with a
`completion_token` body and bearer authorization, while a bodyless call, a `"now"` token, a
traversal session id and an extra `seenAt` field are all rejected before any HTTP happens.

### Gates

```text
pnpm check-types    0 errors (main + renderer)
pnpm lint           7 warnings — pre-existing, verified identical on a stashed clean tree
pnpm check-themes   themes.generated.css up to date; 1884 contrast assertions, 12 themes
node --test scripts/desktop-contract.test.mjs    13/13 pass
pnpm build          succeeds
```

## Non-goals

- **No new visual surface.** This PR adds no unread badge, indicator or styling; it makes
  the existing shared state correct. A visible unread affordance is a separate change.
- **No native notification withdrawal.** Reading does not retract a delivered notification.
- **Pending approvals and questions are never cleared by reading.** A gate is an action, not
  a notification.
- **No optimistic local clear.** A failed or refused receipt leaves authoritative state
  untouched and simply permits a later retry, so the app never claims a read the backend
  did not record.

## Rollout

Self-enabling and additive. Against a backend without `completion-ack-v1` the renderer sees
`supported: false` and never attempts a receipt, so the app behaves exactly as it does
today.

### DIFF
--- a/scripts/desktop-contract.test.mjs
+++ b/scripts/desktop-contract.test.mjs
@@ -188,6 +188,16 @@ test("canonical session operations preserve identity, arguments and main-owned a
 			"POST",
 			{ subscription_id: "a".repeat(32), visible: false, can_notify: true },
 		],
+		[
+			{
+				op: "sessions.seen",
+				sessionId,
+				completionToken: "bbbbbbbb-bbbb-4bbb-8bbb-bbbbbbbbbbbb",
+			},
+			`/${sessionId}/seen`,
+			"POST",
+			{ completion_token: "bbbbbbbb-bbbb-4bbb-8bbb-bbbbbbbbbbbb" },
+		],
 	]) {
 		assert.equal((await requestDesktop(operation, url, token)).status, 200);
 		const actual = seen.at(-1);
@@ -209,6 +219,21 @@ test("canonical session operations preserve identity, arguments and main-owned a
 			text: "hello",
 		},
 		{ op: "sessions.command", sessionId, requestId, command: "goal extra" },
+		// A receipt carries a completion token and nothing else: a timestamp or a
+		// bodyless call is what let a background tab acknowledge an unseen result.
+		{ op: "sessions.seen", sessionId },
+		{ op: "sessions.seen", sessionId, completionToken: "now" },
+		{
+			op: "sessions.seen",
+			sessionId: "../config",
+			completionToken: "bbbbbbbb-bbbb-4bbb-8bbb-bbbbbbbbbbbb",
+		},
+		{
+			op: "sessions.seen",
+			sessionId,
+			completionToken: "bbbbbbbb-bbbb-4bbb-8bbb-bbbbbbbbbbbb",
+			seenAt: 123,
+		},
 		{
 			op: "sessions.watch",
 			sessionId,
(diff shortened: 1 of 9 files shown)


damianvtran:
### UX review — round 1

Reviewer: ux-reviewer (anthropic/claude-opus-5)
Scope: `8c60fd2ac1c2cf3dd462fb4804ccb6f7e3804115..705d01f3f0863562d84e2c51e5ee1e925e04ac09` (UI adapter only), with backend counterpart `3b92560529f17ab0d8f8144d5b61c63b32d8b701..f67c1ac00ca1e55cb9e3aef59bd007b45de58170` (local-operator#697) read as the route it calls.

Reviewing the **interaction contract**, not new pixels — the implementer's claim that this adds no visual surface is accurate (there is no unread affordance anywhere in the renderer; `grep` for `.unseen` outside the two new hooks returns nothing). So the whole feature is invisible plumbing whose only user-perceivable output is *whether a completion is correctly read or unread across TUI / phone / desktop*. That makes the honesty of the state the entire UX.

I did not re-review #687, and I did not audit #696/#83 files.

**What I verified, not assumed.** I re-ran both of the implementer's probes independently in this worktree's own `.venv` (`seen_probe.py`, `sse_probe.py`) — both reproduce exactly as claimed, including `409` on a foreign token, no owner spawned on a cold session, the monotonic `stale A → rev [2,1] / fresh B → rev [2,2] / replay A → rev [2,2]` watermark, and cross-process convergence. `pnpm test:desktop` is 20/20, `pnpm check-types` clean, `pnpm lint` shows 7 warnings that are **identical on the base commit** (pre-existing, none in the diff's files). Backend `tests/unit/server/test_desktop_attention.py` is 6/6.

The positive path is genuinely good, and several things I expected to be wrong are right: the foreground gate lives on the shared IPC door (not the renderer) so no alternate caller routes around it; `elementFromPoint` correctly rejects a *covered* anchor rather than just an offscreen one; `data-completion-complete` is false while an assistant row is still streaming, so a clipped/capped ending stays unread; the error/interrupted anchor is a synthetic `completion-<uuid>`, so the new `notice` row cannot collide with a real message id in `upsert` and silently overwrite transcript content; and `frontend.update` has `attention` stripped so a stale owner delta cannot un-read something another process read. Reads and subscriptions do not mutate the store — confirmed by probe.

My findings are all in one place: **the new receipt polling has no failure handling, and its failures are silent and permanent.** Everything below follows from that.

---

**U1 — BLOCKER — Ordinary write contention permanently kills the receipt poller, and a genuinely unread completion then becomes invisible on the desktop.**

`DesktopSessionBridge._poll_attention` is `while True: await self.refresh_attention(); await asyncio.sleep(1)` with no `try/except`. `AttentionStore` connects with `timeout=2.0` and **no WAL mode**, and every writer takes `BEGIN IMMEDIATE`. So any writer (TUI, relay, another desktop window) holding the write lock longer than 2s raises `OperationalError: database is locked` — which kills the task forever. There is no restart path.

Repro (`/tmp/ux_probe2.py`, `/tmp/ux_probe3.py`, run against this branch):

```
C1. poll alive: True unseen: True
C2. exclusive writer held (simulating a busy TUI/relay writer)...
C3. poll task died during transient lock: True
C4. poll task alive AFTER contention cleared: False
C5. died with: OperationalError('database is locked')
```

The user-visible consequence is the one the design contract most wants to avoid — work that is genuinely unread disappearing:

```
D1. desktop shows read after user viewed A: True
D2. poller dead: True
D3. STORE truth  -> unseen: True  rev: [2, 1]      <- completion B is UNREAD
D4. DESKTOP shows-> unseen: False rev: [1, 1]      <- desktop still says "read"
D5. *** A GENUINELY UNREAD COMPLETION IS INVISIBLE ON THE DESKTOP ***
D6. sidebar list row attention unseen: True (independent read path)
```

Step for a user: read a completion on the desktop; a background writer briefly contends the store; the agent finishes new work while you are away; the desktop stream now reports that session as read forever, while the phone and the sidebar's own list path correctly report it unread. Expected: after contention clears, the next poll recovers and the desktop converges to unread, same as every other interface. Note `mergeCompletionAttention` cannot rescue this — it is monotonic and correctly refuses to regress, but it is never given the newer revision at all because the producer is dead.

This is the one direction the reviewer brief specifically asks about ("could anything make a genuinely unread completion disappear from view"), and the answer is currently yes. Suggested fix: wrap the loop body so a store error is logged and retried on the next tick rather than terminating the task.

**U2 — BLOCKER — A dead poller makes leaving the conversation raise, which aborts `_detach()` and leaks the owner session.**

`_detach()` puts the attention block **first** and suppresses only `CancelledError`:

```python
if self.attention_task is not None:
    self.attention_task.cancel()
    with contextlib.suppress(asyncio.CancelledError):
        await self.attention_task     # re-raises OperationalError if it already died
```

Awaiting an already-failed task re-raises its exception, so the remaining teardown — unsubscribing frontend callbacks and `await remote.dispose()` — never runs. Repro (`/tmp/ux_probe7.py`, `/tmp/ux_probe8.py`):

```
I1. poller died: True
I2. LEAVING THE SESSION RAISES OperationalError: database is locked
I3. reopening the session works; live poller: False

J2. users counter: 0 (0 = believed fully detached)
J3. attention_task cleared: False (False = stuck)
J4. LEAKED owner RemoteSession (dispose skipped): True
J5. LEAKED frontend subscriptions: 2
```

Step for a user: after the contention in U1, close or navigate away from that conversation. The bridge believes it detached (`users == 0`) but the owner runtime is never disposed and two subscriptions leak; reopening the session does not restore a live poller, so receipt sync stays dead for the rest of the app's life. A read-receipt concern must never be able to strand a session owner. Cancel-then-await should suppress the task's own failure too (e.g. `suppress(BaseException)` around the await, or check `task.done()` first), and the attention block should not sit ahead of `dispose()`.

**U3 — MAJOR — A receipt concern can break the two primary read paths: opening a conversation and loading the sidebar both 500 under contention.**

`snapshot()` now begins with an unguarded `await self.refresh_attention()`, and `DesktopSessions.list()` now does an unguarded `state_many(...)`. Neither existed on the base (`git show 3b9256052` → zero `refresh_attention`, zero `attention` in `rows()`). So a store that is merely *busy* takes down functionality that has nothing to do with read receipts:

```
E1. OPENING THE CONVERSATION FAILS under contention: OperationalError: database is locked

G1. GET sessions (sidebar) under contention -> 500
G2. POST seen under contention -> 500 | Internal Server Error
G3. GET sessions after contention clears -> 200
```

`OperationalError` is not in the `errors()` ladder in `desktop_sessions.py`, so it escapes as a bare **500 "Internal Server Error"** rather than one of the vetted sentences. Step for a user: open the app while another interface happens to be writing a receipt — the conversation list fails to load and the transcript fails to open, with a developer-shaped error. Expected: receipt state is decorative on these paths, so a store failure should degrade to "no attention info" and let the snapshot and the list render normally. This is also a straightforward availability regression on paths that previously never touched `attention.db`.

**U4 — MAJOR — The foreground refusal is worded for a user but can never reach one, and it retries silently forever.**

`desktop-ipc.ts` throws a genuinely user-facing sentence — "View this completion in the foreground before marking it read." — but `use-completion-view.ts` swallows every rejection in an empty `.catch()`, and nothing in the renderer renders `attention` at all. So the string is unreachable copy.

That is defensible on its own (this is background sync; a toast saying "couldn't mark as read" would be noise). The real issue is the pairing: `setInterval(check, 500)` never backs off and never stops, so a persistently failing ACK — a backend that predates `completion-ack-v1`, a wedged store, a window state the gate keeps refusing — retries an IPC round trip **twice a second for as long as the conversation is open**, with zero diagnostics anywhere. Note `attention.supported !== true` guards the negotiated case, but not a store or gate that keeps failing.

Step: leave a session open with an unseen completion and a failing ACK; the app quietly issues ~7,200 failed IPC calls per hour and the completion stays unread with no explanation. Expected: back off after repeated failures (or stop after N and resume on a state change), and log once so this is diagnosable. Either surface the sentence or drop it and say in a comment that the throw is a guard, not copy.

**U5 — MINOR — Up to ~1s of cross-interface lag is fine; the fixed 1s poll for every open session is the part worth revisiting.**

Measured convergence when another interface acknowledges: `A. cross-interface convergence latency: 1.001s`. As a *feel* that is good — you look back at the desktop and it has already caught up, no refresh needed, which is exactly the reverse-direction behaviour the feature promises. Flagging only that it is an unconditional 1 Hz SQLite read per attached bridge that continues while the window is hidden or the session idle, which is the same "a lease is not attention" reasoning the contract applies elsewhere. Consider polling lazily (or slower) when the owning window is not foreground.

**U6 — NIT — "Stopped with an error" / "Interrupted" read as states, not outcomes.**

The new durable notice rows are the only strings this diff shows a user. They are consistent with the neighbouring `"Context compacted"` and correctly carry `level: error|warning`, so this is purely a wording nit: the surrounding transcript vocabulary is event-shaped, and these two read like status labels. No action needed if they match the TUI's wording for the same outcomes — worth a glance for cross-interface consistency, since the whole point of the feature is that the three surfaces agree.

---

**Not simulated.** Everything above is from the real backend on this branch (worktree `.venv`, isolated config dir, `CMUX_*` scrubbed) plus the shipped `desktop-contract.test.mjs` main-process harness. One cell I did **not** fake: the native-foreground admission on a genuinely focused, uncovered, non-minimized Electron window. The IPC gate is proven at unit level against a `BrowserWindow` fixture (`isVisible`/`isMinimized`/`isFocused`, each negative refusing individually), and the renderer-side occlusion check via `elementFromPoint` is proven by the same suite — but "a real user really looked at it in a real window" needs a live foreground desktop app, which I cannot obtain here.

**What I would need from the parent to close that cell:** a running packaged/dev `local-operator-ui` against a backend on this branch, with a session holding one unseen completion, and someone to perform four interactions in order — (1) bring the window to the front with the completion's anchor scrolled into view (expect: becomes read within ~1s), (2) scroll the anchor off screen and publish a new completion (expect: stays unread), (3) cover the window with another app while leaving it "visible" (expect: stays unread), (4) minimize, then acknowledge from the phone (expect: desktop converges to read on its own, no refresh). I would mark that cell BLOCKED rather than claim it.

**Verdict: NOT TERMINAL.** Two BLOCKERs (U1, U2) and two MAJORs (U3, U4) remain. The core semantics of the adapter — what counts as a view, what a lease is not, monotonic receipts, capped endings staying unread, no admission on a cold session — are correct and I found no way to acknowledge something the user did not actually see. The failures are all in the new polling layer's lack of error handling, where a routine `database is locked` currently escalates into permanently wrong read state (U1), a leaked owner session (U2), and 500s on unrelated primary paths (U3). Worth noting `test_desktop_attention.py` has no coverage for a store error at all, so a regression test belongs with the fix.

damianvtran:
Superseded by a PR on `dev-viewed-completions`. This repo's CI has no `pull_request` trigger — it runs only on pushes to `main` or `dev-*`, so this branch name could never receive checks (which is why it showed 0). Same commits, renamed branch.

PR #85, opened 2026-09-06
feat(chat): mark a completion read only when it is actually viewed

damianvtran:
# PR Description

## Summary of Changes

Marks a conversation read when the user has actually viewed its completed result, so a
result read in the desktop app stops showing as unread on the phone and in the terminal.
Backend companion: damianvtran/local-operator#697.

Acknowledgement is deliberately hard to earn, because every cheap proxy for "viewed" is
wrong: mounting a screen, holding a stream open, or owning a watch lease all happen while
the window is buried behind another application.

- **Renderer eligibility.** The conversation must be the selected one, the document visible
  and focused, and the **end** of the completed result actually painted and unoccluded —
  hit-tested at its bottom edge, because a result whose tail is below the fold has not been
  read. Streaming, loading and history-backfill states are excluded.
- **Native admission in main.** Renderer evidence cannot prove native foreground: an
  occluded, hidden or minimized window still reports `visibilityState === "visible"` and can
  hold document focus. Main re-checks the real `BrowserWindow` (visible, not minimized,
  focused) and does so on the **shared typed request door**, so no other IPC caller can
  route around it. Ordinary control traffic from a background window is unaffected — only
  receipts are gated.
- **Monotonic receipt merging.** Owner epochs restart and frames arrive reordered after a
  reconnect, so applying whichever payload arrived last would resurrect an already-read
  result or hide a newer unread one. Neither clock may run backwards, and identity is
  namespaced: a persistent agent conversation is not the session that shares its id.
- **Durable outcome anchors.** Error and interrupted endings anchor at their durable marker
  position rather than the current tail, so an old failure is not re-presented as the newest
  thing in the conversation.

## Related Issue(s)

- Backend companion: damianvtran/local-operator#697
- Depends on: #83 (desktop control surface) — **held on backend C5 approval**

## Impact

**Breaking changes:** none. `sessions.seen` is a new contract operation; the transcript
gains data attributes and an optional `frontend` prop.

**Dependency and review scope.** Stacked directly on #83's head (`8c60fd2a`). Review only
the adapter commit above that base. **#83's backend prerequisite is held for human C5
approval and must not be assumed merged.**

**Security.** The receipt cannot be issued by a background window, an occluded window, a
non-selected conversation, or any caller other than the owned main frame. It carries a
completion token only — no timestamp a client could use to acknowledge something it never
saw.

## Testing

There is no test runner in this repository, so verification is typecheck, lint, the theme
gates, a real build, and the contract suite driving the **actual bundled TypeScript**.

### Native foreground admission (`scripts/desktop-contract.test.mjs`)

Drives the real `registerDesktopIPC` against a fixture `BrowserWindow`:

| Window state | Receipt |
| --- | --- |
| visible, not minimized, focused | **admitted** |
| unfocused | **refused** |
| not visible | **refused** |
| minimized | **refused** |
| any background state, `settings.list` | **still works** — the gate is receipt-specific |

### Receipt convergence, on the real merge function

Ten cases bundled from the shipped contract: fresh adoption, newer completion, newer
receipt, stale published clock, stale receipt clock, foreign session id, `agent/<id>`
namespace collision, malformed and negative revisions, missing payload. Every stale or
foreign payload is rejected; no clock runs backwards.

### Request shaping

`sessions.seen` is asserted to reach `/v1/desktop/sessions/{id}/seen` with a
`completion_token` body and bearer authorization, while a bodyless call, a `"now"` token, a
traversal session id and an extra `seenAt` field are all rejected before any HTTP happens.

### Gates

```text
pnpm check-types    0 errors (main + renderer)
pnpm lint           7 warnings — pre-existing, verified identical on a stashed clean tree
pnpm check-themes   themes.generated.css up to date; 1884 contrast assertions, 12 themes
node --test scripts/desktop-contract.test.mjs    13/13 pass
pnpm build          succeeds
```

## Non-goals

- **No new visual surface.** This PR adds no unread badge, indicator or styling; it makes
  the existing shared state correct. A visible unread affordance is a separate change.
- **No native notification withdrawal.** Reading does not retract a delivered notification.
- **Pending approvals and questions are never cleared by reading.** A gate is an action, not
  a notification.
- **No optimistic local clear.** A failed or refused receipt leaves authoritative state
  untouched and simply permits a later retry, so the app never claims a read the backend
  did not record.

## Rollout

Self-enabling and additive. Against a backend without `completion-ack-v1` the renderer sees
`supported: false` and never attempts a receipt, so the app behaves exactly as it does
today.

### DIFF
--- a/scripts/check-edit-diffs.mjs
+++ b/scripts/check-edit-diffs.mjs
@@ -18,46 +18,80 @@ const source = new URL(
 );
 const filename = process.argv[2] || fileURLToPath(source);
 const compiled = ts.transpileModule(readFileSync(filename, "utf8"), {
-	compilerOptions: { module: ts.ModuleKind.CommonJS, target: ts.ScriptTarget.ES2022 },
+	compilerOptions: {
+		module: ts.ModuleKind.CommonJS,
+		target: ts.ScriptTarget.ES2022,
+	},
 }).outputText;
 const module = { exports: {} };
 // Execute exactly the checked-in implementation, not a reimplementation of its
 // ranges. createRequire resolves dependencies from this repository even when a
 // before-fix source snapshot is passed as an explicit argument.
 new Function("require", "module", "exports", compiled)(
-	createRequire(source), module, module.exports,
+	createRequire(source),
+	module,
+	module.exports,
 );
 const { diffHighlight } = module.exports;
 
 for (const document of ["", "Existing trailing text"]) {
-	for (const replacement of ["New line", "First line\nSecond line\nThird line"]) {
+	for (const replacement of [
+		"New line",
+		"First line\nSecond line\nThird line",
+	]) {
 		const diff = { find: "", replace: replacement };
 		const plugin = diffHighlight({
-			diffs: [diff], currentIndex: 0, approvedDiffs: [], originalContent: document,
+			diffs: [diff],
+			currentIndex: 0,
+			approvedDiffs: [],
+			originalContent: document,
 		});
 		const state = EditorState.create({ doc: document });
 		const instance = plugin.create({ state });
 		const cursor = instance.decorations.iter();
 		assert.equal(cursor.from, 0);
 		assert.equal(cursor.to, 0);
-		assert.ok(cursor.value.spec.widget, "insertion must remain inspectable as a widget");
+		assert.ok(
+			cursor.value.spec.widget,
+			"insertion must remain inspectable as a widget",
+		);
 		assert.equal(cursor.value.spec.widget.newText, replacement);
 		assert.equal(cursor.value.spec.widget.oldText, "");
-		assert.equal(state.doc.toString(), document, "preview cannot modify the live buffer");
+		assert.equal(
+			state.doc.toString(),
+			document,
+			"preview cannot modify the live buffer",
+		);
 	}
 }
 
 const normal = diffHighlight({
-	diffs: [{ find: "old", replace: "new" }], currentIndex: 0,
-	approvedDiffs: [], originalContent: "old",
+	diffs: [{ find: "old", replace: "new" }],
+	currentIndex: 0,
+	approvedDiffs: [],
+	originalContent: "old",
 }).create({ state: EditorState.create({ doc: "old" }) });
-assert.equal(normal.decorations.size, 1, "nonempty inline replacement must stay unchanged");
+assert.equal(
+	normal.decorations.size,
+	1,
+	"nonempty inline replacement must stay unchanged",
+);
 assert.equal(normal.decorations.iter().to, 3);
 const multiline = diffHighlight({
-	diffs: [{ find: "old\nline", replace: "new\nline" }], currentIndex: 0,
-	approvedDiffs: [], originalContent: "old\nline",
+	diffs: [{ find: "old\nline", replace: "new\nline" }],
+	currentIndex: 0,
+	approvedDiffs: [],
+	originalContent: "old\nline",
 }).create({ state: EditorState.create({ doc: "old\nline" }) });
-assert.equal(multiline.decorations.size, 2, "nonempty multiline mark + widget must stay unchanged");
-const dismissed = diffHighlight(null).create({ state: EditorState.create({ doc: "" }) });
+assert.equal(
+	multiline.decorations.size,
+	2,
+	"nonempty multiline mark + widget must stay unchanged",
+);
+const dismissed = diffHighlight(null).create({
+	state: EditorState.create({ doc: "" }),
+});
 assert.equal(dismissed.decorations.size, 0);
-console.log("Edit diff regressions: 4 insertion previews + unchanged nonempty/dismissed states passed");
+console.log(
+	"Edit diff regressions: 4 insertion previews + unchanged nonempty/dismissed states passed",
+);
--- a/scripts/desktop-contract.test.mjs
+++ b/scripts/desktop-contract.test.mjs
@@ -188,6 +188,16 @@ test("canonical session operations preserve identity, arguments and main-owned a
 			"POST",
 			{ subscription_id: "a".repeat(32), visible: false, can_notify: true },
 		],
+		[
+			{
+				op: "sessions.seen",
+				sessionId,
+				completionToken: "bbbbbbbb-bbbb-4bbb-8bbb-bbbbbbbbbbbb",
+			},
+			`/${sessionId}/seen`,
+			"POST",
+			{ completion_token: "bbbbbbbb-bbbb-4bbb-8bbb-bbbbbbbbbbbb" },
+		],
 	]) {
 		assert.equal((await requestDesktop(operation, url, token)).status, 200);
 		const actual = seen.at(-1);
@@ -209,6 +219,21 @@ test("canonical session operations preserve identity, arguments and main-owned a
 			text: "hello",
 		},
 		{ op: "sessions.command", sessionId, requestId, command: "goal extra" },
+		// A receipt carries a completion token and nothing else: a timestamp or a
+		// bodyless call is what let a background tab acknowledge an unseen result.
+		{ op: "sessions.seen", sessionId },
+		{ op: "sessions.seen", sessionId, completionToken: "now" },
+		{
+			op: "sessions.seen",
+			sessionId: "../config",
+			completionToken: "bbbbbbbb-bbbb-4bbb-8bbb-bbbbbbbbbbbb",
+		},
+		{
+			op: "sessions.seen",
+			sessionId,
+			completionToken: "bbbbbbbb-bbbb-4bbb-8bbb-bbbbbbbbbbbb",
+			seenAt: 123,
+		},
 		{
 			op: "sessions.watch",
 			sessionId,
@@ -724,4 +749,135 @@ test("IPC rejects other frames and opens only backend-returned authorization onc
 	await assert.rejects(() => open(event, "https://evil.example"));
 	frame.url = "https://evil.example";
 	assert.throws(() => invoke(event, { op: "settings.list" }));
+});
+
+test("a read receipt is admitted only by an actually foreground native window", async () => {
+	// Renderer visibility cannot prove native foreground: an occluded, hidden or
+	// minimized window still reports `visibilityState === "visible"` and can
+	// still hold document focus. Main owns the real BrowserWindow, so this is
+	// the only place the claim can be checked -- and it is enforced on the
+	// shared typed request door so no alternate caller can route around it.
+	globalThis.__desktopHandlers = new Map();
+	const frame = { url: "file:///app/index.html" };
+	const contents = { mainFrame: frame };
+	const state = {
+		destroyed: false,
+		visible: true,
+		minimized: false,
+		focused: true,
+	};
+	const owner = {
+		webContents: contents,
+		isDestroyed: () => state.destroyed,
+		isVisible: () => state.visible,
+		isMinimized: () => state.minimized,
+		isFocused: () => state.focused,
+	};
+	const calls = [];
+	registerDesktopIPC(
+		() => owner,
+		"file:///app/index.html",
+		async (request) => {
+			calls.push(request);
+			return { status: 200, body: { result: { unseen: false } } };
+		},
+	);
+	const invoke = globalThis.__desktopHandlers.get("desktop-request");
+	const event = { sender: contents, senderFrame: frame };
+	const receipt = {
+		op: "sessions.seen",
+		sessionId: "123456abcdef",
+		completionToken: "bbbbbbbb-bbbb-4bbb-8bbb-bbbbbbbbbbbb",
+	};
+
+	await invoke(event, receipt);
+	assert.equal(calls.length, 1, "a focused, visible window may acknowledge");
+
+	// Every individual negative must refuse on its own: a background window is
+	// unfocused, an occluded/hidden one is not visible, and a minimized one can
+	// report both while showing the user nothing.
+	for (const background of [
+		{ focused: false },
+		{ visible: false },
+		{ minimized: true },
+	]) {
+		Object.assign(state, { visible: true, minimized: false, focused: true });
+		Object.assign(state, background);
+		assert.throws(
+			() => invoke(event, receipt),
+			/foreground/,
+			JSON.stringify(background),
+		);
+	}
+	assert.equal(calls.length, 1, "no background state reached the backend");
+
+	// The gate is specific to receipts. Ordinary control traffic from a
+	// background window is legitimate and must not be broken by it.
+	Object.assign(state, { visible: false, minimized: true, focused: false });
+	await invoke(event, { op: "settings.list" });
+	assert.equal(calls.at(-1).op, "settings.list");
+
+	Object.assign(state, { visible: true, minimized: false, focused: true });
+	await invoke(event, receipt);
+	assert.equal(
+		calls.length,
+		3,
+		"the receipt is admitted again once foreground",
+	);
+});
+
+test("receipt revisions converge monotonically across reconnects and reordering", async () => {
+	// Owner epochs restart; the receipt clock does not. Frames arrive reordered
+	// after a reconnect, and a snapshot captured before an acknowledgement can
+	// land after it -- so applying whatever arrived last would resurrect an
+	// already-read result or, worse, hide a newer unread one.
+	const contract = await build({
+		stdin: {
+			contents: 'export * from "./src/shared/desktop-session-contract";',
+			resolveDir: process.cwd(),
+		},
+		bundle: true,
+		format: "esm",
+		platform: "node",
+		write: false,
+	});
+	const { mergeCompletionAttention } = await import(
+		`data:text/javascript;base64,${Buffer.from(contract.outputFiles[0].text).toString("base64")}`
+	);
+	const sessionId = "123456abcdef";
+	const at = (published, acknowledged, extra = {}) => ({
+		conversation_id: `session/${sessionId}`,
+		completion_token: "bbbbbbbb-bbbb-4bbb-8bbb-bbbbbbbbbbbb",
+		anchor_id: "result-1",
+		kind: "complete",
+		unseen: published > acknowledged,
+		revision: [published, acknowledged],
+		supported: true,
+		...extra,
+	});
+	const merge = (current, incoming) =>
+		mergeCompletionAttention(current, incoming, sessionId)?.revision;
+
+	assert.deepEqual(merge(undefined, at(2, 1)), [2, 1]);
+	assert.deepEqual(merge(at(1, 1), at(2, 1)), [2, 1]);
+	assert.deepEqual(merge(at(2, 1), at(2, 2)), [2, 2]);
+	// Neither clock may run backwards, whichever direction the staleness is in.
+	assert.deepEqual(merge(at(2, 2), at(1, 1)), [2, 2]);
+	assert.deepEqual(merge(at(2, 2), at(2, 1)), [2, 2]);
+	// Identity is namespaced: a persistent agent conversation and an unrelated
+	// session are different authorities even when the trailing id matches.
+	for (const foreign of [
+		{ conversation_id: "session/ffffffffffff" },
+		{ conversation_id: `agent/${sessionId}` },
+		{ revision: ["x", 1] },
+		{ revision: [-1, 0] },
+		{ revision: [3] },
+	]) {
+		assert.deepEqual(
+			merge(at(2, 2), at(9, 9, foreign)),
+			[2, 2],
+			JSON.stringify(foreign),
+		);
+	}
+	assert.deepEqual(merge(at(2, 2), undefined), [2, 2]);
 });
--- a/scripts/desktop-contract.test.mjs
+++ b/scripts/desktop-contract.test.mjs
@@ -34,10 +34,14 @@ const bundle = await build({
 		},
 	],
 });
-const { requestDesktop, trustedDesktopFrame, registerDesktopIPC } =
-	await import(
-		`data:text/javascript;base64,${Buffer.from(bundle.outputFiles[0].text).toString("base64")}`
-	);
+const {
+	requestDesktop,
+	trustedDesktopFrame,
+	registerDesktopIPC,
+	guardForegroundReceipts,
+} = await import(
+	`data:text/javascript;base64,${Buffer.from(bundle.outputFiles[0].text).toString("base64")}`
+);
 let server;
 let url;
 const seen = [];
@@ -803,7 +807,7 @@ test("a read receipt is admitted only by an actually foreground native window", 
 	]) {
 		Object.assign(state, { visible: true, minimized: false, focused: true });
 		Object.assign(state, background);
-		assert.throws(
+		await assert.rejects(
 			() => invoke(event, receipt),
 			/foreground/,
 			JSON.stringify(background),
@@ -880,4 +884,57 @@ test("receipt revisions converge monotonically across reconnects and reordering"
 		);
 	}
 	assert.deepEqual(merge(at(2, 2), undefined), [2, 2]);
+});
+
+test("the receipt gate rides the sender, so a non-IPC main caller cannot bypass it", async () => {
+	// `DesktopNotifier` holds its own reference to the same underlying sender
+	// and calls it directly, so a gate living only inside the `desktop-request`
+	// IPC handler would not cover it. Not exploitable while the notifier emits
+	// only `sessions.watch` — but it is precisely how a future main-process
+	// caller would acquire an ungated `sessions.seen`.
+	const state = { visible: false, minimized: true, focused: false };
+	const owner = {
+		isDestroyed: () => false,
+		isVisible: () => state.visible,
+		isMinimized: () => state.minimized,
+		isFocused: () => state.focused,
+	};
+	const calls = [];
+	const guarded = guardForegroundReceipts(
+		() => owner,
+		async (input) => {
+			calls.push(input);
+			return { status: 200, body: { result: {} } };
+		},
+	);
+	const receipt = {
+		op: "sessions.seen",
+		sessionId: "123456abcdef",
+		completionToken: "bbbbbbbb-bbbb-4bbb-8bbb-bbbbbbbbbbbb",
+	};
+
+	// A background window is refused even though this never touches ipcMain.
+	await assert.rejects(() => guarded(receipt), /foreground/);
+	assert.equal(calls.length, 0);
+
+	// The notifier's own traffic is unaffected: a lease is not a read, and it
+	// legitimately reports presence from a background window.
+	await guarded({
+		op: "sessions.watch",
+		sessionId: "123456abcdef",
+		subscriptionId: "a".repeat(32),
+		visible: false,
+		canNotify: true,
+	});
+	assert.equal(calls.length, 1);
+
+	Object.assign(state, { visible: true, minimized: false, focused: true });
+	await guarded(receipt);
+	assert.equal(calls.length, 2);
+
+	// Double application (sender-wrapped AND registered through the IPC entry)
+	// must stay idempotent rather than double-refusing a legitimate receipt.
+	const twice = guardForegroundReceipts(() => owner, guarded);
+	await twice(receipt);
+	assert.equal(calls.length, 3);
 });
--- a/scripts/test-publish-workflow.mjs
+++ b/scripts/test-publish-workflow.mjs
@@ -5,139 +5,296 @@
  * the YAML and CLI parsers; no second parser or dependency install is needed.
  */
 import assert from "node:assert/strict";
-import { test } from "node:test";
-import { createRequire } from "node:module";
-import { readFileSync, writeFileSync, mkdtempSync, mkdirSync, rmSync, existsSync } from "node:fs";
+import { spawnSync } from "node:child_process";
 import { createHash } from "node:crypto";
+import {
+	existsSync,
+	mkdirSync,
+	mkdtempSync,
+	readFileSync,
+	rmSync,
+	writeFileSync,
+} from "node:fs";
+import { createRequire } from "node:module";
 import { tmpdir } from "node:os";
 import { join } from "node:path";
-import { spawnSync } from "node:child_process";
+import { test } from "node:test";
 
 const require = createRequire(import.meta.url);
-const builderRequire = createRequire(require.resolve("electron-builder/package.json"));
+const builderRequire = createRequire(
+	require.resolve("electron-builder/package.json"),
+);
 const appRequire = createRequire(builderRequire.resolve("app-builder-lib"));
 const { load } = appRequire("js-yaml");
 const builder = require("electron-builder/out/builder");
-const workflow = load(readFileSync(new URL("../.github/workflows/publish.yml", import.meta.url), "utf8"));
-const pkg = JSON.parse(readFileSync(new URL("../package.json", import.meta.url), "utf8"));
+const workflow = load(
+	readFileSync(
+		new URL("../.github/workflows/publish.yml", import.meta.url),
+		"utf8",
+	),
+);
+const pkg = JSON.parse(
+	readFileSync(new URL("../package.json", import.meta.url), "utf8"),
+);
 const jobs = workflow.jobs;
 const steps = (job) => jobs[job].steps;
 const step = (job, name) => steps(job).find((s) => s.name === name);
 const needsOf = (job) => [].concat(jobs[job].needs || []);
 function condition(expression, context) {
-  if (!expression) return true;
-  const code = expression.replace(/^\$\{\{\s*|\s*\}\}$/g, "").replace(/needs\.([\w-]+)\.result/g, 'needs["$1"].result');
-  // Expressions come only from the checked-in workflow, not user input. Status
-  // functions are deliberately unsupported: they can bypass GitHub's skip gate.
-  return Boolean(Function(...Object.keys(context), `return (${code})`)(...Object.values(context)));
+	if (!expression) return true;
+	const code = expression
+		.replace(/^\$\{\{\s*|\s*\}\}$/g, "")
+		.replace(/needs\.([\w-]+)\.result/g, 'needs["$1"].result');
+	// Expressions come only from the checked-in workflow, not user input. Status
+	// functions are deliberately unsupported: they can bypass GitHub's skip gate.
+	return Boolean(
+		Function(
+			...Object.keys(context),
+			`return (${code})`,
+		)(...Object.values(context)),
+	);
 }
 function graph(event, prerelease = false, forced = {}) {
-  const result = {};
-  for (const job of Object.keys(jobs)) {
-    const needs = Object.fromEntries(needsOf(job).map((n) => [n, { result: result[n] }]));
-    const success = Object.values(needs).every((n) => n.result === "success");
-    const eligible = success && condition(jobs[job].if, { needs, github: { event_name: event, event: { release: { prerelease } } } });
-    result[job] = eligible ? forced[job] || "success" : "skipped";
-  }
-  return result;
+	const result = {};
+	for (const job of Object.keys(jobs)) {
+		const needs = Object.fromEntries(
+			needsOf(job).map((n) => [n, { result: result[n] }]),
+		);
+		const success = Object.values(needs).every((n) => n.result === "success");
+		const eligible =
+			success &&
+			condition(jobs[job].if, {
+				needs,
+				github: { event_name: event, event: { release: { prerelease } } },
+			});
+		result[job] = eligible ? forced[job] || "success" : "skipped";
+	}
+	return result;
 }
 function sandbox(fn) {
-  const dir = mkdtempSync(join(tmpdir(), "publish-contract-"));
-  try { return fn(dir); } finally { rmSync(dir, { recursive: true, force: true }); }
+	const dir = mkdtempSync(join(tmpdir(), "publish-contract-"));
+	try {
+		return fn(dir);
+	} finally {
+		rmSync(dir, { recursive: true, force: true });
+	}
 }
-const cleanEnv = { PATH: process.env.PATH, HOME: process.env.HOME, NODE_PATH: process.env.NODE_PATH || "" };
+const cleanEnv = {
+	PATH: process.env.PATH,
+	HOME: process.env.HOME,
+	NODE_PATH: process.env.NODE_PATH || "",
+};
 function shell(script, env, cwd) {
-  return spawnSync("bash", ["--noprofile", "--norc", "-e", "-o", "pipefail", "-c", script], { encoding: "utf8", cwd, env: { ...cleanEnv, ...env } });
+	return spawnSync(
+		"bash",
+		["--noprofile", "--norc", "-e", "-o", "pipefail", "-c", script],
+		{ encoding: "utf8", cwd, env: { ...cleanEnv, ...env } },
+	);
 }
 
-for (const [event, prerelease] of [["workflow_dispatch", false], ["release", false], ["release", true]]) {
-  test(`${event} prerelease=${prerelease}: every required job reaches upload`, () => {
-    const result = graph(event, prerelease);
-    assert.ok(Object.values(result).every((r) => r === "success"), JSON.stringify(result));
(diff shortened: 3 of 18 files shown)


damianvtran:
### UX review remediation — round 1

Author: coder implementing agent (`anthropic/claude-opus-5`). The round-1 implementation ran on `openai/gpt-6-astra`; that session died on a provider rate limit and this remediation was completed on Opus, disclosed per the review gate.

Head: `9191f9bc7fb1a2e25ff1151735118dee5c1492ed` on `dev-viewed-completions`. Backend companion: damianvtran/local-operator#697 (`ecb99398`), where F1/F2/F6–F9 and U1–U3/U5 are answered.

Answering the `[ui]` findings from the [code review](https://github.com/damianvtran/local-operator/pull/697#issuecomment-5557439696) and the renderer findings from the [UX review](https://github.com/damianvtran/local-operator-ui/pull/84#issuecomment-5557428963).

**This PR replaces #84.** Same commits; see "CI" below for why the branch was renamed.

| ID | Disposition |
| --- | --- |
| F3 | **Fixed** in `9191f9bc` |
| F4 | **Fixed** in `9191f9bc` — stronger than the finding's minimum |
| F5 | **Fixed** in `9191f9bc` |
| F10 | **Fixed** in `9191f9bc` |
| F11 | **Fixed** in `9191f9bc` |
| U4 | **Fixed** in `9191f9bc` |
| U6 | **Rejected** — wording matches the TUI verbatim; see below |

#### F3 — MAJOR — a crash-recovered outcome was unrenderable, so permanently unclearable

Fixed, and this was the right call to keep in this PR rather than push into core: the desktop adapter is the layer that decides what is renderable, and therefore what is ackable.

`withRecoveredOutcome` synthesizes the notice at `attention.anchor_id` when the kind is `error`/`interrupted` and no row already carries that anchor, mirroring `tui/app.py::_poll_completion_attention` **including** its retry guard — a historical failure is not inserted at the tail of a retry that is already running.

Verified against the real bundled reducer with a genuine `completion-<uuid>` anchor that has no transcript entry:

```text
anchor painted before:            false
synthesized row:                  { kind: 'notice', text: 'Interrupted', level: 'warning', complete: true }
ackable (complete === true):      true
while streaming (retry guard):    false
idempotent on second apply:       1 row
already-read not synthesized:     false
normal 'complete' kind untouched: false
```

#### F4 — MINOR — the comment overclaimed where the guard sits

Fixed by making the claim true rather than by weakening the comment. I confirmed the bypass first: `src/main/index.ts` passed `backendService.requestDesktop` to `DesktopNotifier` and to `registerDesktopIPC` **separately**, so a guard inside the `desktop-request` handler genuinely did not cover the notifier's path.

`guardForegroundReceipts` now wraps the sender itself, and both callers receive the wrapped one. The IPC entry keeps its own application so the gate does not depend on the caller having wrapped it; double application is idempotent, which is pinned by a test. New coverage proves a non-IPC main caller is refused, that `sessions.watch` from a background window still works (a lease is not a read), and that the double-wrap does not double-refuse.

One behavioural note for the reviewer: the guard now **rejects** rather than throwing synchronously, so the existing admission test moved from `assert.throws` to `assert.rejects`. The refusal reaches only a log, so the comment now says that instead of implying user-facing copy.

#### U4 — MAJOR — silent infinite retry

Fixed. The loop backs off after 3 consecutive failures (exponential, capped at ~1/minute instead of ~7,200/hour), logs **once** at the threshold with the error, and stops the interval entirely on success (which also closes F10). No optimistic clear on failure — authoritative state is still the authority.

#### F5 — MINOR — an ACK triggered a history refetch that could contain nothing new

Fixed: reconcile only when the incoming attention names an anchor the transcript has not already painted. The painted-id set is tracked in a ref because this decision is made outside the React updater by design (the existing comment explains why), and it resets on session switch so a previous conversation's anchors cannot suppress the first reconcile of the next one.

#### F11 — NIT

Fixed: the attribute is omitted rather than emitted as `"false"` on every row.

#### U6 — NIT — **rejected**

"Stopped with an error" / "Interrupted" are the exact strings the TUI uses for the same two outcomes (`tui/app.py:11517-11518`). The finding itself says "no action needed if they match the TUI's wording", and cross-interface consistency is the point of the feature — changing them here would make the surfaces disagree.

#### Gates on this head

```text
pnpm check-types    0 errors
pnpm lint           7 warnings — pre-existing, verified identical on a stashed clean tree
pnpm check-themes   1884 contrast assertions, 12 themes
node --test scripts/desktop-contract.test.mjs    14/14 pass (was 13)
pnpm build          succeeds
```

#### CI — why #84 showed 0 checks

Not a fluke, and worth recording: this repo's `ci.yml` has **no `pull_request` trigger**. It runs on `push` to `main` or `dev-*` only, so `fix/desktop-viewed-completions` could never receive checks — #83 gets them solely because its branch is `dev-backend-parity`. Renaming to `dev-viewed-completions` produced a run immediately: **Pre-Release CI Testing — success**.

This means any future branch here that is not `dev-*` silently ships without CI. Flagged rather than fixed: adding a `pull_request` trigger is a workflow change outside this PR's scope.

No merge or release performed. Still stacked on #83, whose backend prerequisite is held for C5 approval.

damianvtran:
### UX review — round 2

Reviewer: ux-reviewer (anthropic/claude-opus-5)
Scope: `705d01f3f0863562d84e2c51e5ee1e925e04ac09..9191f9bc7fb1a2e25ff1151735118dee5c1492ed` (UI remediation diff), with backend counterpart `f67c1ac00ca1e55cb9e3aef59bd007b45de58170..ecb993986b86aae7d9d04b799a20ab304be5028d` (local-operator#697) read as the route it calls.

Remediation round: I audited only the changed interaction flows and re-verified my round-1 findings. Previously approved unchanged files were not re-audited, and I did not go looking for new issues outside the remediation delta. U6 stays closed on the manager's accepted reasoning (strings verbatim identical to the TUI's — changing one surface would make them disagree, which is the opposite of this feature's point); I did not reopen it.

**Round-1 findings: all four resolved, verified from the user's side rather than by reading the try/except.** The round-1 fixture I care most about is that the fix is not merely "stops throwing". Note the shipped regression tests use a dropped-table proxy for contention (`_break_store`), which is honest and documented — but it is not the failure I reported, so I re-ran the *real* one: a competing writer holding the lock past the store's 2.0 s timeout.

**U1 — RESOLVED.** Real `BEGIN IMMEDIATE` writer held 4 s against the live bridge; then new work completes while the user is away:

```
A2. desktop shows A read after viewing: True
A3. real exclusive writer held (past the 2.0s store timeout)
A4. poll task ALIVE during contention: True
A6. STORE truth   -> unseen: True rev: [2, 1]
A7. DESKTOP shows -> unseen: True rev: [2, 1]
A9. CONVERGED to genuinely-unread with no remount: True
```

The desktop converges to the true unread state on its own — no remount, no user action. That is the exact direction that was broken (unread work invisible on desktop while the phone showed it correctly), and it is closed.

**U2 — RESOLVED, including the part that mattered.** "Detach didn't raise" would not have been enough; the round-1 damage was that re-entering never restored a live poller. Leaving with an already-dead poller, then coming back:

```
B3. owner session disposed (not stranded): True
B5. frontend subscriptions leaked: 0
B7. re-entered; LIVE poller restored: True
B8. receipt sync WORKS again after re-entry: True
B9. cross-interface convergence live after re-entry: True
```

Moving the attention block *after* `dispose()` and suppressing the task's own failure is the right shape: a read receipt can no longer strand a session owner.

**U3 — RESOLVED, and I confirmed the guard actually engages.** My first attempt at this was an invalid test and I want to flag the trap: `BEGIN IMMEDIATE` takes a RESERVED lock which still admits readers, so the read-only `state_many`/`revision` connections never failed and the guards were never exercised — the probe passed while proving nothing. Re-run under `BEGIN EXCLUSIVE`, which is what a reader actually meets:

```
D0. fixture valid: raw read raises OperationalError: database is locked
D1. OPENING the conversation under a locked store: OK (d5628d2fbedc)
D2. retains last-known state, does NOT claim read: unseen=True
D3. SIDEBAR under a locked store: OK, 1 row(s)
D4. badge OMITTED rather than fabricated: True
D5. poller still alive while store is locked: True
D6. after unlock, badge returns: True
D7. still correctly UNREAD (nothing acknowledged blind): True
```

Both primary paths stay available, the degraded state omits the badge rather than inventing one, and — the part that matters for this feature's honesty — a degraded store never produces a false *read*. The new 503 (`"Read state is busy right now. It will catch up on its own."`) is a vetted user-shaped sentence that does not echo SQLite's wording or leak a path, and it is accurate: it does catch up, as D6 shows.

**U4 — RESOLVED on the substance.** I pushed on this one because "stops throwing while still telling the user nothing" is not a fix. It genuinely backs off, and it genuinely recovers:

```
schedule: f3=1s f4=2s f5=4s f6=8s f7=16s f8=32s f9=60s (capped)
failing for  10s ->  7 attempts, read registers 6.0s after recovery
failing for 120s -> 10 attempts, read registers 4.0s after recovery
10 min persistent failure: 17 attempts (was 1200)
```

7,200/hour → 60/hour at the ceiling, `stop()` on success, and a *recovering* backend is picked up within seconds rather than the user being stranded at a 60 s ceiling. The refusal string is now correctly documented as a guard rather than copy ("Not user-facing copy: the renderer treats a refusal as 'not read yet' and retries"), which resolves the unreachable-copy half honestly instead of inventing a toast nobody wants in a background sync path. Diagnosability is real on the backend side — `logger.warning("attention poll failed for %s: %s")` propagates to the rotating `local-operator.log`. See U8 below for the renderer half, which is the one soft spot.

**Round-1 invariant re-checked and still holding.** The `data-completion-complete` change from `"false"` to `undefined` is the kind of edit that quietly loosens eligibility, so I re-ran the predicate over the cases that matter:

```
settled assistant with text        -> "true"      (ackable)
STREAMING assistant (clipped)      -> undefined   (stays unread)
capped assistant, empty text       -> undefined   (stays unread)
assistant NOT complete             -> undefined   (stays unread)
record with no complete field      -> undefined   (stays unread)
```

A capped/clipped ending still stays unread, and since the selector only matches `="true"`, an omitted attribute is unmatchable. I found no path in this delta that acknowledges something the user did not actually see. The new synthesized row also cannot overwrite a real message: `state.index.has(anchor)` returns the state unchanged when the anchor collides with an existing durable id (verified against a properly-seeded transcript — my first fixture built zero rows and produced a false positive, which I chased down rather than reporting).

---

**U7 — MAJOR — The crash-recovered outcome row erases itself the moment it is read, so the user sees "Interrupted" flash and vanish, and it is gone from the conversation forever.**

This is a new interaction introduced by this remediation, so it is in scope. `withRecoveredOutcome` is the *only* producer of that row (crash recovery never wrote the durable `completion_attention` entry — that is the premise of the fix), and it is gated on `!attention?.unseen`. The row is also fully ackable, so it acknowledges itself almost immediately:

```
J1. synthesized row carries data-completion-complete="true" -> it IS hit-tested and ACKed
J2. the ACK fires on the very first eligible frame (rAF, then every 500ms)
J3. backend publishes unseen=false; the same memo re-runs:
    rows = [u1]  (row removed: true)
H7. after reopening the conversation later, outcome visible: false
```

Backend confirms the anchor and kind are still supplied after the receipt, so `unseen` is the sole thing removing it:

```
H9.  AFTER read -> unseen=False kind=interrupted anchor=completion-1f8f5be8-...
H10. anchor+kind still supplied after read: True
```

Step for a user: the agent crashes mid-turn; you open the conversation; "Interrupted" appears at the tail; within ~500 ms it disappears while you are looking at it; you reopen that conversation tomorrow and there is no record that the run was ever interrupted — just a user message and a partial reply that looks like the agent simply stopped talking. Expected: reading an outcome marks it read, it does not delete it. An interruption notice is transcript content, not a notification.

This is also a cross-surface disagreement, which is precisely what the feature exists to prevent: **the TUI keeps its notice.** `_poll_completion_attention` calls `_append_block(NoticeBlock(...))` — a real persistent widget that survives the receipt — while the desktop derives the row from `unseen` and drops it. The commit message claims parity with the TUI's guard; the *insertion* condition matches, but the *lifetime* does not. After the ACK the phone/TUI and desktop show different transcripts for the same conversation.

Fix direction (renderer-only, no contract change): remember the anchor once synthesized for the mounted conversation and keep rendering it regardless of `unseen`, so the row's lifetime matches the TUI's. The `unseen` check is doing double duty as both "should I create this?" and "should this still exist?" — only the first is correct.

**U8 — MINOR — The renderer's one diagnostic almost certainly cannot be read in a shipped build.**

The backoff logs once at the threshold via `console.warn`, which is the right instinct. But `devTools: Boolean(process.env.ELECTRON_RENDERER_URL)` — devTools are disabled for `pnpm start` and production builds — and there is no `console-message` forwarder anywhere in `src/main/`, while `src/main/backend/logger.ts` (electron-log) already writes durable files for BACKEND/INSTALLER/OAUTH. So on a user's machine the message lands nowhere reachable. The backoff itself is the important half and it works, so this is MINOR rather than a re-opened U4: routing that one warning through the existing `logger` (or adding a `console-message` bridge) would make the failure genuinely diagnosable from a user's log bundle instead of only in `pnpm dev`.

**U9 — NIT — The `revision()` fast path is a real fix for the wrong half of U5, and is not currently cheaper per call.**

Gating the poll on a change detector is the right design and it works — 6 s idle now costs 0 full reads instead of 6, and a real change still converges (`I4. converges: True`). Flagging only that the two are the same price here:

```
I3. revision(): 83 us   state(): 75 us   (0.9x)
```

The win is avoiding `refresh_attention`'s thread hop and diff/publish, not the query. My round-1 point was that this polls at 1 Hz per attached bridge even while the window is hidden — still true, still fine, still worth revisiting only if bridge counts grow. No action needed in this PR.

---

**Not simulated, and still open.** The native-foreground admission cell remains BLOCKED for me, exactly as in round 1: the IPC gate is proven at unit level against a `BrowserWindow` fixture (14/14 in `desktop-contract.test.mjs` on this head), and `guardForegroundReceipts` moving onto the sender itself is a genuine improvement — `DesktopNotifier` holds its own reference to that sender, so the previous handler-level placement did not have the defence-in-depth property its comment claimed. But "a real user really looked at a real focused window" needs a live foreground app, which I do not have here. That is Q2, and it is the manager's to close with a live window. I did not fake it.

**Verdict: NOT TERMINAL.** One MAJOR (U7) remains. All four round-1 findings — both blockers included — are genuinely resolved under the real failure mode rather than the proxy, and the core honesty invariant holds: I found no way in this delta to acknowledge something the user did not see, and a capped ending still stays unread. U7 is the inverse failure and it is new in this remediation: the desktop now correctly stops hiding unread work, but deletes a crash-recovery outcome the instant it is read, permanently and only on desktop. U8 and U9 are follow-ups and need not gate this round.

damianvtran:
### UX review remediation — round 2

Author: coder implementing agent (`anthropic/claude-opus-5`).
Head: `12b3dc0a0f83bfc9f46f572a550cacd9d449b1b1`. Backend companion: damianvtran/local-operator#697 (`44c8fa00`).

Answering [UX review — round 2](https://github.com/damianvtran/local-operator-ui/pull/85#issuecomment-5557765470) and the `[ui]` findings from [Agent review — round 2](https://github.com/damianvtran/local-operator/pull/697#issuecomment-5557779516). U7 and F12 are the same defect, found independently by both reviewers.

| ID | Disposition |
| --- | --- |
| U7 / F12 | **Fixed** in `12b3dc0a` |
| Q3 | **Fixed** in `12b3dc0a` |
| U8 | **Fixed** in `12b3dc0a` |
| F14, U9, O3/O4 | **No action** — see below |

#### U7 / F12 — MAJOR — the recovered row erased itself the moment it was read

Fixed, and both reviewers were right to call this worse than the defect it came from. The row is itself ackable, so it acknowledged itself within ~500 ms, and since the receipt writes only to the store and never adds a transcript entry, nothing durable replaced it. `unseen` was doing double duty as *"should this be created?"* and *"should this still exist?"* — only the first is correct.

The mounted conversation now remembers the anchors it has synthesized and keeps rendering them regardless of `unseen`. Verified against the real bundled reducer with the exact pre/post-ACK payloads from your repro:

```text
frame 1 (unseen: true)  -> row rendered : true
frame 2 (unseen: false) -> row rendered : true      (was false)
frame 3 (still read)    -> row rendered : true
read on ANOTHER surface, never shown here : false
streaming retry guard (unseen)            : false
idempotent rows after 2nd apply           : 1
collides with real durable id             : user     (durable row wins)
```

The lifetime now matches the TUI's `_append_block(NoticeBlock)`: reading an outcome marks it read, it does not delete it. The three round-1 invariants are unchanged — an outcome acknowledged elsewhere and never shown here still does not appear, a historical failure is still not inserted at a running retry's tail, and a real durable row is never overwritten.

#### Q3 — `withRecoveredOutcome` shipped with zero tests

Fixed. Your 7-cell probe is ported into `scripts/transcript-reducer.test.mjs` (7 → 13 tests) rather than left as a throwaway: the anchor/ackability contract, the lifetime across the ACK, an outcome read elsewhere, the streaming retry guard, idempotency plus durable-id collision, and the kinds it must ignore.

The lifetime test is a real regression guard, not decoration — reverting the one-line U7 fix fails it and nothing else:

```text
✖ the recovered row survives its own acknowledgement
ℹ pass 12  ℹ fail 1
```

#### U8 — MINOR — the renderer diagnostic could not be read in a shipped build

Fixed by adding the `console-message` bridge you identified as missing, routing to the existing electron-log backend file. Only warning and error are forwarded; info and debug would bury the failures the log exists to explain.

Confirmed it survives the real build rather than assuming — the main bundle compiles to V8 bytecode, so checked with `strings`:

```text
console-message   : 1
renderer          : 3
backend-service   : 1
```

#### F14 / U9 / O3 / O4 — no action

F14 (the nine reformatted files) is left alone: the reviewer verified it behaviourally empty and judged reverting riskier, and I agree — a revert would touch release-critical scripts to no behavioural end. Recorded, not fixed. U9 and O3/O4 are observations with no requested change.

#### Gates on this head

```text
pnpm check-types    0 errors
pnpm lint           7 warnings — pre-existing, unchanged
pnpm check-themes   1884 contrast assertions, 12 themes
pnpm test:desktop   27/27 pass (was 21)
pnpm build          succeeds
CI                  Pre-Release CI Testing 12b3dc0a: success
```

Q2 (native foreground on a genuinely focused window) remains BLOCKED for the same reason as round 1 and is the manager's to close with a live window. I did not fake it. No merge or release performed.

damianvtran:
### UX review — round 3

Reviewer: ux-reviewer (`anthropic/claude-opus-5`)
Scope: `9191f9bc7fb1a2e25ff1151735118dee5c1492ed..12b3dc0a0f83bfc9f46f572a550cacd9d449b1b1` (UI remediation delta), with backend counterpart `ecb993986b86..44c8fa000a06` (damianvtran/local-operator#697) read as the route it calls. Reviewed head `12b3dc0a0f83` equals current PR head.

Narrow confirmation round on the round-2 remediation only. I audited the four changed items and re-verified my own U7/U8; I did not re-audit approved unchanged flows and raised no new out-of-scope findings. F14, U6, U9, O3/O4 stay closed on the manager's accepted reasoning, and the `test (3.12, 1)` timing-flake cluster is settled — I did not re-run or touch it. Q2 (native foreground on a live focused window) remains the manager's and is still not simulated here.

---

#### U7 / F12 — MAJOR — **RESOLVED.** The row now outlives its own receipt, and it matches the TUI.

The fix is the right shape: `unseen` gates CREATION only, and a per-conversation `remembered` set keeps the row alive afterwards. I verified this from the user's side against the real bundled reducer, not from the diff.

**The flash-and-vanish is gone.** My round-2 J3 repro, replayed frame by frame:

```text
A1 frame1 unseen=true  -> notice: true | user:u1 | notice:completion-1f8f5be
A2 frame2 unseen=false -> notice: true | user:u1 | notice:completion-1f8f5be   <- survives its own ACK
A3 after 20 more polls -> notice: true | notice rows: 1 | total: 2
A4 text/level: "Interrupted" warning
```

Round 2 dropped the row at A2. It now persists, at one row, with the copy unchanged.

**On the "does it survive re-entering the conversation later" half (my H7 step) — the answer is no, and after checking the TUI I am accepting that as correct, because it is exactly what parity requires.** This is the part I could only settle by driving the other surface, so I did: a real `OperatorApp` under `run_test`, a crash-recovered `interrupted` outcome whose anchor has no durable row, acknowledged, then the app relaunched against the same `attention.db`.

```text
=== TUI PARITY BASELINE (real OperatorApp) ===
T1 notice created                     : True | unseen: True
T2 outcome acknowledged               : True
T3 notice SURVIVES the ACK            : True
T4 still exactly one after 10 polls   : True (count=1)
T5 anchor+kind still supplied post-ACK: interrupted | unseen: False

--- relaunch the TUI on the same conversation (re-entry) ---
T6 notice present after relaunch      : False (count=0)
T7 store state on relaunch: {... 'kind': 'interrupted', 'unseen': False, 'revision': [1, 1]}
```

The TUI's `NoticeBlock` is a mounted widget, never written to durable history, so it too is gone on relaunch. Desktop now behaves identically on both axes — survives the ACK (T3 / A2), absent after re-entry (T6 / B1):

```text
B1 reopened after read -> notice: false | user:u1
```

So the divergence I reported is closed. The residual "an interruption leaves no permanent trace once read" is a property of the shared design — the notice is view state on every surface, not transcript content — and fixing it means giving crash recovery a durable transcript entry in core, which is out of scope for a desktop adapter and would make desktop diverge again if done here alone. Recording it as a follow-up below rather than holding this PR: closing it in the renderer is precisely the mistake round 2 made.

**Nothing I validated earlier was loosened.** Re-ran the full invariant set against the real reducer:

```text
C1 read elsewhere, never shown here      -> notice: false   (no row on a surface that never showed it)
D1 50 polls, unseen true->false          -> notice rows: 1 | total rows: 2   (idempotent, no duplicates)
E1 streaming=true                        -> notice: false | remembered size: 0   (no leak into the set)
E2 after stream, unseen=false, fresh view-> notice: false   (a running retry still gets no historical tail row)
F  complete -> false | error -> true | interrupted -> true
G1 durable row at the anchor             -> kind: assistant  (never overwrites real content)
```

E1 matters most: the streaming guard returns before `remembered.add`, so a capped/streaming/incomplete ending still cannot be remembered into existence later. Cross-conversation isolation and React's double-invoke also hold:

```text
2 ACK lands, still A       -> notices: [anchor-A]        (survives ACK)
3 switch to B              -> notices: [anchor-B]        (no A leak)
4 switch BACK to A, read   -> notices: []                (matches TUI relaunch)
6 StrictMode double-render -> notice rows: 1 | total rows: 2
```

The `useRef` reset keyed on `frontend.session_id` is what buys line 3, and it is reset during render rather than in an effect, so a session switch cannot paint one conversation's outcome into another's first frame.

#### U8 — MINOR — **RESOLVED.** The warning now reaches a durable file in a packaged build.

I did not take the forwarder on trust — the failure mode I reported was environmental, so I reproduced the packaged environment: real Electron 35.5.1, `devTools: Boolean(process.env.ELECTRON_RENDERER_URL)` with the variable unset, `NODE_ENV` unset, the forwarder and `electron-log` wiring copied verbatim from `src/main/index.ts` and `logger.ts`, emitting all four console levels from the renderer.

```text
PROBE devTools enabled: false | can open: true
=== FILE CONTENTS (backend-service.log) ===
[2026-09-06 05:09:35.150] [warn]  [renderer] [attention] could not mark session/abc read after 3 attempts; backing off
[2026-09-06 05:09:35.150] [error] [renderer] [attention] hard failure example
```

The exact backoff string from `use-completion-view.ts:101` lands on disk with devTools off; `console.debug` and `console.log` are correctly filtered by the `level < 2` guard, so the log stays readable. `logger.transports.file.level` is `"debug"` unconditionally, so this is not gated on a dev-only branch. A user's log bundle can now explain a wedged receipt.

Two notes, neither blocking. Electron 35 prints `'console-message' arguments are deprecated and will be removed. Please use Event<WebContentsConsoleMessageEventParams> object instead.` — the positional form still fires correctly here, but it is on a removal path, so this is a follow-up before the next major Electron bump. And the forwarder catches all renderer warnings, so the first thing in the file is Electron's own CSP security warning; that one self-suppresses in a genuinely packaged app, so it is noise only in this probe.

#### Q3 — **CONFIRMED.** `withRecoveredOutcome` is no longer untested: 7 -> 13.

```text
ℹ tests 13   ℹ pass 13   ℹ fail 0
✔ the recovered row survives its own acknowledgement
✔ an outcome already read elsewhere does not appear in a conversation that never showed it
✔ synthesis is idempotent and never overwrites a real durable row
```

The six new tests cover the lifetime rule the round-2 defect violated, which is the one that needed pinning.

#### F13 — **CONFIRMED** (read as the route the renderer depends on, not re-reviewed).

Logging the transition rather than the tick, with recovery logged too, brackets an outage instead of leaving one warning of unknown duration. From the UX angle the operator-facing point is that a persistent store fault no longer buries every other line at ~3,600/hour per bridge.

Gates on this head: `pnpm check-types` clean, `pnpm lint` exit 0 (7 warnings, all pre-existing and none in the touched files).

---

**Verdict: TERMINAL.** No BLOCKER and no MAJOR remains. U7 — the one finding that held round 2 open — is genuinely fixed rather than papered over: the row survives its receipt, and the re-entry behaviour I flagged turns out to be the TUI's behaviour too, verified by driving the real TUI rather than assuming. The honesty invariants I care about all still hold on this head: nothing gets acknowledged that the user did not see, a capped or streaming ending still stays unread, no row appears on a surface that never showed it, and there are no duplicates.

Follow-ups, neither gating this PR:
- A crash-recovered outcome leaves no permanent transcript trace once read, on every surface. Closing it means a durable entry from crash recovery in core; doing it in the renderer alone would re-create the divergence this round just fixed.
- `console-message`'s positional signature is deprecated in Electron 35; migrate to the `Event<WebContentsConsoleMessageEventParams>` object before the next major bump.

damianvtran:
### UX review remediation — round 3 (addendum: rebased onto `main`, synthetic base dropped)

Author: coder implementing agent (`anthropic/claude-opus-5`).

The prerequisite base branch `dev-backend-parity` landed on `main` as `bdf98ffe3aa2d250de7ee42aa6c6804ae396d877` (local-operator-ui v0.15.0), so this PR no longer needs a synthetic base. Rebased directly onto `origin/main` and repointed the PR base from `dev-backend-parity` to `main`.

| | old | new |
| --- | --- | --- |
| base | `8c60fd2ac1c2cf3dd462fb4804ccb6f7e3804115` (`dev-backend-parity`) | `bdf98ffe3aa2d250de7ee42aa6c6804ae396d877` (`main`) |
| head | `12b3dc0a0f83bfc9f46f572a550cacd9d449b1b1` | `739142b97261e50bee3ea865ba5448cd3e848db6` |

The rebase was a clean replay — no conflict, and no semantic conflict to resolve.

#### The reviewed content is unchanged, and here is the proof rather than the assertion

```text
git range-diff 8c60fd2ac..12b3dc0a0 origin/main..739142b97
  1:  SHA = 1:  SHA feat(chat): mark a completion read only when it is actually viewed
  2:  SHA = 2:  SHA fix(chat): give a crash-recovered outcome a row, and stop retrying blind
  3:  SHA = 3:  SHA fix(chat): keep a crash-recovered outcome after it has been read
```

3/3 commits `=`, and the `+`/`-` line sets across the two ranges are byte-identical — **2708 lines**, `diff -q` clean. So the round-3 **UX TERMINAL** verdict (U7 resolved, U8 resolved, Q3 at 13/13) was established on content this rebase does not alter. A reviewer does not have to take that on trust; the two commands above reproduce it.

#### Version

`package.json` stays at `0.15.0` — the same version `main` carries. No bump, and my three commits do not touch it.

#### Gates on `739142b97`

```text
pnpm check-types    0 errors
pnpm lint           7 warnings — pre-existing, unchanged
pnpm test:desktop   27/27 pass
pnpm check-themes   1884 contrast assertions across 12 themes
pnpm build          succeeds
CI                  Pre-Release CI Testing 739142b9: success
```

Still **draft** and not merged. Repointing the base cleared the draft flag as a side effect of the GitHub API call; I set it back immediately, since the hold is the operator's to release. Q2 (native foreground on a genuinely focused window) remains open and unchanged by this rebase.

damianvtran:
### UX review remediation — round 3 (addendum: rebased onto current `main`)

Author: coder implementing agent (`anthropic/claude-opus-5`).

`main` had gained `e766e582d` (#86, CODEOWNERS) since the last rebase. Rebased onto it: head `739142b97` → **`74d3a8c4b132d252d13a60d80bbbad6d54d7c726`**. Clean replay, no conflicts.

#### Equivalence

```text
git range-diff bdf98ffe3..739142b97 origin/main..74d3a8c4b
  1:  SHA = 1:  SHA feat(chat): mark a completion read only when it is actually viewed
  2:  SHA = 2:  SHA fix(chat): give a crash-recovered outcome a row, and stop retrying blind
  3:  SHA = 3:  SHA fix(chat): keep a crash-recovered outcome after it has been read
```

3/3 `=`, and the `+`/`-` line sets are byte-identical across both ranges — **2708 lines**, `diff -q` clean. The round-3 UX TERMINAL verdict (U7 resolved, U8 resolved, Q3 13/13) stands on content this rebase does not alter.

#### Version

```text
git diff origin/main -- package.json   ->  0 lines
"version": "0.15.0"                        (exactly what main carries)
```

#### Gates on `74d3a8c4b`

```text
pnpm check-types    0 errors
pnpm lint           7 warnings — pre-existing, unchanged
pnpm test:desktop   27/27 pass
pnpm check-themes   1884 contrast assertions across 12 themes
pnpm build          succeeds
CI                  Pre-Release CI Testing 74d3a8c4: success
```

Still draft, not merged. Q2 (native foreground on a genuinely focused window) remains open and is unaffected by this rebase.

damianvtran:
### Agent review — round 4

Reviewer: reviewer subagent, anthropic/claude-opus-5 (no effort tier configured on this host; no provider fallback occurred during this round)
Scope: e766e582d..74d3a8c4b132d252d13a60d80bbbad6d54d7c726

Short convergence note. This PR's content already passed its prior rounds; the only question here is whether the rebase onto moved `main` disturbed anything.

**Equivalence — reproduced, holds.** Pre-rebase head recovered from the worktree reflog (`739142b97`, on old base `bdf98ffe3`):

```
$ git range-diff bdf98ffe3..739142b97 e766e582d..74d3a8c4b
1:  f603ab1a5 = 1:  ed2e322f8 feat(chat): mark a completion read only when it is actually viewed
2:  4e762d4c2 = 2:  e35a94602 fix(chat): give a crash-recovered outcome a row, and stop retrying blind
3:  739142b97 = 3:  74d3a8c4b fix(chat): keep a crash-recovered outcome after it has been read

old lines: 2708   new lines: 2708   → IDENTICAL
```

3/3 `=`, **2708/2708 byte-identical** +/- lines. Prior rounds carry over to this head.

**Upstream seam — none exists.** The only commit between the old and new base is `e766e582d` (`chore(repo): add CODEOWNERS and record who may merge`, #86), touching `.github/CODEOWNERS` (+18) and `AGENTS.md` (+42). Zero intersection with this PR's 18 files, none of which is a repo-governance file. There is no seam to audit semantically — unlike the backend, where #707 genuinely moved code under two shared files (analysed in detail on local-operator#697 round 4, verdict clean/TERMINAL).

**Consumer side of the backend seam.** Nothing in this PR caches the runtime capability itself; it consumes `supported` from the attention state the backend publishes, and that field is recomputed on the backend's poll key whenever owner liveness changes. So a runtime refresh underneath a live client cannot strand this UI on a stale `supported` — it re-renders from the next published state.

#### Findings

None. No BLOCKER, no MAJOR, no MINOR, no NIT.

#### Verdict

**clean — TERMINAL.** Content-identical rebase, no upstream overlap, no semantic conflict. Merges on the same basis as damianvtran/local-operator#697.

damianvtran:
### Design review — round 1

Reviewer: designer (`anthropic/claude-opus-5`) — no provider fallback occurred during this round; no effort tier is configured on this host.
Scope: `origin/main..74d3a8c4b132d252d13a60d80bbbad6d54d7c726`, visual surface only — `canonical-transcript.tsx`, `transcript-reducer.ts` (`withRecoveredOutcome`), `chat-content.tsx`, `use-completion-view.ts`. Reviewed head `74d3a8c4b132` equals current PR head.

First design round on this PR; UX rounds 1-3 and code review round 4 are already closed. I judged the **rendered result** of this diff and did not reopen the interaction-flow questions U7/U8/H7 settled, or raise findings on surfaces this diff does not touch.

**How I looked at it.** The Electron app proper needs a live backend, so I rendered the production `CanonicalTranscript` in Storybook — the same harness `canonical-notice.stories.tsx` already uses, which loads `styles/index.css`, the generated theme CSS and the real Geist faces, so type and colour are the product's and not my OS's. Eleven states in the worktree at `74d3a8c4b132`, captured at 3456x1812 on `localOperatorDark`, diffed pairwise with ImageMagick `-metric AE`. Frames in `/tmp/design-r1/`. My capture harness and a local `.env.local` were deleted afterwards; `git status --porcelain` is empty at the review head.

One methodology note worth recording: my first before/after pair for the transition was **invalid** — screenshot latency exceeded the 1200 ms timer, so both frames captured the post-flip state and compared "identical" for the wrong reason. I widened the timer to 15 s and gated each capture on reading the live flag out of the DOM (`unseen=true` confirmed before, `unseen=false` confirmed after). The numbers below are from the corrected pair.

---

#### What the diff actually adds to the screen

Worth stating plainly, because it reframes questions 1 and 2 of the brief: **this app renders no unread affordance at all.** `attention` is consumed only by `use-completion-view` (to send the receipt) and by `withRecoveredOutcome` (to synthesize a row); `chat-sidebar.tsx` has no badge, dot or count, and this PR does not add one. Desktop is the surface that *clears* the flag; the TUI and mobile relay are the surfaces that *display* it.

So the only new pixels here are the crash-recovered outcome rows — `Interrupted` and `Stopped with an error`.

---

#### D1 — NIT — Viewed and unviewed ordinary completions are pixel-identical, and that is correct here

`unseen: true` vs `unseen: false` on the same settled turn: **0 differing pixels of 6,262,272** (`-metric AE` = 0; SHA-256 of both PNGs identical). Nothing in the transcript encodes read state.

I am deliberately not raising this as a defect. There is no element in this app claiming "new", so there is nothing that would need to stop claiming it, and inventing a desktop-only unread treatment inside this PR would put a third rendering of read state next to the two that already own it. The state is legible where it is displayed. Recording it so the round shows the question was asked and answered against frames rather than assumed.

#### D2 — NIT — The transition is imperceptible; no reflow, no jump

The brief asks whether the flip is perceptible or whether it "happens invisibly and leaves them unsure whether their view registered". Measured, with the state change confirmed live in the DOM either side:

```text
G-unseen-TRUE.png   flag read from DOM: unseen=true
G-unseen-FALSE.png  flag read from DOM: unseen=false

full frame           537 differing pixels (8.6e-05)
  difference bbox    51x19+147+56   <- my debug flag text only
transcript region    0 differing pixels   (crop 3456x1712+0+100, flag excluded)
```

Zero pixels move in the transcript. Question 2 is therefore clean in the strong sense — there is no first-frame-vs-settled-frame discrepancy, no reflow, and the recovered row does **not** disappear under the user when its own receipt lands (the round-2 flash-and-vanish U7 closed stays closed at this head).

The invisibility is the same non-issue as D1: the receipt is bookkeeping for other surfaces, and a desktop toast confirming "we marked this read" would be noise about a thing the user never asked for.

#### D3 — NIT — Recovered outcome rows render correctly at both levels

```text
C  interrupted -> "Interrupted",            TriangleAlert, text-ink-dim
D  error       -> "Stopped with an error",  CircleAlert,   text-danger
```

Both sit at the trace tier as one quiet line — correct under branding § 7 ("a completed action is one quiet line, not a card"), and the right call for an outcome the user is reading after the fact rather than being asked to act on.

Contrast against `canvas` `#15130e`, from the palette tokens rather than antialiased subpixels:

| role | token | ratio | floor |
|---|---|---|---|
| `text-ink-dim` (Interrupted) | `#918b7d` | **5.47:1** | 4.5:1 AA |
| `text-danger` (Stopped with an error) | sampled `#e1867c` | **6.98:1** | 4.5:1 AA |

Copy matches the TUI verbatim — `tui/app.py:11652` emits the same two strings — so the two surfaces read identically for the same conversation, which is the divergence this feature exists to remove. Sentence case, no emoji, machine voice in monospace: all consistent with the contract.

#### D4 — NIT — Degraded acknowledgement is visually indistinguishable from success

Question 3's error cell, and the one place I think a real gap exists. With `supported: false` (backend predates the receipt route, so the ack can never land), the frame is identical to the healthy error case bar antialiasing:

```text
D-error.png vs H-degraded2.png -> 2008 px (0.00032), bbox 302x24+993+1043
crop of that region: same glyph, same ink, same string
```

The only signal is a `console.warn` after three failures. The user sees success; their TUI and phone keep showing the session unread indefinitely, with no recourse and no explanation.

I am filing this as a NIT rather than a blocker, deliberately. The failure is **safe-by-construction**: a failed receipt leaves authoritative state intact, so the session stays unread — the conservative direction, and closer to pre-PR behaviour than a false "read" would be. Surfacing it properly means deciding *where* a cross-surface sync failure belongs in this app's information hierarchy, and this app currently has no unread surface to hang it on (D1). That is a design question larger than this diff and should not be answered inside it. Follow-up below.

#### D5 — NIT — Warning-level notices are separated from info-level by glyph shape alone

`level: "warning"` renders `text-ink-dim` — the same ink as an ordinary info trace — with only `TriangleAlert` vs `Info` distinguishing them, and the glyph itself is also `text-ink-dim` (`trace-line.tsx:105`, `:124-125`). Only `level: "error"` moves ink to `text-danger`. So `Interrupted` reads at a glance as an ordinary quiet trace line rather than as an outcome.

**Pre-existing, not introduced here** — `baseline-notices.png` on `main` shows the identical treatment for warning notices — and this diff only newly routes a string through it. Raising it as an observation for the follow-up list, not as a change request against this PR; fixing it would touch every notice in the app and belongs in its own diff with its own contrast round.

---

#### States checked and clean, no finding

| # | State | Result |
|---|---|---|
| A/B | Settled turn, viewed vs unviewed | 0 px diff (D1) |
| C | Recovered `interrupted`, unseen | Correct row, correct tier |
| D | Recovered `error`, unseen | Correct row, `text-danger` |
| E | Recovered outcome, **empty** transcript | Renders alone, correctly placed, no orphan spacing |
| F | No completions, no attention | Blank; owned by the parent, untouched here |
| G | Transition unseen -> seen | 0 px in transcript (D2) |
| H | Ack unreachable / unsupported | Identical to success (D4) |
| I | Narrow view, 420px, `isSmallView` | Wraps cleanly, no clipping, no overflow |
| J | 12-turn transcript + recovered tail | Grouping rhythm holds; tail row correctly last |
| K | `streaming: true` (retry running) | **0 px vs plain viewed** — historical failure correctly suppressed mid-retry |

Alignment and spacing rhythm are unchanged from `main` in every populated frame: the recovered row enters the existing `trace` gap tier, and content bbox growth between A and C is 76 px for one added row at the same measure — the row's own height, no ripple. No console errors at any state.

---

### Verdict

**No BLOCKER and no MAJOR. Round 1 is TERMINAL** — clean from a design standpoint, and this does not require a round 2.

Everything above is a NIT. D1/D2/D3 are confirmations rather than complaints: the new rows are legible, correctly tiered, contrast-clean at both levels, verbatim-consistent with the TUI, stable at narrow width, correctly suppressed during a retry, and the read transition moves zero pixels in the transcript.

Follow-ups, recorded here rather than as issues, per the no-follow-up-issues rule:

- **D4** — decide where a persistently failing cross-surface receipt should surface in this app, once there is an unread surface to attach it to. Blocked on that larger question, not on this diff.
- **D5** — warning-level notices carry their level in glyph shape alone; revisit alongside any future pass on notice hierarchy, with its own contrast round.

damianvtran:
### Released in v0.50.4

- **Tag**: [`v0.50.4`](https://github.com/damianvtran/local-operator/releases/tag/v0.50.4) on `06ffc717083447dfb6d79a5c930de7d7a0b03279` (the bump commit)
- **PyPI**: `local-operator` 0.50.4 — publish.yml run 34062544510 green on all three jobs (`release-build`, `pypi-publish`, `extension-release`)
- **Installed**: `.lop-source` = `06ffc717083447dfb6d79a5c930de7d7a0b03279 main`
- **Smoke**: `lop --version` → `v0.50.4` from outside the repository

Window contents: #713, #697 (plus `local-operator-ui#85`, versioned separately).

Run `/reload` in any running session to pick up the new build.

PR #87, opened 2026-09-06
fix(desktop): onboarding census, badge overlap, and stale-agent 404

damianvtran:
# PR Description

## Summary of Changes

0.15.1 patch for four defects found after 0.15.0 shipped. Companion backend is already v0.50.0; this is UI-only.

- **Onboarding gate uses the provider census**, not the legacy env-key list. A user signed in via backend OAuth was still treated as first-time because `/v1/credentials` does not see the auth DB. Also stops persisting `isModalActive`, which pinned the modal open after one wrong launch.
- **Provider-grid badges no longer overlap the name** at narrow width: cards stack name / badge / method, and the local label shortens to "No key needed".
- **A stale persisted agent id is no longer a dead end.** 404 clears the selection and shows the list; "try again once the local server is running" is reserved for actual connection failures.
- **Regression test** pins readiness labels to a real 0.50.0 census payload.

Round-1 review/QA/design findings (unmanaged first-time hang, hosting-picker census, copy) are being remediated in this same branch before merge.

## Impact

- **No Breaking Changes.** Existing chats, transcripts and the 0.50.0 control surface are unchanged.
- **User:** returning users with OAuth credentials skip onboarding; a deleted agent no longer claims the server is down.

## Testing Details

Coder + independent reviewer/QA/designer rounds. Evidence under the PR comments. `pnpm lint` (7 pre-existing warnings), `check-types`, `check-themes` 1884/12, `test:desktop` 31/31 (13 new) on the first head; remediation will re-run.

## Checklist

- [x] Style guidelines
- [x] Self-review
- [x] Tests for the census gate, readiness labels, and stale-agent 404
- [x] Visual stills for the overlap and placeholders

PR #77, opened 2026-08-04
refactor(ui): Tailwind v4 + shadcn, and a visual system that matches the product

damianvtran:
Migrates the renderer from **MUI 6 + Emotion + styled-components** to **Tailwind v4 + shadcn**, and retunes the visual system to the Local Operator brand.

The app is a desktop tool for non-technical people running a team of AI agents on their own machine. It looked like a developer console. This changes both the stack and how the product reads.

## What changed

**One token source, two consumers.** Twelve user-selectable themes previously lived as twelve standalone `createTheme()` calls (2,686 lines, 67% repeated scaffolding). They are now twelve `ThemePalette` objects behind one `createBaseTheme(palette)` factory, and a generator emits `themes.generated.css` from the same objects — so MUI and Tailwind read the same values and cannot drift. `pnpm check-themes` fails the build if the generated CSS is stale or a contrast floor breaks.

**The contrast contract is executable.** 1,368 assertions across twelve palettes, asserting over component *triples* (ground + fill + border + ink), not just text/background pairs — a control whose fill is invisible but whose border clears 3:1 is legal; one where neither does is not. Zero pinned exceptions.

**MUI is gone** except the date-picker. 1,564 `styled()` call sites across 177 files → 0. `@mui/material` imports 335 → 4. styled-components and Emotion removed entirely.

**The chat surface was redesigned around what an agent run actually is.** Tool traces were the loudest thing on screen; they are now one quiet monospace line each. Streaming text is visible as it arrives instead of being hidden behind a collapsed pill and popped in on completion. The `ASK` question — the one thing that blocks the run on a human answer — was rendering as ordinary assistant prose and is now an accent callout with an icon.

**Startup JS: 15,502 kB → 5,479 kB** (−65%). FontAwesome registered ~1,900 icons to serve 92; routes were all statically imported; KaTeX and mermaid sat on the startup path.

## Testing evidence

96 Storybook frames captured across all twelve palettes in a private headless Chromium over CDP, with animations frozen before each shot: `docs/evidence/`. The capture script is committed (`scripts/capture-evidence.mjs`).

Gates, all green on `ec6112f8d`:

| Gate | Result |
|---|---|
| `pnpm check-types` | 0 errors |
| `pnpm lint` | 373 files, clean |
| `pnpm check-themes` | 1,368 assertions, 12 themes, 0 pinned exceptions |
| `pnpm build` | clean, 9.86s |
| `pnpm startup-closure` | 5,479 kB |

**Measured, not asserted:**

[...]

PR #90, opened 2026-09-06
fix(npm): ship a runnable npx/global package on macOS ARM

### LINKED ISSUE #88: npx/global install fails on macOS ARM with cachedDataRejected
### Summary

`local-operator-ui@0.15.0` fails during startup with `Invalid or incompatible cached data (cachedDataRejected)` when installed or run through npm or Bun.

I reproduced this through:

```sh
npx local-operator-ui
npm install -g local-operator-ui
local-operator-ui
```

I also tried the equivalent automatic and global installation flows with Bun (`bunx` and a global Bun install), with the same result.

### Environment

- macOS 26.6.2
- Apple Silicon (`arm64`)
- Node.js `v24.12.0`
- npm `11.6.2`
- Bun `1.3.5`
- `local-operator-ui` `0.15.0`

### Actual behavior

```text
Downloading Electron binary...
App threw an error during load
Error: Invalid or incompatible cached data (cachedDataRejected)
    at Module._extensions..jsc.Module._extensions..cjsc (.../local-operator-ui/out/main/bytecode-loader.cjs:50:11)
    at Module.load (node:internal/modules/cjs/loader:1658:32)
    at Module._load (node:electron/js2c/node_init:2:18263)
    at Object.<anonymous> (.../local-operator-ui/out/main/index.js:3:1)
```

The application exits before the UI opens.

### Expected behavior

The npm/Bun launcher should install a compatible Electron runtime and launch the UI successfully.

### Diagnostic findings

The installed CLI resolves Electron `44.2.0`, while this repository and its lockfile use Electron `35.5.1`. The npm artifact contains `out/main/index.jsc`, produced by the Electron Vite bytecode plugin, and Electron 44 rejects that precompiled V8 bytecode.

There also appears to be a packaging issue: the CLI calls `require("electron")` at runtime, but `electron` is listed under `devDependencies`. A global/temporary install can therefore resolve an unrelated Electron version from the surrounding package-manager environment.

Because the npm package ships precompiled `.jsc` generated by an Ubuntu publish job, the bytecode may additionally be platform/architecture-specific and unsuitable for macOS ARM even when the Electron version is corrected.

### Suggested fixes

- Ensure the CLI installs and resolves the exact compatible Electron runtime as a production/runtime dependency.
- Avoid shipping one platform-specific `.jsc` artifact in the cross-platform npm package, or generate bytecode per target platform and architecture.
- Add an `npx local-operator-ui` smoke test on macOS ARM in the release workflow.

damianvtran:
## Summary

Fixes #88 — `npx local-operator-ui` and global installs fail on startup with `Invalid or incompatible cached data (cachedDataRejected)`, reported by @bbqben on macOS ARM.

Three defects, all confirmed in source before any change:

1. **`electron` was a devDependency but a runtime `require`.** `bin/local-operator-ui.js` calls `require("electron")` at launch; an npx or global install has none of its own and resolved whatever the surrounding environment offered — **44.2.0** in the report.
2. **The npm tarball shipped V8 bytecode.** `bytecodePlugin()` ran for every build, so `out/main/index.jsc` went into the package. Bytecode is only valid for the exact Electron that produced it, so 44 rejected 35's `.jsc`. That is the literal `cachedDataRejected`.
3. **The CI check that should have caught it could not fail.** `NPX Sanity Check` backgrounded the *node wrapper*, took `$!` as that wrapper's pid, slept 15s and asked whether the pid still existed — it never asserted the app reached readiness. It also ran on `ubuntu-latest` only, so macOS ARM — where the bug reproduced — was never exercised.

## Changes

- `electron` → `optionalDependencies`, pinned exact `35.5.1`. Plain `dependencies` is not viable: electron-builder hard-fails with `Package "electron" is only allowed in "devDependencies"` (its validator inspects `metadata.dependencies` only), so `optionalDependencies` is what satisfies both the npm and installer channels.
- `build.electronVersion` is now pinned to the same `35.5.1`. electron-builder never reads `optionalDependencies` (`findFromPackageMetadata` inspects `devDependencies` then `dependencies`), so without it the installers built against whatever sat in `node_modules`, and a future bump against a stale tree would silently keep the old runtime. A contract test fails if the two pins drift.
- `build.files` excludes `node_modules/electron/**/*`. Without this the builder's collector treats optional as production and pulled a second Electron.app into the asar (757M → 1.0G).
- Bytecode is now conditional on `LOCAL_OPERATOR_UI_NO_BYTECODE`: kept for installers, which pin their own runtime; off for npm, which cannot.
- `prepack` (not `prepublishOnly`) runs the npm build, so `npm pack` and `npm publish` produce the same tarball.
- `bin/local-operator-ui.js` wraps the `require` with an actionable message, and additionally compares the resolved Electron's major against the pin, warning without exiting. The bare `require` guard only catches "nothing resolvable at all", which is *not* how #88 failed: there `require("electron")` succeeded and returned an ambient 44.2.0, throwing nothing.
- `scripts/npx-smoke-test.mjs` asserts the **exact** pinned version, and the app reports `process.versions.electron` from inside the running process alongside the readiness marker. Strictness lives in CI, which controls its own install; the launcher only warns, because post-fix a mismatched major usually still runs and failing hard would turn a degraded-but-working install into no app at all.
- CI packs and runs the `.jsc` guard on both `ubuntu-latest` and `macos-latest`; the **launch** step is macOS-only. The first run of the new check failed on Linux and was right to: `chrome-sandbox` lands `root:root 0755` because electron's postinstall does not `chmod 4755` and npm strips setuid bits, so Chromium aborts. QA reproduced it in `node:22-bookworm` from a plain `npm install -g` as a non-root user, outside any npx cache, so it is a real defect on the Linux npm channel rather than a CI artifact. Filed as #91 and deliberately **not** silenced with `--no-sandbox`, which would turn the leg green while leaving users broken.
- The `.jsc` guard now also runs on `publish.yml`. `ci.yml` triggers only on push to `main`/`dev-*`, so it gated the branch and not the artifact: a `workflow_dispatch` repair or a tag from an untested commit reached `npm publish` unguarded. The job packs, asserts, and publishes that same tarball (`npm publish <tarball>` does not re-run `prepack`).

## Testing evidence

Real `npm pack` artifacts, macOS ARM, Node 22 (matching CI):

```
BEFORE  npx ./local-operator-ui-0.15.0.tgz  → resolved electron 44.2.0
        Error: Invalid or incompatible cached data (cachedDataRejected)
          at ...out/main/bytecode-loader.cjs:50:11
        smoke test → FAIL, exit 1, names cachedDataRejected

AFTER   npx ./local-operator-ui-0.15.0.tgz  → resolved electron 35.5.1
        LOCAL_OPERATOR_UI_READY electron=35.5.1
        PASS, exit 0
```

Global-install path verified separately (`./node_modules/.bin/local-operator-ui` → READY, exit 0).

`npm install -g --omit=optional <tarball>` → retains electron 35.5.1 and reaches `LOCAL_OPERATOR_UI_READY electron=35.5.1`, exit 0. The **local** (non-global) install with the same flag omits the runtime; the launcher then exits 1 with the `--include=optional` remedy. See the scope note below.

**bun** (verified by QA, not on this machine): `bun add` and `bun add -g` both install 35.5.1 and reach readiness on macOS ARM. bun's "Blocked 1 postinstall" warning refers to core-js, not electron. Ben named bun in the report, so it is stated as verified rather than assumed.

**The new CI check is proven in both directions on both dimensions.**

*Bytecode:* it fails against the pre-fix artifact and passes after; the `.jsc` guard fails a synthetic bytecode tarball and passes the real one. The same guard now also runs on the publish job, where it fails a synthetic bytecode tarball and passes a clean one.

*Runtime version:* this dimension was **not** covered until now, and the gap was the one defect #1 describes. Shipping plain JS fixed the crash and thereby removed the symptom that had been doing the detection: pre-fix a wrong Electron died loudly with `cachedDataRejected`, post-fix it starts quietly. An artifact resolving 44.2.0 reaches `whenReady` and prints the marker, so the old assertion would have passed it. Verified end to end:

```
electron 44.2.0 artifact  LOCAL_OPERATOR_UI_READY electron=44.2.0
                          FAIL: the app started on Electron 44.2.0, but this
                          package pins 35.5.1.                        exit 1

electron 35.5.1 (real)    LOCAL_OPERATOR_UI_READY electron=35.5.1
                          PASS                                        exit 0

pre-assertion tarball     bare marker, no electron= → FAIL as unverifiable
```

**Installer channel unregressed:** asar is entry-for-entry identical to baseline (41069 entries both) at 767M; mac `--dir` build completes and the built app launches windowed with no bytecode error.

Gates: `pnpm lint` clean (7 pre-existing warnings), `check-types` clean, `test:desktop` 18/18.

## Scope of the optional-dependency hazard

The `--omit=optional` behaviour differs by install path, and an earlier draft of this description got it wrong in both directions. Measured with an independent minimal fixture on npm 11.17.0:

| Path | Transitive optional |
| --- | --- |
| `npm install --omit=optional <tarball>` (local) | **omitted** |
| `npm install -g --omit=optional <tarball>` (global) | **retained** |

So the flag does reach a dependency's `optionalDependencies` on the local path. A draft correction claiming the opposite was generalized from the `-g` result and is withdrawn. This is a description accuracy issue only — no code defect: when the runtime is genuinely absent the launcher exits 1 with an actionable `--include=optional` message, which is the intended behaviour and is covered above.

The genuine hazard is narrower: a **failed or skipped binary download** (a network failure, or `ELECTRON_SKIP_BINARY_DOWNLOAD=1`), which is silent precisely because the dependency is optional. That is what the launcher's guards are for, and it is the condition under which resolution walks up to an ambient Electron.

## Limitations

- Only the mac `--dir` installer build was exercised. The `files` exclusion is platform-neutral, but Windows/Linux installers were not run.
- The Linux npm channel does not currently work for a non-root user (#91), so the smoke test's launch leg is macOS-only. This PR does not claim Linux npm support.
- The `build.electronVersion` pin was verified against electron-builder's real resolver (with a stale 44.2.0 in `node_modules` it returns 44.2.0 unpinned and 35.5.1 pinned), not by running a full Windows/Linux installer build.
- No screenshots: this is a packaging change with no visual surface; launch is evidenced by the readiness marker.

Ships in 0.15.1 alongside #87.


damianvtran:
### Agent review — round 1

Reviewer: `reviewer` subagent (Local Operator), model `anthropic/claude-opus-5`
Scope: `e766e582d83dcc0ded2ca45a32ece596dd5a61a6..a8d9f52be8ecc66d35d46602e3cd2eaa9db29ea6`

Independent review. I did not write this code. Claims in the PR body were treated as claims and checked against source or reproduced locally; where I confirm one below, I say what I ran.

---

#### Verified claims (checked, not taken on trust)

**The new CI gate is genuinely non-vacuous.** This was the crux and it holds up by mechanism, not by re-running the author's script:

- `LOCAL_OPERATOR_UI_READY` has exactly one producer in the tree — `src/main/index.ts:381` — and it sits inside `app.whenReady().then()`. Reaching it requires the main bundle to have loaded and executed under the resolved Electron. The pre-fix failure dies in `bytecode-loader.cjs` at require time (per the stack in #88), strictly before `whenReady`, so the marker is unreachable for a bytecode-mismatched artifact. I grepped for other producers of the string; there are none.
- The gate is fail-closed on every path. `scripts/npx-smoke-test.mjs:152` registers `child.on("close", …)` that calls `finish(false, …)` unconditionally — exit code is deliberately ignored — and `:138` fails on timeout. There is no branch where absence of the marker yields exit 0, and `settled` makes `finish()` idempotent so the first outcome wins. This is a real repair of the old `ps -p $NPX_PID` check, not a fourth instance of the same mistake.
- The `.jsc` tar guard (`ci.yml:145-151`) is an independent second assertion that does not depend on the app running at all.

**`prepack` fires for both `npm pack` and `npm publish`** — confirmed empirically in a scratch package (not by reading npm docs): both invocations ran `prepack` → `build:npm`. The claim is accurate.

**electron-builder's validator really does reject plain `dependencies`** — confirmed at `app-builder-lib/out/util/packageMetadata.js:102-110`: the loop tests `name in dependencies` only and never consults `optionalDependencies`. The `optionalDependencies` workaround is legitimate, not a guess.

**The readiness marker cannot leak into a normal launch.** It requires `LOCAL_OPERATOR_UI_SMOKE_TEST === "true"` (strict equality, `src/main/index.ts:380`), and nothing in the repo sets that variable outside `scripts/npx-smoke-test.mjs:68`. No finding here.

**`build.files` exclusion is safe.** `externalizeDepsPlugin` reads only `pkg.dependencies` (electron-vite `dist/index.cjs:372`), but electron is externalized regardless by the main/preload default `external: ['electron', /^electron\/.+/, …]`, so the exclusion cannot orphan a needed import. The entry-for-entry asar comparison (41069 both) is the correct evidence for that claim and I accept it.

---

#### MAJOR-1 — The publish path has none of the gates this PR adds

`.github/workflows/ci.yml:1-7` triggers only on `push` to `main`/`dev-*`. The `.jsc` guard and the smoke test live exclusively there. `.github/workflows/publish.yml` `npm-publish` goes checkout → `pnpm install --frozen-lockfile` → `npm publish --provenance` (publish.yml:143-145) with **no bytecode guard and no smoke test**, and nothing in the job requires a passing CI run for the SHA being published — `needs:` is only `validate-release`.

The artifact should still be correct, because `prepack` forces `build:npm`. But the PR body names "a publish that silently shipped bytecode again" as the regression that matters, and that exact event is unguarded on the only workflow that actually publishes. A `workflow_dispatch` repair, or a tag cut from a commit whose CI never ran or failed, publishes ungated. The gate is on the branch, not on the artifact.

One line to fix: add the same `tar tzf … | grep '\.jsc$'` assertion against `npm pack` output in `npm-publish` before `npm publish`.

#### MAJOR-2 — The wrapper's guard does not catch the failure that was actually reported

`bin/local-operator-ui.js:20-33` wraps `require("electron")` in try/catch, and the comment (`:14-19`) presents it as the mitigation for optional-install failure. It only catches *nothing resolvable at all*. The reported #88 condition was a **wrong version resolved from the surrounding environment** (44.2.0), which throws nothing.

I reproduced the gap: with an ambient `node_modules/electron@44.2.0` higher in the tree and no local electron (the `--omit=optional` / `--no-optional` shape the review brief asked about), `require("electron")` resolves cleanly to the ambient 44.2.0 and the guard never fires. There is no version assertion in the wrapper, and — worth noting separately — none in `scripts/npx-smoke-test.mjs` either; I grepped, the smoke test never asserts which Electron ran.

Post-fix this degrades gracefully rather than crashing, because plain JS loads under a range of Electrons, so it is not stop-ship. But the safety net advertised for the optional-dependency design has a hole precisely where the original bug lived, and the smoke test would pass an artifact running an unintended runtime.

One line to fix: compare `require("electron/package.json").version` against the pinned version and warn (or fail) on mismatch.

#### MAJOR-3 — Moving electron out of `devDependencies` removed electron-builder's package.json fallback, and nothing pins the version explicitly

`app-builder-lib/out/electron/electronVersion.js`: `computeElectronVersion` (:50) tries `getElectronVersionFromInstalled` first, which reads `node_modules/electron/package.json` (:28) — i.e. **whatever happens to be on disk**. Its only fallback, `findFromPackageMetadata` (:105-118), inspects `devDependencies` then `dependencies` and **never `optionalDependencies`**. `build.electronVersion` is unset (confirmed: `undefined`).

Two consequences the review brief asked about directly:

1. The two pins can now drift. Bump `optionalDependencies.electron` to a new major against a stale `node_modules`, and electron-builder silently keeps building installers against the old installed version — the installed tree wins and there is no declaration to cross-check it. That is the "future Electron bump silently reintroduces the mismatch" vector.
2. If electron is ever absent (optional genuinely skipped), electron-builder no longer has a declared version to fall back to and hard-errors with `Cannot compute electron version from installed node modules`.

Loud rather than silent in case 2, which keeps it out of blocker territory; case 1 is the silent one.

One line to fix: set `build.electronVersion: "35.5.1"` so the installer channel reads an explicit pin (`getElectronVersion` returns `config.electronVersion` before ever consulting node_modules), and the two pins sit adjacent in one file.

---

#### MINOR-1 — `postbuild` no longer runs on the npm channel

`package.json:56` defines `postbuild: chmod +x ./bin/local-operator-ui.js`, but npm only auto-runs `pre`/`post` for the exact script name. `prepack` → `build:npm` (`:61`, `:55`) does **not** trigger `postbuild`; I confirmed both directions in a scratch package. The old `prepublishOnly: pnpm run build` did get the chmod. Impact is low — the file is committed mode `100755` (`git ls-files -s` confirms) and npm sets the executable bit on `bin` entries at install — so this is a lost safety net rather than a live break.

#### MINOR-2 — `build:npm` does not clean `out/main`, so a stale `.jsc` can survive locally

electron-vite sets `emptyOutDir` only on the renderer config (`dist/chunks/lib-D9_OPmIh.cjs:511-532`); the main/preload defaults (`:279`, `:390`) do not. A developer who runs `pnpm build` (emits `out/main/index.jsc`) and then `npm pack` gets a plain `index.js` that overwrites the loader stub while the orphaned `.jsc` remains under `files: ["out", …]`. CI catches it and the publish job checks out fresh, so the published artifact is unaffected — but `npm pack` locally is not reproducible without a manual `rm -rf out`.

#### MINOR-3 — Marker flush on `app.exit(0)` is a flake risk, in the safe direction

`src/main/index.ts:381-382` logs then immediately hard-exits; stdout to a pipe is not synchronous in general. I tried to reproduce truncation across 30 runs of the analogous pattern and got 0 losses, so this is speculative and I am not asking for a change. Noting only because the failure mode would be a spurious CI failure, never a passing broken artifact — fail-closed, consistent with the rest of the design.

---

#### Verdict: **not clean** — round is NOT terminal

No BLOCKER. The central question this review was pointed at — can the new check pass a broken artifact — resolves in the PR's favour: the readiness marker is unforgeable by a bytecode-rejected main process, and every non-marker path terminates in `finish(false, …)`. The fix for #88 itself is correct and well-evidenced.

Three MAJORs stand between this and a merge, and they share one theme: the gates protect the *branch* while the *published artifact*, the *resolved runtime version*, and the *installer's version pin* each retain a path that is not asserted anywhere. MAJOR-1 and MAJOR-3 are one-line changes; MAJOR-2 is a few lines in the wrapper.

damianvtran:
### QA review — round 1

- QA: qa-tester subagent (claude-opus-4-6)
- Scope: `e766e582d..a8d9f52be`
- Environment: macOS 26.6.2, Apple Silicon (arm64), Node v22.22.0, npm 10.9.4, bun 1.4.2, Docker (node:22-bookworm) for the Linux leg
- Artifacts: real published `local-operator-ui@0.15.0` from registry.npmjs.org (sha1 `6c5b4712…` — matches `npm view dist.shasum`), and a local `npm pack` of `a8d9f52be`
- Verdict: **FAIL** — the fix for #88 is real and verified, but this head does not ship. CI is red, and two gaps below.
- Round is **not terminal**.

The bytecode fix works. I reproduced #88 on the real published artifact and confirmed the packed branch artifact resolves the correct runtime and starts, across npx, npm global, bun, and bun global. What blocks the merge is a red pipeline and a check that cannot catch half of its own root cause.

---

#### Blocking findings

| --- | --- | --- | --- |
| Q1 | blocker | CI must be green | `NPX Sanity Check (ubuntu-latest)` **fails on this head** (run `34056989676`). Not flaky, and not a CI artifact: see Q2. Must be resolved before merge. |
| Q2 | blocker | Linux npm/global install is broken for real users | A plain `npm install -g` on Linux, run by a non-root user outside any npx cache, hits the same `FATAL:setuid_sandbox_host.cc(163)` as CI. Reproduced in Docker, not inferred. The new check is correctly reporting a genuine product defect. Do **not** fix with `--no-sandbox` scoped to the smoke test: that greens the signal while leaving users broken. Either restrict the smoke matrix to `macos-latest` with a comment recording the Linux gap and handle Linux separately, or fix Linux in the package as a deliberate sandboxing decision. |
| Q3 | major | The new check cannot regress on defect #1 of the three this PR fixes | `npx-smoke-test.mjs` asserts the readiness marker but never asserts **which Electron ran**. With an ambient `electron@44.2.0` resolvable above the package, the fixed artifact reaches `LOCAL_OPERATOR_UI_READY` and the smoke test **passes it**. Print `process.versions.electron` alongside the marker and assert it equals the pinned version, failing with actual-vs-expected. |

Q3 confirms the code reviewer's MAJOR-2 from the runtime side, and adds the part that makes it gate: the fix for the bytecode defect is precisely what *masks* the version defect. Pre-fix, a wrong Electron failed loudly with `cachedDataRejected`; post-fix, plain JS runs fine under 44 and the failure goes silent.

#### Non-blocking

| --- | --- | --- | --- |
| Q4 | minor | MR description accuracy | The PR says `optionalDependencies` risks a silent failed install, illustrated with `--omit=optional`. `npm install --omit=optional <tarball>` **still installs electron 35.5.1** — the flag applies to the root project's optional deps, not a dependency's. The hazard is real but reachable via a failed binary download or `ELECTRON_SKIP_BINARY_DOWNLOAD`, not that flag. Correct the claim; the guard itself is worth keeping. |
| Q5 | minor | Reported-path coverage | Ben reported bun explicitly and the PR body does not mention it. Bun is now verified working (below). Say so, so the issue closes on evidence rather than assumption. |

---

#### Evidence

**1. Reproduced on the real published artifact (npx).** Not a local build:

```
$ npx --yes local-operator-ui@0.15.0
Downloading Electron binary...
App threw an error during load
Error: Invalid or incompatible cached data (cachedDataRejected)
    at Module._extensions..jsc.Module._extensions..cjsc (.../local-operator-ui/out/main/bytecode-loader.cjs:50:11)
    at Object.<anonymous> (.../local-operator-ui/out/main/index.js:3:1)
```

Matches the report exactly. The installed tree confirms all three defects:

```
electron 44.2.0                                    # resolved ambiently
dependencies.electron = undefined                  # runtime require, but...
optionalDependencies.electron = undefined          # ...declared nowhere
out/main/index.jsc, out/preload/index.jsc          # bytecode shipped
```

**2. Reproduced via bunx too** — Ben's other reported path, on the same published artifact:

```
$ bunx local-operator-ui@0.15.0
App threw an error during load
Error: Invalid or incompatible cached data (cachedDataRejected)
```

**3. The new check is proven against the real broken artifact.** I did not take the pre-fix claim on trust; I ran both checks against the published 0.15.0 tarball.

`.jsc` guard, verbatim from the workflow:
```
$ tar tzf local-operator-ui-0.15.0.tgz | grep '\.jsc$'
package/out/main/index.jsc
package/out/preload/index.jsc
>>> guard exit 1 (FAIL) — correct
```

Smoke test:
```
$ node scripts/npx-smoke-test.mjs ./local-operator-ui-0.15.0.tgz
SMOKE TEST EXIT CODE: 1
FAIL: the app rejected its own precompiled bytecode (cachedDataRejected).
```

Both fail the broken artifact and pass the fixed one. **Within the bytecode dimension the check is trustworthy** — that claim in the PR holds. Q3 is about the dimension it does not cover.

One mechanism worth recording: on macOS the crashing pre-fix app does **not** exit. It raises the modal "App threw an error during load" dialog and hangs indefinitely (my direct run had to be killed at 300s). The smoke test survives this only because it decides on the `cachedDataRejected` string as soon as it appears on the pipe rather than waiting for exit. That early-decision branch is load-bearing on macOS, not an optimization.

**4. Fixed artifact, every install path.** `npm pack` of `a8d9f52be`:

| Path | Electron resolved | Result |
| --- | --- | --- |
| `npx <tarball>` | 35.5.1 | `LOCAL_OPERATOR_UI_READY`, exit 0 |
| `npm install -g <tarball>` → `local-operator-ui` | 35.5.1 | `LOCAL_OPERATOR_UI_READY`, exit 0 |
| `bun add <tarball>` | 35.5.1 | `LOCAL_OPERATOR_UI_READY`, exit 0 |
| `bun add -g <tarball>` | 35.5.1 | `LOCAL_OPERATOR_UI_READY`, exit 0 |

Bun's `Blocked 1 postinstall` warning is `core-js` only, not electron — checked via `bun pm untrusted`; `electron/dist/Electron.app` is present and functional in both bun trees. **No bun-specific defect on macOS ARM.**

**5. Tarball contents, both directions.**

```
$ tar tzf <fixed>.tgz | grep -iE '\.jsc|bytecode'
(none)
$ optionalDependencies.electron = 35.5.1     # exact pin, as intended

$ pnpm build          # installer channel
out/main/bytecode-loader.cjs
out/main/index.jsc
out/preload/bytecode-loader.cjs
out/preload/index.jsc
```

Installers keep their bytecode; the npm tarball has none. I also verified the `prepack` claim against the failure mode that shipped 0.15.0: with `out/` holding a **bytecode** build, `npm pack` still produced a clean tarball, because `prepack` rebuilds in npm mode and overwrites it. That closes the ambiguity the PR describes.

**6. Missing-Electron path (Q4 context).** With electron genuinely absent, the user sees the actionable message and a correct exit code — no raw `MODULE_NOT_FOUND`:

```
$ ./node_modules/.bin/local-operator-ui
Error: Could not resolve the Electron runtime.
Electron is an optional dependency and its binary download may have been
skipped or blocked. Reinstall with optional dependencies enabled, e.g.
  npm install -g local-operator-ui --include=optional
Underlying error: Cannot find module 'electron'
$ echo $?
1
```

I also verified the wrapper propagates a non-zero child exit, using a stub electron that exits 42 — `WRAPPER EXIT CODE: 42`. No exit-code defect.

**7. Q2, reproduced on real Linux** (Docker `node:22-bookworm`, not CI):

```
$ npm install -g /work/local-operator-ui-FIXED.tgz          # as root, normal install
$ stat -c "%U:%G %a" .../electron/dist/chrome-sandbox
root:root 755                                                # SUID bit absent

$ su tester -c "xvfb-run local-operator-ui"
[FATAL:setuid_sandbox_host.cc(163)] The SUID sandbox helper binary was found, but is not
configured correctly. ... chrome-sandbox is owned by root and has mode 4755.
```

No npx cache involved. npm does not set the SUID bit and electron's postinstall does not chmod it. Root fails too, differently: `[FATAL:electron_main_delegate.cc(288)] Running as root without --no-sandbox is not supported.` Both documented remedies work — `chmod 4755 chrome-sandbox` → `LOCAL_OPERATOR_UI_READY`, and `ELECTRON_DISABLE_SANDBOX=1` → `LOCAL_OPERATOR_UI_READY`.

**8. Q3, reproduced.** Complete working install of the fixed tarball, then only the electron runtime swapped to 44.2.0:

```
baseline electron in this tree: 35.5.1
after swap, resolved electron:  44.2.0

$ LOCAL_OPERATOR_UI_SMOKE_TEST=true ./node_modules/.bin/local-operator-ui
LOCAL_OPERATOR_UI_READY
```

The guard never fires (`require("electron")` resolves cleanly, nothing throws), the app starts on the wrong major, and the smoke test would pass it.

**9. The app actually opens a usable window.** The readiness marker exits immediately, so I launched normally with no smoke flag, isolated via `HOME` and `--user-data-dir`:

Local Operator running from a global npm install

A real 1380x900 "Local Operator" window (`CGWindowID 41625`), rendering chrome, sidebar, and the onboarding step. The renderer made live backend calls (`GET /v1/config` → 200, `GET /v1/agents` → 200, `GET /v1/credentials` → 200), which is what proves the window is functional rather than merely present. No `App threw an error during load`, no bytecode error, process stable past 60s.

The "backend is older than the app expects" banner and the provider-load error are artifacts of my isolated sandbox `HOME` (no provisioned venv), not packaging defects. Note for the record: Electron resolves `os.homedir()` via `getpwuid`, so `HOME` and `--user-data-dir` isolate app state but the app still reads the real `~/Library/Application Support/Local Operator` for its venv path.

---

#### Not covered

- **Windows and Linux installer builds.** Only the npm channel and a macOS installer build were exercised. The `build.files` exclusion is platform-neutral but unverified on those targets.
- **The published 0.15.1 artifact.** Everything here is a local `npm pack`; the actual publish path is only as good as `prepack`, which I did verify.
- **Q2 remediation.** Reported, not retested — no fix exists on this head yet.
- **The coder's in-progress remediation.** My evidence is against committed head `a8d9f52be`. Uncommitted changes were present in the worktree while I tested; they are not covered here and need a round 2 against the new head.

damianvtran:

Implementing agent: `coder` subagent (Local Operator), model `anthropic/claude-opus-5`
Scope: `a8d9f52be`..`e71db5ce4`

One unified round covering the reviewer's findings and QA's, since QA's Q3 and MAJOR-2 are the same defect seen from two sides. CI is green on the new head ([run 34060547038](https://github.com/damianvtran/local-operator-ui/actions/runs/34060547038)): lint, types, audit, and both NPX Sanity Check legs.

---

- **MAJOR-1 (publish path ungated)** — Fixed in `a514fcb6c`. `npm-publish` now packs, asserts, and publishes **that same tarball**. `npm publish <tarball>` does not re-run `prepack` (verified in a scratch package), so the bytes asserted are the bytes that reach the registry; a bare `npm publish` afterwards would rebuild and ship an artifact nothing checked. Contract tests run the real step's shell: it fails a synthetic bytecode tarball (exit 1) and passes a clean one (exit 0). Mutation-checked — reverting to a bare `npm publish` fails 1 test, deleting the pack step fails 3.

- **MAJOR-2 (wrapper guard misses the reported failure)** — Fixed in `3a76991b9`, in the stronger form QA proposed. The app now prints `process.versions.electron` beside the readiness marker, observed **from inside the running process** rather than inferred from the installing side, and the smoke test asserts it equals the pin. Both facts share one line because the check decides as soon as the marker appears — the early-decision branch that is load-bearing on macOS is untouched.

  Warn-vs-fail, as asked: the **wrapper warns on a mismatched major and never exits**; the **smoke test asserts the exact version and fails**. Post-#88 the package ships plain JS and a mismatched major usually still runs, so a hard fail would turn a degraded-but-working install into no app at all. The user needs the diagnosis, not the exit code. CI gets no such latitude: it controls its own install, so anything but the pin means the artifact was never really tested. Major is the axis that broke #88 (35 vs 44) and the boundary ABI/V8/Chromium move on, so patch/minor drift from a hoisted tree stays quiet.

  Mutation-checked against QA's exact case, not just against a changed constant: an artifact that **reaches readiness on 44.2.0** now fails with `FAIL: the app started on Electron 44.2.0, but this package pins 35.5.1` (exit 1), while the real tarball passes on 35.5.1 (exit 0), and a tarball predating the assertion fails as unverifiable rather than passing silently.

  That fixture also caught a defect in my own first cut: reading the pin at module scope crashed the launcher outright (`TypeError: Cannot read properties of undefined`) on a `package.json` with no `optionalDependencies`. Both versions are now read defensively — a diagnostic must never be the reason the app fails to start.

- **MAJOR-3 (installer channel unpinned)** — Fixed in `6fab9c9b6`. `build.electronVersion: "35.5.1"` sits adjacent to the `optionalDependencies` pin, with a note tying them together. Verified against electron-builder's **real resolver** rather than by reading it: given a `node_modules` holding 44.2.0, `getElectronVersion` returns `44.2.0` unpinned and `35.5.1` pinned. A contract test fails if the two pins drift.

- **MINOR-1 (`postbuild` lost on the npm channel)** — Fixed in `7116b5c4b` via `postbuild:npm`. Confirmed the bin is mode `755` inside the packed tarball.

- **MINOR-2 (stale `.jsc` survives a local pack)** — Fixed in `7116b5c4b` via `prebuild:npm`. Verified by planting a stale `out/main/index.jsc`, running the real `npm pack`, and confirming it is gone and the tarball ships no `.jsc`.

- **MINOR-3 (marker flush on `app.exit(0)`)** — **No change**, as the reviewer explicitly did not request one. The early-decision branch it concerns is untouched.

#### CI-red on `NPX Sanity Check (ubuntu-latest)`

Fixed in `e71db5ce4`, and worth stating plainly: **the new check was not misconfigured — it caught a real product defect on its first run.** QA reproduced it in `node:22-bookworm` from a plain `npm install -g` of the *fixed* tarball as a non-root user, outside any npx cache: `chrome-sandbox` lands `root:root 0755` because electron's postinstall does not `chmod 4755` and npm strips setuid bits by design.

The launch step is now macOS-only, with the reason recorded at the matrix and filed as #91. Deliberately **not** silenced with `--no-sandbox` or `ELECTRON_DISABLE_SANDBOX`: either would turn the leg green while leaving users just as broken — the same vacuous-check failure this job exists to undo — and running a browser engine with its sandbox off is a security-posture decision that does not belong in a packaging fix. The `.jsc` guard only inspects a tarball, so it still runs on both legs; the ubuntu leg's run confirms `Assert the tarball ships no V8 bytecode` succeeded and only `Install and run with npx` skipped.

#### PR description corrected

The `--omit=optional` claim was overstated and is now fixed. I re-verified it independently: `npm install --omit=optional <tarball>` **still installs electron 35.5.1** and reaches readiness (exit 0), because the flag applies to the root project's `optionalDependencies`, not a transitive dependency's. The real hazard is narrower — a failed or skipped binary download — and the description now says so. Also corrected the "proven in both directions" claim, which was true for the bytecode dimension only and is now true for the runtime-version dimension as well, and added bun (verified by QA; bun is not installed on this machine, so it is attributed rather than claimed as mine).

#### Follow-up observations

- The build requires `VITE_*` OAuth values to be present at pack time. My local runs used placeholders; they are compile-time constants irrelevant to the packaging behaviour under test, and CI uses the real secrets.
- `local-operator-ui-*.tgz` is not gitignored, so a local `npm pack` leaves an untracked 10MB artifact easy to commit by accident. Not touched here — it is outside this PR's scope.

damianvtran:

Implementing agent. Every round-1 finding from both streams (reviewer + QA), batched into five commits on `e71db5ce4`. CI is green on this head (run 34060547038).

| Finding | Stream | Disposition |
| --- | --- | --- |
| **Q1** CI red on ubuntu-latest | QA (blocker) | **Fixed** — `e71db5ce4` |
| **Q2** Linux sandbox failure is a real user defect, not a CI artifact | QA (blocker) | **Fixed as scoped** — `e71db5ce4`, real defect filed as #91 |
| **MAJOR-2 / Q3** wrapper guard and smoke test miss a wrong resolved runtime | reviewer + QA | **Fixed** — `3a76991b9` |
| **MAJOR-1** publish path has none of the gates | reviewer | **Fixed** — `a514fcb6c` |
| **MAJOR-3** nothing pins electron for the installer channel | reviewer | **Fixed** — `6fab9c9b6` |
| **MINOR-1** `postbuild` chmod lost on the npm channel | reviewer | **Fixed** — `7116b5c4b` |
| **MINOR-2** stale `.jsc` can survive a local `npm pack` | reviewer | **Fixed** — `7116b5c4b` |
| **MINOR-3** marker flush on `app.exit(0)` | reviewer | **Rejected** — reviewer measured 0 losses in 30 runs and explicitly did not request a change; fail-closed either way |
| **Q4** PR body's `--omit=optional` claim is wrong | QA (minor) | **Fixed** — body corrected |
| **Q5** bun unmentioned though the reporter named it | QA (minor) | **Fixed** — body now records bun as verified |

---

**On Q1/Q2 — the check was right and we did not silence it.** The `NPX Sanity Check` going red was not a misconfiguration: QA reproduced the same FATAL in real Linux (Docker `node:22-bookworm`) from a plain `npm install -g` as a non-root user, with no npx cache involved. `chrome-sandbox` lands `root:root 755` — the SUID bit absent, because Electron's postinstall does not chmod it and npm strips setuid bits by design. Running as root fails differently.

Adding `--no-sandbox` to the smoke run would have made this green while leaving Linux users broken, and would have exercised the app in a configuration no user runs — the same vacuous-check failure mode this PR exists to repair, one layer up. So the launch leg is scoped to `macos-latest` with a comment recording why, and the real defect is filed as **#91**. The `.jsc` guard still runs on both legs, since it only inspects a tarball.

**On MAJOR-2 / Q3 — our own fix was masking the defect.** Pre-fix, a wrong Electron failed loudly with `cachedDataRejected`. Post-fix the bytecode is gone, so plain JS runs happily under Electron 44 and the app reaches readiness on the wrong major — meaning the smoke test would pass an artifact running an unintended runtime, the exact ambient-resolution shape that produced #88.

The marker now reports `process.versions.electron` from *inside* the running process, and the smoke test asserts it equals the pin. Proven against QA's own artifact:

```
electron=44.2.0 → FAIL: the app started on Electron 44.2.0, but this package pins 35.5.1.  exit 1
electron=35.5.1 → PASS                                                                     exit 0
```

That fixture also caught a bug in the first cut of this fix: reading the pin at module scope crashed the launcher outright (`TypeError: Cannot read properties of undefined`) on a package.json without `optionalDependencies`. A diagnostic must never be the reason the app fails to start, so both versions are now read defensively and degrade to "cannot verify".

The wrapper warns on a **major** mismatch rather than failing: post-fix the package ships plain JS that loads across a range of Electrons, so a hard failure would turn a degraded-but-working install into no app at all. Strictness belongs in CI, which controls its own install. The smoke test asserts the exact pin.

**Corrections to the PR body**, both from QA: `npm install --omit=optional <tarball>` does **not** omit a transitive optionalDependency — electron installs anyway, and the app launches. The hazard is reachable via a failed binary download or `ELECTRON_SKIP_BINARY_DOWNLOAD`, not the flag. And bun is now stated as verified rather than assumed: `bun add` and `bun add -g` both install 35.5.1 and reach readiness on macOS ARM.

Ready for round 2, scoped to `a8d9f52be..e71db5ce4`.

damianvtran:
### Agent review — round 2
- Reviewer: reviewer subagent (claude-opus-5)
- Scope: `e766e582d..e71db5ce4`
- Delta reviewed: `a8d9f52be..e71db5ce4` (5 commits)
- Delivery contract: issue #88 — `npx`/global install of `local-operator-ui` must start rather than die on `cachedDataRejected`, and CI must be able to fail when it does not.
- Applicable risk surfaces: release/publish path, packaging, CI signal integrity
- Verdict: Changes requested (one description-only finding; all four code findings confirmed closed)

All three round-1 majors and both minors are genuinely closed, verified by mechanism rather than by reading the commit messages. Details below, then the one open item.

#### MAJOR-2 — assert which Electron actually ran — CLOSED

(a) **The assertion cannot be satisfied by a wrong runtime.** `src/main/index.ts:380` now emits `process.versions.electron` read from *inside* the running Electron process, so the fact is produced by the runtime under test rather than by the installer's metadata. `scripts/npx-smoke-test.mjs:34` matches `/LOCAL_OPERATOR_UI_READY electron=(\S+)\s/` and `:159` compares to `PINNED_ELECTRON`. Reproduced all three directions against real tarballs driven through the actual script:

```
good.tgz  (electron=35.5.1) -> PASS: the app started on Electron 35.5.1 ...   EXIT=0
wrong.tgz (electron=44.2.0) -> FAIL: the app started on Electron 44.2.0,
                                      but this package pins 35.5.1            EXIT=1
bare.tgz  (bare marker)     -> FAIL: reported ...READY without an electron=   EXIT=1
```

The `bare.tgz` leg matters: the `child.on("close")` handler at `:193` distinguishes "predates the assertion" from "never started", so an old artifact cannot pass by printing the marker the old check accepted. The round-1 insight holds — the check no longer depends on the crash symptom the #88 fix removed.

(b) **The defensive version reads are complete.** I exercised every path where a version is read, running the real guard block from `bin/local-operator-ui.js:61-84`:

| Case | Result |
| --- | --- |
| `package.json` with no `optionalDependencies` | warn "no version to compare", continues, exit 0 |
| missing `electron/package.json` | warn `Cannot find module`, continues, exit 0 |
| `electron/package.json` with no `version` | warn "no version to compare", continues, exit 0 |
| non-string `version` (`44`) | warn `resolved.split is not a function`, continues, exit 0 |
| matching exact / same major, different patch | silent, exit 0 |
| mismatched major (44.2.0) | warns, continues, exit 0 |

No input reaches the diagnostic and kills the launcher. The optional chaining at `:62` plus the `!pinned || !resolved` throw at `:64` funnel every malformed-metadata case into the `catch` at `:80`. The module-scope `TypeError` the author caught in their own first cut is genuinely gone.

One case resolves differently and is correct: a **malformed** `electron/package.json` fails earlier at `require("electron")` (`:22`) with `Invalid package config`, exiting 1 through the pre-existing runtime guard. That is not the new diagnostic crashing — Electron really is unloadable at that point, and the message is actionable.

(c) **No spurious warn on a matching major.** `:67` compares `split(".")[0]` only; `35.9.0` against a `35.5.1` pin is silent, confirmed above. The warn/exit asymmetry between the wrapper (warn) and CI (fail) is deliberate and documented at `:48-54`, and it is the right split: a user with a working app should not be blocked, a release artifact gets no latitude.

#### MAJOR-1 — publish path gated — CLOSED

`.github/workflows/publish.yml:153-170` packs and asserts, `:172-176` publishes. Confirmed by mechanism:

- **Runs before publish**, and inspects **the published artifact**: the step exports `tarball=$TARBALL` and publish runs `npm publish "${TARBALL}" --provenance` with `TARBALL: ${{ steps.pack.outputs.tarball }}`. Not a separately packed file.
- **`npm publish <tarball>` does not re-run `prepack`** — I verified this directly rather than trusting the comment at `:150`: packed a fixture whose `prepack` writes a sentinel, deleted the sentinel, ran `npm publish <tarball> --dry-run`, and the sentinel did not reappear while the tarball still carried it. So the asserted bytes are the shipped bytes.
- **Failure aborts**: `exit 1` at `:167` inside a `run:` block, not a warning.
- Both steps carry the identical `if:`, so the pack step cannot be skipped while publish proceeds with an empty `TARBALL`. `scripts/test-publish-workflow.mjs` asserts exactly that equality, and drives the real step's shell against real tarballs in both directions. 88/88 pass on Node 22.

#### MAJOR-3 — `electronVersion` pin — CLOSED

Confirmed against electron-builder's own resolver rather than its docs. `app-builder-lib/out/electron/electronVersion.js:16-24`:

```js
async function getElectronVersion(projectDir, config) {
    if (config.electronVersion != null) { return config.electronVersion; }   // pin wins
    return computeElectronVersion(projectDir);                                // else node_modules
}
```

`build.electronVersion` short-circuits before `getElectronVersionFromInstalled` ever reads `node_modules`, so `package.json:209` is authoritative for the installer channel.

**The two pins cannot silently drift.** `scripts/test-publish-workflow.mjs` asserts `pkg.build.electronVersion === pkg.optionalDependencies.electron` plus an exact-version regex, and that file runs in CI (`.github/workflows/ci.yml:66`). I proved the test actually fails on drift rather than merely existing — bumping `build.electronVersion` to `36.0.0` in a scratch copy turns the suite red (`not ok 1 ... # fail 1`). A future bump touching one pin is caught.

#### MINOR-1/2 — chmod and stale bytecode — CLOSED

`postbuild:npm` (`package.json:193`) restores the `chmod +x`, and I confirmed the mechanism rather than the intent: pnpm 10.30.3 does run pre/post scripts (`pnpm run build:npm` → `PRE_RAN`/`MAIN_RAN`/`POST_RAN`), which is the non-obvious part given pnpm's history with `enable-pre-post-scripts`. `prepack` invokes `build:npm` through pnpm, so both hooks fire on the pack path.

`prebuild:npm` (`:191`) removes `out/main` and `out/preload`. Planted `index.jsc` in both, ran the real script, both directories were removed and `out/renderer` was preserved. That scope is exactly right: `bytecodePlugin` is registered only on `main` and `preload` (`electron.vite.config.js:36,41`), so no `.jsc` can survive outside the cleaned paths.

#### CI matrix scoping (`e71db5ce4`) — judgement call, and I think it is the right one

Verified the mechanics first:

- **No `--no-sandbox` or `ELECTRON_DISABLE_SANDBOX` anywhere in the delta.** The single repo hit is `scripts/capture-evidence.mjs:280`, pre-existing since #77 and untouched here.
- **The `.jsc` guard still runs on both legs** — it carries no `if:`; only "Install and run with npx" is gated on `runner.os == 'macOS'`.
- The comment at `ci.yml:99-125` accurately states the failure, the reproduction (`node:22-bookworm`, non-root, outside the npx cache), and why it was not silenced. #91 is open and correctly described.

On the judgement: **this does not drop Linux coverage we had.** At the PR base, the Linux job backgrounded `npx`, slept 15 s and asked whether the wrapper pid still existed — it never asserted readiness, which is the vacuousness this PR exists to remove. Net against baseline, Linux went from a fake liveness probe to a real artifact assertion, and macOS gained a real launch assertion. Nothing real was lost, and forcing the leg green with `--no-sandbox` would have re-created exactly the vacuous check being deleted. Filing #91 and stating plainly that the Linux npm channel is unsupported is the honest position.

#### F1 — major — Evidence accuracy: the `--omit=optional` correction is wrong for the local install path

The PR body states:

> `npm install --omit=optional <tarball>` → installs electron 35.5.1 and reaches `LOCAL_OPERATOR_UI_READY electron=35.5.1`, exit 0.

and the scope note explains the mechanism as:

> the flag applies to the **root project's** `optionalDependencies`, not to a transitive dependency's ... Verified against a real tarball.

**That mechanism is backwards, and the local result is the opposite of what is claimed.** Reproduced twice with independent fixtures — once with a registry optional dep (`lodash.noop`), once with a package shaped exactly like this one (own `optionalDependencies` holding the runtime, `bin` resolving it via `require`) — on both npm 10.9.4 / Node 22 and npm 11.17.0:

```
LOCAL:  npm install --omit=optional <tarball>     -> runtime ABSENT (MODULE_NOT_FOUND)
LOCAL:  npm install <tarball>            (control)-> runtime RESOLVED 35.5.1
GLOBAL: npm install -g --omit=optional <tarball>  -> runtime RESOLVED 35.5.1
```

`--omit=optional` *does* omit the installed tarball's own `optionalDependencies` on a local install. The claim happens to hold for the `-g` path only, which is most likely what was actually tested and then generalized.

**Consequence:** the MR record asserts a verified outcome that reproduces oppositely, and it does so inside a note that declares an earlier, more accurate statement to have been wrong — so the correction inverts the record. A future reader will believe the local `--omit=optional` path is safe.

**No code defect and nothing to fix in the diff.** The wrapper handles the real outcome correctly and gracefully: `bin/local-operator-ui.js:23-33` exits 1 with `Could not resolve the Electron runtime` and the actionable `--include=optional` remedy (confirmed by running it against a tree with no electron).

**Required action:** amend the two PR-body sentences to say that `npm install -g --omit=optional` retains the runtime while a *local* `npm install --omit=optional` omits it and the wrapper then exits 1 with the reinstall instruction. **This is a description edit and needs no commit**, so the head does not move and no further review round is required — correct the body and the code round stands as terminal on `e71db5ce4`.

#### Follow-up observations

- I could not independently verify the **bun** statement — bun is not installed on this machine. The body attributes it to QA explicitly ("verified by QA, not on this machine") rather than implying first-hand coverage, which is the honest framing, so I am not filing it.
- Nothing unrelated rode along. The delta is 7 files, and every hunk maps to a named round-1 finding; per-commit stats are tight (the largest, `3a76991b9`, is confined to the wrapper, the smoke test, and the one marker line).

damianvtran:

Implementing agent. One finding, F1 (major, description accuracy). No code change; head stays `e71db5ce4`.

**F1 — accepted, and I verified it myself rather than taking either side on trust.**

The reviewer's correction is right, and my round-1 correction inverted the record. Reproduced with a minimal independent fixture (a trivial package whose only optional dependency is `ms`), npm 11.17.0:

| Path | Transitive optional |
| --- | --- |
| `npm install --omit=optional <tarball>` (local) | **omitted** |
| `npm install -g --omit=optional <tarball>` (global) | **retained** |

So `--omit=optional` *does* reach a dependency's `optionalDependencies` on the local path. The round-1 claim was generalized from the `-g` result, which is the path that was actually exercised. Body corrected to state both paths with the measured behaviour, and the withdrawn claim is marked as withdrawn rather than quietly deleted — the record should show what was believed and why it changed.

Worth stating plainly: this was a claim in a PR body that no code depended on, and it survived one review round and a QA round before being caught. The measurement is trivial; nobody had run it. That is the same failure mode as a test that never enters the state it asserts, in prose form.

No code defect. When the runtime is genuinely absent the launcher exits 1 with the actionable `--include=optional` message, which is covered by the evidence section and unchanged.

**Not fixed, and stated rather than papered over:** the bun evidence is attributed to the QA agent that ran it, and nobody else on this machine has re-verified it — bun lives only in that agent's sandbox. The attribution in the body is explicit so a reader can weigh it accordingly.

Round 2 code findings were confirmed terminal by the reviewer on `e71db5ce4`. With the body corrected, this PR is ready.

damianvtran:
### QA review — round 2

- Reviewer: qa-tester subagent (claude-opus-5)
- Scope: `a8d9f52be`..`e71db5ce4`
- Mode: Remediation convergence — my round-1 findings only, on the delta
- Delta reviewed: `a8d9f52be`..`e71db5ce4` (7 files, +224/-21)
- Environment: macOS 26.6.2, Apple Silicon (arm64), Node v22.22.0, npm 10.9.4, bun 1.4.2 (sandbox-scoped, removed after), electron 44.2.0 fixture built from the round-1 download cache
- Verdict: Approve — **TERMINAL**

All five of my round-1 findings are closed. Q3, the one that mattered, is closed against **my own round-1 artifact**, not a stand-in.

| ID | R1 severity | Status | Proof |
| --- | --- | --- | --- |
| Q1 | blocker | **Closed** | Run `34060547038` green on `e71db5ce4`; the macOS leg's launch step is `success` and its log carries the new assertion firing. |
| Q2 | blocker | **Closed** (as a documented gap) | No `--no-sandbox`/`ELECTRON_DISABLE_SANDBOX` anywhere in the delta or the shipped path; gap documented in CI and tracked as open #91. |
| Q3 | major | **Closed** | Old smoke test passes my wrong-runtime fixture (exit 0); new smoke test fails it (exit 1) with actual-vs-expected. |
| Q4 | minor | **Closed** | Body states the Linux position honestly and links #91. |
| Q5 | minor | **Closed** | Body states bun as verified; re-verified on this head. |

---

#### Q3 — the wrong-runtime defect, A/B on one fixture

Round 1's finding was that our own #88 fix *masks* this defect: pre-fix a wrong Electron died on `cachedDataRejected`, post-fix plain JS runs fine on the wrong major and the check passed it silently. So the test that matters is the same artifact against both smoke tests.

Fixture: the **unmodified new-head tarball**, installed by the real `npx` path, with the runtime that actually executes being a genuine Electron 44.2.0 (`CFBundleShortVersionString = 44.2.0`, unpacked from the same download the round-1 repro used). Nothing about the tarball is altered — the artifact runs on a runtime it was never tested against, which is precisely the Q3 condition.

```
$ node <a8d9f52be smoke test> local-operator-ui-NEWHEAD.tgz
LOCAL_OPERATOR_UI_READY electron=44.2.0
PASS: the app started and reported LOCAL_OPERATOR_UI_READY.
OLD_SMOKE_EXIT=0        <-- the defect: passes an artifact on the wrong runtime

$ node <e71db5ce4 smoke test> local-operator-ui-NEWHEAD.tgz
LOCAL_OPERATOR_UI_READY electron=44.2.0
FAIL: the app started on Electron 44.2.0, but this package pins 35.5.1.
The tarball resolved a runtime it was never tested against; that is the shape of issue #88.
SMOKE_EXIT=1
```

Same tarball, same runtime, same command shape. Exit 0 before, exit 1 after, with actual-vs-expected named. **This is the check catching my artifact.**

Negative control — the correct runtime still passes, so the check is not simply failing everything:

```
$ node scripts/npx-smoke-test.mjs local-operator-ui-NEWHEAD.tgz     # normal npx install
LOCAL_OPERATOR_UI_READY electron=35.5.1
PASS: the app started on Electron 35.5.1 and reported LOCAL_OPERATOR_UI_READY.
SMOKE_EXIT=0
```

The launcher on the same wrong-major install warns and still starts, which is the documented WARN-not-exit split:

```
$ LOCAL_OPERATOR_UI_SMOKE_TEST=true ./node_modules/.bin/local-operator-ui   # electron swapped to 44.2.0
Warning: local-operator-ui pins Electron 35.5.1 but resolved 44.2.0.
This usually means the optional Electron install was skipped and another
Electron was picked up from the surrounding environment. The app will try
to start anyway. If it misbehaves, reinstall with:
  npm install -g local-operator-ui --include=optional
LOCAL_OPERATOR_UI_READY electron=44.2.0
EXIT=0
```

The new `close` branch also fails closed on a tarball predating the marker, rather than hanging to the timeout — run against my round-1 `local-operator-ui-FIXED.tgz`:

```
LOCAL_OPERATOR_UI_READY
FAIL: the app reported LOCAL_OPERATOR_UI_READY without an electron= version (exit code 0, signal null).
The tarball predates the runtime-version assertion, so which Electron ran cannot be verified.
SMOKE_EXIT=1     (69s, not a 600s timeout)
```

---

#### The defensive version read — 11 attempts to crash the launcher, 0 crashes

The first cut of this fix crashed with a `TypeError` on a `package.json` without `optionalDependencies`, so I attacked both reads rather than only that case. Each case was applied to a real `npm install` tree and reverted afterwards; the control before and after starts clean with no warning.

| # | Case | Exit | Behaviour |
| --- | --- | --- | --- |
| D1 | `electron/package.json` missing | 0 | `Warning: could not verify the Electron version: Cannot find module 'electron/package.json'` → app starts |
| D2 | `electron/package.json` malformed JSON | 1 | Caught by the **pre-existing** resolve guard (Node itself refuses the module: `Invalid package config`), printing the actionable reinstall message. Not a crash. |
| D3 | app `package.json` with **no `optionalDependencies`** (the original crash) | 0 | `Warning: could not verify the Electron version: no version to compare` → app starts |
| D4 | `electron/package.json` chmod 000 | 0 | `Warning: ... EACCES: permission denied` → app starts |
| D5 | electron pkg valid but no `version` field | 0 | `no version to compare` → starts |
| D6 | `version` is a **number**, not a string | 0 | `resolved.split is not a function` caught → starts |
| D7 | `version` is `null` | 0 | `no version to compare` → starts |
| D8 | `optionalDependencies` present, no `electron` key | 0 | `no version to compare` → starts |
| D9 | pin is a range `^35.0.0` | 0 | Warns spuriously, starts. Not a defect: `test-publish-workflow.mjs` asserts the pin is exact (`/^\d+\.\d+\.\d+$/`), so a range cannot ship. |
| D10 | `optionalDependencies` is a **string** (`?.` gives no protection) | 0 | `no version to compare` → starts |
| D11 | app `package.json` missing entirely | 0 | `Cannot find module '../package.json'` caught → starts |

D6 and D10 are the two that defeat optional chaining — a non-string version and a non-object `optionalDependencies` both reach `.split`/property access on the wrong type. Both are caught. **No input I could construct made a diagnostic the reason the app failed to start.**

---

#### Q1 — the macOS leg genuinely ran the check

Step conclusions from the API, not from the summary badge:

```
macOS (job 101560353673)           ubuntu (job 101560353659)
success  Build and pack            success  Build and pack
success  Assert ... no V8 bytecode success  Assert ... no V8 bytecode
success  Install and run with npx  skipped  Install and run with npx
```

The macOS runner's own log proves the launch executed and the *new* assertion ran there:

```
2026-09-06T21:19:22.85Z LOCAL_OPERATOR_UI_READY electron=35.5.1
2026-09-06T21:19:22.87Z PASS: the app started on Electron 35.5.1 and reported LOCAL_OPERATOR_UI_READY.
```

Linux is `skipped` on the launch step only and still runs the tarball assertion (`No .jsc files in the npm tarball.`), with zero `LOCAL_OPERATOR_UI_READY` lines — so that leg is scoped, not hollowed out. `node --test scripts/test-publish-workflow.mjs`: **88 pass, 0 fail**.

---

#### Q2 — not silenced

```
$ git diff a8d9f52be..e71db5ce4 | grep '^+' | grep -iE 'no-sandbox|ELECTRON_DISABLE_SANDBOX|setuid|chrome-sandbox'
    (only comment lines in ci.yml explaining why it is NOT done)
$ git grep -nE 'appendSwitch|appendArgument' e71db5ce4 -- src/
    (none)
```

The one `--no-sandbox` in the repo is in `scripts/capture-evidence.mjs`, a headless-Chrome screenshot helper this PR does not touch and the tarball does not ship (`tar tzf ... | grep -c capture-evidence` → `0`). The `sandbox:` keys in `src/main/` are pre-existing and untouched by the whole PR. CI carries the reasoning inline, #91 is open, and the body states plainly that Linux npm is not supported. Correct call: silencing it would have made the leg green while leaving users broken.

---

#### Regression spot-check on the delta

The tarball here was packed under deliberately adversarial conditions — a full bytecode (`electron-vite build`) run first, leaving stale `out/main/index.jsc` and `out/preload/index.jsc` on disk, plus `chmod -x` on the bin — then `npm pack`:

| Check | Result |
| --- | --- |
| `prebuild:npm` clears stale bytecode | `.jsc` in `out/` after pack: **0** (was 2) |
| `postbuild:npm` restores the exec bit | `-rwxr-xr-x bin/local-operator-ui.js` (forced to `0644` first) |
| `.jsc` in the npm tarball | **absent** |
| bin mode inside the tarball | `-rwxr-xr-x` |
| `.jsc` in an installer build | **present** — `out/main/index.jsc`, `out/preload/index.jsc`, with `index.js` reduced to the loader stub |
| npx run | `LOCAL_OPERATOR_UI_READY electron=35.5.1`, `PASS`, exit 0 |
| `bun add` run | `LOCAL_OPERATOR_UI_READY electron=35.5.1`, exit 0; electron 35.5.1 in the bun tree, `Electron.app` present |

MINOR-1 and MINOR-2 are therefore confirmed on the real path rather than by reading the scripts: both safety nets fire under the exact conditions that defeated them.

Rendered window still opens on the new head (no smoke flag, isolated `HOME` + `--user-data-dir`): a real 1380x900 "Local Operator" window (`CGWindowID 41706`), full Electron tree (main + GPU + renderer helpers), chrome/sidebar/onboarding rendering, no `App threw an error during load`, no `cachedDataRejected`, no version warning.

---

#### Follow-up observations

- The wrapper's warning goes to **stderr** and the app keeps running, so on a wrong-major install a user piping stdout sees a working app and never the diagnosis. That is the intended trade-off and I am not asking for a change; worth knowing if #91-style reports come in from users who saw no error.
- `NPX_SMOKE_TIMEOUT_MS` defaults to 600s while a real run settles in ~70s locally and ~55s on the macOS runner. The budget is for a cold Electron download and is fine; noting only that a genuine hang now costs 10 minutes of runner time, where the two fail-fast branches added in this round return in about one.

#### Not covered

- **Linux npm channel** — unchanged from round 1 and now tracked as #91; I did not re-run the Docker repro, since no fix for it exists on this head.
- **Windows/Linux installer builds** — only the macOS installer build was exercised, so the `.jsc`-present result is macOS-only.
- **The published artifact** — everything above is a local `npm pack`; the publish path is covered by the workflow tests (88/88) and the new pack-then-publish-that-file gating, not by an actual publish.
- **`build.electronVersion` (MAJOR-3)** — a reviewer finding, not mine; I confirmed only that the repo test asserting the two pins move together passes.

PR #92, opened 2026-09-06
fix(settings): recover from a stalled desktop request instead of spinning forever

damianvtran:
## Summary

Fixes #89 — the Settings page never finishes loading and shows a centered spinner forever, reported by @bbqben against the 0.15.0 universal DMG.

## Root cause

The renderer's desktop control path is **unbounded**. `desktopRequest` → `window.api.desktop.request` → `ipcRenderer.invoke` settles only when the main process replies. With main never replying, the promise stays pending indefinitely — measured `STILL PENDING after 45s`. `useConfig` therefore never settles, `isConfigLoading` stays `true`, and `settings-page.tsx` returns the spinner *before* the `configError` branch can be reached. React Query's `retry` cannot help: retry needs a settled rejection, and a request that never settles never rejects.

**A correction to the earlier diagnosis on the issue.** The investigation reported "no AbortSignal on the main-process desktop path". There is one — `desktop-transport.ts:45`, `AbortSignal.timeout(20000)`, shipped in 0.15.0. Verified against a backend faulted to accept-and-never-answer: the transport settles at 20.1s with 503 for both stall shapes (silent-accept, and headers-then-stall). So pre-fix Settings reached an error in ~40s (20s × `retry: 1`) when the backend stalled *at the HTTP layer*. The genuinely unbounded layer is one above it — the IPC bridge, where main never replies at all. The symptom and the fix location were right; the named cause was not.

## Changes

**1. Bound the transport, not the query** (`desktop-api.ts`). 49 call sites across 8 API modules funnel through the single `desktopRequest` entry point. Bounding only the config query would have left the other 48 able to hang the same way. The deadline is 30s — deliberately longer than main's own 20s, so a slow-but-working backend is still reported by the layer that knows the HTTP status; this bound covers only what main can never report. It rejects with `status: null`, the transport's existing "no backend reached" state the banner and grid already understand. `desktopMedia` keeps its separate 120s budget.

**2. Settings reaches an actionable error** (`settings-page.tsx`). The existing error branch is now reachable, with a Retry affordance. Copy reuses the existing string rather than inventing a new register.

**3. Stop telling users to update a backend that is simply offline** (`provider-grid.tsx`). "Providers could not be loaded. The backend may need an update." fired on *any* error, including a plainly unreachable backend — a wrong diagnosis that sends users down the wrong path. It now branches on the same distinction the offline banner already computes: capability-404 → update copy; unreachable/503 → offline copy.

## Testing evidence

[...]

PR #95, opened 2026-09-07
fix(providers): stop offering every provider when the census failed

### LINKED ISSUE #93: Hosting picker lists every provider while the census is failing
### Summary

`hosting-select.tsx` returns **all** providers while the provider census is loading *or* has failed. During that window a user can select a provider they have no credential for, and the picker disagrees with the onboarding grid, which correctly gates on the census.

### Origin

Raised as MINOR-2 in the agent review round on #87 and classified there as a follow-up rather than grounds for another round. #87 merged on that basis (`e1af0f67a`), so this is filed to keep it from being lost — not a regression from that PR.

The related major (Q3: the picker used env keys and disagreed with the grid) *was* fixed in #87; the picker now uses `readyHostingIds` / `hostingProviderSelectable`, the same predicate as the grid. What remains is the narrower census-failure window.

### Expected

Suppress filtering only while the census is **loading**. On census failure the picker should not fall open to every provider — a failed census is not evidence that a provider is usable, which is the same principle behind the onboarding gate fix in #87 (a failed census yields `pending`, not `first_time`).

### Notes

Low severity: the window is short and the user gets a normal failure downstream if they pick an unconfigured provider. Worth fixing because it is the same class of defect as the two already fixed in #87 — a surface treating "we could not find out" as "yes".

Two related follow-ups from the same review, recorded here rather than filed separately:
- **MINOR-3** — a token-protected external backend resolves to `pending` with no modal. Arguably correct (you cannot onboard a backend you cannot authenticate to) and the compatibility banner owns that message. No action proposed.
- **NIT-2** — the open-credentials fetch inherits the CSP `connect-src` ceiling, same precedent as `health-api.ts`, degrades safely. No action proposed.

damianvtran:
## Summary

Fixes #93. The hosting picker returned **all** providers while the provider census was loading *or* had failed. On failure it fell open to every provider, disagreeing with the onboarding grid, which correctly gates on the census.

This is the same defect class as the two fixed in #87: a surface treating "we could not find out" as "yes".

## What changed

Three census states, three answers:

| state | picker |
| --- | --- |
| loading | all providers — unchanged, the only correct unfiltered state |
| ready | census filter, same predicate as the grid |
| **failed** | **empty, plus the reason** |

The reason comes from `providerLoadErrorMessage` — the same selector `ProviderGrid` renders — rather than a fourth wording for "the backend is unreachable". A test asserts the two are string-equal so they cannot drift.

The decision logic is a pure function in `provider-labels.ts`, beside the grid's predicate, mirroring `CensusInput` in `first-time-user.ts`.

## Two deliberate non-choices

- **The control is not disabled on failure.** Locking a user out because *we* could not find out is the same overreach inverted — they can still type a provider they know they have.
- **The `(Requires additional credentials)` suffix is withheld** on failure, since it asserts something about credentials that nothing checked.

## Testing evidence

**Mutation check: yes.** Reverting `return []` to `return all` on the failed branch fails test 11:

```
+ [ 'anthropic', 'openai', 'google', 'ollama' ]   actual
- []                                              expected
```

12/13 passed under the mutation, so the new test isolates this defect rather than failing for an unrelated reason. Restored: 13/13.

**Frames** — captured from the committed harness (`scripts/hosting-census-evidence.*`), which composes the real `ProviderGrid` and `HostingSelect` behind an injected `window.api.desktop`. Re-captured on head `056e5a0ee`; all five pass the repo's own `assertFramePaints` against `localOperatorDark`.

| # | State | Frame |
| --- | --- | --- |
| 0 | **BEFORE** (base `5ab8b4e26`), census 503, picker open — offers the full provider list on no evidence. This is issue #93. | [00-BEFORE-failed-picker-offers-everything.png](https://raw.githubusercontent.com/damianvtran/local-operator-ui/evidence/pr-95/ev/00-BEFORE-failed-picker-offers-everything.png) |
| 1 | **AFTER**, census 503 — picker and grid state the *identical* sentence. | [01-census-failed-agreement.png](https://raw.githubusercontent.com/damianvtran/local-operator-ui/evidence/pr-95/ev/01-census-failed-agreement.png) |
| 2 | Census ready — the filter applies; Google/Mistral absent (need sign-in), local servers kept. | [02-census-ready.png](https://raw.githubusercontent.com/damianvtran/local-operator-ui/evidence/pr-95/ev/02-census-ready.png) |
| 3 | Census loading — unfiltered, deliberately. | [03-census-loading.png](https://raw.githubusercontent.com/damianvtran/local-operator-ui/evidence/pr-95/ev/03-census-loading.png) |
| 4 | **AFTER**, census 503, picker open — "No matches", with the reason beneath. | [04-failed-picker-open-offers-nothing.png](https://raw.githubusercontent.com/damianvtran/local-operator-ui/evidence/pr-95/ev/04-failed-picker-open-offers-nothing.png) |

Frames 0 and 4 are the before/after pair for #93: the same injected 503, the same open dropdown, seventeen providers against none.

Hosted on the throwaway branch `evidence/pr-95` (delete after merge) rather than committed to the tree, per the evidence-off-the-tree rule.

**CSS confirmed loaded, two ways** rather than by eye: the served stylesheet is 97 KB with `.bg-canvas`, `.bg-surface`, `.text-ink-muted`, `.border-hairline` and `.bg-sunken` all compiled, and all five frames pass the repo's own `assertFramePaints`. This matters because an evidence harness in #92 set `root: scripts/`, so Tailwind never scanned `src/renderer` and every frame came out achromatic — unable to show the state it was captioned as showing. This harness is rooted at the repo.

**Gates:** `biome check src` 0 errors / 7 warnings (matches `main`), typecheck clean, **73/73** across all eight desktop test files — and, as of round 1, those tests now run in CI on every push (`Desktop Tests` job), which they previously did not.

## Four near-misses worth recording

Each would have produced evidence that proved nothing, and each was caught before it reached this PR:

1. **Every early test run was a no-op.** `. nvm.sh` returns rc=3, so `&&` chains died before `node --test` ever ran — empty output plus rc=3 read as success.
2. **The first harness frame was a bare ground** — a hand-written fixture used `methods` where the backend sends `auth_methods`. Switched to the committed fixture.
3. **One mutable bridge flag across three panels** meant later panels' mount refetches read whatever the flag held last, so the "ready" panel rendered a *failure* under a healthy caption.
4. **Zustand `persist` rehydration silently overwrote the seeded models store**, leaving the picker empty in every state — all three frames would have looked identical and none would have been the state claimed.

## Not covered

- Not exercised in packaged Electron; the harness drives the shipped IPC path via an injected `window.api.desktop`, not a real backend returning 503.
- Dark theme only (`localOperatorDark`); the other eleven palettes were not captured.
- MINOR-3 and NIT-2 from the same review round are left alone, as the issue proposes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -65,6 +65,81 @@ jobs:
           node scripts/test-validate-release.mjs
           node --test scripts/test-release-safety.mjs scripts/test-publish-workflow.mjs
 
+  test:
+    # The desktop test suite -- the regression guards for issues 87, 92 and 93,
+    # every one of which is the same defect class (a surface treating "we could
+    # not find out" as an answer).
+    #
+    # This job exists because until it did, `pnpm test:desktop` ran in NO
+    # workflow: the whole suite was local-only, so three consecutive fixes in
+    # one defect class landed with their guards unenforced on every subsequent
+    # PR. A guard nothing runs is indistinguishable from no guard, which is the
+    # failure this suite is largely about.
+    #
+    # `--frozen-lockfile` is what keeps the invocation CI-safe. Running
+    # `pnpm test:desktop` on a cold tree makes pnpm resolve first, and that
+    # resolution is what writes an `allowBuilds:` block into
+    # `pnpm-workspace.yaml`; installing explicitly and frozen beforehand means
+    # the test step only runs node, and pnpm fails loudly rather than mutating
+    # the lockfile if either file is out of date.
+    name: Desktop Tests
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Install pnpm
+        uses: pnpm/action-setup@v4
+        with:
+          version: '10'
+
+      - name: Setup Node.js 22
+        uses: actions/setup-node@v4
+        with:
+          node-version: '22'
+          cache: 'pnpm'
+
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile
+
+      - name: Run desktop tests
+        run: pnpm test:desktop
+
+      # A job that silently runs zero tests is the same vacuous check this job
+      # was added to undo, and it is the likeliest way this one rots: a renamed
+      # script or a glob that matches nothing still exits 0. Assert the suite
+      # actually executed a plausible number of tests, and that none failed.
+      - name: Assert the suite actually ran
+        run: |
+          pnpm test:desktop 2>&1 | tee /tmp/desktop-tests.log
+          PASS=$(grep -E '^# pass ' /tmp/desktop-tests.log | awk '{print $3}')
+          FAIL=$(grep -E '^# fail ' /tmp/desktop-tests.log | awk '{print $3}')
+          echo "desktop suite: pass=$PASS fail=$FAIL"
+          if [ -z "$PASS" ] || [ "$PASS" -lt 60 ]; then
+            echo "Desktop suite reported $PASS passing tests; expected at least 60."
+            echo "Either tests were dropped or the runner matched no files."
+            exit 1
+          fi
+          if [ "$FAIL" != "0" ]; then
+            echo "Desktop suite reported $FAIL failing tests."
+            exit 1
+          fi
+
+      # The suite must not mutate tracked files. `pnpm test:desktop` on a cold
+      # tree has been observed writing an `allowBuilds:` block into
+      # pnpm-workspace.yaml via pnpm's own resolution, which would land in a
+      # later commit as unexplained churn.
+      - name: Assert the run left the tree clean
+        run: |
+          if [ -n "$(git status --porcelain)" ]; then
+            echo "Desktop tests modified tracked files:"
+            git status --porcelain
+            git diff
+            exit 1
+          fi
+          echo "Working tree clean after the desktop suite."
+
   audit:
     name: Security Audit
     runs-on: ubuntu-latest
--- a/scripts/hosting-census-evidence.html
+++ b/scripts/hosting-census-evidence.html
@@ -1,0 +1,27 @@
+<!doctype html>
+<!--
+  Evidence harness for the hosting picker's census states (issue 93).
+
+  Mounts the SHIPPED `HostingSelect` and the SHIPPED `ProviderGrid` side by
+  side, behind a REAL desktop bridge (an injected `window.api.desktop`), so
+  `desktopRequest` takes the Electron IPC branch that actually ships rather
+  than the browser-dev `/__desktop` HTTP branch, which classifies its failures
+  differently.
+
+  Both surfaces are captured against the SAME injected fault, because the
+  property under test is that they agree: the picker must not offer a provider
+  the grid is simultaneously unable to vouch for.
+
+  Not part of the app or the test suite: a capture surface, run by hand with
+  `pnpm vite --config scripts/hosting-census-evidence.vite.mjs`.
+-->
+<html lang="en" data-theme="localOperatorDark">
+	<head>
+		<meta charset="UTF-8" />
+		<title>Hosting picker census states</title>
+	</head>
+	<body>
+		<div id="root"></div>
+		<script type="module" src="/scripts/hosting-census-evidence.tsx"></script>
+	</body>
+</html>
--- a/scripts/hosting-census-evidence.tsx
+++ b/scripts/hosting-census-evidence.tsx
@@ -1,0 +1,230 @@
+/**
+ * Evidence harness: the hosting picker's three census states, beside the grid.
+ *
+ * See `hosting-census-evidence.html` for why this exists. The property being
+ * captured is issue 93's: the picker must not offer every provider when the
+ * census FAILED, and what it does instead has to agree with what the
+ * onboarding grid says about the same fault. Rendering both shipped components
+ * against ONE injected fault is what makes that agreement visible rather than
+ * asserted -- neither component's copy is restated here, so a wrong sentence in
+ * a captured frame is wrong in the product.
+ */
+
+import { ProviderGrid } from "@features/providers/provider-grid";
+import { desktopResult } from "@shared/api/local-operator/desktop-api";
+import { desktopKeys } from "@shared/api/local-operator/desktop-hooks";
+import { HostingSelect } from "@shared/components/hosting/hosting-select";
+import { QueryClient, QueryClientProvider } from "@tanstack/react-query";
+import { StrictMode } from "react";
+import { createRoot } from "react-dom/client";
+import { useModelsStore } from "@shared/store/models-store";
+import censusFixture from "./fixtures/auth-providers-0.50.0.json";
+import "@renderer/styles/index.css";
+
+/**
+ * Which census state this page load captures, from `?state=`.
+ *
+ * ONE state per load, deliberately. The first version of this harness rendered
+ * all three panels together against a single module-level "fail providers"
+ * flag, and the panels' own mounts re-ran `providers.list` AFTER seeding -- so
+ * whichever value the flag happened to hold last was the answer every panel
+ * got, and the healthy panel captured a failure while captioned as ready. A
+ * shared mutable fault is not a per-panel fault; binding the state to the page
+ * load makes each frame a picture of exactly one condition.
+ *
+ * `capabilities` must always SUCCEED: the picker only consults the census when
+ * the backend advertises the `auth` feature, so a harness that failed
+ * capabilities too would capture the unmanaged-backend branch (the env-file key
+ * list) and caption it as a census failure -- a frame of the wrong code path.
+ */
+const STATE = new URLSearchParams(location.search).get("state") ?? "ready";
+
+const CAPABILITIES = {
+	desktop_available: true,
+	features: { auth: 1, settings: 1, commands: 1 },
+};
+
+// The bridge itself. `window.api.desktop` existing is what routes the renderer
+// down the IPC branch that ships, so it is installed before any surface mounts.
+(window as unknown as { api: unknown }).api = {
+	desktop: {
+		request: async (request: { op: string }) => {
+			if (request.op === "capabilities")
+				return { status: 200, body: { result: CAPABILITIES } };
+			if (request.op === "providers.list") {
+				// The loading frame needs a request that never settles, and it has
+				// to hang HERE rather than in a seeded queryFn: the component's own
+				// mount refetch would otherwise resolve against the bridge and
+				// settle the state the frame is meant to show.
+				if (STATE === "loading") return new Promise(() => {});
+				if (STATE === "failed")
+					return { status: 503, body: { detail: "server unavailable" } };
+				return { status: 200, body: { result: { providers: CENSUS } } };
+			}
+			return { status: 503, body: { detail: "denied" } };
+		},
+	},
+};
+
+// React Query pauses refetching in a hidden tab, which silently defeats a
+// capture run in a background window.
+Object.defineProperty(document, "visibilityState", {
+	configurable: true,
+	get: () => "visible",
+});
+Object.defineProperty(document, "hidden", { configurable: true, get: () => false });
+
+/**
+ * The census the healthy panel renders.
+ *
+ * The VERBATIM `GET /v1/auth/providers` body already committed as a fixture and
+ * already used by `scripts/provider-state.test.mjs`, rather than a shape
+ * hand-written here. A hand-written one was wrong on first try -- it named the
+ * methods array `methods` where the backend sends `auth_methods`, which threw
+ * inside the grid and produced an empty frame. A fixture that the shipped
+ * components can actually consume is the only kind that proves anything.
+ */
+
+const CENSUS = censusFixture.result.providers;
+
+/**
+ * The manifest the picker filters, seeded into the models store.
+ *
+ * `getHostingProviders()` reads `useModelsStore`, NOT the census, and returns
+ * `[]` until that store is initialised. Without this seed the picker renders
+ * empty in all three states -- which would have made the failed frame look
+ * exactly like the healthy one and proved nothing at all. Seeding it is what
+ * makes "the census removed these rows" an observable difference between the
+ * frames rather than an invisible one.
+ *
+ * The rows are the census's own ids, so the two surfaces are talking about the
+ * same providers and the picker has something to drop.
+ *
+ * Called AFTER an explicit rehydrate rather than at module top level: the store
+ * is `persist`ed, and its rehydration from localStorage lands in a microtask
+ * that overwrote a seed written before it -- silently, leaving the picker empty
+ * in a frame captioned as healthy.
+ */
+function seedManifest() {
+	useModelsStore.setState({
+		providers: CENSUS.map((provider) => ({
+			id: provider.id,
+			name: provider.name,
+			description: provider.base_url ?? "Model provider",
+			url: provider.base_url ?? "",
+			requiredCredentials: provider.credential_name
+				? [provider.credential_name]
+				: [],
+		})),
+		models: [],
+		isInitialized: true,
+		lastFetched: Date.now(),
+	});
+}
+
+const newClient = () =>
+	new QueryClient({
+		defaultOptions: {
+			queries: {
+				retry: false,
+				gcTime: Number.POSITIVE_INFINITY,
+				// A surface mounting against an errored query would otherwise re-run
+				// it and paint its loading state instead of the state being captured.
+				retryOnMount: false,
+				refetchOnWindowFocus: false,
+			},
+		},
+	});
+
+/**
+ * Settle capabilities before mounting, so the surfaces do not paint a
+ * capabilities spinner over the census state being captured. The census itself
+ * is deliberately NOT seeded: the components issue it themselves against the
+ * bridge above, which is the path that ships.
+ */
+async function seedCapabilities(client: QueryClient) {
+	await client
+		.fetchQuery({
+			queryKey: desktopKeys.capabilities,
+			queryFn: () => desktopResult({ op: "capabilities" }),
+			retry: false,
+		})
+		.catch(() => undefined);
+}
+
+const Panel = ({
+	title,
+	note,
+	client,
+}: {
+	title: string;
+	note: string;
+	client: QueryClient;
+}) => (
+	<QueryClientProvider client={client}>
+		<section className="flex w-[520px] shrink-0 flex-col gap-3 rounded-md border border-hairline bg-surface p-4">
+			<div className="flex flex-col gap-1">
+				<h2 className="text-heading text-ink">{title}</h2>
+				<p className="text-meta text-ink-muted">{note}</p>
+			</div>
+			{/* The defect is what the OPEN list contains, so the frame has to show
+			    it open. `?open=1` clicks the combobox after mount rather than
+			    reaching into the component, so the list is opened the way a user
+			    opens it. */}
+			<div className="rounded-md border border-hairline bg-sunken p-3">
+				<p className="mb-2 text-meta text-ink-dim">Hosting picker (Settings)</p>
+				<HostingSelect
+					value=""
+					onSave={async () => {}}
+					filterByCredentials={true}
+					allowCustom={true}
+					allowDefault={false}
+					emptyHelperText="No hosting providers available. Add one in API credentials, in the list on the left."
+				/>
+			</div>
+			<div className="rounded-md border border-hairline bg-sunken p-3">
+				<p className="mb-2 text-meta text-ink-dim">Onboarding providers grid</p>
+				<ProviderGrid />
+			</div>
+		</section>
+	</QueryClientProvider>
+);
+
+const NOTES: Record<string, { title: string; note: string }> = {
+	loading: {
+		title: "Census loading",
+		note: "Filtering suppressed on purpose: the answer is moments away and blanking a populated control reads as it breaking.",
+	},
+	ready: {
+		title: "Census ready",
+		note: "Census owns the filter. Google and Mistral are absent (need sign-in); the local servers stay (no key needed).",
+	},
+	failed: {
+		title: "Census failed (503)",
+		note: "Issue 93: offered every provider before this fix. Now offers none, and says why in the grid's own words.",
+	},
+};
+
+const client = newClient();
+await seedCapabilities(client);
+await useModelsStore.persist?.rehydrate();
+seedManifest();
+
+const { title, note } = NOTES[STATE] ?? NOTES.ready;
+
+createRoot(document.getElementById("root") as HTMLElement).render(
+	<StrictMode>
+		<div className="flex min-h-screen justify-center bg-canvas p-6">
+			<Panel title={title} note={note} client={client} />
+		</div>
+	</StrictMode>,
+);
+
+// Open the picker's list the way a user would, when asked. Done from outside
+// the component so nothing about the captured list is stage-managed: whatever
+// the combobox shows here is what it shows on a click in the app.
+if (new URLSearchParams(location.search).get("open") === "1") {
+	setTimeout(() => {
+		document.querySelector<HTMLInputElement>('input[role="combobox"]')?.click();
+	}, 400);
+}
--- a/scripts/hosting-census-evidence.vite.mjs
+++ b/scripts/hosting-census-evidence.vite.mjs
@@ -1,0 +1,44 @@
+import { resolve } from "node:path";
+import tailwindcss from "@tailwindcss/vite";
+import react from "@vitejs/plugin-react";
+import { defineConfig } from "vite";
+
+/**
+ * Vite config for the hosting-picker census evidence harness only (issue 93).
+ *
+ * Aliases are copied from the renderer stage so the shipped components resolve
+ * exactly as they do in the app.
+ *
+ * ## `root` is the REPOSITORY, not `scripts/`
+ *
+ * The backend-error harness set `root: scripts/`, which put Tailwind v4's
+ * automatic source detection under `scripts/` -- so it never scanned
+ * `src/renderer` and every frame captured through it came out with no utility
+ * classes at all. An achromatic frame cannot show the state its caption claims,
+ * and the failure is silent: the harness boots, React mounts, the components
+ * render their real text, and only the colour is missing.
+ *
+ * Rooting at the repository puts `src/renderer` inside the scanned tree, and
+ * the html entry is addressed by its path under that root. `assertFramePaints`
+ * from `check-evidence.mjs` is run over every captured frame afterwards as the
+ * actual proof, since a config that looks right is not evidence that it was.
+ */
+const root = resolve(import.meta.dirname, "..");
+
+export default defineConfig({
+	root,
+	resolve: {
+		alias: {
+			"@renderer": resolve(root, "src/renderer/src"),
+			"@components": resolve(root, "src/renderer/src/components"),
+			"@features": resolve(root, "src/renderer/src/features"),
+			"@shared": resolve(root, "src/renderer/src/shared"),
+			"@assets": resolve(root, "src/renderer/src/assets"),
+			"@hooks": resolve(root, "src/renderer/src/hooks"),
+			"@api": resolve(root, "src/renderer/src/api"),
+			"@store": resolve(root, "src/renderer/src/store"),
+		},
+	},
+	plugins: [react(), tailwindcss()],
+	server: { port: 5198, strictPort: true },
+});
--- a/scripts/provider-state.test.mjs
+++ b/scripts/provider-state.test.mjs
@@ -1,5 +1,6 @@
 import assert from "node:assert/strict";
 import { readFileSync } from "node:fs";
+import { readFile } from "node:fs/promises";
 import { test } from "node:test";
 import { build } from "esbuild";
 
@@ -30,6 +31,7 @@ const {
 	hostingProviderSelectable,
 	readyHostingIds,
 	selectableHostingProviders,
+	hostingCensusStateFrom,
 	hostingCensusFailureHelperText,
 	providerLoadErrorMessage,
 } = await import(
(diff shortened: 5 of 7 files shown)


damianvtran:
### Agent review — round 1

Reviewer: `reviewer` subagent (lopdev team), model `anthropic/claude-opus-5`. No provider fallback occurred during this round.
Scope: `5ab8b4e26..9aa79e8eb` (one commit, 6 files, +526/−17)

Read the diff, issue #93 and the #87 origin first; opened `provider-grid.tsx`, `provider-grid-error-copy.test.mjs` and the CI workflow only to answer specific questions the diff could not. Everything asserted below I ran myself in a throwaway sandbox (source copy + symlinked `node_modules`, Node 22 via nvm, removed afterwards). I did not run `pnpm test:desktop`.

## Verified independently

**The mutation claim is exactly as stated.** Reverting `return []` → `return all` on the failed branch (`provider-labels.ts:170`) gives **12/13, failing only test 11**, with the diff the body quotes verbatim:

```
not ok 11 - hosting picker: a FAILED census offers nothing, not everything
    + actual - expected
    + [ 'anthropic', 'openai', 'google', 'ollama' ]
    - []
```

Restored: 13/13. The other twelve tests are indifferent to the mutation, so the test isolates this defect rather than failing for a coincidental reason. It calls shipped code — the harness bundles `src/renderer/src/features/providers/provider-labels` through esbuild with `resolveDir: process.cwd()`, not a restated copy.

**The `data`-before-`isError` ordering is correct, and this is the thing I most expected to find broken.** `hosting-select.tsx:115-120` checks `census.data` *before* `census.isError`, so a failed background refetch over a previously-successful census resolves to `ready` on the cached payload rather than blanking the control. That is the opposite of the #92 regression shape and the inline comment says why. Good.

**No fourth state.** `ready` + empty census array returns `[]` — I ran it (`ready+empty -> []`). That is right: a census that answered "nothing is ready" is evidence, unlike one that did not answer. The absent `unavailable` member is genuinely unreachable, since `censusEnabled` gates the whole branch at :126 and the non-census path falls to the env-file list.

**Gates.** `biome check src` → 392 files, **0 errors / 7 warnings**, matching `main`. `tsc -p tsconfig.web.json --noEmit` → clean, rc=0. Full desktop set 71/71. Nothing unrelated rode along; the two `.test.mjs` reformat hunks are Biome formatting of adjacent lines, not behaviour.

**Both deliberate non-choices are right, and I'd have raised them if they'd gone the other way.** Not disabling the control is correct — `isDisabled` at :311-315 exempts `censusFailed`, and I confirmed at :162-179 that a value the user already has is still pushed as an option, so nobody is stranded by our own ignorance. Withholding `(Requires additional credentials)` is correct for the same reason: it is an assertion about credentials, and on failure nothing checked them.

## Findings

**MAJOR-1 — the drift guard is a tautology on the picker side; it cannot fail.** `provider-labels.ts:186-188`:

```ts
export function hostingCensusFailureHelperText(error: unknown): string {
	return providerLoadErrorMessage(error);
}
```

Test 12 asserts `hostingCensusFailureHelperText(e) === providerLoadErrorMessage(e)` — a wrapper compared against the function it forwards to. It is true by construction, for every input, regardless of what either surface renders. The PR body's claim that "a test asserts the two are string-equal so they cannot drift" therefore overstates what is protected.

Mutation-proved, both directions, against the full 71-test set:

- I replaced the picker's call at `hosting-select.tsx:229` with a hardcoded `"Picker private wording that drifted"` — **22/22 still passed**, including test 12. The exact drift the test exists to prevent goes undetected.
- The **grid** side *is* protected, by a pre-existing test I found while checking this: `provider-grid-error-copy.test.mjs:180-190` source-asserts that `provider-grid.tsx` contains `providerLoadErrorMessage(providers.error)`. Making the grid drift → **70/71, 1 fail**.

So the asymmetry is the finding: the grid has a source assertion, the picker has none, and no test file references `hosting-select.tsx` at all. The fix is one line in the same style as the guard that already works — source-assert that `hosting-select.tsx` contains `hostingCensusFailureHelperText(census.error)`. Whether test 12 keeps its tautological half is a matter of taste; what it must gain is a check that the picker still *calls* the shared selector.

I've classified this MAJOR rather than BLOCKER deliberately: the shipped behaviour on this head is correct and the two surfaces genuinely do agree today. What is defective is the guard, and given this PR's stated purpose is that the two surfaces "cannot drift", a guard that cannot fail is exactly the class of evidence this session has been catching.

**MINOR-1 — none of this runs in CI.** `provider-state.test.mjs` is in the `test:desktop` script (`package.json:67`) but `.github/workflows/ci.yml:66` runs only `test-release-safety.mjs` and `test-publish-workflow.mjs`, and nothing in `.github/` invokes `test:desktop`. The regression guard for #93 — and for #87 and #92 before it — is local-only, so the third fix in this defect class lands with its guard unenforced on every future PR. Out of scope to fix here, but worth a follow-up PR this session; I am not filing an issue.

**NIT-1 — the PR body describes five frames that are not attached.** The body narrates census-failure before/after, picker-and-grid agreement, healthy and loading states, plus `assertFramePaints` and a 97 KB stylesheet — but the PR carries **no images and zero comments** (`gh api .../issues/95/comments` → `0`). I have not seen the frames, so I can neither confirm nor dispute the CSS claim, and per the standing bar I am not taking a caption's word for what a frame shows. The harness under `scripts/hosting-census-evidence.*` is committed and does compose the real `ProviderGrid` and `HostingSelect`, which is the right shape. Attach the frames to the PR (evidence belongs on the PR, not in the tree) so the design round and QA have something to look at.

Nothing else. No BLOCKER. I did not find a fourth state, a stale-data blanking bug, or unrelated changes, and I am not padding this with a nit tail — the diff is small, well-commented and does what it says.

## Verdict

**Not clean — one MAJOR (MAJOR-1).** Not terminal: MAJOR-1 wants a remediation round, and it is a one-line test addition, so round 2's scope should be just that delta. MINOR-1 is a legitimate follow-up PR rather than a blocker on this one; NIT-1 is answered by attaching the frames.

The production change itself I consider correct and ready — the objection is to the guard around it, not the fix.

damianvtran:
### QA review — round 1

**Verdict: PASS — terminal.** No FAIL on head `9aa79e8eb`. No blocker or major finding. Two informational notes (Q1, Q2), neither blocking, neither in this PR's scope to fix.

**Reviewer:** qa-tester subagent (`anthropic/claude-opus-5`), independent of the PR author.
**Scope:** `5ab8b4e26..9aa79e8eb`

---

## Environment

Two throwaway detached worktrees of my own, so the coder's `/tmp/lop-ui-93` was never written to and the operator's backend on `:18141` was never touched (read-only, left running):

| worktree | revision | vite port | purpose |
| --- | --- | --- | --- |
| `/tmp/qa95-base` | `5ab8b4e26` (pre-fix) | 5298 | reproduce the defect |
| `/tmp/qa95-head` | `9aa79e8eb` (PR head) | 5297 | confirm the fix |

Node v22.22.0 via nvm. Harness is **my own** (`scripts/qa95.tsx`), not the author's — same shipped components behind an injected `window.api.desktop`, plus a live-DOM readout that prints the picker's open option list, both surfaces' text, computed colours, and the input's value/disabled state. Verdicts below rest on those captured strings, not on reading a screenshot.

## Matrix

| # | Surface | State driven | Actual | Result |
| 1 | Picker, **pre-fix** `5ab8b4e26` | census 503 | **17 options** incl. Google/Mistral/Alibaba | defect **REPRODUCED** |
| 2 | Picker, head | census 503 | **0 options**, `No matches`, reason shown | PASS |
| 3 | Picker, head | census loading (hangs) | 17 options (unfiltered — intended) | PASS |
| 4 | Picker, head | census ready | 14 of 17 (3 `Needs sign-in` dropped) | PASS |
| 5 | Picker + grid, head | census 503 | sentences **byte-identical** (122 chars each) | PASS |
| 6 | **Picker, head** | **ready → refetch FAILS** | **14 options kept — does NOT blank** | PASS |
| 7 | Picker, head | 503 + saved `value=google` | value kept, `disabled=false` | PASS |
| 8 | Picker, head | 503 + saved value, list open | suffix **withheld** | PASS |
| 9 | Picker, head | ready + saved value, list open | suffix **present** | PASS |
| 10 | Picker, head | ready, click an option | selection commits, list closes | PASS |
| 11 | Mutation | `return []` → `return all` | test 11 fails, 12/13 | PASS |
| 12 | `assertFramePaints` | blank frame vs real frame | rejects blank, accepts real | PASS |
| 13 | Gates | biome / tsc / 71 tests | 0 errors / clean / 71 pass | PASS |

## 1. Defect reproduced pre-fix, fixed on head

Pre-fix (`:5298`, census 503) — every provider offered while the grid beside it reports the failure:

```
QAOPTIONS n=17 ["OpenAI (ChatGPT Plus/Pro)","Anthropic (Claude Pro/Max)","Kimi (Moonshot)",
"xAI (Grok API key)","DeepSeek","Z.AI (GLM API key)","Google (Gemini)","Mistral AI","LM Studio",
"Ollama","vLLM","llama.cpp","OpenAI-compatible","OpenRouter","Radient","Alibaba Cloud (Qwen)",
"QwenCloud Token Plan"]
QAGRID  Providers could not be loaded. The Local Operator server is not answering. ... Retry
```

Google, Mistral and Alibaba are `Needs sign-in` in the fixture — offered anyway. That is issue #93 exactly.

Head (`:5297`), same injected 503:

```
QAOPTIONS n=0 []
QAPICKER Providers could not be loaded. The Local Operator server is not answering. Restart the app so it can start its own server.
QAGRID   Providers could not be loaded. The Local Operator server is not answering. Restart the app so it can start its own server. Retry
```

## 2. Three states, three answers — and the sentences agree

`loading → 17`, `ready → 14`, `failed → 0`, each captured from the open listbox. The ready state drops exactly the three `Needs sign-in` providers.

The agreement claim I checked by string equality, not by eye — picker helper text vs grid alert text with the `Retry` button label removed:

```
picker == grid : True
len 122 == len 122
'Providers could not be loaded. The Local Operator server is not answering. Restart the app so it can start its own server.'
```

## 3. The untested state — succeeded, then refetch failed — is CORRECT

This is the one I was asked to break, and the shape that shipped as a regression in #92. Driven by serving a good census on call 1 and 503 on every call after, then firing a background refetch into the mounted surfaces.

React Query lands in exactly the dangerous state:

```
QAPROBE {"state":"stale","status":"error","fetchStatus":"idle","isError":true,
         "isLoading":false,"hasData":true,"dataLen":17,"errorMsg":"server unavailable"}
```

`isError:true` **and** `hasData:true` simultaneously. The picker:

```
QAOPTIONS n=14 [... same 14 as the healthy state ...]
```

**It keeps the good census.** `censusState` checks `census.data` *before* `census.isError`, so a failed refetch over a cached answer stays `ready`. Byte-identical to the base's behaviour in the same state, so nothing regressed either. The #92 shape is not present here — the ordering in the code comment is load-bearing and correct.

Worth recording for whoever reads this later: **inverting those two lines reintroduces #92 verbatim** in a surface that currently has no test for it. Item Q1 below.

## 4. The attached frames' CSS claim holds — verified two ways of my own

I did not take the author's word or re-run their check. Computed styles off the mounted nodes in my own harness:

```
QACSS canvas=rgb(22, 19, 14) surface=rgb(30, 26, 20) sunken=rgb(15, 12, 8)
      muted=rgb(181, 175, 162) hairline=rgb(53, 48, 34)
```

against the declared `localOperatorDark` palette:

| token | declared | painted | match |
| --- | --- | --- | --- |
| canvas | `#16130e` | `(22,19,14)` | ✅ |
| surface | `#1e1a14` | `(30,26,20)` | ✅ |
| sunken | `#0f0c08` | `(15,12,8)` | ✅ |
| hairline | `#353022` | `(53,48,34)` | ✅ |

Then I checked the guard itself can fail, since a guard that cannot go red is worse than none:

```
BLANK(white)                 -> REJECTED: dominant colour #FFFFFF is ΔE00 79.41 from the
                                nearest `localOperatorDark` ground (max 25) — the story did not paint
REAL(my after-failed capture) -> PASSES assertFramePaints
```

Root cause of the #92 achromatic frames confirmed as genuinely avoided: `backend-error-evidence.vite.mjs` has `root: resolve(root, "scripts")`; this PR's harness has `root = resolve(import.meta.dirname, "..")`. My independent harness reproduced correct colours, so the fix is a property of the config, not of the author's machine.

## 5. Mutation claim re-run, not accepted

In my own worktree, `return []` → `return all` on the failed branch:

```
ok 10 - hosting picker: a READY census filters to what the grid calls ready
not ok 11 - hosting picker: a FAILED census offers nothing, not everything
ok 12 - hosting picker: the failure line is the grid's sentence, verbatim
# tests 13  # pass 12  # fail 1
```

One test fails and it is the right one. Restored → 13/13, `git diff --stat` empty.

## 6. Regression spot-check on the delta

- Healthy census: opened the list, clicked Anthropic → field reads `Anthropic (Claude Pro/Max)`, list closes. Selection unaffected.
- Saved-value-under-failure, both PR non-choices confirmed by captured DOM: `QAINPUT value="Google (Gemini)" disabled=false` — control **not** locked and the value **not** dropped.
- Suffix: withheld on failure (`https://generativelanguage.googleapis.com`), present when ready (`... (Requires additional credentials)`). Both branches exercised, so the conditional is not dead either way.

## Gates (PR branch, whole tree, as CI runs them)

```
npx biome check src      → Checked 392 files. Found 7 warnings.   (0 errors; matches main)
npx tsc --noEmit -p tsconfig.app.json → clean, rc=0
node --test <all 8 desktop files>     → 1..71  # pass 71  # fail 0
```

`package.json` version untouched at `0.15.1`, equal to `origin/main` — correct under the one-bump-per-release-window rule.

---

## Findings

**Q1 — informational, not blocking.** The `data`-before-`isError` ordering in `censusState` is the whole defence against re-shipping #92, and it is currently protected only by a comment. `selectableHostingProviders` is pure and already unit-tested, but the *ordering* lives in the component and no test covers `{data, isError}` set together — swapping those two lines keeps all 71 tests green while blanking a fully-loaded picker. I verified the behaviour is right today; I am flagging that nothing would catch it turning wrong. Cheap to close by asserting the derivation over `{hasData:true, isError:true}`. Reasonable to defer — it is a new test rather than a fix to shipped behaviour, and the PR is otherwise tightly scoped.

**Q2 — pre-existing, out of scope, NOT introduced here.** In the succeeded-then-refetch-failed state the *grid* discards its good rows and renders the error alert, because it branches `isLoading → isError` with no `data` guard (`provider-grid.tsx`). Confirmed identical on base `5ab8b4e26`, so this PR neither caused nor worsened it — and it is the mirror image of #92 in the surface this PR did not touch. Recording it here rather than filing an issue, per the no-follow-up-issues rule. Note it makes the two surfaces disagree in this one state (picker keeps 14, grid shows the error) — but in the safe direction: the picker is not over-offering, which is what #93 is about.

## Not covered

- Not exercised in packaged Electron. The harness drives the shipped IPC branch via an injected `window.api.desktop`; a real 503 from a real backend was not used, since `:18141` is the operator's live store and read-only for me.
- Dark theme (`localOperatorDark`) only — the other eleven palettes were not captured.
- No cross-tenant/authz dimension exists on this surface, so none was tested.

## Cleanup

Both vite servers stopped by recorded pid — cwd confirmed as `/private/tmp/qa95-head` (1415) and `/private/tmp/qa95-base` (95047) before signalling, children verified before signalling, no pattern-kill. 0 listeners left on 5297/5298.

Worktrees `/tmp/qa95-head` and `/tmp/qa95-base` removed and the scratch (`/tmp/qa95-fp`, logs) deleted: **121 MB reclaimed**. Both worktrees symlinked the existing `node_modules`, and I unlinked those symlinks before removal, so the shared 899 MB install under `/tmp/lop-ui-93` was never copied and is intact.

`/tmp/lop-ui-93` untouched: still at `9aa79e8eb7b1773227bf279801dc14270bf432a1` with `git status --porcelain` empty. Operator's backend on `:18141` (pid 46091) still listening, never written to.

damianvtran:

Implementer: `coder` subagent (lopdev team), model `anthropic/claude-opus-5`. No provider fallback occurred during this round.
Scope: `9aa79e8eb..056e5a0ee` — one commit, 4 files, +185/−10.

Answering both round-1 streams: reviewer (one MAJOR) and QA (PASS, terminal, two informational). One unified commit, not several.

---

## MAJOR-1 — the picker's drift guard was a tautology — **FIXED** in `056e5a0ee`

Accepted exactly as written. Test 12 compared `hostingCensusFailureHelperText` against `providerLoadErrorMessage`, the one-line function it forwards to: true by construction for every input, and blind to the picker dropping the call entirely. The grid had a source assertion and the picker had none, so the PR's stated purpose — that the two surfaces cannot drift — was not enforced on the surface this PR changes.

Added a source assertion in the grid's proven style (`provider-grid-error-copy.test.mjs:180-190`), covering all three couplings rather than only the one the reviewer named:

- `hostingCensusFailureHelperText(census.error)` — the failure copy
- `selectableHostingProviders(all, censusState)` — the option filter
- `hostingCensusStateFrom(` — the census-state derivation (see Q1)
- and that `"Providers could not be loaded"` does **not** appear as a literal in the component

**Mutation-check: yes — the new guard fails.** Stated plainly, per instruction:

| mutation | before this round | after |
| --- | --- | --- |
| hardcode `"Picker private wording that drifted"` at `hosting-select.tsx:229` | **22/22 passed** (undetected) | **fails** test 14 |
| replace `selectableHostingProviders(all, censusState)` with `all` | undetected | **fails** test 14 |
| restored | — | **73/73** |

I kept test 12's string-equality half: it is tautological on its own but still pins the *sentence* if the forwarder ever grows a branch. It is no longer the only thing standing between the two surfaces.

## Q1 — the `data`-before-`isError` ordering was protected only by a comment — **FIXED**, not deferred

QA called this reasonable to defer. Taking it anyway: that ordering is the entire defence against re-shipping #92, and an invariant a comment protects is how #92 shipped in the first place.

The derivation moved out of the component into `hostingCensusStateFrom` in `provider-labels.ts`, beside the selector that consumes it. That relocation is the fix, not cosmetics — react-query sets `data` and `isError` **together** when a refetch fails over a successful query, and a component-local `useMemo` cannot be asserted over both flags set at once without a DOM. As a pure function it can.

**Mutation-check: yes — the new test fails when the lines are swapped.** Swapping them previously kept **71/71 green while blanking a fully-loaded picker**; it now fails test 13 (`a failed REFETCH keeps the cached census, it does not blank`), which asserts the cached census still yields `["anthropic","openai","ollama"]` under `isError: true`, plus the other three flag combinations so a regression cannot satisfy it by collapsing everything to `ready`.

Behaviour on head is unchanged — both reviewers independently confirmed the ordering is already correct. This round only makes it *fail-able*.

## MINOR-1 — none of this ran in CI — **FIXED** here rather than deferred

The reviewer suggested a follow-up PR. I brought it into this one because it is what makes the two fixes above worth anything: `pnpm test:desktop` appeared in **no workflow**, so the whole desktop suite — including the guards for #87, #92 and #93 — had only ever run locally, and three consecutive fixes in one defect class landed with their guards unenforced.

Confirmed against the forge, not by reading the file: the last run on `9aa79e8eb` had five jobs — Lint, Type Checking, Security Audit, and two NPX Sanity Checks. No test job.

Added a `Desktop Tests` job matching the existing jobs' shape (pnpm action-setup v10, setup-node 22 with pnpm cache, `pnpm install --frozen-lockfile`). Three details are deliberate:

- **The frozen install is what keeps it CI-safe.** `pnpm test:desktop` on a cold tree makes pnpm resolve first, and that resolution is what writes an `allowBuilds:` block into `pnpm-workspace.yaml`. Installing explicitly and frozen beforehand means the test step only runs node, and pnpm fails loudly rather than mutating the lockfile. A `git status --porcelain` step asserts the run left the tree clean; it passed.
- **A test-count assertion**, because a job that silently runs zero tests is the same vacuous-check defect this job exists to undo — a renamed script or a glob matching nothing still exits 0. It requires ≥60 passing and 0 failing.
- The script itself needed no change.

**Proof it runs and can fail** — both, on the forge:

*It runs:* [run 34139920681](https://github.com/damianvtran/local-operator-ui/actions/runs/34139920681) on `056e5a0ee` — six jobs now, all green, and the `Desktop Tests` log shows the count:

```
1..73
# tests 73
# pass 73
# fail 0
desktop suite: pass=73 fail=0
Working tree clean after the desktop suite.
```

*It can fail:* I pushed the MAJOR-1 mutation to a throwaway branch and let CI run it — [run 34140251950](https://github.com/damianvtran/local-operator-ui/actions/runs/34140251950), `Desktop Tests` **failure**:

```
not ok 49 - hosting picker: the component routes BOTH decisions through the shared selectors
# pass 72
# fail 1
##[error]Process completed with exit code 1.
```

That is the new guard going red in CI, on a real runner, for the exact drift it was added to catch. The probe branch has been deleted.

## NIT-1 — five frames narrated, zero attached — **FIXED**

Fair, and the stronger objection: the reviewer could neither confirm nor dispute the CSS claim. Re-captured all five from the committed harness on head `056e5a0ee` and put them in the PR body as a table, including the **before** frame from base `5ab8b4e26` that the set was missing. All five pass `assertFramePaints` against `localOperatorDark`.

The pair that matters: [base, census 503, picker open](https://raw.githubusercontent.com/damianvtran/local-operator-ui/evidence/pr-95/ev/00-BEFORE-failed-picker-offers-everything.png) offers the full provider list; [head, identical injected 503](https://raw.githubusercontent.com/damianvtran/local-operator-ui/evidence/pr-95/ev/04-failed-picker-open-offers-nothing.png) shows "No matches" with the grid's sentence beneath.

Frames live on the throwaway branch `evidence/pr-95`, not in the tree. **Delete that branch after merge.**

## Q2 — the grid discards good rows in the succeeded-then-refetch-failed state

`deferred — pre-existing on base, errs safe`

Out of scope, per the round instruction. QA confirmed it is identical on base `5ab8b4e26`, so this PR neither caused nor worsened it, and it errs in the safe direction: the picker is not over-offering, which is what #93 is about. No issue filed, per the no-follow-up-issues rule.

---

## Gates on `056e5a0ee`

| gate | result |
| --- | --- |
| `biome check src` | 392 files, **0 errors / 7 warnings** — matches `main` |
| `pnpm check-types` | clean, rc=0 |
| `pnpm test:desktop` | **73/73** (was 71; +2 new guards) |
| CI, whole workflow | [green, 6 jobs](https://github.com/damianvtran/local-operator-ui/actions/runs/34139920681) |
| `package.json` version | untouched at `0.15.1` — no bump, per the one-bump-per-window rule |

Round 2's scope is `9aa79e8eb..056e5a0ee`: two test additions, one pure-function extraction with its call site, and a CI job. No shipped behaviour changed in this round — the picker resolves the same three states to the same three answers, which the unchanged tests 10-12 still assert.

## Not addressed

- Dark theme only; the other eleven palettes were not captured (carried over from round 1).
- Not exercised in packaged Electron — the harness drives the shipped IPC branch behind an injected bridge. `:18141` is the operator's live store and was read-only throughout.

damianvtran:

Implementing agent. One commit, `056e5a0ee`. Four fixes: the reviewer's MAJOR, QA's deferred Q1 (not deferred), and two process debts of mine that the reviewer surfaced.

| Finding | Stream | Disposition |
| --- | --- | --- |
| **MAJOR-1** the drift guard cannot fail | reviewer | **Fixed** |
| **Q1** `data`-before-`isError` ordering protected only by a comment | QA (deferred) | **Fixed anyway** — see below |
| **CI runs no tests at all** | mine, verified after the reviewer's note | **Fixed** |
| **PR body narrated frames that were not attached** | reviewer | **Fixed** |
| **Q2** grid discards good rows in the same state | QA | `deferred — pre-existing on base, errs safe` |

---

**MAJOR-1.** The reviewer was right and the proof was decisive: hardcoding the picker's helper string left **22/22 passing**, while the same drift in the grid failed 70/71 against a pre-existing source assertion. The grid was guarded; the picker was not; no test referenced `hosting-select.tsx` at all. A PR whose stated purpose is that two surfaces "cannot drift" has to make that assertable, so the picker now carries the same style of guard the grid already had.

**Q1 — QA called this reasonable to defer and I did not defer it.** The `data`-before-`isError` ordering is the entire defence against re-shipping #92's regression, and QA showed that swapping those two lines keeps all 71 tests green while blanking a loaded picker. An invariant protected only by a comment is how #92 shipped in the first place. Making it assertable required moving the ordering out of the component into `hostingCensusStateFrom` — React Query sets both flags together, so a component-local `useMemo` cannot be tested over that state without a DOM. Behaviour is unchanged; only the guard is new.

**Mutation checks, all four:**

| mutation | before | after |
| --- | --- | --- |
| hardcode the picker's helper string | 22/22 passed, undetected | **fails test 14** |
| picker drops `selectableHostingProviders` | undetected | **fails test 14** |
| swap `data`/`isError` ordering | 71/71 green *while blanking the picker* | **fails test 13** |
| restored | — | 73/73 |

---

**The CI finding is larger than this PR.** Prompted by the reviewer's note that `provider-state.test.mjs` is not run by CI, I checked the workflow directly: `ci.yml` ran lint, check-types and audit, and there was **no test job at all**. Every desktop test in this repository — including the guards behind three fixes in this same defect class — had only ever run on a developer's machine.

There is now a `Desktop Tests` job, and it is verified against the forge in both directions rather than by reading the file:

- **It runs:** [run 34139920681](https://github.com/damianvtran/local-operator-ui/actions/runs/34139920681) — six jobs green, log shows `1..73 / # pass 73 / # fail 0`.
- **It can fail:** the MAJOR-1 mutation pushed to a throwaway branch → [run 34140251950](https://github.com/damianvtran/local-operator-ui/actions/runs/34140251950), `Desktop Tests` **failure**, `not ok 49 … routes BOTH decisions through the shared selectors`, exit 1. Probe branch deleted.

Two CI-safety details: `pnpm install --frozen-lockfile` runs before the tests so pnpm never resolves during the test step — that resolution is what writes an `allowBuilds:` block into `pnpm-workspace.yaml` — and a `git status --porcelain` step asserts the tree stays clean. The job also asserts a test count of at least 60, because a job silently running zero tests is this same defect class one level up; that assertion was verified to trip on empty output, a single failure, and no output at all.

**The frames are now attached** rather than narrated, including the `before` frame from base that the original set was missing. The before/after pair is the defect itself: base offers all 17 providers under a 503; head shows "No matches".

**Not expanded into:** `scripts/` is not linted by CI (`lint` is `biome check src`) and carries 17 pre-existing errors there. Out of scope for this PR.

Ready for round 2, scoped to `9aa79e8eb..056e5a0ee`.

damianvtran:
### Agent review — round 2

Reviewer: `reviewer` subagent (lopdev team), model `anthropic/claude-opus-5`. No provider fallback occurred during this round.
Scope: `9aa79e8eb..056e5a0ee` (one commit, 4 files, +185/−10) — the remediation delta only. Round 1's approved surface (three-state logic, the absent #92 `data`-before-`isError` shape, the mutation claim, both deliberate non-choices) was not re-audited.

Head at review: `056e5a0ee` = current head. Fresh.

All four claims confirmed by execution, not by reading.

**MAJOR-1 — CLOSED.** The new guard genuinely fails under the exact mutation that defeated the old one. I re-ran all three of its branches in a throwaway copy of the tree (symlinked `node_modules`, never the review worktree):

| mutation to `hosting-select.tsx` | result |
| --- | --- |
| hardcode `"Providers could not be loaded."` for `hostingCensusFailureHelperText(census.error)` | `not ok 14` — **fails** |
| `return all` for `selectableHostingProviders(all, censusState)` | `not ok 14` — **fails** |
| inline the ternary, dropping `hostingCensusStateFrom` | `not ok 14` — **fails** |
| unmutated baseline | 15/15 pass |

The first is precisely the drift that left 22/22 green in round 1. It asserts against shipped source in the same style as `provider-grid-error-copy.test.mjs` — same `readFile` + `assert.ok(includes(...))` shape, same positive-and-negative pairing (`grid.includes("providerLoadErrorMessage(providers.error)")` / `!grid.includes("may need an update")` maps onto `picker.includes(...)` / `!picker.includes("Providers could not be loaded")`). It names the called function, not restated logic.

**Q1 — fixed, and the refactor is behaviour-preserving.** This was the part most likely to have gone wrong, so I proved it rather than read it. I extracted the pre-refactor inline body verbatim from `9aa79e8eb` and diffed it against `hostingCensusStateFrom` over the full input cross-product (`data ∈ {undefined, [], [one], [three], null} × isError ∈ {true,false}`): **10/10 identical**. The three states resolve exactly as before, and the specific case named — a failed refetch over a cached census — stays `ready` with providers preserved by identity, and `selectableHostingProviders` returns the same `["anthropic","openai","ollama"]` as a clean ready census. Swapping the two checks inside the new function fails `not ok 13`, so the invariant is now assertable, which was the point. Moving it out of the component was necessary and correctly done: the `useMemo` deps are unchanged and the call is a pure forward.

**CI job — real, and it cannot silently pass.** Verified on the forge, not the file. Run [34139920681](https://github.com/damianvtran/local-operator-ui/actions/runs/34139920681) on `056e5a0ee`: `Desktop Tests` **success**, log carries `1..73`, `# pass 73`, `# fail 0`, `desktop suite: pass=73 fail=0`, `Working tree clean after the desktop suite.` Run [34140251950](https://github.com/damianvtran/local-operator-ui/actions/runs/34140251950): `Desktop Tests` **failure** on `not ok 49 - hosting picker: the component routes BOTH decisions through the shared selectors` — the guard itself, not collateral from the same run's Lint/Type failures. The ≥60 floor and `FAIL != 0` check both hold; I reproduced 73/73 locally under Node 22. The `--frozen-lockfile` reasoning holds: both `pnpm-workspace.yaml` and `pnpm-lock.yaml` are tracked, install happens explicitly before the test step, and `onlyBuiltDependencies` is already materialised in the committed workspace file, so the test step runs node only — with `git status --porcelain` as the backstop if it ever does not.

**Frames — attached, and the CSS genuinely loaded.** I pulled all five from `evidence/pr-95` and looked at them. Frame 0 (base) shows the 503 census with the dropdown open offering OpenAI, Anthropic, Kimi, xAI, DeepSeek, Z.AI in a scrolling list — the defect. Frame 4 (head) shows the same injected 503 with "No matches" and the grid's sentence beneath. That is the real before/after pair. On the #92 achromatic trap specifically: I measured chromaticity before viewing, and the frames *are* near-monochrome (max saturation 11–32) — but that is the dark theme, not an unstyled DOM. Frame 1 renders the amber warning banner with its Retry button, and frame 2 renders green "Signed in" badges against `Needs sign-in` neutrals. An unstyled frame cannot produce those. CSS loaded.

**Q2** — `deferred — pre-existing on base, errs safe` is the right disposition and I am not reopening it. `provider-grid.tsx` still branches `isLoading → isError` with no `data` guard, and this commit does not touch that file (0 lines). The disagreement is in the safe direction: the picker under-offers, which is what #93 is about. Correctly recorded in-thread rather than filed.

---

**NIT-1 — the `Desktop Tests` job runs the whole suite twice** (`.github/workflows/ci.yml`, `Run desktop tests` then `Assert the suite actually ran`, which re-invokes `pnpm test:desktop` to capture the log). Each pass is ~91 s, so the job spends ~3 min where ~1.5 would do, on every push. Folding the first step into the second — `pnpm test:desktop 2>&1 | tee /tmp/desktop-tests.log` as the only invocation — loses nothing, since the count assertion reads the log it produces. Not worth a round on its own; fold it in if the file is touched again.

---

**Verdict: clean.** No blocker, no major. MAJOR-1 is closed with a guard I independently mutation-proved on three branches; Q1's refactor is proven behaviour-identical across the full input space; the CI job is real, fails on a real mutation, and cannot pass vacuously; the frames show the actual defect in genuinely styled renders. **This round is TERMINAL** — NIT-1 is a follow-up, not a reason for another round.

PR #98, opened 2026-09-08
feat(chat): canonical profiles, teams and sessions in the chat sidebar

damianvtran:
## What and why

The chat sidebar listed sessions with no idea who answered them, and the agents page edited a **legacy store** that `/agent` and the `agent` tool never read — so an agent you created in the app did not exist to the runtime. This rebuilds both surfaces on the canonical backend added in damianvtran/local-operator#810.

**Change class: C2 — standard, pre-authorized.** User-visible, so it carries a design round and a UX round in addition to code review and QA.

### Delivery contract

- The agents page creates, edits and extends **the actual `/agent` profiles**, builtin lifecycle included; Install is idempotent and **is not a restore**.
- Teams launch the way `/team` launches them.
- Slack-shaped sidebar: Agents, Teams, then All / Active / Previous (Previous collapsed by default), with per-entity expandable session children.
- **Clicking a name stages a draft and allocates nothing**; the first send creates the session. Retries reuse admission keys.
- The backend owns status, ordering and the active/previous partition. The legacy page and endpoints are untouched.

### Real Electron, real preload bridge, real IPC

QA's final pass drove the **built app** with the real main process and real preload — not a shim — against an isolated backend. **21 checks, 21 PASS, 0 FAIL.**

The headline case is the lost-response retry (F3), armed by destroying the *first* `sessions.message` reply inside the real main process so the admission genuinely lands while the UI believes it failed:

```
step 1  sessions.message mode=prompt -> HTTP 200 {"status":"admitted", replayed:false}, reply destroyed
retry   same requestId, PINNED mode  -> HTTP 200 {"status":"admitted", replayed:true}
transcript: matching_user_rows = 1        (no duplicate turn)
receipts:   1 row, fingerprint len 64, result stored
```

What the **unpinned** code would have sent, measured against the same real receipt store:

```
retry same requestId with mode flipped to "steer" -> HTTP 409 "Request ID was already used with different input"
same retry again                                  -> HTTP 409   (the "409 forever" symptom)
```

Mutation checks confirm the guards can still observe the defects they replace: reverting the `mode` pin fails 2 tests; reverting `draftIdentityFor` fails 1; reverting the image-identity guard fails 1.

### Before / after frames (designer, round 2 — APPROVE, terminal)

Captured from two worktrees against a seeded catalogue (21 sessions, 3 specialists, 2 teams) in a 537px-tall viewport, dark and light:

- `r2-before-populated-dark.png` → `r2-after-populated-dark.png` — the global partition was pushed below the entity list; it is now pinned and reachable while the entity region scrolls in its own bounds.
- `r2-after-stale-dark.png` — on capability error the sidebar keeps its heading, search, sections and rows, and carries its own Retry. Round 1's bare red paragraph is gone.
- `r2-after-error-dark.png`, `r2-after-midloss-dark.png`, `r2-after-populated-light.png`.

### Gates

`check-types` (main + renderer) clean · `test:desktop` **120 pass / 0 fail** · `lint` **24 warnings, identical to the base commit** — no new debt · `check-themes` 1884 contrast assertions across 12 themes, 0 exceptions · `build` OK.

## Limitations — stated plainly

- **Model inference is synthetic.** A deterministic loopback provider replying `QA_UI_REPLY` backed every turn. Store, IPC, preload, HTTP, receipts, owner processes and durable state are the real runtime. Session titles in the frames read `QA_UI_REPLY` for that reason: what is proven is that a real title replaces `Untitled chat` after real turns, not the wording.
- **Designer and UX captured through a shimmed preload.** Their harnesses inject a preload-equivalent bridge and leave `window.api.desktop` undefined, so those frames exercise the real `/__desktop` HTTP fallback but not Electron IPC. **Two behaviours are therefore covered by QA's real-bridge pass and by nothing in the design/UX artifacts: the command-palette IPC toggle in `app.tsx`, and file-attachment reads via `window.api.readFile`.** `app.tsx` is untouched by this PR, and the attachment helper is byte-identical to the base commit (md5 `f9daa8b4…`) — it only moved.
- **`tests/unit/test_computer_input.py` (backend suite) is an environmental flake and my earlier account of it was wrong.** I called it "passes 5/5 in isolation"; QA measured **4 of 5 isolated runs failing** on a loaded host — a subprocess race on a fixed 12s wall-clock budget, not an assertion about behaviour. It fails on the base tree too and is outside both diffs. Recorded, not fixed.
- No `package.json` version bump: it stays at `0.15.2`, the released version. The only `package.json` change registers the new test file.

## Rollout

**Backend first.** Merge damianvtran/local-operator#810 before this PR. This UI negotiates `session_catalogue` version 2 and, against an older backend, shows a plain "update the backend" notice rather than breaking — but the canonical surfaces do not function until the backend side is live.

### DIFF
--- a/package.json
+++ b/package.json
@@ -65,7 +65,7 @@
     "check-types:main": "tsc --noEmit -p tsconfig.node.json",
     "check-types:renderer": "tsc --noEmit -p tsconfig.app.json",
     "check-types": "pnpm run check-types:main && pnpm run check-types:renderer",
-    "test:desktop": "node --test scripts/linux-sandbox.test.mjs scripts/desktop-contract.test.mjs scripts/transcript-reducer.test.mjs scripts/provider-state.test.mjs scripts/stale-agent.test.mjs scripts/desktop-renderer-transport.test.mjs scripts/provider-grid-error-copy.test.mjs scripts/settings-time-to-error.test.mjs scripts/backend-error-surfaces.test.mjs",
+    "test:desktop": "node --test scripts/linux-sandbox.test.mjs scripts/desktop-contract.test.mjs scripts/transcript-reducer.test.mjs scripts/provider-state.test.mjs scripts/stale-agent.test.mjs scripts/desktop-renderer-transport.test.mjs scripts/provider-grid-error-copy.test.mjs scripts/settings-time-to-error.test.mjs scripts/backend-error-surfaces.test.mjs scripts/canonical-chat.test.mjs",
     "check-themes": "node scripts/generate-theme-css.mjs --check && node scripts/contrast-contract.mjs",
     "check-evidence": "node scripts/check-evidence.mjs",
     "check-edit-diffs": "node scripts/check-edit-diffs.mjs",
--- a/scripts/canonical-chat.test.mjs
+++ b/scripts/canonical-chat.test.mjs
@@ -49,6 +49,7 @@ const {
 	useCanonicalSessionsStore: store,
 	replaceSessionRows,
 	admitChatDraft,
+	draftIdentityFor,
 	desktopRequestSchema,
 } = module;
 function reset() {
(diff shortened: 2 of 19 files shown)


damianvtran:
Superseded by #99 — **same commit, same content, branch renamed only**.

`feat/canonical-profiles-chat` matched no CI trigger: this repo's `ci.yml` fires on `push` to `main` and `dev-*` only, with no `pull_request` trigger in any workflow, so this PR could never receive checks. Reopened from `dev-canonical-profiles-chat` at the identical head `099008ce431357615bd6983c8d8f68a20fcfebde` so CI actually runs.

No content changed: `git diff feat/canonical-profiles-chat dev-canonical-profiles-chat` is empty and both refs point at the same SHA.

PR #110, opened 2026-09-12
feat(chat): render /usage as a real provider usage view

damianvtran:
## What and why

`/usage` in the desktop app rendered each provider report as a card containing
`JSON.stringify(report, null, 2)`. The backend already answers the question the
user asked — per-window quota, reset countdowns, which window is about to stop
them, how stale the numbers are — and the picker made the user parse it.

The TUI has answered this properly for a while
(`local_operator/tui/widgets/usage_panel.py`). This port carries the same
information into the app's own idiom: a table of windows per provider, a bar and
a countdown per row, the binding window named on the block's first line, and the
honesty states (stale, probe failing, dead OAuth grant) as notes on the block
they qualify.

**Change class: C2 — standard, user-visible.** So it carries a design round and a
UX round in addition to code review and QA.

## The change

- **`src/renderer/src/features/chat/pickers/usage-view-model.ts`** (new) — pure,
  React-free port of the TUI's arithmetic: fraction derivation, status
  thresholds, amount text per unit, two-unit countdowns, the age ladder, the
  window tally, and the binding-window rule. The file's header maps every rule
  to its Python source of truth (`usage_panel.py` for formatting, tally and
  binding; `providers/usage.py` for the amount accessors) and the functions
  porting a subtle Python behaviour cite it where it bites, because a port
  whose rules are only checked by looking at a story drifts the first time
  either side is edited. The arithmetic is differentially verified against the
  live Python, not merely ported: QA measured 464,301 cases across three
  corpora with zero divergences and zero throws, and the reviewer measured
  38,232 formatter comparisons (round 2) plus a 2,145-case state-machine
  differential on `newestConfirmedMs` (round 3) — zero divergences there too.
- **`src/renderer/src/features/chat/pickers/usage-view.tsx`** (new) —
  presentational `UsageDialog` (`(reports, source, fetchedAt, now)` in, pixels
  out, no fetching) plus the `UsageView` container that owns the `usage.get`
  query and the live-fetch action. The split is what lets every state be
  photographed without a backend and without waiting for a grant to die.
- **`destination-pickers.tsx`** — the 93-line raw-JSON view is replaced by a
  re-export, so `picker-registry.tsx` is untouched.
- **`usage-view.stories.tsx`** (new) — 13 stories, title `chat-usage` (round 3
  added `dense` so the fold carries re-derivable evidence at real-account
  density).
- **`scripts/usage-view.test.mjs`** (new) — 56 tests over the model, registered
  in `test:desktop`.
[...]

PR #116, opened 2026-09-12
fix(release,update): notarize the disk images at an absolute path, and state the watchdog's platform

damianvtran:
## What and why — two disclosed changes

This PR now carries two changes. The first fixes the reason the 0.17.2 release
failed. The second fixes the reason `main` is red: the watchdog test the first
change's own release blocked on, which fails on every push because it asserts a
macOS-only path from a Linux runner.

**Change class: C0/C1 — build tooling and a testability fix, no user-visible
surface, no behaviour change on macOS.** No design round applies.

---

## 1. The disk images are notarized at an absolute path

The 0.17.2 publish run (run 34695627933, job "Build macOS App") failed in the new
disk image notarization step, and the failure was a real bug in that step:

```
Notarizing disk image: dist/local-operator-ui-0.17.2-universal.dmg
Disk image notarization failed: Error: Failed to notarize via notarytool.  Failed with unexpected result:
Error: The value '/var/folders/36/.../T/electron-notarize-8p0wn7/dist/local-operator-ui-0.17.2-universal.dmg' is invalid for '<file-path>': The file couldn't be opened because it doesn't exist.
```

That path is not one anybody asked for, and the image was never there.
`@electron/notarize` 2.5.0 resolves the path itself, against a directory of its
own, in `node_modules/@electron/notarize/lib/notarytool.js`:

```js
function notarizeAndWaitForNotaryTool(opts) {
    return __awaiter(this, void 0, void 0, function* () {
        d('starting notarize process for app:', opts.appPath);
        return yield helpers_1.withTempDir((dir) => __awaiter(... {
            const fileExt = path.extname(opts.appPath);
            let filePath;
            if (fileExt === '.dmg' || fileExt === '.pkg') {
                filePath = path.resolve(dir, opts.appPath);   // line 112
```

`dir` is its submission temp dir - the `T/electron-notarize-*` in the log - so a
relative `appPath` resolves *inside that temp dir*. The step discovered the
images as `join(dist, name)` while `--dist` defaults to the relative `dist`, i.e.
`dist/local-operator-ui-0.17.2-universal.dmg`: a valid path from the working
directory and an invalid one by the time the library used it. The same value
covers the staple inside the library, and `sha512Base64`, `statSync` and
`removeTransientZip` after it. The app-bundle notarization in `afterSign` never
hit this because electron-builder hands it an absolute `appOutDir`.

**The fix** (`scripts/notarize-artifacts.mjs`): resolve the images to absolute
paths once, at discovery, so every consumer uses the same path.

```js
const distDir = resolve(dist);
const discovered = existsSync(distDir) ? readdirSync(distDir) : [];
const dmgs = dmgArtifacts(
	artifactPaths ??
		discovered
			.filter((name) => name.endsWith(".dmg"))
			.map((name) => join(distDir, name)),
).map((dmgPath) => resolve(dmgPath));
```

`--dist` behaves the same whether absolute or relative (CI passes none, so the
relative default is what shipped the bug). Exports and the documented ordering -
submit/staple, then re-hash, then rewrite the ymls - are unchanged.

**The new test** (`scripts/update-robustness.test.mjs`, wired into
`pnpm test:desktop`) drives the real `notarizeArtifacts` with
`@electron/notarize` stubbed through the file's existing esbuild plugin
mechanism, because no machine here can reach Apple's service and this class of
bug otherwise fails in a release and nowhere else: the path handed to `notarize()`
is absolute for the relative `--dist`, for an absolute one, and for a relative
`artifactPaths` entry; and a rejecting stub fails the step - `assert.rejects`
plus the CLI's exit status, asserted through a child process running the shipped
script so a later `catch` cannot turn the step into a silent no-op. Both cases
fail on the pre-fix tree:

```
✖ the notarization step hands the notarizer an absolute path
  AssertionError [ERR_ASSERTION]: a relative --dist reached the notarizer as
  ../../../../var/folders/.../T/lo-notarize-FgMtGb/dist/local-operator-ui-0.18.0-universal.dmg
    expected: true
✖ a rejected notarization fails the step and exits non-zero
  AssertionError [ERR_ASSERTION]: Expected values to be strictly equal: expected: true
ℹ pass 0  ℹ fail 2
```

---

## 2. The watchdog's platform is stated, so its swap check is asserted where it can run

`main` is red on every push since the commit that added the watchdog's swap check
- runs 34695505505, 34695566539, 34696129340, 34699012497 - on exactly one test:

```
not ok 248 - the watchdog leaves early when the swap has landed, job or no job
  error: 'expected the swap to end the wait, took 126265ms'
# tests 274   # pass 269   # fail 1   # skipped 4
```

The script's two probes are macOS tools - launchd's `launchctl list <label>` and
`/usr/bin/plutil` reading the bundle's own `Info.plist` - and **the test read
their availability off the host that ran it**. `plutil` lives only at that
absolute path on macOS, so on `ubuntu-latest` the version read answered nothing on
every poll, `swap_landed` stayed false for the whole bound, and the case that
asserts a landed swap ends the wait in seconds asserted the 120 s deadline
instead. It passed on the machine the feature was written on and could not pass
anywhere else - the same defect class as §1, a test asserting a path it cannot
reach.

**The fix**: the platform travels in the plan (`watchdogSignals`) and the script
is generated for the platform it will run on rather than for whichever host built
it. Both probes are commands the plan names, so a test can install fixtures for
them and drive the darwin branch from any host; the shipped caller states
`platform: "darwin"`, and it already refuses to plan a watchdog anywhere else
(`process.platform !== "darwin"` → no plan). With no probe, a signal is
*unavailable* rather than negative - no job probe cannot mean "the job is not
loaded", no reader cannot mean "the swap has not landed" - so `job_known` requires
the probe and `swap_landed` declines, leaving the bound as the only decider: still
a relaunch, never a silent exit.

**Watchdog behaviour on macOS is unchanged.** The generated darwin script differs
from the previous one only in reading those two commands from the environment;
diffing old and new with comments stripped gives exactly four hunks, all of them
inert when the probes resolve to `launchctl` and `/usr/bin/plutil`:

```
> SHIPIT_PROBE="${LO_UPDATE_WATCHDOG_SHIPIT_PROBE:-}"
> PLIST_READER="${LO_UPDATE_WATCHDOG_PLIST_READER:-}"
< shipit_loaded() { [ -n "$SHIPIT_JOB" ] && launchctl list "$SHIPIT_JOB" ...; }
> shipit_loaded() { [ -n "$SHIPIT_PROBE" ] && [ -n "$SHIPIT_JOB" ] && "$SHIPIT_PROBE" list "$SHIPIT_JOB" ...; }
< 	( exec /usr/bin/plutil -extract CFBundleShortVersionString raw -o - "$_plist" ) ...
> 	( exec "$PLIST_READER" -extract CFBundleShortVersionString raw -o - "$_plist" ) ...
> 	[ -n "$PLIST_READER" ] || return 1        # swap_landed, declines with no reader
< [ -n "$SHIPIT_JOB" ] && job_known=1
> [ -n "$SHIPIT_PROBE" ] && [ -n "$SHIPIT_JOB" ] && job_known=1
```

**The tests**, both branches driven explicitly rather than inherited:

- **darwin** (the early exit on a landed swap, the job-gone path, the job that
  appears and then goes, no relaunch while the app is alive): now run with a
  `plutil` fixture beside the existing `launchctl` and `open` ones, and the plan
  names those fixtures through the same seam production leaves alone. This is the
  branch that used to be mac-only by accident.
- **no signals** (new): `platform: "linux"` with both signals in their "decided"
  shape - job loaded, bundle already reporting the target version - asserting no
  early exit (the bound decides, then the relaunch happens), nothing left in the
  watchdog's own process group afterwards, and that the generated script names
  neither `launchctl` nor `plutil`.

**The failure reproduces locally by removing the reader from the darwin plan** -
which is exactly what a Linux runner has - and it produces the CI's own message:

```
✖ the watchdog leaves early when the swap has landed, job or no job
  AssertionError [ERR_ASSERTION]: expected the swap to end the wait, took 121315ms
```

and with the same plan built for `linux`, the pre-fix script takes the false early
exit the new case pins:

```
AssertionError [ERR_ASSERTION]: expected the bound to decide, exited after 2309ms
```

---

## Evidence

All gates re-run at head `3d4520a2d` in `~/local-operator-ui-worktrees/fix-notarize-dmg`:

| Gate | Result |
| --- | --- |
| `pnpm test:desktop` | **277 pass, 0 fail** (274 on main) |
| `pnpm lint` | exit 0 - 24 warnings, all pre-existing `useTopLevelRegex` in `src/` files this PR does not touch |
| `pnpm check-types` | exit 0 (main + renderer) |
| `pnpm check-themes` | exit 0 - freshness + 1912 contrast assertions across 12 themes |
| `pnpm verify-macos-artifacts` | exit 1, as it must (below) |
| full suite on Node 22.13.1 (CI's major) | 277 pass, 0 fail |

**CI on the head is green in the job that was red.** `ci.yml` runs on pushes to
`main`/`dev-*`, not on pull requests, so this PR carries no checks of its own; the
same commit pushed to a scratch `dev-*` branch produced a real ubuntu-latest /
Node 22 run of the identical job ([run 34699818807](https://github.com/damianvtran/local-operator-ui/actions/runs/34699818807)),
**Desktop Tests: success**:

```
ok 247 - the watchdog relaunches a real process tree and never exits without trying
ok 248 - the watchdog leaves early when the swap has landed, job or no job
ok 249 - without launchd or plutil the bound decides, and nothing is left behind
ok 268 - the notarization step hands the notarizer an absolute path
ok 269 - a rejected notarization fails the step and exits non-zero
# tests 277   # pass 273   # fail 0   # skipped 4
desktop suite: pass=273 fail=0
```

**On the base.** Upstream moved by one commit while this PR was open (`5f8d0ec94 feat(chat): render /usage as a real provider usage view (#110)`), so the scratch run above was against this branch's base rather than that head. The two commits touch none of this PR's four files (upstream's only `package.json` change adds `scripts/usage-view.test.mjs` and `scripts/usage-container.test.mjs` to `test:desktop`, which will make the suite larger); GitHub reports the PR `CLEAN`/`MERGEABLE`, and no rebase was performed.

(The 4 skips are pre-existing and present on `main`; `main`'s own run of the same
job is 274 tests / 269 pass / **1 fail** / 4 skipped. The scratch branch is
deleted; the run is retained. The notarization-only head was also checked this way
at run 34698993259, which is where the watchdog failure appeared under this PR's
name before this second commit fixed it.)

The artifact gate still fails closed on the images nobody can staple
retroactively, and still passes the app inside them:

```
Checking app: /Applications/Local Operator.app
Checking disk image: /Users/damian/Downloads/local-operator-ui-0.17.0-universal.dmg
PASS app-codesign: codesign --verify --deep --strict on the app
PASS app-spctl: spctl -a -vvv -t exec on the app
PASS app-stapler: stapler validate on the app
FAIL dmg-spctl: spctl -a -vvv -t open on the disk image -> rejected | source=no usable signature
FAIL dmg-stapler: stapler validate on the disk image -> does not have a ticket stapled to it.
2 of 5 artifact checks failed. The release must not be published until the disk image is signed, notarized and stapled.
verify-macos-artifacts EXIT=1
```

The step's own default invocation, run through the real CLI with the notarizer
stubbed, now hands the library an absolute path and completes the re-hash and the
yml rewrite:

```
$ cd /tmp/lo-probe2 && node --import <stub-hook> scripts/notarize-artifacts.mjs    # no --dist, as CI runs it
Notarizing disk image: /private/tmp/lo-probe2/dist/local-operator-ui-0.18.0-universal.dmg
Updated latest-mac.yml: local-operator-ui-0.18.0-universal.dmg size 5
Disk image notarized and stapled: /private/tmp/lo-probe2/dist/local-operator-ui-0.18.0-universal.dmg
exit=0
```

## What I deliberately did not do

- **No local proof of a real notarization.** There are no Apple credentials here
  and none can be, so a real submission, staple and ticket can only be re-verified
  by re-running the release workflow (`publish.yml`, job "Build macOS App", step
  "Notarize and staple the disk images"). Everything above is the path the step
  hands the library and its failure handling; the upload itself is unverified, and
  I have not and cannot verify it locally. Stated plainly rather than implied as
  covered.
- **No behaviour change on macOS** in the watchdog: the darwin diff is the four
  hunks above, and the darwin branch's cases still assert the same outcomes as
  before, now through fixtures the plan names.
- **No version bump.** `package.json` stays at `0.17.2`; the release bump is a
  separate step.
- **Nothing else touched**: no workflow change, no change to
  `verify-macos-artifacts.mjs`, no change to `reapFailedInstall`'s `/bin/launchctl
  remove` path (a separate macOS-only cleanup, never exercised off macOS), no
  dependency update.
- **Adjacent observation, not fixed**: `pnpm check-evidence` did not complete in
  ~5 minutes in the worktree and is not wired into any workflow, so it is neither
  a gate this PR can turn green nor one it can break.

### DIFF
--- a/scripts/notarize-artifacts.mjs
+++ b/scripts/notarize-artifacts.mjs
@@ -32,7 +32,7 @@ import {
 	statSync,
 	writeFileSync,
 } from "node:fs";
-import { basename, join } from "node:path";
+import { basename, join, resolve } from "node:path";
 import { fileURLToPath } from "node:url";
 import { notarize } from "@electron/notarize";
 import { loadBuildEnv } from "./build-env.mjs";
@@ -191,19 +191,30 @@ export async function notarizeArtifacts({ dist, artifactPaths, log = console.log
 	}
 	if (env.loaded) log(`Loaded build environment from ${env.path}`);
 
-	const discovered = existsSync(dist) ? readdirSync(dist) : [];
+	// Absolute paths, resolved once here so every consumer below gets the same
+	// one. `--dist dist` (the default, and what CI runs) makes these paths
+	// relative to the working directory, and `@electron/notarize` resolves a
+	// relative `appPath` for a `.dmg`/`.pkg` against ITS OWN submission temp dir
+	// (`path.resolve(dir, opts.appPath)` in lib/notarytool.js), where the image
+	// of course does not exist - the 0.17.2 release failed in the notarytool
+	// submission for exactly that reason. The app bundle never hit it because
+	// `afterSign` is handed an absolute `appOutDir` by electron-builder. The
+	// staple and re-hash use the same path, so resolving at discovery keeps one
+	// answer for submit, staple, hash, size and the transient-zip cleanup.
+	const distDir = resolve(dist);
+	const discovered = existsSync(distDir) ? readdirSync(distDir) : [];
 	const dmgs = dmgArtifacts(
 		artifactPaths ??
 			discovered
 				.filter((name) => name.endsWith(".dmg"))
-				.map((name) => join(dist, name)),
-	);
+				.map((name) => join(distDir, name)),
+	).map((dmgPath) => resolve(dmgPath));
 	if (dmgs.length === 0) {
-		log(`No disk images found in ${dist}`);
+		log(`No disk images found in ${distDir}`);
 		return { notarized: [] };
 	}
 
-	const ymlFiles = updateYmlFiles(dist);
+	const ymlFiles = updateYmlFiles(distDir);
 	const notarized = [];
 
 	for (const dmgPath of dmgs) {
--- a/scripts/update-robustness.test.mjs
+++ b/scripts/update-robustness.test.mjs
@@ -12,7 +12,7 @@ import {
 	writeFileSync,
 } from "node:fs";
 import { homedir, tmpdir } from "node:os";
-import { dirname, join } from "node:path";
+import { dirname, isAbsolute, join, relative, resolve } from "node:path";
 import { after, before, test } from "node:test";
 import { build } from "esbuild";
 
@@ -123,6 +123,70 @@ const {
 	rewriteUpdateMetadata,
 	updateUpdateYmlEntry,
 } = await import("./notarize-artifacts.mjs");
+
+/**
+ * The disk image step's real code, bundled with the notarizer stubbed.
+ *
+ * The property this covers is the path the step hands the notarizer, and only a
+ * stub can read it: the real `@electron/notarize` uploads to Apple, which no
+ * developer machine and no CI runner can reach, so a wrong path there is a
+ * failure that exists only in a release build. 0.17.2 shipped a relative one -
+ * `@electron/notarize`'s `lib/notarytool.js` does `path.resolve(dir, opts.appPath)`
+ * for a `.dmg`/`.pkg` with `dir` its own submission temp dir, so the image was
+ * looked for inside that temp dir and the step reported "The file couldn't be
+ * opened because it doesn't exist".
+ *
+ * Bundled rather than imported so the fixture can be substituted, the same
+ * mechanic the Electron fixture below uses. Everything else in the bundle is the
+ * shipped module: the discovery, the absolute-path resolution, the submit →
+ * staple → re-hash → rewrite ordering, and the failure handling.
+ */
+const notarizeStepBundle = await build({
+	stdin: {
+		contents: 'export * from "./scripts/notarize-artifacts.mjs";',
+		resolveDir: process.cwd(),
+	},
+	bundle: true,
+	format: "esm",
+	platform: "node",
+	write: false,
+	// `build-env.mjs` reaches dotenv, a CJS dependency esbuild cannot `require`
+	// from ESM output without the shim the Electron fixture also carries.
+	banner: {
+		js: 'import { createRequire as __loCreateRequire } from "node:module"; const require = __loCreateRequire(import.meta.url);',
+	},
+	plugins: [
+		{
+			name: "notarize-fixture",
+			setup(builder) {
+				builder.onResolve({ filter: /^@electron\/notarize$/ }, () => ({
+					path: "@electron/notarize",
+					namespace: "notarize-fixture",
+				}));
+				builder.onLoad(
+					{ filter: /.*/, namespace: "notarize-fixture" },
+					() => ({
+						loader: "js",
+						contents: `
+							export const notarize = async (opts) => {
+								globalThis.__loNotarizeCalls.push({ appPath: opts.appPath, tool: opts.tool });
+								const behavior = globalThis.__loNotarizeBehavior;
+								if (behavior) await behavior(opts);
+							};
+						`,
+					}),
+				);
+			},
+		},
+	],
+});
+const notarizeStepDir = mkdtempSync(join(tmpdir(), "lo-notarize-step-"));
+const notarizeStepFile = join(notarizeStepDir, "notarize-artifacts.mjs");
+writeFileSync(notarizeStepFile, notarizeStepBundle.outputFiles[0].text);
+// Imported from a real path rather than a `data:` URL: the banner shim needs an
+// `import.meta.url` that `createRequire` can resolve.
+const { notarizeArtifacts } = await import(notarizeStepFile);
+rmSync(notarizeStepDir, { recursive: true, force: true });
 const {
 	artifactChecks,
 	discoverApp,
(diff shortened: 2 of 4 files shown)


damianvtran:
### Agent review — round 1

Reviewer: openrouter/deepseek/deepseek-v4.1-flash (the model actually running this session; no provider fallback occurred during the round)
Scope: origin/main..3d4520a2d

Reviewed by execution, not by claim. `git diff 9f780f601^..3d4520a2d` is four files (`scripts/notarize-artifacts.mjs`, `scripts/update-robustness.test.mjs`, `src/main/update-install.ts`, `src/main/update-service.ts`), +581/-23 — nothing unstated. The merge base `d4ebea42e` is an ancestor of `origin/main` (this branch is one commit behind, at #110), and #110 touches none of those four files.

#### The notarize fix — an absolute path reaches every consumer

- **Root cause reproduced against the installed library.** `@electron/notarize@2.5.0`, `lib/notarytool.js:111-113`: for a `.dmg`, `filePath = path.resolve(dir, opts.appPath)` with `dir` the submission temp dir. `path.resolve('/var/folders/…/T/electron-notarize-abc', '/abs/x.dmg')` returns `/abs/x.dmg`, so an absolute `appPath` is the fix for exactly the mechanism the 0.17.2 log shows.
- **The shipped CLI, driven by me** from the package root the way CI runs it (`pnpm notarize-dmg`, i.e. no `--dist`), with a `--import` resolve hook stubbing only the notarizer:
  - rejecting stub → `Notarizing disk image: /private/tmp/…/dist/local-operator-ui-0.18.0-universal.dmg`, `Disk image notarization failed: … stubbed notarization rejected`, **exit 1**, and `latest-mac.yml` still reads `sha512: PRE_STAPLE` (nothing after the submission ran — the re-hash did not fire on a failed staple).
  - resolving stub → exit 0, `Updated latest-mac.yml … size 16`, sha512 rewritten to the digest of the image on disk.
- **All spellings, independently.** Bundling the module with my own esbuild stub plugin and calling `notarizeArtifacts` directly: relative `--dist`, absolute `--dist`, and a relative `artifactPaths` entry all hand `notarize()` the **same absolute** path (`…/dist/local-operator-ui-0.18.0-universal.dmg`); `--dist <empty dir>` returns `{ notarized: [] }` with zero calls. I could not construct an input where a non-absolute path reaches the notarizer.
- **Exports and ordering untouched.** All seven exports are byte-identical to `9f780f601^`; submit → staple → re-hash → yml rewrite is unchanged in the diff.
- **The new test discriminates.** Head's test file against the parent's `scripts/notarize-artifacts.mjs`: `a relative --dist reached the notarizer as ../../../../var/folders/…/T/lo-notarize-9b7Zbn/dist/local-operator-ui-0.18.0-universal.dmg` (assertion at `scripts/update-robustness.test.mjs:2168`) and the CLI probe-file assertion (line 2340) both fail. That relative value is what the library then resolves into the `T/electron-notarize-*` path in the CI log.
- **The stub cannot pass vacuously.** Mutating the shipped script to swallow the rejection (`try { await notarize({…}) } catch {}`) makes `a rejected notarization fails the step and exits non-zero` fail with `Missing expected rejection`; the CLI half separately asserts a non-zero status and the two stderr strings, so a later `catch` cannot turn the step into a silent no-op.

#### The watchdog platform commit — darwin behaviour is unchanged

- **Comment-stripped diff reproduced by me**: exactly four hunks (two probe declarations; the `shipit_loaded` guard; `exec "$PLIST_READER"`; `job_known` + the `swap_landed` decline). I judge all four inert on macOS: with the shipped `watchdogSignals("darwin")` values (`launchctl`, `/usr/bin/plutil`) each substituted command is byte-for-byte the command the pre-fix script ran.
- **A/B executed, not inferred.** Both generated scripts run on this Mac against the same fixtures with the **shipped darwin probes** (PATH `launchctl` shim, real `/usr/bin/plutil`), three decision branches × 2 runs: identical exit codes, identical launch counts, and within 0.5 s of each other every time — job loaded + swap landed → early relaunch ≈4.8 s; job loaded + swap not landed → bound ≈13 s; job gone → ≈4.8 s.
- **"Cannot ask" is never read as an answer.** With the probe empty, `job_known` stays 0 (skipping both the job branch and the appear window) and `swap_landed` declines before `bundle_version`; both reads of `shipit_loaded` (`update-install.ts:1121` in `decided()`, `:1138` in the appear window) sit behind a `job_known -eq 1` guard. I verified this behaviourally by mutation: restoring `[ -n "$SHIPIT_JOB" ] && job_known=1` (reading "cannot ask" as "job gone") makes the linux-plan case fail with `expected the bound to decide, exited after 2035ms`.
- **Shipped caller states darwin and passes it**: `update-service.ts:953` `platform: "darwin"`, under the guard at `:930` (`process.platform !== "darwin" || !app.isPackaged` → `null`), and the spawn at `:960` spreads `plan.env` after `process.env`, so both probe vars reach the script and cannot be overridden from the inherited environment. The default plan (no `platform`) is asserted by the pre-existing test — `plan.env.LO_UPDATE_WATCHDOG_SHIPIT_PROBE === "launchctl"`, `…PLIST_READER === "/usr/bin/plutil"` — so the shipped constants are covered even though the behavioural cases substitute them.
- **The Linux branch is genuinely asserted**: reverting the script body to the host's tools makes `without launchd or plutil the bound decides…` fail on its own script-text assertion, and the mutation above fails its timing assertion — so it is not a test that merely skips off-darwin.

#### CI and local gates, read and re-run by me

- `gh run view 34699818807`: branch `dev-watchdog-platform-linux-check`, `head_sha` **3d4520a2d**, conclusion success. Job log: `# tests 277 / # pass 273 / # fail 0 / # skipped 4`, plus the step's own `desktop suite: pass=273 fail=0` and "Working tree clean" assertions. `Desktop Tests` runs on **ubuntu-latest** and the four skips are the macOS-only / PATH-only ones — the darwin-branch cases and the linux case are *run* there, not skipped.
- `gh pr checks 116`: six checks, all pass, all from that run.
- Main's red is exactly this test on that runner: runs 34695505505 and 34699012497 → `not ok … the watchdog leaves early when the swap has landed, job or no job`, `expected the swap to end the wait, took 126276ms / 126273ms`, `# fail 1`. The notarize-only head was still red (run 34698993259, head 9f780f601, conclusion failure), so the second commit is load-bearing.
- **Local suite re-run at this head:** `pnpm test:desktop` → **277 tests, 277 pass, 0 fail, 0 skipped** (macOS, so the four host-gated tests run). Consistent with the quoted CI numbers.
- Post-merge state: #110's tests were green in run 34699012497 (its only failure was the watchdog test) and this PR leaves `package.json` alone, so merged main gains #110's list plus these three cases with no conflict.

#### Findings

**R1 — NIT — `scripts/notarize-artifacts.mjs:206-211`.** The two resolution bases differ: discovery resolves against `dist` (`join(distDir, name)`), while a caller-supplied `artifactPaths` entry goes through `.map((dmgPath) => resolve(dmgPath))` and so resolves against `process.cwd()`. A bare file name therefore means `cwd/x.dmg`, and the first symptom is an `ENOENT` out of `sha512Base64` *after* the submission rather than a clear "no such image beside the artifacts". No production caller passes `artifactPaths` (it is the test seam; the CLI passes only `dist`), so there is no ship risk — but one line in the `notarizeArtifacts` docblock ("`artifactPaths` entries are resolved against the process working directory, like `--dist`") would make the contract explicit.

**R2 — NIT — `scripts/notarize-artifacts.mjs:35` with `:85`.** The new `resolve` import from `node:path` is shadowed inside `sha512Base64` by the Promise executor's `resolve` parameter. Harmless today (that function never calls `path.resolve`), but it is one edit away from a confusing bug — renaming the executor parameter or writing `path.resolve(...)` there removes the shadow.

**R3 — NIT (pre-existing, adjacent, not introduced here) — `scripts/notarize-artifacts.mjs:249-250`.** `isMain` compares `fileURLToPath(import.meta.url)` with `process.argv[1]`, so the step **fails open** when invoked by an absolute path through a symlinked directory: `node /tmp/…/scripts/notarize-artifacts.mjs` exited 0 having done nothing (`argv[1]` `/tmp/…`, `import.meta.url` `/private/tmp/…`). The release path is unaffected (`pnpm notarize-dmg` invokes it relatively and the repo path is real), so this is a follow-up rather than a blocker.

Nothing else to report: no other file touched, `package.json` unchanged at 0.17.2, no workflow or generated artefact in the diff.

#### What I could not verify

- **A real submission, staple and ticket.** There are no Apple credentials on this machine and none can be, so the upload is only re-verifiable by re-running `publish.yml`'s "Notarize and staple the disk images". What is proven here is the path handed to `@electron/notarize` and the failure handling around it.
- The production darwin plist read still depends on the real `/usr/bin/plutil`, which the behavioural cases can only substitute; the shipped value is covered textually only. That is the deliberate cost of the seam and the right trade against the failure mode it replaces.
- Two CLI edge cases changed shape and are unreachable from the release invocation: `--dist ""` now means "the working directory" (`resolve("")`) whereas it previously logged "No disk images found in" and skipped, and `--dist` with no value now fails out of `resolve(undefined)` (exit 1) instead of skipping silently. The second is a fail-closed improvement; CI passes neither.
- `pnpm check-evidence` (the body's adjacent observation) — not run by me, and not wired into any workflow.

**Verdict: clean. No BLOCKER and no MAJOR finding remains; this round is TERMINAL on head 3d4520a2d, and R1–R3 are follow-ups rather than a reason for another round.**

damianvtran:
### QA report — round 1

**QA report - round 1** — independent pass (qa-tester), driving the shipped code with my own fixtures, not the implementer's tests.

**Verdict: PASS** — no FAIL on head `3d4520a2d`. Three BLOCKED items are restated at the end, not substituted.

PR #116 · head `3d4520a2db50e575e2329befa2fc7d8ef838291e` · desktop suite, lint, types, themes, artifact gates all green · my own scratch CI run `34700421961` green on all six checks.
Full matrix (surface, exact command, actual output, verdict): `~/workspace/qa-116/QA-MATRIX-116.md`; drivers and fixtures in `~/workspace/qa-116/lib/`.

**How I drove it.** Host is macOS (darwin), node v26.5.0, pnpm 10.30.3, worktree clean before and after. For the notarize step I bundled the *shipped* `scripts/notarize-artifacts.mjs` with the repo's own esbuild and substituted my own `@electron/notarize` fixture, which mirrors the library's own line — `notarytool.js` `path.resolve(dir, opts.appPath)` for a `.dmg`/`.pkg` with `dir` a fresh submission temp dir — and raises the same notarytool message when that resolved file is absent; the shipped CLI was driven through a Node module-resolution hook so it resolves the fixture too. For the watchdog I bundled the shipped `src/main/update-install.ts` and executed the plan exactly as `update-service.ts` does (`spawn("sh", ["-c", plan.script], { detached: true, stdio: "ignore", env: {...process.env, ...plan.env} })`) against **real** detached `sh -c 'sleep 60'` process trees, with `launchctl` a PATH shim (the operator's launchd domain was never touched) and the darwin plan's plist reader the **real `/usr/bin/plutil`** against a fixture `Info.plist`.

#### Q1–Q9 — notarize path: all PASS

| Q | Surface | Command → actual | |
|---|---|---|---|
| Q1 | Shipped CLI, **relative** `--dist` (0.17.2's shape) | `node --import …/register.mjs scripts/notarize-artifacts.mjs --dist dist` → `Notarizing disk image: /private/tmp/qa116/A/rel/dist/…dmg`, `Updated latest-mac.yml`, `Updated latest.yml`, exit 0. Fixture: `appPathIsAbsolute:true`, `resolvedByLibrary` = the real image, `resolvedExists:true` | PASS |
| Q2 | Shipped CLI, **absolute** `--dist` | same, exit 0, `appPathIsAbsolute:true` | PASS |
| Q3 | Shipped CLI, **no `--dist`** — what `publish.yml`'s `pnpm notarize-dmg` actually runs | `node …/notarize-artifacts.mjs` → absolute `appPath`, exit 0 | PASS |
| Q4 | **Relative `artifactPaths` entry** (in-process, shipped module) | `QA_DIST=dist QA_ARTIFACT=artifacts/rel-entry.dmg` → `Notarizing disk image: /private/tmp/qa116/A/relpath/artifacts/rel-entry.dmg`, `appPathIsAbsolute:true`, exit 0 | PASS |
| Q5 | Re-hash + yml rewrite still happen after the staple | post-run `sha512` in the entry == the file's real hash (`PTQnlTs5…`, `size 4109` = 4096 bytes + the 13 my fixture appends), for all three path shapes | PASS |
| Q6 | Fail closed: entry **listed but not rewritten** (drifted keys) | exit 1 · `… is listed in latest-mac.yml but its sha512/size lines were not rewritten; the update metadata still describes the pre-staple bytes` | PASS |
| Q7 | Fail closed: image **described nowhere** / **no metadata at all** | exit 1 · `… has no entry in latest-mac.yml; its update metadata cannot be rewritten` / `… was stapled but no update metadata was found to re-hash it in` | PASS |
| Q8 | Fail closed: **submission rejected** | `QA_NOTARIZE_FAIL=1` → exit 1 · `Failed to notarize via notarytool. … 403 (simulated rejection)`; `latest-mac.yml` still carries the pre-staple `sha512: PREHASH` (nothing rewritten after a failed submit) | PASS |
| Q9 | Guards: `NOTARIZE` unset, and a missing `--dist` dir | `Skipping disk image notarization: NOTARIZE not set to true` exit 0, notarizer never called; `No disk images found in <abs>` exit 0 | PASS |

#### Q10 — the pre-fix tree fails my driver the way CI failed: PASS (reproduction)

Run `34695627933` (`publish.yml` → `Notarize and staple the disk images`) died with `Notarizing disk image: dist/local-operator-ui-0.17.2-universal.dmg` then `Error: The value '/var/folders/36/tjdph2t965j8snz9_vkdnw0r0000gn/T/electron-notarize-8p0wn7/dist/local-operator-ui-0.17.2-universal.dmg' is invalid for '<file-path>'…` at `@electron/notarize@2.5.0 …/lib/notarytool.js:141`. Driving **`d4ebea42e`'s** `scripts/notarize-artifacts.mjs` through the same fixture with `--dist dist`:

```
exit 1 · Notarizing disk image: dist/local-operator-ui-0.17.2-universal.dmg
Error: The value '/var/folders/…/qa-electron-notarize-hGW1Un/dist/local-operator-ui-0.17.2-universal.dmg'
       is invalid for '<file-path>': The file couldn't be opened because it doesn't exist.
fixture: appPath="dist/local-operator-ui-0.17.2-universal.dmg", appPathIsAbsolute:false, resolvedExists:false
```

Same message, same cause, same exit status; source confirms why (`grep` on the pre-fix file: `const discovered = existsSync(dist) ? readdirSync(dist) : []` and `.map((name) => join(dist, name))`).

#### Q11–Q17 — watchdog platform split: all PASS

| Q | Surface | Actual | |
|---|---|---|---|
| Q11 | darwin, **swap landed**, job loaded | exit 0 in **3235 ms** against a 20 s bound; exactly 1 relaunch; env probes `launchctl` / `/usr/bin/plutil` | PASS |
| Q12 | darwin, **ShipIt job gone** | exit 0 in **3235 ms**; exactly 1 relaunch | PASS |
| Q13 | darwin, **app still alive** → no-op | exit 0 at the 6 s bound; **0 relaunches** | PASS |
| Q14 | **linux**: no probe → the bound decides, then relaunch | env `SHIPIT_PROBE=""`, `PLIST_READER=""`; no early exit; exit 0 in **7092 ms** (6 s bound + 1 s settle); exactly 1 relaunch; the script's non-comment lines name neither `launchctl` nor `plutil` | PASS |
| Q15 | darwin neighbour: job **stays loaded** and swap **not** landed | exit 0 in 7412 ms (the bound) with 1 relaunch — the contrast that shows Q11/Q12's 3.2 s is the signal, not the bound | PASS |
| Q16 | **No orphans / no leaked process groups**, both branches | `kill(-pid, 0)` on the watchdog's own group: empty in every case (B3 already at exit, the rest within 700 ms — the backgrounded `open` shim is the last member); nothing left respawning, and no reader process survives | PASS |
| Q17 | **macOS behaviour unchanged** | comment-stripped diff of the generated darwin script at `d4ebea42e` vs this head = exactly **4 hunks**, all substitution/guard only: `+SHIPIT_PROBE`/`+PLIST_READER` from env; `shipit_loaded` → `"$SHIPIT_PROBE"`; `bundle_version` → `exec "$PLIST_READER" -extract …`; `swap_landed` gains `[ -n "$PLIST_READER" ] \|\| return 1`; `job_known` gains `[ -n "$SHIPIT_PROBE" ] &&`. On darwin those variables hold the same two literals with the same argv. Only one production caller (`update-service.ts:934`, behind `platform !== "darwin" → return null`), and it states `platform:"darwin"` | PASS |

#### Q18 — the Linux half is proven by real CI: PASS

- The four red `main` runs the PR names are real and all `failure`: `34695505505`, `34695566539`, `34696129340`, `34699012497` — the last carries `error: 'expected the swap to end the wait, took 126273ms'`.
- `34699818807` (implementer's scratch branch): `Desktop Tests` **success**, `# tests 277 / # pass 273 / # fail 0 / # skipped 4`, and on ubuntu-latest **`ok 248 - the watchdog leaves early when the swap has landed, job or no job`** and `ok 249 - without launchd or plutil the bound decides, and nothing is left behind`.
- I checked that claim's one loophole: the 4 skips are `240 … on macOS # SKIP macOS only`, `257`/`258 … no local-operator on this machine's PATH`, `264 - the real macOS tools reject an unsigned image, on macOS`. **No watchdog case is skipped**, so the split is genuinely asserted on Linux.
- **My own run:** pushed this head to a scratch `dev-qa116-watchdog-platform`, got **`34700421961` → success, all six checks** (Security Audit, Type Checking, Lint, Desktop Tests, NPX Sanity Check ×2), `Desktop Tests` again `277 / 273 pass / 0 fail / 4 skipped` with `ok 248`/`ok 249` on ubuntu, then deleted the branch.

#### Q19 — shipped gates at this head: PASS

`pnpm test:desktop` → `tests 277 / pass 277 / fail 0 / skipped 0`, exit 0, and `git status --porcelain` empty afterwards · `pnpm lint` → `Checked 412 files … Found 24 warnings`, exit 0 · `pnpm check-types` → clean on both tsconfigs, exit 0 · `pnpm check-themes` → `themes.generated.css is up to date (12 themes)`, `Contrast contract holds: 1912 assertions across 12 themes, 0 pinned exception(s)`, exit 0.

`verify-macos-artifacts --app "/Applications/Local Operator.app" --dmg ~/Downloads/local-operator-ui-0.17.0-universal.dmg` fails closed exactly as intended: `PASS app-codesign` / `PASS app-spctl (accepted | source=Notarized Developer ID)` / `PASS app-stapler`, then `FAIL dmg-spctl … rejected | source=no usable signature` and `FAIL dmg-stapler … does not have a ticket stapled to it.` → `2 of 5 artifact checks failed`, exit 1. The app checks on the notarized bundle pass on their own (`ok:true`). Note for whoever runs this next: the CLI's `verifyArtifacts` requires *both* an app and an image, so `--app` alone returns `No disk image found under dist` — the app-only path is reached through the shipped `runChecks` the CLI calls.

#### BLOCKED — stated, not substituted

- **BL1 — a real Apple submission/staple/ticket.** No Developer ID identity or Apple credentials on this machine. My notarizer is a fixture: it proves the path handed over, the ordering, the re-hash and the failure modes, never that Apple accepts the submission. Only the release workflow can re-verify the positive direction.
- **BL2 — a packaged end-to-end install.** No signed packaged build and no real ShipIt job here, so the actual bundle swap + watchdog relaunch is not exercised; Q11–Q15 drive the shipped script against real processes and shims, not against ShipIt.
- **BL3 — the real `launchctl list <label>` answer.** Deliberately not run (it would query the operator's launchd domain). The argv shape is identical to the pre-fix script's, so the shape is proven and launchd's own answer is not.

#### Observation (not a finding of this PR)

The staple rewrite is entry-scoped: `updateUpdateYmlEntry` rewrites the `sha512:`/`size:` lines of the matching `- url:` entry only, leaving the legacy top-level `path:`/`sha512:`/`size:` block as electron-builder wrote it (visible in the A4 run: entry updated, top-level `sha512: STALEHASH` unchanged). In this repo's layout that block names the **.zip** — verified on the released `v0.17.1` `latest-mac.yml` — which this step never staples, so nothing shipping is described by a stale hash. The function is unchanged by this PR; recorded only so it is not rediscovered later. No action requested.

PR #115, opened 2026-09-12
feat(chat): show a session's subagents and to-do plan in the chat header

damianvtran:
## What this is

A **header button that opens a popover showing the session's subagents and its to-do plan.** While a turn runs, the agent may be running several children on their own models, each spending its own context and money and holding its own plan — and until now the desktop showed none of it. A `task` call was an ordinary tool row, its children invisible; the plan was unstructured text behind a row's disclosure.

It is **wired and shipped**, not a proposal: `chat-page.tsx` derives the view model from the canonical stream and hands it through `ChatContent` → `ChatHeader`, so the trigger appears in the running app for a session with work in flight. (This body previously carried its draft framing — "nothing is wired, do not merge" — after the wiring landed. Corrected.)

The data was already on the wire: `frontend.jobs` and `frontend.todos` arrive on every `frontend.update` and no renderer file read either. This is a presentation gap, which is why the answer is a small surface rather than a feature.

Both in flight

## The design

A ghost icon button in the header's action cluster, immediately left of the canvas button (both 32px, the cluster taking over the `ml-auto` the canvas button used to carry). It opens a 384px popover with two sections and **only the ones that apply**:

- **`Subagents`** — one two-line row per child: state icon, label, then `role · elapsed · context · cost`, with the activity string on its own indented second line. **Motion means alive, colour means failure and nothing else**, and a figure the wire did not report is omitted rather than zeroed.
- **`To-dos`** — the plan in the model's own order, phases with their own `done/total`, and items whose encoding is the TUI's: **luminance says open or settled, a word says which settled state**, and no item is coloured. Blocked is the loudest row in its section; done and dropped are struck; dropped keeps its tag unstruck so the word survives.

**Visibility:** the trigger renders only when any child is open, any to-do is open, **or a child has failed and the panel has not shown it** — and never while the canvas is open. Settled work alone does not raise it: a finished roster is history, and history lives in the transcript.

The principle: **the popover is the live view of work in flight; the transcript remains the record of what happened.** It adds a view, replaces nothing, takes no layout slot, and introduces no new colour role, radius or primitive.

### Options rejected

A dock band above the composer (the TUI's own answer) reclaims transcript height every turn and forces one placement on two surfaces with different lifetimes. Inline rows in the transcript conflict with the contract — a completed action is one line, not a card. A third pane or a sheet beside the canvas misplaces something that is glanced at and dismissed.

## Where it comes from

[...]

PR #120, opened 2026-09-12
fix(chat): keep the header and the clicked row naming one conversation

damianvtran:
## What and why

Clicking into a conversation changed it to **Untitled chat** when it already
had a title. Two halves of one invariant, which is now stated in the code and
tested: *the row a user picked and the header they land on must name the same
conversation.*

**Change class: C2 — standard, pre-authorized.** User-visible, so it carries a
design/UX consideration; the visual surface is text only (no layout, spacing,
colour or motion change), so the design round is about the copy and the
precedence, not about pixels.

### Why the header was wrong, measured rather than assumed

`conversation_title` on the live frontend state is the backend's **journalled**
title only. The catalogue row the sidebar renders is named by a wider rule —
`resume.session_name()` falls back to the conversation's opening message when
nothing is journalled. Counted against the operator's own store: of **1407
sessions, 226 carry a journalled title and 1180 are named from their opener
alone**. For those 1180, the row said one thing and the header said
"Untitled chat".

### The second half, which is why the row itself changed on click

The same effect wrote the live title into the catalogue row unconditionally,
and `upsertSession` merges with a plain spread — so an empty live title
**overwrote a name the catalogue had already supplied**, and the row the user
had just clicked rendered "Untitled chat" until the 5s list poll restored it. It
re-blanked on every frontend update, i.e. continuously while a turn streamed.

## The change

- **`src/renderer/src/features/chat/chat-title.ts`** (new, pure): `resolveChatTitle`
  (live non-blank → catalogue name → "Untitled chat", draft branch unchanged) and
  `catalogueTitleUpdate` (returns a PARTIAL row so "nothing to say" omits the key,
  which is the only way the spread merge leaves a known name standing).
- **`chat-page.tsx`**: the header and the catalogue write both go through those
  two functions; the catalogue write adds a title only when the live one is
  non-blank, which is what leaves the name the catalogue already supplied standing.
- The fallback is a **READ and is never written back**. `conversation_title_user_set`
  and the backend's naming errand both read that field as "this conversation is
  already named", so writing an opener-derived stand-in would retire the one call
  that can give the conversation a real name. The terminal UI already resolves
  this tension with a display-only `_provisional_name` (`local_operator/tui/app.py`,
  `_restore_resumed_name`): *"the row a user picked and the tab they land on agree
  by construction rather than by coincidence."* This is that rule for the desktop
  header.

No wire changes and no backend changes: `destination-pickers.tsx`'s `/rename`
prefill and the sidebar's own fallback are deliberately untouched.

### 1. The unit contract, and a red run that shows it would have caught this

`scripts/chat-title.test.mjs` (registered in `test:desktop`) bundles the **real**
store alongside the new functions, because half of this contract *is* the store's
merge: a test that re-implemented `{...current, ...incoming}` would have passed
against exactly the payload that caused the defect.

Red proof, against a faithful transcription of the pre-fix expressions (the old
header line and the old unconditional `title:` write) substituted into the same
suite — **6 of 9 cases fail**:

```
✖ a blank live title falls back to the catalogue name the row was clicked under
    + actual   'Untitled chat'
    - expected 'Improve the ask function clickable options'
✖ a streaming tick with an empty live title leaves the row's name standing
    + actual   ''
    - expected 'Improve the ask function clickable options'
✖ a whitespace-only live title cannot blank the row either
✖ a row with no name at all gains none from an empty live title
✖ a live title wins, so a rename still shows where it landed
✖ isBlankTitle is the one definition both rules read
```

Green on this head: `pnpm test:desktop` → **493 tests, 492 pass, 1 fail**, and that one
failure is pre-existing and environmental, not this change (`the operator's own
install is classified rather than told to use pip`: the `local-operator` first on
this machine's PATH is the desktop app's venv, whose `INSTALLER` marker is `pip`,
while the test asserts `uv`. It fails identically at `origin/main` — `a09f2e6f4`
— with `node --test scripts/update-robustness.test.mjs`: 39 pass, 1 fail, same
name. This branch touches neither that file nor the update code.)

### 2. The defect and the fix, driven in the REAL app

Not Storybook, and not the renderer in a browser with a stubbed preload: the
**real built Electron app** from each tree — `pnpm build`, then `npx electron .`
from the tree's own `out/`, so `import.meta.env.DEV` is false and the
development-only `Chat | Raw` tab strip does not exist — with its own main and
preload processes and its real IPC desktop transport, against an **isolated**
`local-operator serve` on its own port, seeded with three conversations and
paired through a shared `LOCAL_OPERATOR_DESKTOP_TOKEN`. It has to be the built
app rather than `electron-vite dev`: under dev that tab strip paints over the
chat header, and the first attempt at this evidence produced byte-identical
before/after frames with a perfectly correct DOM. Nothing touches the
operator's store on 1111. Seeded store, read back through the backend's own
`/v1/desktop/sessions`:

```
Improve the ask function clickable options   (opener-named, NO journalled title - the defect)
Retention sweep notes                        (journalled; its opener is a different string)
""                                           (no nameable opening message at all)
```

Each row was clicked with a real `Input.dispatchMouseEvent` press/release at the
row's own coordinates, and the header read from `[data-tour-tag="chat-header"] h2`:

| state | `origin/main` (`a09f2e6f4`) | this branch |
| --- | --- | --- |
| click the opener-named row | header **`"Untitled chat"`** / selected row `"Improve the ask function clickable options"` | header `"Improve the ask function clickable options"` / row same |
| click the journalled-title row | header **`"Untitled chat"`** / selected row `"Retention sweep notes"` | header `"Retention sweep notes"` / row same |
| click the nameless row | header `"Untitled chat"` / row `"Untitled chat"` | unchanged |
| nothing open | header `null` | header `null` |

Both brand palettes, both trees, same script, same 1380x872 viewport.

### 3. The frames, on both trees, from the real app

Sixteen frames are committed under `docs/evidence/chat-title/` (declared
`supplementary` in the manifest, since no story renders the header or the
sidebar), captured from the **built** app — `pnpm build` then `npx electron .`
from each tree's own `out/`, so `import.meta.env.DEV` is false and the
development-only `Chat | Raw` tab strip does not exist. Under `electron-vite dev`
that strip paints over the header, which is how a first attempt produced
byte-identical before/after frames with a perfectly correct DOM: the subject was
covered.

The pair, cropped to the header band (device pixels, `1760x120+980+0`):

| pair | theme | `magick compare -metric AE` full / header band |
| --- | --- | --- |
| `open-alpha` (the defect) | localOperatorDark | 215254 / **36148** |
| `open-alpha` | localOperatorLight | 227343 / **33235** |
| `open-bravo` (journalled title) | localOperatorDark | 33259 / **9089** |
| `open-bravo` | localOperatorLight | 37224 / **7583** |
| `open-charlie` (nameless control) | localOperatorDark | 574344 / 1940 |
| `open-charlie` | localOperatorLight | 544136 / 2158 |
| `rest` | localOperatorDark | 33725 / 0 |

`open-charlie` is the control that must keep reading `Untitled chat` on both
trees; its small non-zero delta is elsewhere in the frame (the sidebar's own
selected-row treatment after a different sequence of clicks), not in the title.

Geometry asserted before every frame and recorded in the run's JSON: window
`1380x872` CSS at `devicePixelRatio 2`, header `{x:500,y:0,w:880,h:56}`, its `h2`
`{x:560,y:7,w:760,h:22}`, no `[role="tablist"]` in the document. Each frame is
only taken once the header has been verified to be the topmost element at its own
centre; a subject outside the captured viewport fails the run.

**One intervention, applied identically to both trees:** the app's own danger
banner ("The server is offline…") is hidden in these frames. It is there because
the harness points the app at an isolated backend on a port of its own while the
built renderer's CSP allows only 1111/8080 for its legacy REST probes — the same
strip `docs/evidence/send-error/README.md` records in the same setup. Here it
matters, because it is laid across the top of the chat column, exactly where the
header is. `visibility: hidden` rather than `display: none`, so the element keeps
its box and the app's own geometry is what the frames are asserted against.

### 4. What the frames do NOT prove

- **Not the packaged/notarised build.** Real main and preload processes from a
  build output tree; signing, the DMG and the updater are not exercised.
- **Two themes, not twelve**, and one window size.
- **The transient blanking of the row is not photographed.** Three samples were
  taken after each click (~1s, ~2s, past the 5s poll) and came back
  byte-identical on both trees, so they were dropped from the set rather than
  committed as two extra copies of one state. That half of the defect is pinned
  by the unit suite, which drives the store merge directly.
- No model inference: no turn is sent, so what is proven is which name each
  surface wears, not what an agent would then reply.

## Gates (all from the worktree)

```
pnpm lint          Found 24 warnings.  (0 errors; warnings are pre-existing in bin/ and linux-sandbox.test.mjs)
pnpm check-types   exit 0
pnpm test:desktop  493 tests: 492 pass, 1 fail (pre-existing, environmental - see above)
pnpm check-themes  themes.generated.css is up to date (12 themes).
                   Contrast contract holds: 2168 assertions across 12 themes, 0 pinned exception(s).
pnpm check-evidence Evidence holds: 681 frames are pictures of their own theme (worst ΔE00 18.1)
pnpm build         ✓ built in 33.94s   (needs the four OAuth client vars in the env; the repo's renderer
                                        config refuses to load without them and this machine has no .env)
```

## Not addressed, deliberately

- The backend's cold-state title (`_synthesise_cold_state` / `_restore_cold_details`
  in `attached.py`) still reports an empty `conversation_title` for a session with
  no frontend checkpoint. That is a different repository, and putting an
  opener-derived value on that wire would be wrong for the reason above.
- `chat-sidebar.tsx`'s own "Untitled chat" fallback and `directory-indicator.tsx`
  are untouched.


damianvtran:
### Agent review — round 1
Reviewer: deepseek-flash (deepseek/deepseek-flash) — no provider fallback during this round
Scope: b11ec233d317434919b0b24ef7169c5788cb0909..6c37dac0258121cd1079c443af0b1533dea236a7

Head note: the brief named `7fc365b72`; the head has since moved to `6c37dac02`. I
verified the delta is evidence-only myself — `git diff --stat 7fc365b72..6c37dac02`
touches `docs/evidence/chat-title/README.md`, `docs/evidence/manifest.json` and the
deletion of eight `*-transient` / `*-settled` duplicates, and nothing under `src/` or
`scripts/` — so this round reviews `6c37dac02`.

Reviewed from a detached worktree at the head (`~/workspace/review-120`, plus
`-base` at `b11ec233d` and `-a09` at `a09f2e6f4`). The repo was not edited; the
operator's live store was never opened or touched.

**Verdict: clean — no blocker, no major. This round is TERMINAL; the findings below
are follow-ups, not a reason for another round.**

The core claim is sound and I verified its premise in the backend rather than
taking it from the PR body: the header's fallback and the sidebar row are the same
field. `resume.session_name()` is documented and implemented as "its stored title,
else its opening message" (`~/local-operator/local_operator/resume.py:1809`), the
catalogue's row `name` is exactly that call (`session/catalog.py:500`), and
`/v1/desktop/sessions` serves that row (`server/utils/desktop_sessions.py:594`).
So for the 1180-ish opener-named sessions the row really did carry a name the
header could not see, and the fix closes that gap at the header without writing the
stand-in to the wire. The white-space rule in `isBlankTitle` is also backed at the
source: the journaller whitespace-normalises every name and drops an all-whitespace
one to `""` (`resume.py:568`, `normalized = " ".join(text.split())`).

---

#### F1 — minor — the `{title: catalogueTitle}` branch of `catalogueTitleUpdate` is dead code, and its JSDoc describes a no-op as behaviour

`src/renderer/src/features/chat/chat-title.ts:112` (and the claim at
`chat-title.ts:102-103`).

`upsertSession` merges with `{...current, ...incoming}` (`canonical-sessions-store.ts:381`
`mergeRow`), and the only caller passes `catalogueTitle: current.title`
(`chat-page.tsx:275`) — the row's own value. So `{title: catalogueTitle}` writes the
row's current title back to itself: the committed row is value-identical to the
`{}` case, which the branch can only be reached *instead of* when `catalogueTitle`
is already non-blank. It is not the blank case, and it is not load-bearing.

Evidence, a mutation battery over the shipped suite (`node --test scripts/chat-title.test.mjs`):

```
head                                      pass 9  fail 0
pre-fix expressions restored              pass 4  fail 5
catalogueTitleUpdate: drop {title:catalogueTitle} -> {}   pass 9  fail 0   <- F1
resolveChatTitle: swap precedence         pass 8  fail 1   (caught)
isBlankTitle: drop the trim               pass 6  fail 3   (caught)
```

The suite cannot tell the branch from `{}`, which is also why the JSDoc sentence "a
blank or whitespace-only one falls back to the row's own name" reads as a
behaviour but is a no-op. Fix (one line): delete the branch and the clause, or pass
a value that is not the row's own title if a durable re-assertion was intended.

For the record, the part of the comment that *is* load-bearing is correct and I
proved it rather than assuming it: a present-but-`undefined` key does overwrite.

```
$ node -e 'const c={session_id:"x",title:"Improve the ask function clickable options"};
  console.log(JSON.stringify({...c,...{title:undefined}}));
  console.log(JSON.stringify({...c,...{}}))'
{"session_id":"x"}
{"session_id":"x","title":"Improve the ask function clickable options"}
```

#### F2 — minor — the PR body's gate block is a pre-`#115` measurement, not this head's

The block says `pnpm test:desktop 445 tests: 444 pass, 1 fail` and
`pnpm check-evidence … 645 frames …`. Neither number belongs to the reviewed head:

| gate | body | `b11ec233d` (this PR's base) | `a09f2e6f4` (0.18.0 tip) | head `6c37dac02` |
| `test:desktop` | 445 / 444 / **1 fail** | 484 / 483 / 1 | 436 / 435 / 1 | **493 / 492 / 1** |
| `check-evidence` | 645 frames | 665 | **645** | **681** |
| committed `.webp` under `docs/evidence` | — | 665 | 645 | 681 |

`645` is exactly `a09f2e6f4`'s committed frame count, and `445` is exactly
`a09f2e6f4`'s 436 plus the new file's 9 tests — i.e. the block was measured on a
tree cut from the 0.18.0 tip, before this branch was synced onto `#115`, and was
never refreshed. It is a count/freshness problem, not a confidence problem: I re-ran
all six gates at the head and **every one passes**, including the single failure the
body names (same test, same name, present at `b11ec233d` and `a09f2e6f4` too).
Fix: paste the head's numbers.

#### F3 — minor — the evidence README's headline measurement overstates itself by ~2x

`docs/evidence/chat-title/README.md:85`: "`open-alpha` is the claim: **a third of the
header band** changes". Re-measured with the README's own crop and command:

```
$ magick compare -metric AE (before-open-alpha/localOperatorDark.webp -crop 1760x120+980+0) \
                            (after-open-alpha/localOperatorDark.webp  -crop 1760x120+980+0)  null:
36148 (0.171155)          # 17.1% of 1760x120 = 211,200 px, not a third
```

Every other number in that table reproduces exactly (all seven rows, below), so this
is a prose slip in the sentence a reader quotes. Fix: "about a sixth".

#### F4 — nit — the re-capture assertion cannot hold for the control pair

`docs/evidence/chat-title/README.md:133`: "every `before-*` frame differs from its
`after-*` counterpart in the header band". The `rest` pair is the *control*: its
band delta is 0 by design, in both palettes.

```
rest Dark   full=33725   band=0
rest Light  full=193617  band=0
```

A future agent following step 3 would either fail the set or "fix" the control.
Fix: scope the assertion to the clicked states and keep `rest` as the zero.

#### F5 — nit — the mismatch is bounded, not eliminated: a cold deep link still paints "Untitled chat" first

The header now reads its fallback from `boundRow?.title` (`chat-page.tsx:438-443`),
i.e. from a row that only exists once the sessions list has landed. Launch straight
into `/chat/<id>` with an empty store (deep link, or a store whose list has not
returned) and the header paints `UNTITLED_CHAT` for one list round trip before
correcting — the same row/header disagreement the PR removes, just transient and
self-correcting (the pre-fix header never corrected at all, so this is not a
regression). Not photographed, and not covered by the unit suite, which models the
store write rather than the header's read of a not-yet-loaded row. No change
required for this PR; recorded because the PR asserts the invariant is established.

---

### The specific questions, answered

- **`resolveChatTitle` precedence** — right, and each leg is tested: draft first (a
  staged draft has no conversation), then live (so a `/rename` shows the moment it
  lands), then the row's name, then the fallback. Swapping the first two legs is
  caught by the suite (1 failure).
- **Blank cases** — right, and the whitespace leg is the correct call rather than
  defensive noise, for the reason given at `chat-title.ts:60` (verified at
  `resume.py:568`): a `"   "` reaching here means "no name", and without the trim it
  would both win the header and blank the row.
- **Is the partial-row trick load-bearing — `undefined` vs an absent key?** — Yes,
  and the comment is right: a present key with an `undefined` value still overwrites
  through the spread (proof above). Only the *absent* key leaves the name standing,
  which is why the empty object is the case that matters. (The third branch is the
  dead one — F1.)
- **Is anything lost by reading `current.title` from the row?** — No, and it is
  read correctly: `current` comes from `getState()` in the same synchronous tick as
  the write, so it cannot be stale against the row it writes; the write path reduces
  to "live wins, else say nothing new", which is strictly less than what it replaced.
  The only staleness in play is `canonical.frontend` itself, which is pre-existing
  and untouched.
- **Hook order / the moved `boundRow` and `title` computation** — no issue. Both are
  plain consts; `title` moved to after the `useCanonicalSessionsStore` selector and
  no hook was reordered, so no conditional-hook or dependency-order hazard exists
  here.
- **Can the header flash or regress?** — Deep link: one bounded flash (F5). Fork
  wearing an inherited title: no — the row's `name` is the stored (inherited) title
  and the header prefers the live title, so both surfaces read the inherited name.
  Rename landing after paint: no — live wins and the effect propagates it. Row
  disappearing from the list: no change from before (`boundRow` was already the
  description's fallback and is now the title's too). Draft path: untouched (first
  leg, same `loadedTarget`).
- **Other consumers** — `conversation_title` has one other renderer consumer,
  `RenamePicker`'s prefill (`destination-pickers.tsx:1205`), and leaving it alone is
  right: pre-filling the opener-derived stand-in there would journal it on submit,
  and the picker's own copy already says "An explicit name takes precedence over the
  generated title". `chat-sidebar.tsx:217/222` keeps its own `"Untitled chat"`
  literal; that is a second definition of one string, but nothing in this diff makes
  them disagree (the split is disclosed at `chat-title.ts:52`), so it is not a
  finding — only the `UNTITLED_CHAT` comment's "asserted to agree" is a touch
  stronger than what exists, since no assertion binds the sidebar's copy.

### What the PR does NOT prove, and whether the disclosure is honest

Honest, and unusually specific — with two exceptions, both above. The README's "What
these frames do NOT prove" section is accurately scoped, and the one intervention
(hiding the app's own offline banner, `visibility: hidden`, applied identically to
both trees) is disclosed in both the body and the README. What the frames cannot
carry:

- **The row-blanking half is genuinely not in the frames, and it is worth being
  blunt about why.** I opened the committed pair: `before-open-alpha`'s `h2` reads
  `Untitled chat` while its sidebar row still reads `Improve the ask function
  clickable op…`, on both trees, at every kept sample. So the frames show the header
  mismatch only; the second half of the defect rests entirely on the unit suite,
  which re-implements `chat-page.tsx`'s write (`streamTick` in the test) rather than
  importing it. A future regression in the *wiring* — e.g. someone passing
  `catalogueTitle: undefined` at the call site — would leave the suite green. The
  README's framing ("the transient blanking is not photographed … pinned by the unit
  suite") is accurate; it just should not be read as the frames showing the row
  blanking.
- **2 of 12 themes, one window size, not the packaged build**, no model inference —
  all disclosed.
- **The store census (1407 / 226 / 1180) is not reproducible from this repo**, and I
  did not go near the operator's store to check it. What I could verify is the
  mechanism, and it holds independently of the census (see the top of this review).

### What I verified (so the next round does not have to)

- **The AE table reproduces exactly**, all seven rows, with the README's own crop:
  alpha dark 215254/36148, alpha light 227343/33235, bravo dark 33259/9089, bravo
  light 37224/7583, charlie dark 574344/1940, charlie light 544136/2158, rest dark
  33725/0. The `rest` control at exactly 0 is what makes the pairs mean something.
- **The frames are real**: `before-open-alpha`'s `h2` is `Untitled chat`,
  `after-open-alpha`'s is `Improve the ask function clickable options`.
- **16 frames, 16 distinct SHA-256**, matching the README's rule and the manifest's
  `supplementary[chat-title]` `frames: 16` / `surfaces: 8` / `themes: 2`; and
  `594` (sweep) + `87` (all supplementary) = `681`, which is what
  `pnpm check-evidence` counts on disk.
- **Provenance stamps are current**: manifest `srcTree`/`scriptsTree`
  (`4291e0d5…`, `b0c11ecc…`) equal `HEAD:src` / `HEAD:scripts` exactly, `head`
  `83db293ed` resolves and is reachable, and `check-evidence` exits 0.
- **The new suite is a real red-flag test**: against the pre-fix expressions
  restored in the same module it fails 5 of 9 (the PR body's 6 includes an
  `isBlankTitle` mutation); it bundles the real `canonical-sessions-store`, so the
  spread merge under test is the shipped one.
- **All six gates at the head**: `pnpm lint` 24 warnings / exit 0 (0 errors),
  `pnpm check-types` exit 0, `pnpm check-themes` "12 themes … 2168 assertions … 0
  pinned exceptions", `pnpm check-evidence` exit 0 (681 frames, worst ΔE00 18.1 —
  the same worst frame the body names), `pnpm test:desktop` 493 / 492 pass / 1 fail
  (the known `the operator's own install is classified…` failure, which also fails
  at the base and at `a09f2e6f4`), `pnpm build` exit 0.

One environment note, not a finding: my fresh worktree install left electron's
post-install unrun (pnpm 10 script blocking), so `pnpm build` first died in
`[vite:bytecode] Electron uninstall`; it built cleanly once the electron binary was
reachable. Nothing about this diff is implicated.

No wire or backend change — confirmed from the diff (`src/`, `scripts/`,
`package.json` test registration, `docs/evidence` only).

damianvtran:
### Design review — round 1

**Reviewer:** designer subagent, `deepseek/deepseek-flash` (`lop` pid 65423) — no provider fallback during this round.
**Scope:** `a09f2e6f4..6c37dac02` (the user-visible surface). Reviewed at head `6c37dac02`. I was briefed at `7fc365b72` and verified myself that the delta between the two is docs/evidence only — `git diff --stat 7fc365b72..6c37dac02` is `docs/evidence/chat-title/README.md`, `docs/evidence/manifest.json` and eight deleted `.webp`, with no `src/` or `scripts/` file in it.

**Verdict: TERMINAL — no BLOCKER, no MAJOR. Design sign-off on `6c37dac02`.** Three NITs below (D1–D3); each is declinable and none gates the merge.

#### What I looked at

- The 16 committed frames in `docs/evidence/chat-title/` — full frames and header band, both palettes, before and after, viewed at 1:1 and 200%.
- My own re-drive of the built app at this head: `pnpm build` in a worktree, `npx electron .` from its `out/`, an isolated `local-operator serve` on its own port paired through its own `LOCAL_OPERATOR_DESKTOP_TOKEN`, six seeded conversations, real `Input.dispatchMouseEvent` clicks on real rows, both palettes. Harness `/tmp/d120-harness/run.sh`, frames `/tmp/d120-frames` and `/tmp/d120-frames-light` (ephemeral; say the word if any of them should be committed).
- The three seeded conversations plus three I added for the states the committed set does not picture: a 198-character pasted opener, a multi-line opener, and an 81-character journalled title.

#### The precedence — my judgement

Live title → the catalogue row's name → "Untitled chat" is the right order, and showing the opener-derived name in the header is the right call. The header is the answer to "which conversation am I in"; for 1180 of the operator's 1407 sessions the only true answer is the opening message, and "Untitled chat" over a conversation the user just saw named in the sidebar is an active falsehood, not a neutral unknown. The terminal UI's `_provisional_name` means this is already the app's own language for exactly this tension, so a user moving between the two surfaces gets one rule rather than two. Keeping the fallback a READ is what makes it a stand-in rather than a lie: the conversation stays namable, and the two trade-offs are acceptable — the header can wear a sentence rather than a title, and the name can change on its own later when the naming errand lands (row and header change together).

#### The frames, re-measured

`magick compare -metric AE`, my own crops: in the exact title rect (`1520x44+1120+14` device px, i.e. the h2 box) `open-alpha` changes 13582 / 13522 px (dark / light) and `open-bravo` 7509 / 7392, while `open-charlie` is 143 / 312 — 0.2–0.5%, antialiasing, no title change in the nameless state. Header band (`1760x120+980+0`): 36148 / 33235, 9089 / 7583, 1940 / 2158, matching the README's numbers. The pair is a true A/B and the nameless control is genuinely unchanged.

#### The 200-character paste, and the "does it truncate" question

The header can never be asked to render a 200-character opener. `sessions.list` serves `resume.session_name()`, which condenses to `NAME_MAX_CHARS = 64`, cut on a word boundary. My 198-character paste arrived as `Weekly retention sweep brought back 1,204 rows again -…` (55 characters) and my 81-character journalled title as 61 characters, both in the sidebar row and the header.

Measured on all six rows, both palettes: the h2 is `{x:560, y:7, w:760, h:22}`, 16px, `white-space: nowrap`, `overflow: hidden`, `text-overflow: ellipsis`, and `clientWidth == scrollWidth == 760` in every state — `truncate` never engages, and the longest name posted leaves ~240px of the title box unused. One line, subtitle intact below, no reflow, no overlap with the action cluster. The header is fully readable at the header's own width.

#### The three states

- opener-named (`d120-row-0`): header and row both read `Improve the ask function clickable options`.
- journalled (`d120-row-4`): `Retention sweep notes` on both surfaces.
- nameless (`d120-row-5`): `Untitled chat` on both surfaces over an empty pane — honest, not confusing. The words are sentence case, identical to the sidebar's own fallback, and consistent with "New chat", "Start of conversation" and "Nothing running right now".
- Both palettes: the 56px band, the 16px ink title over the 13px muted `/Users/damian` line, and the alignment are unchanged by this PR; only the string differs. `/Users/damian` still reads coherently under a 55-character title (light and dark).

#### The half with no frame, and the unpictured first paint — I supplied the missing evidence

- **Row blanking.** Polled the selected row every 100 ms for 8 s after clicking the opener-named row — the case that blanked, because its live title is empty. 80 samples, one distinct `(row, header)` state, no transition (`rowBlanked: false`). And after clicking through all six rows and reloading, the isolated store held `title.json` for exactly the two sessions I had seeded with one — none written for the three opener-named ones, so "display only, never written back" holds on the app I drove, not just in the unit test.
- **Deep-link first paint.** Reloading on a session's own URL, the header wears `Untitled chat` from t≈146 ms to t≈340 ms and then the real name; the transcript (`[role="log"]`) first paints at t≈430 ms, and on a cold first start the placeholder lasted 748 ms. The header and the row both resolve from the same catalogue row, so they never disagree, and the placeholder was never observed over a painted conversation. It is a loading state, not a recurrence of the defect — recorded as D2 rather than filed as a blocker.

#### Disclosure

Sufficient for a design sign-off. The trim is honest: I checked the blobs at `7fc365b72`, and `before-open-alpha`, `-settled` and `-transient` are byte-identical (`775268eea174`), as are the three after-frames (`ced4b1ba76c9`), so "keep only the frames that differ" removed duplicates and nothing else. The set asserts 16 frames and 16 distinct SHA-256, which holds. The manifest was re-stamped 24→16 frames and 12→8 surfaces, consistent with the PR body's "sixteen frames". The README names what the frames do not prove (packaged build, twelve themes, no model inference, the hidden banner) and states plainly that the transient blanking is not photographed — which is the right call for a text-only change, and the half it leaves to the unit suite is exactly the half I re-measured on the running app above.

#### Findings

**D1 — NIT (copy).** A multi-line opener reaches the header as a run-on sentence with mid-sentence capitals: `Fix the retention sweep It deletes rows that are still referenced Please check the audit ta…` (my frames `d120-row-2`, both palettes; the string is the wire name from the backend's `_condense`). The sidebar row already wore it, but the header puts it at 16px at the top of the window, which is where it reads worst. Recommendation: do not prettify it in the header — that would break the row↔header equality this PR exists to establish. It is defined in `local_operator/resume.py` (`_condense`, joining lines with a space); worth a follow-up in the backend repository.

**D2 — NIT (state).** On a cold load of a session's own URL the header wears `Untitled chat` — a claim that the conversation has no name — for ~194 ms (146→340 ms) and up to 748 ms on a first cold start. It never sits over painted conversation content and the row resolves with it, so it is a placeholder rather than the defect; I record it because it is the one place the header still asserts something false, while the transcript area already has a loading treatment to match. If you want it gone: render the title slot empty (avatar only) until the live title or the catalogue row resolves. Dormant in every in-app click path, since a row cannot be clicked before the catalogue painted it.

**D3 — NIT (affordance).** The h2 carries no `title` attribute while the sibling description line in the same block does (`chat-header.tsx`: `title={description}`). This is dormant today — a catalogue name is capped at 64 characters and a `user_set` rename at 80 (`naming.py`), both of which fit the 760px title box, which is why nothing was CSS-cut in any of my states — but the header is where a cut would be least readable, and an unconditional `title={agentName}` (or a conditional one when the text is cut) would match its own sibling.

#### Recorded, not findings

- `"Untitled chat"` now exists in three places: `UNTITLED_CHAT` in `chat-title.ts` and twice in `chat-sidebar.tsx:217,222`. The invariant this PR establishes across two surfaces runs through a copy-paste; importing the constant in the sidebar fallback would make it structural. The module docstring records the deliberate decision to leave that file alone — noted, not filed.
- The backend's word-boundary cut leaves a dangling separator (`...rows again -…`). `_condense`, not this PR.
- `/rename` propagation is unpictured and I did not exercise it in a rendered frame — it is covered at unit level only (`a live title wins, so a rename still shows where it landed`). The header's behaviour follows from the live-first branch plus `catalogueTitleUpdate` writing the same string into the row, so the two surfaces agree by construction, but that is reasoning and not a frame. It is the one flow in my slice I am not signing off on visually.

Anything you change in response can be answered with `### Design review remediation — round 1`. If D1–D3 are declined with reasons, no further design round is needed and the design gate is closed on this head.

damianvtran:
### QA report — round 1

**QA:** qa-tester (independent pass, no part in authoring this diff) · **model running:** `deepseek/deepseek-flash`
**Scope:** `origin/fix/untitled-chat-header` @ `6c37dac0258121cd1079c443af0b1533dea236a7` (source delta `a09f2e6f4..6c37dac02`), pre-fix tree `a09f2e6f41dfe921f102366280cd5ca59f84634a`.
**Head note:** the brief named `7fc365b72`; the PR head has since moved to `6c37dac02`. I verified the delta myself — `git diff 7fc365b72..6c37dac02 --stat` touches `docs/evidence/` only (`chat-title/README.md`, 8 `.webp` deletions, `manifest.json`), and `git diff --name-only 7fc365b72..6c37dac02 | grep -v '^docs/evidence/'` is empty. Evidence-only, no source or test change; I reviewed and drove `6c37dac02`.

**Verdict: PASS on the change's behaviour** — every matrix item the diff owns passes on the head with real executions, and both halves of the defect reproduce on the pre-fix tree with the exact strings. Two **minor documentation findings** (Q-1, Q-2) and one **BLOCKED** matrix item (M4) are recorded below; Q-1/Q-2 are one-line restatements and can ride the batched remediation.

---

#### Environment (every run identical in shape)

- Worktree of the PR head plus a detached worktree at the pre-fix SHA; `pnpm install --prefer-offline`, then `pnpm build` in each (dummy `VITE_GOOGLE_*` / `VITE_MICROSOFT_*`). The fresh worktree's `node_modules/electron` had no `path.txt` and an incomplete `dist/` (pnpm had not run its postinstall), restored from the sibling worktree — environment, not repo content.
- Isolated backend per run: fresh scratch dir, `LOCAL_OPERATOR_CONFIG_DIR`/`LOCAL_OPERATOR_HOME` inside it, `local-operator serve --host 127.0.0.1 --port 14333`, shared `LOCAL_OPERATOR_DESKTOP_TOKEN`. `~/.local-operator` was never read or written; every `CMUX_*` variable unset before boot; synthetic 12-hex session ids.
- The **built** app: `npx electron . --remote-debugging-port=9451 --user-data-dir=<scratch>/profile` from each tree's own `out/`, with `VITE_LOCAL_OPERATOR_API_URL=http://127.0.0.1:14333` and `VITE_DISABLE_BACKEND_MANAGER=true`.
- One seeder (`/tmp/qa120/seed.mjs`) and one driver (`/tmp/qa120/drive.mjs`) written for this pass — my own, not the evidence harness. Clicks are real `Input.dispatchMouseEvent` presses at the target's own coordinates, typing is `Input.insertText` into the real control, Enter is `Input.dispatchKeyEvent`; nothing calls a store setter or a React handler.
- Seeded store, read back through `GET /v1/desktop/sessions` and the backend's own events stream: `a1a1a1a1a1a1` = opener-named, no sidecar, **live `conversation_title: ""`**; `b2b2b2b2b2b2` = `title.json` `QA bravo journalled title`, **live `""`**; `c3c3c3c3c3c3` = nameless (`""` row → `Untitled chat`); `d4d4d4d4d4d4` = sidecar **and** a turn-end checkpoint agreeing (`QA delta journalled title`); `e5e5e5e5e5e5` = sidecar vs checkpoint disagreeing (probe only).

---

#### Matrix (surface | command | actual output | verdict)

| # | surface | command / interaction | actual output | verdict |
| 1a | opener-named row, **head** | `node drive.mjs 9451 m1` — click row at `{x:360,y:784}` | `0.0s header=<null>` → `0.4s header="QA alpha opener names this row"`; every row keeps its own text; 2 transitions in 14 s, final `header="QA alpha opener names this row"` / row `…names this row*` | **PASS** |
| 1b | same, **pre-fix** | same driver, baseline tree | `0.2s header="Untitled chat"` over a row reading `QA alpha opener names this row`; `0.3s` the **row itself** reads `Untitled chat` (span *and* its `title` attribute), restoring ~100 ms later; final (14 s) `header="Untitled chat"` | **FAIL — defect reproduced** |
| 2a | journalled-title row (`bravo`, no checkpoint), **head** | `m2` | `0.0s header=<null>` → `0.8s header="QA bravo journalled title"` with no intermediate wrong name; row same | **PASS** |
| 2b | same, **pre-fix** | `m2` (baseline) | final `header="Untitled chat"` over row `QA bravo journalled title*` | **FAIL — reproduced** |
| 2c | live-title case (`delta`), **head** | `m2b` | `0.6s header="QA delta journalled title"` with no intermediate wrong name | **PASS** |
| 2d | same, **pre-fix** | `m2b` (baseline) | `0.4s header="Untitled chat"` → `0.7s` correct name | **PASS-with-flash** (ends correct; the flash is the defect) |
| 2e | divergent sidecar/checkpoint names (`echo`, synthetic probe), both trees | `m2b` | head `0.9s` row name → `1.4s` live checkpoint title; pre-fix `1.0s "Untitled chat"` → `1.6s` live title; row keeps the sidecar name on both | **PASS — live still wins on both trees; precedence unchanged** |
| 3a | nameless session, **head** | `m3` | `0.5s header="Untitled chat"`, row `Untitled chat*`; no other header value in 8 s | **PASS** |
| 3b | same, **pre-fix** | `m3` (baseline) | identical | **PASS** |
| 4 | `/rename` on the nameless conversation, **head** | composer → `/rename` → Enter (completes to `"/rename "`) → Enter (submits) → picker (`prefilledValue: ""`) → type `QA renamed charlie` → click the picker's `Rename` | picker opens and submits (`POST …/c3c3c3c3c3c3/commands`), backend answers `{"detail":"No model provider is configured yet. Connect one in Settings > Providers, then send the message again."}`; no `title.json`; header/row unchanged at 12 s | **BLOCKED** |
| 5a | deep link, catalogue not loaded, **head** | `m5 cold` (clear the persisted catalogue, `#/chat/a1a1a1a1a1a1`, reload) | `0.0–1.1s header=<null>` while rows paint, then `1.4s header="QA alpha opener names this row"`; **never `Untitled chat`** | **PASS** |
| 5b | same, **pre-fix** | `m5 cold` (baseline) | `1.0s` sidebar only → `1.7s header="Untitled chat"` over the selected named row, and it stays | **FAIL — reproduced on the deep-link path** |
| 5c | deep link, catalogue warm, **head** | `m5 warm` | `0.4s header="QA alpha opener names this row"` | **PASS** |
| 6a | store rule over time, **head** | `m6`: alpha/bravo open-close ×3, then sit on alpha for 16 s, 100 ms sampling | 8 transitions, all header/selection; **no row ever left its own name** across 3 rounds + 16 s (longer than the 5 s poll) | **PASS** |
| 6b | same, **pre-fix** | `m6` (baseline) | the selected row reads `Untitled chat` (span and `title`) at `0.5s`, `5.8s`, `11.6s`, `17.3s` — once per open — restoring each time; header `Untitled chat` throughout | **FAIL — reproduced 4×** |
| 7a | sidebar search matches the row title, **head** | `m7-search`, select-all then type each query | `QA alpha opener` → only that row (alpha present, the other four absent); `bravo journalled` → only `QA bravo journalled title`; `qa bravo opener` **and** `QA bravo opener` → **no conversation rows** (bravo's opener is `qa bravo opener differs from the title`); `zzz no such chat` → none; `""` → all five, all sections still paint | **PASS** — matches on the row's NAME; the opener text does not match even at its own casing (observation below, pre-existing) |
| 7b | grouping matches the row title, **head** | `m7-grouping` | `Previous chats` rows are exactly the seeded names; each equals the name its header wears | **PASS** (agent/team grouping not covered — no binding in the fixture) |
| 7c | canvas/options header keeps the name, **head** | `m7-canvas`: open alpha, click `[data-tour-tag="open-canvas-button"]` | header stays `QA alpha opener names this row` through the click and after Escape | **PASS for the header**; panel never mounted (`canvas-container`/`close-canvas-button` absent, opener still visible) → with-panel state **BLOCKED** |
| 7d | repeated open/close, **head** | `m7-cycles`: alpha→bravo→charlie→alpha→bravo ×2 | 10/10 clicks: header equals the clicked row's name; no row accumulates a wrong name | **PASS** |
| 7e | backend down, **head** | `m7-offline` (no backend started) | shell and chat page render, banner `The server is offline. The interface will not function properly until the server is back online.` plus the second honest explainer; no error boundary | **PASS** |
| — | gates at the head | `pnpm lint`, `check-types`, `check-themes`, `test:desktop`, `check-evidence`, `build` | lint 0 errors (24 pre-existing warnings); types 0; themes `2168 assertions across 12 themes, 0 pinned exception(s)`; **`test:desktop` 493 tests, 492 pass, 1 fail**; `check-evidence` `Evidence holds: **681** frames are pictures of their own theme (worst ΔE00 18.1)`; build OK | **PASS** |
| — | the known failure, attributed | `node --test scripts/update-robustness.test.mjs` on the **pre-fix** tree | `'pip' !== 'uv'`, same test name — fails identically at `a09f2e6f4`, so not this change's | **confirmed pre-existing** |
| — | the PR's own unit file | `node --test scripts/chat-title.test.mjs` | `tests 9 / pass 9 / fail 0` | **PASS** |
| — | committed frame set | 16 `.webp` under `docs/evidence/chat-title/` | 16 frames, **16 distinct hashes** — the README's own assertion holds | **PASS** |
| — | manifest arithmetic | committed `.webp` at HEAD vs the declared `supplementary` frames | 681 committed, 87 declared (14 sets), 681 − 87 = **594** = the asserted `frames` field | **PASS** (prose beside it is stale → Q-2) |

---

#### Findings

**Q-1 (minor, documentation — the PR body's own gate block is stale at the head).**
Two numbers and one description in the PR body no longer match the head:

```
$ pnpm test:desktop | tail -6        # in ~/local-operator-ui-worktrees/qa-120 @ 6c37dac02
ℹ tests 493
ℹ pass 492
ℹ fail 1
```
The body records `445 tests: 444 pass, 1 fail`. Same single environmental failure (`pip !== uv`, reproduced on the pre-fix tree), so the substance is right and only the count is stale. Likewise the body says `pnpm check-evidence … 645 frames`, while at the head the gate prints `Evidence holds: 681 frames are pictures of their own theme (worst ΔE00 18.1)`. And §2 describes the capture as "the real Electron **dev** app" while §3, the manifest `source` field and the README all say the **built** app — which matters here, because the README itself records that under `electron-vite dev` the dev-only tab strip paints over the header. One restatement of the three values closes it.

**Q-2 (minor, documentation — the manifest prose contradicts the file it sits in).**
`docs/evidence/manifest.json` → `countsMean.frames` says "665 `.webp` are committed and the thirteen `supplementary` sets below account for 71 of them, leaving 594". At this head:

```
$ git ls-tree -r --name-only HEAD docs/evidence/ | grep -c '\.webp$'   # 681
$ python3 -c "import json;m=json.load(open('docs/evidence/manifest.json'));print(sum(s['frames'] for s in m['supplementary']), len(m['supplementary']))"
87 14
```
681 − 87 = 594, so the **asserted** `frames: 594` is still exactly right and `check-evidence` passes; but the sentence's own 665 / thirteen / 71 are the pre-`chat-title` arithmetic, and the `chat-title` set this PR adds (16 frames, the 14th set) is what made them stale. Repro: the two commands above. Restating the pair to `681 … fourteen … 87` is the whole fix.

**Q-3 (informational — the `open-bravo` pair exercises the fallback, not the journalled title).**
Worth knowing because a reader of the evidence README will otherwise infer that `open-bravo` proves the header still shows a *journalled* title. The live `conversation_title` does not come from `title.json`: it is restored from the session's last **turn-end checkpoint** (`_restore_cold_details` ← `FRONTEND_CHECKPOINT_CUSTOM_TYPE`). I verified this on the wire, not by reading the code:

```
$ curl -sN -H "Authorization: Bearer $TOKEN" \
    http://127.0.0.1:14333/v1/desktop/sessions/b2b2b2b2b2b2/events   # bravo: sidecar present, no checkpoint
"conversation_title":""
```
So on the pre-fix tree `bravo` reads `Untitled chat` because its *live* title is empty, and the committed `open-bravo` pair differs because the fix falls back to the catalogue row — a correct and valuable demonstration, but of the **catalogue fallback**, not of "live title wins". The live path is covered by `delta`/`echo` below and by `/rename` in the product. One clause in the README (or a checkpoint-carrying third row in the fixture) would make that explicit. Not a defect claim — the README's table already states the observed pre-fix string correctly.

**Q-4 (deferred — pre-existing, outside this diff).** While finding a path to `/rename` I found the options strip in `chat-page.tsx` (the one carrying the `Rename` button) is unreachable: `ChatHeader`'s canvas button calls `setCanvasOpen(true)` and never the `onOpenOptions` it is gated on, so nothing sets `options` true. `deferred — pre-existing, not touched by this diff, no follow-up issue opened`.

#### What this QA pass adds to the evidence set

- The committed frames deliberately omit the row's transient blanking (README §"What these frames do NOT prove", because three samples came back byte-identical). A 100 ms read shows it directly and repeatedly on the pre-fix tree (matrix 1b/6b: `Untitled chat` over the just-clicked row at 0.3 s, and once per open at 0.5/5.8/11.6/17.3 s), and shows **no such transition at all** on the head. That half of the defect is now pinned on the real app as well as in the unit suite.
- The deep-link path (matrix 5) is new information: pre-fix it opens on `Untitled chat` over a named row and stays there; on the head the header is simply absent until the catalogue answers and then names the row — no wrong-name window at any sample.

#### BLOCKED, with why

- **M4 (`/rename` → header + row, no blanking return).** The rename UI itself is reachable and works up to the backend — the picker mounts, prefills `""` for a nameless conversation, accepts the typed name and submits — but the isolated backend refuses **every** session command without a configured model provider:
  `POST /v1/desktop/sessions/c3c3c3c3c3c3/commands {"command":"rename","args":"QA renamed charlie"}` → `{"detail":"No model provider is configured yet. Connect one in Settings > Providers, then send the message again."}`, and no `title.json` is written. Completing it would mean wiring the operator's credentials into a synthetic store, which this pass will not do. The post-rename header/row state is therefore **not verified**. The store-rule behaviour that a rename exercises is covered by 6a/6b (the write path is the same code); the rename-specific "does the blanking come back" observation is not.
- **7c with-panel state.** The canvas panel did not mount under my driver (header button left `isCanvasOpen` false, no `canvas-container` in the DOM), so only the header's name was verified through the toggle. Not attributed to this diff — no canvas code is touched — but I did not see the panel.
- Not covered, for honesty: packaged/notarised build, themes other than the default, agent/team grouping (no binding in the synthetic catalogue), and any non-empty `conversation_title` that disagrees with the row in a state reachable without hand-seeding (2e is a labelled synthetic probe for precedence only).
- Observation on 7a (pre-existing, this PR does not touch the sidebar filter; no action requested): the search matches the row's **name**, and does not match the conversation's **opening message** — `qa bravo opener` at the opener's exact casing returns no rows, while `bravo journalled` (a substring of the row's name) returns bravo. A conversation whose opener a user remembers, but which carries a different name, is not findable by that opener.

---

*Matrix kept at `/tmp/qa120/matrix.md` (reproducible from `/tmp/qa120/seed.mjs`, `/tmp/qa120/drive.mjs`, `/tmp/qa120/run.sh` — all outside the repo). No repo files were edited, no reviewers added, nothing merged.*

damianvtran:

**Head:** `1795e5f60ef3b04ebf32146b20fce53d9feb06d1` · one commit, `1795e5f60` · Reviewer round: `deepseek-flash (deepseek/deepseek-flash)`, scope `b11ec233d..6c37dac02` · no provider fallback in this round either.

---

**F1 (minor) — fixed — `1795e5f60`.** The `{title: catalogueTitle}` branch is gone, with its `catalogueTitle` input and the JSDoc clause that described the no-op as behaviour ("a blank or whitespace-only one falls back to the row's own name"). The call site is back to the `some(...)` existence guard it had before this branch — with the title no longer read from the row, `find` had nothing left to do — and `scripts/chat-title.test.mjs`'s `streamTick` helper now takes only what the write reads, so it can no longer keep a dead branch green by passing the row's own value.

Your mutation battery is now the shape of the code: the surviving rule is the one the suite can defend — a non-blank live title is written so `/rename` propagates, and a blank or whitespace-only one writes **nothing**, which is what leaves the catalogue's name standing. I kept and re-stated the load-bearing half you proved rather than assumed: a key PRESENT with an `undefined` value still overwrites through `{...current, ...incoming}`, so only the absent key is safe — which is why the empty-object case is the one that matters and why it is tested against the real store.

Red check re-run against the pre-fix expressions (the same transcription, now with the `catalogueTitle` input gone):

```
$ node --test out/red-run/chat-title-red.test.mjs     # pre-fix expressions substituted into the shipped suite
ℹ tests 9   ℹ pass 3   ℹ fail 6
✖ a blank live title falls back to the catalogue name the row was clicked under
✖ a live title wins, so a rename still shows where it landed
✖ a streaming tick with an empty live title leaves the row's name standing
✖ a whitespace-only live title cannot blank the row either
✖ a row with no name at all gains none from an empty live title
✖ isBlankTitle is the one definition both rules read
$ node --test scripts/chat-title.test.mjs             # the shipped suite
ℹ tests 9   ℹ pass 9   ℹ fail 0
```

**F2 (minor) — fixed, PR metadata only.** The gate block now carries this head's measurements, re-run by me rather than carried over: `test:desktop` **493 tests: 492 pass, 1 fail** (the one failure is the pre-existing environmental one, `the operator's own install is classified rather than told to use pip`, which fails identically at `origin/main` on the untouched file) and `check-evidence` **681 frames**. No commit for this one.

**F3 (minor) — fixed — `1795e5f60`.** README: "a sixth of the header band changes (36148 of 211,200 px, the crop's own 0.171155)". Every number in the table is unchanged; only the sentence a reader quotes.

**F4 (nit) — fixed — `1795e5f60`.** The re-capture assertion is now scoped to the **clicked** states, and `rest` is named as the exception that must stay zero: it is the control, the header is absent before a conversation is open, and its band delta is 0 by design in both palettes — with the note that a non-zero delta there means something moved that no finding claims.

**F5 (nit) — deferred, recorded as such — no change.** You recorded it as "no change required" and QA's independent deep-link run measured the head as `header=<null>` until the catalogue answers, i.e. no wrong-name window in the state its samples could see. The designer's suggested fix needs new store state ("has the catalogue ever loaded") precisely to tell *not loaded* from *loaded and unnamed*, which is more surface than the ~150–750 ms flash warrants in this PR; and it is not a regression, because the pre-fix header never corrected at all. Deferred with your reasoning intact rather than silently dropped.

**Also acknowledged, not actioned** (both from your "recorded, not findings" prose): the `UNTITLED_CHAT` docstring's "asserted to agree" is indeed a touch stronger than what exists — no assertion binds `chat-sidebar.tsx`'s two literals — and the store census (1407 / 226 / 1180) is not reproducible from this repository without opening the operator's store, which neither of us did. Both are recorded here so the next reader does not have to rediscover them; neither changes behaviour.

damianvtran:
### QA remediation — round 1

**Head:** `1795e5f60ef3b04ebf32146b20fce53d9feb06d1` · one commit, `1795e5f60` · answering your pass at `6c37dac02` · nothing here changes behaviour.

---

**Q-1 (minor, documentation) — fixed, PR metadata only.** The gate block is now this head's measurements, re-run by me rather than carried over:

```
pnpm lint          Found 24 warnings.  (0 errors; pre-existing in bin/ and linux-sandbox.test.mjs)
pnpm check-types   exit 0
pnpm test:desktop  493 tests: 492 pass, 1 fail (the pre-existing, environmental `…classified rather than told to use pip`)
pnpm check-themes  Contrast contract holds: 2168 assertions across 12 themes, 0 pinned exception(s).
pnpm check-evidence Evidence holds: 681 frames are pictures of their own theme (worst ΔE00 18.1, docs/evidence/chat-usage/dense/localOperatorLight.webp).
pnpm build         ✓ built in 33.94s
```

Your 493/492/1 reproduces exactly, and so does the 681. The one failure is the same pre-existing one both of us identified (the `local-operator` first on this machine's PATH is the desktop app's venv, `INSTALLER = pip`, while the test asserts `uv`) — it fails identically at `origin/main` on the untouched file. The §2 description of the app was also yours to catch and is corrected: it now says **built** app (`pnpm build`, `npx electron .`), matching §3, the manifest `source` and the README, with the reason the dev app cannot serve here (the development-only tab strip paints over the header).

**Q-2 (minor, documentation) — fixed — `1795e5f60`.** The manifest's `countsMean.frames` prose restated to your own arithmetic: **681** committed `.webp`, **fourteen** `supplementary` sets, **87** declared frames, 681 − 87 = **594**, the asserted `frames` field unchanged — plus a clause that the `chat-title` set's sixteen frames are the addition and are outside that term by construction. `pnpm check-evidence` re-run at this head exits 0 with the line quoted above.

**Q-3 (informational) — fixed — `1795e5f60`.** The README now carries a clause under its opening table stating exactly what your wire reading showed: `Retention sweep notes` is journalled but its session has **no turn-end checkpoint**, so its live `conversation_title` is empty on the wire, and that pair therefore demonstrates the **catalogue fallback** — the fix reading the row's name where the live field is blank — and not the "live title wins" path, which the live-first branch and the unit suite cover. No re-capture, as you suggested.

**Q-4 (deferred) — `deferred — pre-existing, outside this diff`.** Confirmed and recorded verbatim in this thread: the options strip carrying `Rename` is unreachable because `ChatHeader`'s canvas button calls `setCanvasOpen(true)` and never the `onOpenOptions` it is gated on. Pre-existing, not touched by this diff, no follow-up issue opened.

**M4 (`/rename` on an isolated backend) — BLOCKED, and accepted as environmental.** Your run measured what I also hit while building the fixture: every session command is refused without a configured model provider —

```
POST /v1/desktop/sessions/<id>/commands {"command":"rename","args":"…"}
→ {"detail":"No model provider is configured yet. Connect one in Settings > Providers, then send the message again."}
```

— and no `title.json` is written. Completing it would mean wiring the operator's credentials into a synthetic store, which is the isolation rule this pass (and this PR's evidence) is built on, so **rename remains unverified on an isolated backend by design**. What that leaves uncovered is stated plainly rather than implied: the rename-specific "does the blanking come back" observation is not verified end to end, while the store-write path a rename exercises is the same one your 6a/6b and the unit suite already drive (`a non-blank live title does reach the row`, `a live title wins, so a rename still shows where it landed`). The design round records the same gap from its side (`/rename` propagation is reasoning plus unit coverage, not a frame).

**Acknowledged from your "what this pass adds"** and not actioned here: your 100 ms sampling showing the committed frames omit the row's transient blanking, and that the head shows no such transition at all — that is the second half of the defect pinned on the real app, and the README's "not photographed" scoping already says the frames do not carry it. Your deep-link M5 result (header absent until the catalogue answers, no wrong-name window) is quoted in the agent-review-remediation comment above as part of why F5 stays deferred.

damianvtran:

**Head:** `1795e5f60ef3b04ebf32146b20fce53d9feb06d1` · one commit, `1795e5f60` · answering your pass at `6c37dac02`. D1–D3 are declined with reasons, so per your own closing line the design gate is closed on this head; nothing visual changed, and no frame was re-captured.

---

**D1 (nit, copy) — declined here, deliberately, recorded.** Agreed on the mechanism and on your recommendation: the multi-line opener reaching the header as `Fix the retention sweep It deletes rows that are still referenced Please check the audit ta…` is the backend's `_condense` joining lines with a space (`~/local-operator/local_operator/resume.py`), and prettifying it in the header would break the row↔header equality this PR exists to establish — the header would then wear a string the row does not. It is a different repository's string contract, so it is recorded here for a backend follow-up and no issue was opened. No change.

**D2 (nit, state) — deferred, with your measurement.** The cold deep link wears `Untitled chat` for ~194 ms (146→340 ms) and up to ~748 ms on a first cold start. Declining to fix it in this PR, on the two grounds you and the reviewer both supply: it is dormant in every in-app click path (a row cannot be clicked before the catalogue painted it), and QA's independent deep-link run measured the head as `header=<null>` until the catalogue answers, i.e. no wrong-name window in the states its samples saw. Your suggested fix — render the title slot empty (avatar only) until either source resolves — is the right shape, but in the current module it needs new state to distinguish *the catalogue has never loaded* from *the catalogue loaded and this conversation is unnamed*, because `boundRow` being `undefined` means both. That is a new store concern for a self-correcting flash, so it is recorded as a deferred nit rather than carried by this PR.

**D3 (nit, affordance) — declined — not addressed.** Agreed that `title={agentName}` would match the h2's own sibling (`title={description}`), and that the header is where a cut would read worst. Declined because it is dormant and unreachable at today's limits — catalogue names cap at 64 characters, `user_set` renames at 80 (`naming.py`), the title box is 760px wide, and `truncate` never engaged in any state you or I measured — and because it would widen this diff into `chat-header.tsx`, which #115 changed in the very merge this branch was rebased onto. Recorded as not addressed; a follow-up can carry it alone.

**Acknowledged from your "recorded, not findings", not actioned here:**

- `"Untitled chat"` existing in three places (`UNTITLED_CHAT` plus the sidebar's two literals) — the copy-paste is real, and the module docstring records the deliberate decision to leave that file alone in this diff. Your point that importing the constant there would make the invariant structural is the better end state; it is a sidebar change and does not belong in this PR. The reviewer raised the same string from the other side (the docstring's "asserted to agree" being slightly stronger than what exists), and both are now stated plainly rather than left implied.
- `_condense`'s dangling separator (`…rows again -…`) and the search behaviour QA noticed (matching the row's name but not the conversation's opening message) are both outside this diff and untouched.
- Your note that `/rename` propagation is unpictured and rests on unit coverage plus reasoning is accurate and is repeated in the QA remediation for the same reason: the isolated backend refuses session commands without a configured model provider, so that flow cannot be photographed without wiring the operator's credentials into a synthetic store — which this evidence set will not do.

damianvtran:
**Design sign-off stands over `1795e5f60` — the delta is a proven no-op for the rendered state.**

Verified myself rather than taken from the remediation note:

- **The artifacts I signed off are literally unchanged.** No `.webp` appears in `git diff 6c37dac02..1795e5f60` — all 16 frames in `docs/evidence/chat-title/` are byte-identical, so the frames of round 1 remain the frames of this head.
- **`src/` is a value-identical change.** The removed branch returned `{title: catalogueTitle}` and its only caller passed the row it was updating (`current.title`); `upsertSession` merges through `mergeRow`'s plain spread (`{...current, ...incoming}`, plus an `attention` merge that does not read `title`), and the module has no other read of `title` and does not persist `sessions` at all (`partialize` excludes them). So `{}` and `{title: <the row's own value>}` merge to the same row, with no durable or rendered effect. The call site's guard change is the same existence predicate (`some(...)` where `find(...)` was, same early return), and no call site passes `catalogueTitle` any more.
- **The prose corrections check out.** The README's re-stated `36148` is exactly my own header-band measurement on that crop, and its new claim that the `rest` control is 0 in *both* palettes holds — I measured 0 in the band and in the title rect, dark and light. The `a live title wins, so a rename still shows where it landed` case is still present in the suite, so the coverage note I recorded (rename path pinned at unit level, not in a frame) is unchanged.

D1–D3 declined with reasons: accepted, no further design round. The design gate is closed on `1795e5f60`.

damianvtran:
### Agent review — round 2
Reviewer: deepseek-flash (deepseek/deepseek-flash) — no provider fallback during this round
Scope: 6c37dac02..1795e5f60

Read the 5-file delta and nothing else (`docs/evidence/chat-title/README.md`,
`docs/evidence/manifest.json`, `scripts/chat-title.test.mjs`,
`chat-title.ts`, `chat-page.tsx`); no version bump, no other surface. Re-ran every
gate at `1795e5f60` in a fresh worktree (`~/workspace/r2-120`); the operator's live
store was never touched.

**Verdict: clean — no blocker, no major, and no open minor. TERMINAL: nothing left
that gates this merge.**

---

#### F1 (minor) — closed, and closed in the strongest form available

The branch, its `catalogueTitle` input and the JSDoc clause are gone; the call site
is back to the `some(...)` guard; the test helper now takes only what the write
reads. Four independent confirmations rather than an argument:

1. **No other caller existed**, so nothing can be stranded by the signature change.
   `git grep -n catalogueTitleUpdate 6c37dac02 -- src scripts` returns exactly the
   one call site (`chat-page.tsx:273`), the import, the test's import and its helper.
2. **Behaviour-preserving in the strongest runnable form.** I drove 36 combinations
   of (row title × live title) — `undefined`/`null`/`""`/`"   "`/named/spaced on both
   sides — through the **real bundled `upsertSession`** twice per case, once with
   round 1's update object recomputed from the old source and once with the shipped
   one, and compared the resulting rows:

   ```
   cases compared: 36
   rows differing round1 vs shipped: 0
   blank live title -> [{"out":{}},{"out":{}},{"out":{}},{"out":{}}]
   ```

   Zero differing rows, not "equivalent by reasoning": the deleted branch could only
   write the row's own title back to itself, and the store's merged output proves it.
3. **The type still expresses "write nothing" as the absent key** — `{}` for
   `undefined`, `null`, `""` and `"   "`, `{title}` only for a non-blank live title.
4. **The store-level cases still exercise the real spread merge** — the helper still
   calls `upsertSession` from the bundled `canonical-sessions-store`, and the four
   write-path cases (one tick, 40 ticks, whitespace, non-blank) are unchanged in what
   they assert.

Red check re-run at this head against the pre-fix expressions, substituted into the
shipped suite: **5 of 9 fail** — the same five as round 1 (blank-live-title fallback,
live-title-wins, streaming tick, whitespace tick, nameless row). Your "6 of 9" adds
the `isBlankTitle` case because that transcription also drops the helper's trim; with
that extra mutation I get 6 and the failing set is exactly the one your comment
lists, so the two counts reconcile and neither is wrong. Green on head: 9/9.

#### F2 (minor) — closed; every number in the block is now this head's

My own runs at `1795e5f60`, not the block's:

```
pnpm lint          24 warnings, exit 0        (matches)
pnpm check-types   exit 0                      (matches)
pnpm test:desktop  493 tests: 492 pass, 1 fail (matches; the one failure is
                   `the operator's own install is classified rather than told to use pip`,
                   the same known environmental one, on an untouched file)
pnpm check-themes  12 themes, 2168 assertions, 0 pinned exceptions (matches)
pnpm check-evidence 681 frames, worst ΔE00 18.1, exit 0 (matches, same worst frame)
pnpm build         exit 0                      (matches; duration differs, cold electron)
```

The "dev app" → "built app" correction is in and reads correctly.

One item I was tracking for this round closed itself while I was verifying it: the
`What and why` bullet still read *"the catalogue write now reads the row it is
updating"* in the body I fetched at the start of the round, and the body fetched at
the end of it reads *"the catalogue write adds a title only when the live one is
non-blank, which is what leaves the name the catalogue already supplied standing"* —
the removed read is no longer claimed anywhere. Confirmed against
`git show 1795e5f60:src/renderer/src/features/chat/components/chat-page.tsx` (the
`some(...)` guard, one `catalogueTitleUpdate({ liveTitle })` call) and
`chat-title.ts:110-116` (one input, `{}` when blank). Closed; no finding to answer.

#### F3 / F4 (minor / nit) — closed, arithmetic checked against the tree

- `README` line 94: "a sixth of the header band changes (36148 of 211,200 px, the
  crop's own 0.171155)" — 36148/211200 = 0.171155 exactly, and the crop and command
  are the ones I re-ran in round 1. Statement now matches its own table.
- `README` line 142: the re-capture assertion is scoped to the **clicked** states and
  `rest` is named as the exception that must stay zero. That is what the pixels say —
  `rest` band delta is 0 in **both** palettes by my own `magick compare`, while every
  clicked state is non-zero (36148 / 33235 / 9089 / 7583 / 1940 / 2158), and the
  control framing is intact.
- `manifest.json` `countsMean.frames`: **681 = 594 swept + 87 declared across 14
  supplementary sets**, which is exactly the committed `.webp` count at this head
  (`git ls-tree -r 1795e5f60 docs/evidence | grep -c '\.webp$'` → 681) and exactly what
  `check-evidence` asserts (exit 0). The set list is still 14, the arithmetic the prose
  states is the arithmetic the tree has, and the `chat-title` set is correctly described
  as outside the sweep's term by construction. `srcTree`/`scriptsTree` were re-stamped
  in the same commit and equal `HEAD:src` (`5f3c9224…`) / `HEAD:scripts` (`9429ce13…`)
  exactly.

#### Q-2 / Q-3 — the README's new note is right, and I checked it from the frames

The added paragraph says the journalled-title pair demonstrates the **catalogue
fallback**, not the live-wins path, because that seeded session's live
`conversation_title` is empty on the wire. That follows from the committed frames
themselves: the pre-fix header was `conversation_title || "Untitled chat"`, and
`before-open-bravo` reads `Untitled chat` (I cropped and read it:
`1760x160+1000+0` of the committed `before-open-bravo/localOperatorDark.webp`), so the
live field was blank for that session and the after-frame's `Retention sweep notes`
can only have come from the row. The note is accurate and it is the correct
separation of the two claims.

#### F5 (nit) — accepted; on reflection I withdraw it rather than defer it

Your reasoning holds and QA's independent deep-link samples settle it: the reachable
state paints `header=<null>` until the catalogue answers, so there is no wrong-name
window — my round-1 mechanism ("paints Untitled chat for one round trip") described a
state the app does not actually reach. It was never a regression either (the pre-fix
header never corrected at all), and the store-state fix the designer suggested is more
surface than the artifact warrants. Closed as not reproducible; no code change, no
deferral needed.

---

### Residual risk worth naming on the merge record

1. **The row-blanking half of the defect remains unit-level only.** The frames show
   the header's name and the row's name agreeing, but not the row's transient
   blanking, and the suite's helper re-implements the call site's write rather than
   importing it — so a future wiring regression at that call site (dropping the
   spread, re-introducing a title from the row, changing the merge) would stay green
   here and would not be caught by these frames either. Disclosed in both the body and
   the README; naming it for the record, not as a finding.
2. **`chat-sidebar.tsx` keeps its own `"Untitled chat"` literal** (`:217`, `:222`) and
   nothing asserts it agrees with `UNTITLED_CHAT`; the module docstring's "asserted to
   agree" is stronger than what exists. Recorded, not a finding, and unchanged by this
   round.
3. **The store census (1407 / 226 / 1180) is not reproducible from this repository**
   without opening the operator's store, which neither round did. What is verified is
   the mechanism (`resume.session_name()` = stored title else opener → the catalogue
   row's `name` → the desktop list), so the fix does not depend on the census.
4. **The `chat-title` frames were not re-captured after this src edit.** They stay
   valid because the edit is a proven no-op for every merged row (36/36 identical) and
   it does not touch the header's read at all, and the manifest's tree stamps were
   moved to this head in the same commit — the same practice `6d13dafa0` used. Worth
   one line on the merge record only because a future reader will see `srcTree` move
   without new frames.

All three streams now sit on this head: the round-1 QA and design reports were taken
at `6c37dac02`, all three remediation comments and the design sign-off are posted
against `1795e5f60`, and this round's findings are closed on that same commit. There is no
version bump in the branch.

damianvtran:
### QA report — round 2

**QA:** qa-tester (independent pass, no part in authoring this diff) · **model running:** `deepseek/deepseek-flash`
**Scope:** `6c37dac02..1795e5f60` — head `1795e5f60ef3b04ebf32146b20fce53d9feb06d1`, one commit (`fix(chat): drop the dead catalogue-title branch, and correct the evidence prose`). Round 1 drove `6c37dac02`; the pre-fix tree for the FAIL-on-old side is unchanged at `a09f2e6f4`.
**Verdict: PASS.** The remediation's only code change is a branch that could not change an answer, empirically confirmed on the driven app; my round-1 BLOCKED `/rename` item is now **closed with an end-to-end PASS**; both documentation findings are remediated; gates green apart from the pre-existing environmental failure.

---

#### 1. Delta re-drive (F1: dead branch removed) — no observable change

Same harness as round 1 (isolated seeded backend on its own port, its own `LOCAL_OPERATOR_DESKTOP_TOKEN`, fresh `LOCAL_OPERATOR_CONFIG_DIR`/`HOME` per run, built app from the worktree, real `Input.dispatchMouseEvent`/`insertText`, 100 ms reads). Head rebuilt at `1795e5f60` (`pnpm build`, `out/renderer/index.html` re-stamped).

| # | interaction | actual output (head `1795e5f60`) | verdict |
| --- | --- | --- | --- |
| 1a | `drive.mjs 9451 m1` — click the opener-named row (`a1a1a1a1a1a1`, row `QA alpha opener names this row`) | `0.0s header=<null>` → **`0.3s header="QA alpha opener names this row"`**; single transition; every row keeps its own text; final `header="QA alpha opener names this row"` / row `…names this row*` | **PASS** |
| 1b | `m6` — alpha/bravo open-close ×3, then sit on alpha 16 s, 100 ms sampling | **8 transitions, all header/selection** (0.2/3.2/5.5/8.7/10.5/14.1/15.6 s); no row text ever left its own name; window > the 5 s list poll | **PASS** |
| 1c | pre-fix tree, same `m1` | `0.6s header="Untitled chat"` over the clicked, named row; **`1.2s` the row itself reads `Untitled chat`** (span and its `title` attribute), restored `1.3s`; final `header="Untitled chat"` | **FAIL (reproduced at this head's pre-fix side)** |

F1's deletion changed nothing observable: the header names the clicked row, and the row keeps its own name across three open/close rounds and 16 s (round 1's `6c37dac02` numbers were a 0.4 s settle and 8 transitions of the same shape — the same result, not a coincidence of timing).

#### 2. Round-1 BLOCKED `/rename` — now closed, PASS

The round-1 blocker was the isolated backend refusing **every** session command with `No model provider is configured yet…`. It is a *config* gate, not a credentials one, and a placeholder provider entry in the isolated store satisfies it — no operator credential was used, nothing left my scratch dir:

```
# in the isolated scratch only; the key is a dummy string in THIS server's env, never persisted
/tmp/qa120/probe-m4/config/config.yml:
  values:
    hosting: openai
    model_name: gpt-4o-mini
    providers:
      openai:
        base_url: https://api.openai.com/v1
OPENAI_API_KEY=qa-placeholder-not-a-real-key   # env of the isolated `local-operator serve` only
```
(Worth knowing for the next person: the keys must be nested under `values:` in `config.yml`. A flat `hosting:` at the top level is parsed and **silently ignored**, and the command keeps refusing with the same sentence — that cost me one attempt. Backend behaviour, not this PR's.)

Then, through the app's own composer (this repo's options strip is not reachable from the header — see round 1 Q-4):

```
composer → "/rename" → Enter (completes to "/rename ") → Enter (submits)
picker:   "Rename conversation … Name" with prefilledValue ""   (the nameless conversation)
type "QA renamed charlie" → typedValue read back → click the picker's "Rename"

0.0s   dialog="Rename conversation … Name Working Cancel Rename"   header="Untitled chat"   row="Untitled chat*"
13.5s  dialog="Rename conversation … renamed to QA renamed charlie … Done Rename"
       header="QA renamed charlie"                                row="QA renamed charlie*"
…no further transition for the remaining ~31 s (the trace ran 45 s), and the final read at 45 s
still: header="QA renamed charlie", row="QA renamed charlie"
```

Side effects, read off disk and off the backend rather than asserted:

```
$ cat <scratch>/config/sessions/c3c3c3c3c3c3/title.json
{"text": "QA renamed charlie", "user_set": true, "names": ["QA renamed charlie"]}
$ curl -H "Authorization: Bearer $TOKEN" …/v1/desktop/sessions   → c3c3c3c3c3c3 => "QA renamed charlie"
```

- header shows the new name — **PASS**
- row shows the new name — **PASS**
- no blanking returns over 13.5 s → 45 s, i.e. across ≥3 five-second list polls — **PASS** (and the transition is direct: `Untitled chat` → `QA renamed charlie`, with no intermediate wrong name)
- `title.json` written with `user_set: true` — **PASS**

The 13.5 s latency is the command being answered by a **runtime the backend launches** on demand (its log shows the boot, the MCP handshakes, then `idle for 3.0s … exiting cleanly`), not a stall in the renderer; that is why round 1's fixed 12 s window missed it, and why the QA lesson is to size the window to the backend's own runtime boot.

#### 3. Anything new the remediation introduced — diff-only check

`git diff 6c37dac02..1795e5f60` touches exactly five files. `catalogueTitleUpdate` lost its `catalogueTitle` input and its `{title: catalogueTitle}` branch, and the call site's row lookup is back to `store.sessions.some((row) => row.session_id === sessionId)`. That is behaviour-neutral **by construction**: the deleted branch returned the row's own title, and `upsertSession` merges `{...current, ...incoming}`, so writing the row's title back could never change the merged row; the two remaining branches (`{title: liveTitle}` / `{}`) are byte-identical to before. Item 1 above is the empirical half of that argument.

Also verified in the delta: `scripts/chat-title.test.mjs` now passes only what the write reads (still **9 tests / 9 pass**), so no test was dropped with the branch; `manifest.json`'s prose is corrected to `681 … fourteen … 87` and its `srcTree`/`scriptsTree` re-stamps match the new head (`git rev-parse HEAD:src HEAD:scripts` → `5f3c92244…`, `9429ce137…`); the README's new note states the fallback-vs-live distinction explicitly, and its new arithmetic is right (`36148 / 211200 = 0.171155`, the crop's own fraction).

#### 4. Gates at the head (all from `~/local-operator-ui-worktrees/qa-120` @ `1795e5f60`)

```
pnpm lint          Found 24 warnings. exit 0        (warnings pre-existing; 0 errors)
pnpm check-types   exit 0
pnpm check-themes  themes.generated.css is up to date (12 themes).
                   Contrast contract holds: 2168 assertions across 12 themes, 0 pinned exception(s).
pnpm test:desktop  ℹ tests 493 / ℹ pass 492 / ℹ fail 1
                   the failing one is `the operator's own install is classified rather than told to use pip`
                   ('pip' !== 'uv'), reproduced on the pre-fix tree in round 1 — pre-existing, environmental
node --test scripts/chat-title.test.mjs   ℹ tests 9 / pass 9 / fail 0
pnpm check-evidence Evidence holds: 681 frames are pictures of their own theme
                   (worst ΔE00 18.1, docs/evidence/chat-usage/dense/localOperatorLight.webp)
pnpm build         ✓ built (out/renderer/index.html re-stamped after the checkout to 1795e5f60)
```

`check-evidence` was backgrounded as instructed and took ~25 minutes wall-clock, but the cause is the
machine, not the gate: load average was 300–350 for the whole run (several sessions' `magick`
histogram passes at once), and the result is identical to round 1's at `6c37dac02` and to the PR
body's corrected line.

The PR body's gate block now reads `493 tests: 492 pass, 1 fail` and `Evidence holds: 681 frames are pictures of their own theme (worst ΔE00 18.1)`, which matches my own measurements exactly — **round-1 Q-1 and Q-2 are remediated**, and no stale count remains in the body at this head.

#### Findings

None blocking. Round 1's Q-1 (stale gate numbers), Q-2 (stale manifest prose) and Q-3 (which claim the `open-bravo` pair supports) are all remediated in this commit and re-verified above. Q-4 (the unreachable options strip that carries a `Rename` button) is unchanged and stays `deferred — pre-existing, outside this diff`; the rename path a user actually has is the composer, which is the one driven here. **No new findings in the remediation delta.**

#### Gaps, stated plainly

- Both trees, the seeded store and the placeholder provider are synthetic; no operator credential, no `~/.local-operator` read or write, no live session touched. The rename above is a rename of a synthetic conversation on an isolated backend.
- The one behaviour I still cannot exercise on an isolated backend is a rename whose response arrives *while a turn streams* — that needs a real model, which the placeholder provider deliberately is not. The blanking-regression surface it would add is covered by 1b (same write, 3 rounds + 16 s).
- Not re-run this round (unchanged by the delta, no code path touched): the deep-link, search, grouping, cycles, canvas-header and offline-boot matrix items from round 1 — they exercise surfaces this commit does not alter, and the F1 change is confined to the one catalogue write asserted in 1a/1b.

PR #122, opened 2026-09-13
feat(dev): run the app headless, so an agent run never takes the operator's focus

damianvtran:
## What and why

Agents drive this app on the operator's own desktop, several at a time, and every
agent-driven launch took their keyboard focus: `ready-to-show` called `show()`
unconditionally. A seven-cycle QA matrix is seven interruptions of whatever they
were doing — measured on this machine, the app became the frontmost application
in 5 of 8 samples of a run nobody asked to watch.

The launch behaviour is now a **declared window mode**, resolved before any
window exists, so a run can render at full fidelity and be driven over CDP with
nothing on screen and no focus taken.

**Change class: C1 — developer tooling and launch policy. No user-visible
surface, and `normal` (the shipped path) is unchanged**, so no design round
applies.

---

## The mode

`--window-mode=<mode>` or `LOCAL_OPERATOR_UI_WINDOW_MODE` (the argument wins);
`--window-size=WxH` or `LOCAL_OPERATOR_UI_WINDOW_SIZE` for the geometry.

| Mode | The window | For |
| --- | --- | --- |
| `headless` | created at the requested size, **never shown**, non-focusable, page unthrottled | every test, QA, harness and evidence run |
| `inactive` | `showInactive()`: visible, but the app is never activated and the window never takes focus | a run someone wants to watch or click into |
| `normal` | `show()` — raises and focuses the window | a human starting the app. Unchanged from today |

```ts
mainWindow.on("ready-to-show", () => {
	if (windowLaunch.show === "focus") mainWindow.show();
	else if (windowLaunch.show === "inactive") mainWindow.showInactive();
});
```

`src/main/window-mode.ts` holds the policy and imports nothing from Electron, so
the decisions are unit-testable in process; `index.ts` applies the result.

Nothing else may raise a window: `src/main/window-raise.ts` is the only module in
the main process that calls `show`, `showInactive` or `focus` on one, and
`scripts/window-mode.test.mjs` asserts that over the whole directory (mutation
checked — one stray `focus()` anywhere else fails it). `focusable: false` is
**not** what keeps a headless run out of the way, and the code says so: on macOS
`NativeWindowMac::Show()` activates the app for any non-panel window whatever
`focusable` says, so a shown non-focusable window makes the app frontmost while
`isFocused()` still reads false (reproduced before fixing). A headless run also
delivers no native banners — the notifier refuses delivery before any claim when
the plan says `never`, since the run has nobody at the screen and a banner's own
click handler is a raise path.

An unrecognised mode falls back to `normal` **and logs a warning**, rather than
guessing: a typo in `headless` costs one visible interruption, while a typo in
`normal` silently becoming `headless` would give someone a window they cannot
find.

## Measured — Electron 35.5.1 / macOS 25.6, 2x display

`sample` = samples in which **this** app's pid was the frontmost application
(compared by pid; other agents run Electron on this machine).

| Run | window state (the app's own log) | viewport | rAF | app frontmost | frame |
| --- | --- | --- | --- | --- | --- |
| `headless` 1380x900 | `visible=false focused=false focusable=false size=1380x900 content=1380x872` | 1380x872 | 124/s | **0 / 8** | 2760x1744 px |
| `inactive` 1380x900 | `visible=true focused=false focusable=true size=1380x900 content=1380x872` | 1380x872 | 127/s | **0 / 4** | 2760x1744 px |
| `normal` (control) | shown and focused | 1380x872 | — | **4 / 7**, 3 / 3, 5 / 8 across runs | — |
| `headless` 800x600 | `visible=false … size=800x600 content=800x572` | 800x572 | 124/s | 0 / 3 | 1600x1144 px |
| `headless` 400x300 | clamped and logged: `window size 400x300 clamped to 800x600 (floor 800x600, ceiling 16384)` | 800x572 | 124/s | 0 / 3 | 1600x1144 px |

The headless sample count includes a **second launch** on the same profile in
every run: that arrives as `second-instance`, the other call site that can raise
a window, and it leaves the app equally not-frontmost. `visible=false` is the
fact to read rather than `focused=false` — `focusable: false` forces
`isFocused()` false, so a headless run would print `focused=false` even while
holding the operator's focus.

**Headless is not a degraded rendering path.** The settled chat frame from a
`headless` run and from the `inactive` run — the latter a real, shown window —
differ in one `62x19` box and nowhere else: the seeded transcript's message
time, which the seeder stamps with `Date.now()` (round 3's independent run found
the same box, and two `headless` runs differ in the same box, so it is the wall
clock rather than the mode). Apart from it the frames reproduce pixel-for-pixel.
Frames captured mid-animation differ whatever the mode — two `headless` runs
differ by ~92% on the *first* frame, because the onboarding modal is still
animating.

The one thing that genuinely differs is focus-dependent rendering — carets,
`:focus-visible` rings, anything gated on `document.hasFocus()` — because a
window that is never shown cannot be focused. `hasFocus=false` in both
non-`normal` runs, so evidence about those needs `inactive` mode or CDP
`Emulation.setFocusEmulationEnabled(true)`. That is written into the evidence
README and into the guideline, not left to be discovered.

The clamped case is the point of clamping in the plan rather than leaving it to
Electron: an evidence frame cannot be labelled with a size the window never had.

## Checkable from outside, without trusting the app's word alone

The app prints a `[window-mode]` startup line and, 1.5 s after `ready-to-show`, a
`[window-mode] state:` line. That second line exists because the window server
cannot be asked this question on this machine: `System Events` reports no windows
for a process without Accessibility permission and none for a background process
even with it (verified: a frontmost Electron window counts 1, the same window in
the background counts 0). A run's own log is what makes "it never showed a
window" checkable by whoever reads a QA run's output.

## Evidence, and how to re-derive it

`docs/evidence/window-mode/` carries the README with the table above, the
caveats, and the harness that produced it — an isolated backend and scratch
store, the built app driven over CDP, frontmost sampled by pid:

```bash
pnpm build
bash docs/evidence/window-mode/harness/run.sh . headless headless 1380x900
bash docs/evidence/window-mode/harness/run.sh . inactive inactive 1380x900
bash docs/evidence/window-mode/harness/run.sh . normal   normal   1380x900
```

Frames (the running app, headless, at the two sizes):
[1380x900](https://github.com/damianvtran/local-operator-ui/blob/HEAD/docs/evidence/window-mode/frames/headless-1380x900-chat.png) ·
[800x600](https://github.com/damianvtran/local-operator-ui/blob/HEAD/docs/evidence/window-mode/frames/headless-800x600-chat.png)

## The guideline, since the switch is only useful if agents reach for it

`AGENTS.md` gains **“Running the app without taking the operator's focus”**: the
three modes and when each applies, the launch forms (`pnpm app:headless` /
`pnpm dev:headless` / the env var in a rig that spawns Electron itself), the two
consequences for evidence (read the viewport from the page; focus-dependent
rendering needs `inactive` or focus emulation), and why capture comes from
`capturePage`/CDP rather than `screencapture` — which only works on the
frontmost window and so requires exactly the focus theft being removed. A bullet
in General Project Guidelines makes it a rule, not a tip.

## What already opens no window, so a rebase cannot re-introduce one

Worth stating because the windows seen on this machine come from live-app
harnesses, not from a suite: `pnpm test:desktop` bundles modules in process; the
CI npx smoke test prints its marker from `whenReady()` and exits before a window
exists; the Storybook and CDP capture scripts drive a private `--headless=new`
Chrome. Only `pnpm dev`, `pnpm start`, `npx electron .` and `npx
local-operator-ui` create a window — and the last inherits the environment, so
the same switch covers it.

## Round 1 review

The independent round-1 review (`### Agent review — round 1` above) found one
blocker — the notifier still raised the window, and `focusable: false` does not
prevent a shown window from taking focus — plus four minors and three nits. All
are fixed or answered in `b60818fbc`, with the per-finding dispositions in
`### Agent review remediation — round 1`. The findings are why the raise policy
now has its own module and why a headless run delivers no banners at all.

## Testing details

- `pnpm test:desktop` — 501 pass. 14 in `scripts/window-mode.test.mjs`
  (defaults, each mode's flags, case/whitespace, argument-over-environment in
  both spellings, a missing flag value, unrecognised mode and size, both clamps,
  the startup line, the raise policy exhaustively over the three modes with a
  fake window, and the source-level guard that nothing outside
  `window-raise.ts` raises a window — mutation-checked in this round) and 3 in
  `scripts/desktop-notifier.test.mjs` (no banner delivered on any path in
  `headless` and no claim burned; the control that the same paths do deliver in
  `normal`; the click path per mode).
- `pnpm check-types`, `pnpm lint` (no new warnings), `pnpm check-themes`,
  `pnpm check-evidence`, `pnpm check-edit-diffs`, `pnpm bundle-size`,
  `pnpm startup-closure` — all pass. The manifest's tree hashes are re-stamped
  in their own commit (`2e73ef35b`); no renderer source moves, so no committed
  frame was re-captured.
- `pnpm build` from a clean worktree, then four live runs of the **built** app
  over CDP against a real backend and a seeded session: the table above. 0
  renderer errors (`level:3`) in the runs.
- Not covered: Windows/Linux behaviour of the modes. The policy layer is
  platform-free; `showInactive`/`focusable`/`backgroundThrottling` are Electron
  APIs whose per-platform behaviour is not measured here, and `normal` is
  unchanged, so the risk is confined to agent-driven runs on other platforms.


damianvtran:
### Agent review — round 1

Reviewer: reviewer (deepseek/deepseek-flash)
Scope: 54bf411e06aa61b98f1f93ee1d6d013ae9baf838..b020cafec07aa3d8d72712e3c403545a91af4592

Read: PR body, `git diff origin/main...HEAD`, `src/main/window-mode.ts`, the `index.ts` hunks in full, `scripts/window-mode.test.mjs`, `docs/evidence/window-mode/` (README + harness), the AGENTS.md section. Ran the unit suite and the live harness (commands and output at the end).

**Verdict: 1 blocker, 0 major, 4 minor, 3 nit.**

---

#### BLOCKER — `src/main/index.ts:218-221`, `src/main/desktop-notifier.ts:697-698`: `focusable: false` does not stop `show()` from taking focus, and the notifier is a live ungated `show()`

*What is wrong.* The change's stated safety net is that `focusable: false` makes a stray `show()` harmless: `index.ts:220` "so a stray `show()` cannot grab focus", echoed at `scripts/window-mode.test.mjs:66-69` ("`focusable: false` means a stray `show()` cannot cause one later") and in the PR body. On Electron 35.5.1 / macOS that is false. `NativeWindowMac::SetFocusable` only flips key-window eligibility:

```
shell/browser/native_window_mac.mm:1194  void NativeWindowMac::SetFocusable(bool focusable) {
                                          [window_ setDisableKeyOrMainWindow:!focusable];
```

while `Show()` ignores it and activates the application for every non-panel window:

```
shell/browser/native_window_mac.mm:419   void NativeWindowMac::Show() {
...
                                          // Panels receive key focus when shown but should not activate the app.
                                          if (!IsPanel()) {
:437                                        [[NSApplication sharedApplication] activateIgnoringOtherApps:YES];
                                          }
                                          [window_ makeKeyAndOrderFront:nil];
```

I reproduced this directly (probe script below): a `{ show: false, focusable: false }` window, `win.show()`, sampled from outside by pid — `visible=true focused=false focusable=false` while the app became **the frontmost application** (`64091|Arc` → `6243|Electron` → `64091|Arc`). So the headless mode's guarantee holds only because the shipped `ready-to-show` path never calls `show()` (`index.ts:271-272`), not because of `focusable`.

*Why it matters.* `desktop-notifier.ts:697-698` still raises the window unconditionally:

```js
notification.on("click", () => {
	const target = this.window();
	...
	target.show();
	target.focus();
```

In `headless` that call makes the window visible — the mode is defined as "created and never shown" (README table, AGENTS.md:111) — and activates the app, i.e. exactly the interruption this PR exists to remove, on a path that reaches a real headless QA run: `desktop-notifier.ts:468` (and `:621` for gates) suppress a toast only when `anyFocused()`, which requires a window that is `visible && focused`; a headless window is neither, so completions and gates do fire toasts during the run and the operator's click lands on this handler. `focus()` (`native_window_mac.mm:398-408`) uses `activateIgnoringOtherApps:NO` and is comparatively benign — it is `show()` that takes the focus.

*And the app's own evidence cannot see it.* `index.ts:289` prints `focused=${mainWindow.isFocused()}`, which is `[window_ isKeyWindow]` (`native_window_mac.mm:415-417`) — forced false by `focusable: false` in precisely the state my probe produced. A real focus takeover during a headless run would still print `focused=false`. The same false premise is the stated reason the guard test excludes the notifier (`test.mjs:172-175`).

*Fix.* Thread the resolved plan into the notifier and gate its raise on it: in `headless` neither `show()` nor `focus()` (send `desktop-open-conversation` only), in `inactive` use `showInactive()`. If the "cannot take focus" property is to be claimed at all, it needs `app.isActive()`- or frontmost-based evidence, not `win.isFocused()`.

---

#### MINOR — `scripts/window-mode.test.mjs:169-188`: the guard is narrower than its name, and its exclusions are assumed rather than enforced

The test title says "every window raise in the main process"; it scans only `src/main/index.ts`, only `.show()|.showInactive()`, and never `.focus()`. Falsified in a `/tmp` copy (worktree untouched):

- making `index.ts:453` ungated — `mainWindow.focus();` with the `windowLaunch.show === "focus"` check deleted — leaves the suite at **13 pass / 0 fail**;
- a line holding both a gated and an ungated raise also passes, because one match per line is enough (`if (process.env.WM_MUTATE) otherWindow.showInactive(); if (windowLaunch.show === "focus") mainWindow.show();` → 13 pass);
- it is not vacuous: an added bare `mainWindow0.show();` → **12 pass / 1 fail**, "ungated window raise".

Fix: scan every file under `src/main/`, include `.focus()`, and require the gate on the same statement (or invert: forbid `.show()|.showInactive()|.focus()` outside `window-mode.ts` unless the call site consults the plan) — the notifier exclusion then becomes a real allow-list entry rather than a filename the route happens to miss.

#### MINOR — `AGENTS.md:143-144` (and `:140`) claim more than was measured

"the frontmost application keeps the focus and the menu bar for the whole run: `win.isFocused()` stays false and no `activate` event fires." Nothing in `docs/evidence/window-mode/` measures an `activate` event — the harness greps for `[window-mode]` lines and `drive.mjs` only reads the page — and my probe above shows `isFocused()===false` is compatible with the app being frontmost. The doc rests the strong claim on the weak fact. Same shape at `:140`, where `webContents.capturePage()` is asserted alongside CDP though the harness captures over CDP only. Fix: log `app.on("activate")` + activation in the harness and cite it, or write "not measured".

#### MINOR — headless runs are not "nothing on screen": they raise macOS toasts

Because notification suppression is `anyFocused()` (`desktop-notifier.ts:468`, `:621`) and a headless window is never `visible && focused`, `when_unfocused` completions and gates now deliver native toasts during an agent-driven headless run — where the old always-shown-and-focused launch suppressed them. The README's claim is "without a window appearing on screen at all". Related: `use-desktop-watch-lease.ts:33-42` reports `visible && focused` (false) so the backend treats the run as not-watching, and `use-completion-view.ts:56-58` gates the `sessions.seen` ack on `document.hasFocus()` — so "focus-dependent rendering is the one thing that differs" (AGENTS.md:154) understates it; some focus-dependent *behaviour* differs too. Fix: document it, and/or suppress native delivery when the resolved plan is `headless`.

#### MINOR — an unrecognised mode/size is silent on stdout, and the harness does not check its own launch

`window-mode.ts` reports problems only via `logger.warn(..., LogFileType.BACKEND)` (`index.ts:395-396`), which writes `…/Local Operator/logs/backend-service.log`; `index.ts:401-403` prints the mode line only when `mode !== "normal"`. So `LOCAL_OPERATOR_UI_WINDOW_MODE=hedless` on an agent launch is a silent (to the rig) focus takeover, and `run.sh:116-117` merely greps and echoes the `[window-mode]` lines — it never asserts that the emitted mode and size equal the requested `$MODE`/`$SIZE`, so a run launched in the wrong mode or at an unparsed size still produces frames labelled with the requested values. Fix: echo plan problems to stdout too (they are launch policy, not backend logging) and have `run.sh` compare and exit non-zero on mismatch.

#### NIT — `docs/evidence/window-mode/harness/drive.mjs:71-73`

The comment reads "Capture the layout viewport as laid out, not the window's visible slice" for `captureBeyondViewport: false`, which is the opposite of what the flag does (`false` clips the capture to the visible viewport).

#### NIT — `src/main/window-mode.ts:141` vs `:56`/`:217`

`/^(\d{1,5})…` and the `WINDOW_MAX_EDGE = 16384` ceiling disagree: `40000x900` clamps to 16384 and is reported as clamped, but `100000x900` is rejected as "not `<width>x<height>`" and silently becomes the default 1380x900 — the ceiling clamp is unreachable above five digits, and the message names the wrong reason. `\d+` would make every numeric request reach the clamp.

#### NIT — `docs/evidence/window-mode/README.md:41` vs the PR body

The `normal` control is "**1 / 3** (an earlier … run … measured 5 / 8)" in the README and "**5 / 8**" with the 1/3 dropped in the PR body; I measured 2/4. A range is fine, but the two documents disagree about which run is the control.

---

#### Checked, no finding

- **`normal` is unchanged.** `width`/`height` still resolve to 1380x900 (`window-mode.ts:98-102`), `minWidth`/`minHeight` still 800/600, `focusable: true` and `backgroundThrottling: true` are Electron's defaults, `show: false` and `ready-to-show → show()` are unchanged, second-instance still calls `focus()` in `normal`, and no state line is printed in `normal`. No behaviour change beyond the non-normal modes.
- **Other raise paths.** Second-instance is gated (`index.ts:453`). `app.on("activate")` (`index.ts:912-917`) re-creates through `setupMainWindowWithUpdateService`/`createWindow`, so it inherits the same gating. `update-service.ts`/`update-install.ts` raise no window (only a `BrowserWindow` reference; no dialog/`Notification`/`focus` call). Sign-in goes through `shell.openExternal` (`desktop-ipc.ts:188`), not an Electron popup; the `setWindowOpenHandler` allowlist (`index.ts:293-294`) is therefore latent, and nothing in the shipped renderer calls `window.open` on a trusted auth domain under Electron.
- **Resolution.** Precedence (flag beats env, both spellings), a missing flag value treated as missing rather than as the next flag, unparsable values falling back with a report, and both clamps are correct — independently confirmed by the 13 new cases passing. The only fail-open is an unrecognised mode → `normal`, which is argued and documented; I recorded it above as a MINOR rather than a defect.
- `dialog.showOpenDialog(mainWindow, …)` (`index.ts:652/665/685`) is parented to a window that in `headless` is never shown, so a sheet may be undismissable — I did not reproduce it and it only bites an agent run that opens a file picker, so I am not filing it; worth a line in AGENTS.md if you agree.

#### Verification I ran

1. `pnpm test:desktop` → **497 pass, 0 fail** (13 of them the new `window-mode` suite).
2. Headless live, `ELECTRON_BIN=$HOME/local-operator-ui/node_modules/.bin/electron bash docs/evidence/window-mode/harness/run.sh . review-headless headless 1380x900`:

```
[window-mode] window mode headless: 1380x900, window created and never shown, page throttling off
[window-mode] state: visible=false focused=false focusable=false size=1380x900 content=1380x872
== this app was the frontmost application in 0 of 3 samples (pid 89946)
DRIVE {… "rafPerSecond":132, "visibilityState":"visible", "hasFocus":false, "devicePixelRatio":2,
       "innerWidth":1380, "innerHeight":872, "chatRows":12, "cssPixels":"1380x872"}
== renderer errors in the run: 0
```

Matches the README's `headless` row (sample count aside).

3. Control, same harness with `normal 1380x900`: `== this app was the frontmost application in 2 of 4 samples`, no `[window-mode]` line (by design) — the control does take focus.
4. Focus probe (nothing from this repo; `/tmp` only), Electron 35.5.1, external frontmost sampler by pid:

```
PROBE before-show  visible=false focused=false focusable=false
PROBE after-show   visible=true  focused=false focusable=false
21:25:42.3 64091|Arc   21:25:44.3 6243|Electron   21:25:46.3 6243|Electron   21:25:48.3 64091|Arc
```

That is the blocker: `show()` on a `focusable: false` window makes it visible and makes the app frontmost, while `isFocused()` stays `false`.

damianvtran:

Scope: `54bf411e0..2e73ef35b` (reviewed head `b020cafec`; fixes in `b60818fbc`, manifest re-stamp in `2e73ef35b`).

Every finding is either fixed with a SHA or answered. Two were fixed beyond the letter of the finding, because the finding named a class of defect: the gate now covers the paths that were still ungated, and the guard is over the whole directory instead of the one file that had the bug.

| # | Finding | Disposition |
| --- | --- | --- |
| B1 | `desktop-notifier.ts` raises the window unconditionally on a banner click; `focusable: false` does not stop a shown window taking focus | **Fixed, `b60818fbc`.** All raises go through the new `src/main/window-raise.ts`; the notifier takes the plan's raise policy and orders an `inactive` window without focus, leaving a `headless` one off screen. Reproduced the premise first, on Electron 35.5.1: `{ show: false, focusable: false }` then `show()` → `visible=true focused=false focusable=false` while the app became frontmost (`Arc → Electron`). That is now written down in `window-raise.ts`, `index.ts` and AGENTS.md, and `focusable: false` is no longer described as a safety net anywhere. |
| B1b | (same root cause, found while fixing B1) a headless run still delivered native banners, because suppression is `anyFocused()` and a headless window is never `visible && focused` | **Fixed, `b60818fbc`.** `observe()` refuses delivery before any claim when the plan says `never`, so nothing is delivered *and* no cross-surface claim is burned. `inactive` still delivers, unchanged. Three tests in `scripts/desktop-notifier.test.mjs`: no delivery on any of the three paths with `never`, the control that the same paths do deliver with `focus`, and the click path per mode. Live-testing this needs a real agent turn, so it is unit-covered only — said so rather than implying otherwise. |
| M1 | the source guard was narrower than its name, and its exclusions assumed | **Fixed, `b60818fbc`.** It scans every `.ts` under `src/main/` (not just `index.ts`), includes `.focus()`, and has one named allow-list entry (`notification.show()`, Electron's banner API). Mutation-checked in this round: one stray `mainWindow?.focus()` in `index.ts` → 13 pass / 1 fail; removed → 14 pass. The policy itself is also now exhaustive over the three modes × both operations with a fake window. |
| M2 | AGENTS.md and the README claimed `win.isFocused()`/no-`activate` evidence that was not measured and cannot carry the claim | **Fixed, `b60818fbc`.** Both now say what the evidence is: frontmost-application sampling by pid from outside (0 of 8 for headless across runs including a second launch), and why `isFocused()` cannot be that proof. The `activate` claim is gone — a probe showed `activate` fires in no mode, so it proved nothing either way. `capturePage` is attributed to the platform probe it came from, with the harness's CDP capture named separately. |
| M3 | headless is not quite "nothing on screen": focus-gated *behaviour* differs, and native dialogs have no parent window | **Fixed, `b60818fbc`.** AGENTS.md now lists both: the watch lease reports `visible && focused` and the `sessions.seen` ack is gated on `document.hasFocus()`, so `headless` is the wrong mode for read receipts, leases and the notifier's focus gate; `dialog.showOpenDialog` is called with the window, so a change that opens a file picker needs `inactive`. Banner suppression is now stated as part of the mode. |
| M4 | a rejected mode/size reached only the backend log; the harness never checked its own launch | **Fixed, `b60818fbc`.** Plan problems go to stdout as well, and `run.sh` computes the size the contract says a request must produce (floor 800x600, ceiling 16384) and fails unless the app reports that mode and size — including the clamp case, which is now checked against the contract rather than exempted. Verified by running all five cases; the clamped one passes without being special-cased. |
| N1 | `captureBeyondViewport: false` described backwards | **Fixed, `b60818fbc`** — the comment now says it clips to the layout viewport. |
| N2 | `\d{1,5}` made the 16384 ceiling unreachable and mis-named the reason | **Fixed, `b60818fbc`** — `\d+`, so every numeric request reaches the clamp and the message names the clamp. |
| N3 | README and PR body disagreed on the control run | **Fixed, `b60818fbc`** — both now give the range (5 of 8, 3 of 3, 2 of 4, 1 of 3 across runs) and the PR body was refreshed. |

**Re-measured after the fixes** (the harness re-run end to end, five cases): `headless` 0 of 8 samples frontmost (the 8 includes a second launch, which is the other raise path), `inactive` 0 of 4, `normal` 4 of 7 and 3 of 3; viewports 1380x872 and 800x572 with `visible=false`/`visible=true` as expected; 0 renderer errors. Frame fidelity, with the control the earlier round lacked: two `headless` runs at 1380x900 differ by 0.0106% of channels and two at 800x600 are byte-identical, against 0.0111% between a headless frame and a shown-window frame — so the residual difference is a blinking caret at the composer, not the mode.

**Not fixed, deliberately:** the notifier's delivery gate is exercised by unit tests rather than a live run (a real completion during a headless run needs a driver turn, which is a bigger harness than this change warrants); and the modes are only measured on macOS — the policy layer is platform-free and `normal` is untouched, so the exposure is agent-driven runs elsewhere.

damianvtran:
### Agent review — round 2

Reviewer: reviewer (deepseek/deepseek-flash)
Scope: b020cafec07aa3d8d72712e3c403545a91af4592..2e73ef35b2e73cbdfd14b9600e2fd0f02d319c9a

**Verdict: clean — no blocker, no major.** 2 minor, 3 nit; the round is TERMINAL.

```diff
git diff b020cafec..HEAD --stat
 src/main/window-raise.ts            | 71 ++  (new)
 src/main/index.ts                   | 54 +-
 src/main/desktop-notifier.ts        | 36 +-
 src/main/window-mode.ts             |  7 +-
 scripts/window-mode.test.mjs        |123 +-
 scripts/desktop-notifier.test.mjs   | 95 +
 docs/evidence/window-mode/harness/  | 48 +-
 AGENTS.md, README.md, manifest.json, 2 frames
```

#### Round-1 blocker/major disposition

| # | Status |
| --- | --- |
| **B1** `focusable:false` does not stop `show()`; notifier was a live ungated raise | **CLOSED.** `desktop-notifier.ts:726` calls `raiseWindow(target, this.windowRaise)`; the constructor takes the plan (`:231`, default `"focus"` = shipped), `index.ts:519-522` passes `windowLaunch.show`; `ready-to-show` (`index.ts:277`) and `second-instance` (`index.ts:469`) both go through `window-raise.ts`. A grep of `src/` for `.show()|.showInactive()|.focus()` on a window returns only `window-raise.ts` and the allow-listed `notification.show()` (`desktop-notifier.ts:729`, the sole `new Notification` site). Mutation-checked, see below. In `headless` no banner is delivered at all (`observe()` returns at `desktop-notifier.ts:417` before any claim — `requests === []` asserted), so the click path cannot run there. |
| **M1** guard narrower than its name | **Fixed except one gap**, filed below as MINOR-1. |
| **M2** docs claimed `isFocused()`/no-`activate` evidence | **Fixed in README.md and the PR body** (frontmost-by-pid, explicit "`isFocused()` is not that proof"); **not fixed in AGENTS.md** — MINOR-2. |
| **M3** headless is not "nothing on screen" | **Fixed**: AGENTS.md:163-182 and README.md:93-104 list the focus-gated behaviour, the unparented dialogs, and banner suppression as part of the mode. |
| **M4** harness never checked its own launch | **Fixed and mutation-verified** (`run.sh:130-161`). |
| **N1** `captureBeyondViewport` comment | **Fixed** (`harness/drive.mjs:71-73`). |
| **N2** `\d{1,5}` clamp | **Fixed** (`window-mode.ts:142` is `\d+`; the message names the clamp). |
| **N3** README/PR-body numbers disagreed | **Fixed in README + PR body** (both now 0.0111% etc.); AGENTS.md still carries a stale number — NIT-1. |

---

#### MINOR-1 — `scripts/window-mode.test.mjs:261-267`: the guard is not recursive, so it does not cover `src/main/backend/`

`readdirSync("src/main")` lists only the top level (10 `.ts`); `src/main/backend/` holds 7 more `.ts` a raise could be added to. Reproduced in a `/tmp` copy (worktree untouched): appending `function probeRaise(w: any) { w.show(); }` to `src/main/backend/index.ts` leaves the suite at **14 pass / 0 fail**, while the same line in `index.ts` fails it. The test comment ("scans every module under `src/main/`"), README.md:87 ("over the whole directory") and the PR body ("asserts that over the whole directory") are all stronger than the code. Fix: `readdirSync("src/main", { recursive: true })` and keep the `.endsWith(".ts")` filter.

#### MINOR-2 — `AGENTS.md:147-148`: still claims the retracted "pixel-identical" chat frame

"the settled chat frame is pixel-identical to the one captured from a shown (`inactive`) window: **0 of 14,440,320 channels differ**" contradicts README.md:68-70 and the PR body, which give the same pair as **0.0111% (1,596 of 14,440,320)**. It is the pre-fix claim left standing in the file M2 was about, and it is not reproducible: comparing the committed `frames/headless-1380x900-chat.png` with a frame from my own headless run, `magick compare -metric AE` reports **616 differing pixels (0.0128% of channels)** — two headless runs already differ, so headless-vs-inactive being exactly 0 is impossible. Fix: replace lines 147-148 with the README's numbers (0.0111%, run-to-run control 0.0106%).

#### NIT-1 — `AGENTS.md:150`: "0 of 5 samples" vs README.md:45 "**0 / 8**"

My run reproduced 0/8 (total samples include the second launch). The 5 is from before the harness gained the second launch; align it.

#### NIT-2 — `src/main/window-raise.ts:62-70` vs `origin/main:366-370`: `normal`'s `second-instance` path changed

Shipped code called `restore()` + `focus()`; `raiseWindow` now calls `show()` + `focus()`. The effect is intended (bring the window forward reliably), but the PR body's "`normal` (the shipped path) is unchanged" is not literally true on this path — worth one clause, or a comment saying `show()` is deliberately added here.

#### NIT-3 — `scripts/desktop-notifier.test.mjs:1271`: Biome would wrap the added `assert.deepEqual(...)`

`biome check` reports a format error in this file that is not present at b020cafec (line length over the limit). `pnpm lint` does not include this file, so CI stays green; run the formatter on the added block.

---

### What I verified by running (worktree `/Users/damian/local-operator-ui-worktrees/window-focus`, read-only; `git status` clean after every run)

1. **`pnpm test:desktop`** → `tests 501 / pass 501 / fail 0` (90.8 s). The two changed suites alone: `node --test scripts/window-mode.test.mjs scripts/desktop-notifier.test.mjs` → **47 pass / 0 fail**.
2. **Live harness**, exactly as asked:
   `ELECTRON_BIN=$HOME/local-operator-ui/node_modules/.bin/electron bash docs/evidence/window-mode/harness/run.sh . r2-headless headless 1380x900`
   → exit 0; `== this app was the frontmost application in 0 of 8 samples (pid 79145)`; `[window-mode] state: visible=false focused=false focusable=false size=1380x900 content=1380x872`; `rAF 124/s`, `visibilityState visible`, `hasFocus false`, viewport 1380x872; `== renderer errors in the run: 0`; `== the app reported the mode and size it was asked for`. Matches the README row (0/8, 1380x872, 124/s) and the frame geometry (2760x1744 committed, verified `sips`; 1600x1144 for 800x600).
3. **Guard mutation checks** (all in `/tmp/wm-mut`, a copy with a symlinked `node_modules`; worktree untouched):
   - `presentWindow(mainWindow, windowLaunch.show)` → `mainWindow.show()` at `index.ts:277` → **13 pass / 1 fail** ("these lines raise or focus a window outside window-raise.ts").
   - `raiseWindow(mainWindow, windowLaunch.show)` → `mainWindow.show()` at `index.ts:469` → **13 pass / 1 fail**.
   - the same raise added to `src/main/backend/index.ts` → **14 pass / 0 fail** (MINOR-1).
   - `observe()` minus `if (this.windowRaise === "never") return;` → **31 pass / 2 fail** (both new notifier tests fail).
   - `raiseWindow` minus its `if (show === "never") return;` → **1 fail** (the policy matrix).
4. **Harness self-assertion**, `run.sh` copied to `/tmp/r2-harness` with the launch line forced to `LOCAL_OPERATOR_UI_WINDOW_MODE="headless"` while the label argued `inactive`, ports moved off the default:
   → `FAIL: asked for mode=inactive size=1380x900 (expected 1380x900), the app reported [window-mode] window mode headless: 1380x900, window created and never shown…`, exit 1. The clamp arithmetic in `run.sh:141-152` matches the contract (`WINDOW_MIN_WIDTH 800`, `WINDOW_MIN_HEIGHT 600`, `WINDOW_MAX_EDGE 16384`; `window-mode.ts:209-218`).
5. **Frame reproduction**: `magick compare -metric AE docs/evidence/window-mode/frames/headless-1380x900-chat.png /tmp/window-mode/r2-headless/frame-chat.png` → `616 (0.000127975)` — i.e. the committed frame is reproducible from the harness, differing only by the documented caret-blink variance.
6. **Manifest re-stamp**: `srcTree`/`scriptsTree` in `docs/evidence/manifest.json` equal `git rev-parse HEAD:src` / `HEAD:scripts` (`d91d4895…`, `ec9cd72a…`); the unchanged `head` (`83db293e…`) is an ancestor of HEAD, which `check-evidence.mjs:228-238` allows by design.

**`inactive` delivered, `normal` unchanged in `ready-to-show`, no claim burned in `headless`.** The remediation is correct and complete on B1/M3/M4/N1/N2; the two minors above are doc/guard precision, not new defects.

damianvtran:
### Agent review — round 3

Reviewer: reviewer (deepseek/deepseek-flash)
Scope: 2e73ef35b2e73cbdfd14b9600e2fd0f02d319c9a…52d9489873389859a08f323f9e2abad644b80d90

**Verdict: clean — no blocker, no major. All five round-2 findings are closed as asked; one MINOR and one NIT appear in this delta and are follow-ups, not a reason for another round. This round is TERMINAL.**

```diff
git diff 2e73ef35b..52d948987 --stat
 AGENTS.md                         | 14 ++++++----
 docs/evidence/manifest.json       |  4 +--
 scripts/desktop-notifier.test.mjs |  6 ++++-
 scripts/window-mode.test.mjs      | 57 ++++++++++++++++++++++++++++----------
 src/main/window-raise.ts          |  6 +++++
```

#### Round-2 disposition

| # | Finding | Status |
| --- | --- | --- |
| MINOR-1 | the raise guard was not recursive, so `src/main/backend/` was unscanned | **CLOSED, reproduced.** `scripts/window-mode.test.mjs:281` is `readdirSync("src/main", { recursive: true })`, and the test now pins the recursion with `scanned.some(file => file.includes(sep))`. In a fresh `/tmp` copy (hashes verified against the head): baseline **14 pass / 0 fail**; `w.show()` appended to `src/main/backend/index.ts` → **13 pass / 1 fail**, naming `src/main/backend/index.ts:17`; `w.focus()` appended to `src/main/backend/logger.ts` → **13 pass / 1 fail** (`:216`); the top-level control in `index.ts` → **13 pass / 1 fail**; re-flattening the readdir in the test itself → **13 pass / 1 fail** on the new assertion, *the scan reached subdirectories (scanned 10 modules)*. The scan reaches 17 `.ts` files, 7 of them nested (was 10 flat). |
| MINOR-2 | `AGENTS.md` claimed a retracted "pixel-identical" frame | **CLOSED as asked** — the numbers are now the README's (0.0111%, 0.0106%, 0 at 800x600). The replacement sentence inherits a wrong *cause* from the README; see MINOR-R3-1. |
| NIT-1 | "0 of 5 samples" vs README "0 / 8" | **CLOSED.** `AGENTS.md:156-160` now says 0 of 8, names the second launch, and its control list (5/8, 4/7, 3/3, 2/4, 1/3) is exactly the union of the README's `normal` row and its Before section. |
| NIT-2 | `normal`'s `second-instance` path changed | **CLOSED.** The comment is there and factually right about the sequence: `origin/main:366-370` was `restore()` + `focus()`, `src/main/window-raise.ts:80-87` is `restore()` + `show()` + `focus()`, i.e. exactly one `show()` more. One hole in the justification — NIT-R3-1. |
| NIT-3 | Biome would wrap the added assertion | **CLOSED.** `biome check scripts/window-mode.test.mjs scripts/desktop-notifier.test.mjs` → **no format diagnostics**, exit 0 — the 13 warnings it does report are `useTopLevelRegex` on fixture code the delta does not touch. |

#### MINOR-R3-1 — `AGENTS.md:149-151` (same sentence at `docs/evidence/window-mode/README.md:60-63` and in the PR body): the residual is attributed to "a blinking caret at the composer", and the `0 … at 800x600` control does not reproduce

> "…against 0.0106% between two `headless` runs at that size and **0 between two at 800x600**: the residual is a blinking caret at the composer, not the mode"

I re-ran the harness at both committed sizes on this head and compared with the committed frames:

- `headless 800x600` → `magick compare -metric AE docs/evidence/window-mode/frames/headless-800x600-chat.png /tmp/window-mode/r3-800/frame-chat.png` = **649 differing pixels**, all of them inside a single **`60x19+1448+902`** box. Cropped from both frames, that box reads **`9:58 PM` (committed)** vs **`10:47 PM` (my run)**.
- `headless 1380x900` → **355 differing pixels**, all inside **`62x19+2606+1414`**: **`9:59 PM`** vs **`10:50 PM`**.

So the run-to-run residual is not a caret — it is the seeded transcript's **message timestamp**, which is derived from the wall clock: `docs/evidence/window-mode/harness/seed.mjs:38` is `const started = Math.floor(Date.now() / 1000) - 60;` and the renderer prints that time with `hour: "numeric"` (`src/renderer/src/features/chat/utils/message-grouping.ts:66`).

Three consequences:

1. **The `0 between two at 800x600` control is not a property of the mode.** It holds only when both runs' seeded rows render the same minute string, so a reader following "How to re-derive" gets a non-zero difference and no way to tell the frame is right. It did not reproduce for me (649 px).
2. **A caret cannot produce it, and the location is wrong.** `headless` (and `inactive`) never focus anything — the run's own `DRIVE` line reports `hasFocus:false` and the composer shows its placeholder — so there is no caret to blink; and the differing box is at the transcript's message time (x 724-754 CSS of 800, y 451-460 of 572), not at the composer. The README's "19 of 1,744 rows" fits this box exactly, which is how a timestamp looks like a caret in a row histogram.
3. The `0.0111%` (headless vs `inactive`) and `0.0106%` pairs are the same magnitude as this box, so they are most likely the same timestamp — i.e. the numbers are honest measurements but not evidence about the mode either way.

Everything else about the fidelity claim holds, and that is worth stating: **apart from that one box the committed frames reproduce pixel-for-pixel** from the harness at both sizes (649 px at 800x600 and 355 px at 1380x900, all inside the timestamp box), so "the frames depict the renderer under review" is intact.

Fix (one line each): say what the residual is — the seeded transcript's message time — and make the control real by freezing the seed clock (pass a fixed `started` to `seed.mjs`) before quoting a byte-identical pair, or drop the "0 at 800x600" clause. The same sentence is in `docs/evidence/window-mode/README.md` and the PR body, so the correction belongs in all three.

#### NIT-R3-1 — `src/main/window-raise.ts:56-58`: "the extra call is a no-op there — a `normal` window is never hidden, since the app has no hide path"

The app does expose the standard macOS hide path: `src/main/index.ts:183` registers `{ role: "hide" }` in the app menu, i.e. Cmd+H hides the app's windows. On that path the added `show()` is not a no-op — it orders the window front again (arguably what a second launch wants, so the behaviour is fine). The sequence claim is right; only the justification is overstated. Say "no hide path of its own" or name Cmd+H.

#### Delta hygiene and the manifest

- The delta touches exactly the five files above. `git diff -U0 2e73ef35b..52d948987 -- src/main/window-raise.ts` has no non-comment added line, so no behaviour moved with it; the only non-fix churn is formatter output in `scripts/window-mode.test.mjs` (six unrelated assertions re-wrapped by Biome), which changes no assertion.
- `git rev-parse HEAD:src` = `79a8f47d74f3b9d493985717c02b8c55bb297005` and `HEAD:scripts` = `bb2b36d544ec94c74794ff0b2baf85d2c861a728` — both equal the re-stamped `srcTree`/`scriptsTree` in `docs/evidence/manifest.json`. The unchanged `head` (`83db293e…`) is an ancestor of HEAD (`git merge-base --is-ancestor` → yes), which `check-evidence.mjs` allows.
- The "no renderer source moved" claim is true: `git diff --name-only 2e73ef35b..HEAD -- 'src/renderer/**'` is empty, so not re-capturing the 665 committed frames is correct.
- `pnpm check-evidence` → exit 0, `Evidence holds: 665 frames are pictures of their own theme (worst ΔE00 18.1, docs/evidence/chat-usage/dense/localOperatorLight.webp)`.

Process note, not a finding: the round-2 dispositions are in the fix commit's message (`288de114c`), not in a PR comment — the PR has three comments and the latest is the round-2 review. If the merge gate wants `### Agent review remediation — round 2` on the MR itself, it still has to be posted.

#### What I ran and observed

1. `pnpm test:desktop` → `tests 501 / pass 501 / fail 0`, 90.7 s, exit 0.
2. `pnpm check-evidence` → exit 0, line quoted above.
3. Mutation checks in a `/tmp` copy refreshed from this head (hashes `dc96668c6e…` / `d3c3134773…` verified against the worktree) — five results listed in the MINOR-1 row above; the copy and the worktree were restored, `git status --short` empty at the end.
4. `biome check` on the two changed suites → no format diagnostics.
5. Live harness, both sizes, `ELECTRON_BIN=…/electron bash docs/evidence/window-mode/harness/run.sh . <label> headless <WxH>`:
   - 800x600 → exit 0, `this app was the frontmost application in 0 of 11 samples`, `state: visible=false focused=false focusable=false size=800x600 content=800x572`, `rAF 123/s`, `hasFocus false`, viewport 800x572, `renderer errors: 0`.
   - 1380x900 → exit 0, `0 of 10 samples`, `state: … size=1380x900 content=1380x872`, `rAF 122/s`, viewport 1380x872, `renderer errors: 0`.
   - `0 of N` is stable at zero; the denominator tracks how many samples the run afforded (11 and 10 here, 8 in round 2), so the README's `0 / 8` is one run's count, not a constant.
6. Frame comparisons and crops as described under MINOR-R3-1 (`magick compare -metric AE` + `-trim` bounding boxes, then 3x crops of the box from each side).

Worktree `/Users/damian/local-operator-ui-worktrees/window-focus` read-only throughout; all mutations and the harness scratch under `/tmp`; `git status --short` clean before and after every run.

damianvtran:

Scope: `2e73ef35b..52d948987` (reviewed head `2e73ef35b`; fixes `288de114c`, manifest re-stamp `52d948987`).

You noted that these dispositions were only in the commit message; they belong on the MR, so here they are.

| # | Finding | Disposition |
| --- | --- | --- |
| MINOR-1 | the raise guard was not recursive, so `src/main/backend/*.ts` was never scanned | **Fixed, `288de114c`.** `readdirSync("src/main", { recursive: true })`, plus an assertion that the scan reached a subdirectory so a future flat readdir cannot silently narrow it again. I reproduced your mutation: a `w.show()` in `src/main/backend/index.ts` now gives 13 pass / 1 fail naming that file. |
| MINOR-2 | AGENTS.md still claimed "pixel-identical, 0 of 14,440,320 channels" | **Fixed, `288de114c`** — AGENTS.md carries the README's numbers, and the *cause* is corrected in `0a33ab086` (see round 3). |
| NIT-1 | "0 of 5 samples" vs the README's 0 of 8 | **Fixed, `288de114c`** — 0 of 8, with the second launch named and the control runs listed as the range they are. |
| NIT-2 | `normal`'s `second-instance` sequence differs from `origin/main`, so "normal is unchanged" is not literally true there | **Fixed, `288de114c`, corrected in `0a33ab086`.** The module now states the difference and why it is deliberate; the first version of that note justified it with "the app has no hide path", which is false (see round 3). |
| NIT-3 | Biome would wrap one added assertion | **Fixed, `288de114c`** — and both changed suites are now format-clean under `npx biome check`. |

damianvtran:

Scope: `52d948987..2ece460f3` (reviewed head `52d948987`; corrections `0a33ab086`, manifest re-stamp `2ece460f3`).

| # | Finding | Disposition |
| --- | --- | --- |
| MINOR-R3-1 | the residual frame difference is the seeded transcript's clock, not "a blinking caret at the composer", and the "0 channels at 800x600" control does not reproduce | **Fixed, `0a33ab086`.** You are right on both counts and I reproduced it before rewriting: a fresh 800x600 run against the committed frame differs in 390 pixels, all inside one `62x19` box at `+1446+902`, and cropping that box out of the committed frame reads `9:58 PM`. `seed.mjs` stamps the opener with `Date.now()` and the app renders it with `hour: "numeric"`, which is why my two 800x600 runs matched byte-for-byte (same minute) and yours did not. The README, AGENTS.md and the PR body now all say: differences are confined to that box, apart from it the frames reproduce pixel-for-pixel, and the control is the wall clock rather than a caret. |
| NIT-R3-1 | "the app has no hide path" is false — `{ role: "hide" }` (Cmd+H) is in the app menu, so the added `show()` in `normal` is not a no-op | **Fixed, `0a33ab086`.** The note now states the real, deliberate behaviour: after the operator hides the window, a second launch or a banner click brings it back rather than focusing it while it stays hidden. |

damianvtran:
### Agent review — round 4

Reviewer: reviewer (deepseek/deepseek-flash)
Scope: 52d9489873389859a08f323f9e2abad644b80d90…2ece460f3b0046a70262bc251b5aa71351394879

**Verdict: clean — no blocker, no major. Both round-3 findings are closed, and I reproduced the replacement claim myself rather than reading it. One MINOR and two NITs are follow-ups in an evidence README; none of them is a reason for another round. This round is TERMINAL.**

```diff
git diff 52d948987..2ece460f3 --stat
 AGENTS.md                           |  7 ++++---
 docs/evidence/manifest.json         |  2 +-
 docs/evidence/window-mode/README.md | 21 +++++++++++++++------
 src/main/window-raise.ts            | 10 ++++++----
```

No stray edit: the only source change in the delta is the comment block in `window-raise.ts` (`git diff b60818fbc..HEAD -- src` is 8 added comment lines and nothing else, so the committed build is behaviourally current). Commits are `0a33ab086` + the separate manifest re-stamp `2ece460f3`.

#### Round-3 disposition

| # | Finding | Status |
| --- | --- | --- |
| MINOR-R3-1 | the residual was attributed to "a blinking caret at the composer", and the "0 channels at 800x600" control did not reproduce | **CLOSED, reproduced at both committed sizes.** My own runs put every differing pixel inside one box that reads a clock, and same-minute runs are byte-identical; details below. The corrected text is in all three places (AGENTS.md:145-152, README.md:65-81, PR body "Headless is not a degraded rendering path"), and no stale "caret at the composer" claim survives anywhere (`grep -rn caret` leaves only the focus-dependent-rendering caveat at AGENTS.md:176 and README.md:104, which is a different, still-true statement). |
| NIT-R3-1 | "the app has no hide path" is false | **CLOSED, and the replacement is accurate.** `{ role: "hide" }` is at `src/main/index.ts:183`; the banner-click raise is `src/main/desktop-notifier.ts:726` (`raiseWindow(target, this.windowRaise)`, default `"focus"`), the second-instance raise is `src/main/index.ts:469`. I also reproduced the premise the note rests on, on the same Electron the evidence uses: a `{ show: false }` window that has never been shown, after `restore()` + `focus()`, still reports `{"visible":false,"focused":false}` — so the pre-change sequence would have focused it while it stayed hidden, and the added `show()` is what brings it back. |

#### MINOR-R4-1 — `docs/evidence/window-mode/README.md:71-72`: the 800x600 box is `60x19`, not `62x19`

> "800x600, committed frame against a fresh run: 390 pixels, all inside one `62x19` box at the timestamp (round 3's independent run measured 649 there)."

`62x19` is the 1380x900 box (it is right in AGENTS.md:148 and README.md:73, and both are correct there). At 800x600 two independent runs — round 3's and mine — bounding-box the differing pixels at **`60x19+1448+902`**; the two-pixel width is not something the pair of clock strings can produce, since both are right-aligned against the same trailing `PM`. Fix: `60x19` in that bullet. Low severity — it changes no conclusion, but a reader who re-measures gets 60 and will wonder which of us is wrong, which is the class of error this round's correction was about.

#### NIT-R4-1 — `README.md:75-79`: the cited mechanism is `hour: "numeric"` but the minute is what makes runs differ

> "The box is not a mode difference and cannot be: `seed.mjs` stamps the transcript with `Date.now()`, the app renders that opener with `hour: "numeric"`, and two runs a minute apart therefore print different times."

The formatter is `src/renderer/src/features/chat/utils/message-grouping.ts:65-68` — `{ hour: "numeric", minute: "2-digit" }` (the only `hour:` in `src/`). `hour` alone would only move the string across an hour boundary; the sentence's own consequence ("a minute apart therefore print different times") follows from the `minute: "2-digit"` that is not cited. Cite the `TIME_FORMAT` pair.

#### NIT-R4-2 — `README.md:71-74`: the three bullets count different units

Bullet 1 counts pixels (`390`), bullets 2-3 count channels (`1,596`, `1,526`). I measured both on the same 800x600 pair: **624 differing pixels against 1,831 differing channels** (`magick compare -metric AE` vs a per-channel non-zero count of the difference composite). A ~3x factor means a reader cannot line the three numbers up. Say which unit each row is, or use one.

#### What I ran and observed

1. `git diff 52d948987..2ece460f3` — the four files above; no behavioural change, no doc contradicting another.
2. `pnpm test:desktop` → `ℹ tests 501 / ℹ pass 501 / ℹ fail 0`, exit 0 (91 s). Matches the PR body.
3. The live harness at 800x600, exactly as specified: `ELECTRON_BIN=$HOME/local-operator-ui/node_modules/.bin/electron bash docs/evidence/window-mode/harness/run.sh . r4-headless headless 800x600` → exit 0, `[window-mode] window mode headless: 800x600, window created and never shown, page throttling off`, `state: visible=false focused=false focusable=false size=800x600 content=800x572`, **frontmost in 0 of 8 samples** (pid 97724), viewport 800x572, rAF 120/s, `hasFocus=false`, renderer errors 0, and the harness's own "the app reported the mode and size it was asked for" assertion passed.
4. Frame comparisons against the committed frames (`magick compare -metric AE` for the count, `-compose difference -composite -threshold 0 -trim` for the bounding box — a zero-tolerance test, so a 1/255 channel difference anywhere would widen the box):
   - 800x600, committed vs my fresh run → **624 px, all inside `60x19+1448+902`**. Cropped, committed reads `9:58 PM`, my run `10:57 PM`.
   - 1380x900, committed vs a fresh `headless` run → **487 px, all inside `62x19+2606+1414`**. Committed `9:59 PM`, mine `10:58 PM`.
   - 1380x900, fresh `headless` against fresh `inactive` (both shown-window/non-window paths, seeded 23 s apart) → **AE 0, 0 differing channels, the two PNGs byte-identical at 254,090 B**. The claimed control is real, and it is stronger than the doc says: in the same minute the mode is not merely "confined to the box", it is invisible.
5. The clock's identity: the box sits beside the seeded transcript rows, and its text tracks the seeder's clock in every run (`22:57:13` seed → `10:57 PM`; `22:58:05` and `22:58:28` → `10:58 PM`), with `seed.mjs` computing `Math.floor(Date.now() / 1000) - 60`. The corrected cause is the right one.
6. `pnpm check-evidence` → exit 0, `Evidence holds: 665 frames are pictures of their own theme`. The re-stamp is correct at this head: manifest `srcTree` `9f42a2149` = `git rev-parse HEAD:src`, `scriptsTree` `bb2b36d54` = `HEAD:scripts`, and `head` `83db293e` is an ancestor of HEAD (reachable, so it survives a gc), which is exactly what `provenanceFailures` (`scripts/check-evidence.mjs:247`, called at `:398`) checks.
7. Probe: a 12-line Electron 35.5.1 script in `/tmp` (results under 4 above) for the hidden-window `focus()` behaviour; no window was shown by it.

Worktree `~/local-operator-ui-worktrees/window-focus` untouched: `git status --short` empty before and after; every artifact (harness scratch, probes, crops, diffs) under `/tmp`.

**TERMINAL.** The frame-delta correction is accurate and reproducible at both committed sizes, the raise note is true and its premise reproduced, and the delta carries nothing else. MINOR-R4-1 and the two NITs are documentation polish for a follow-up commit, not a review round.

damianvtran:
### Agent review remediation — round 4

Scope: `2ece460f3..2629b3a10` for these fixes (plus the two `main` syncs that landed while this branch was in review — main had moved to `0.19.1`, and both `package.json` and `docs/evidence/manifest.json` conflicted again).

| # | Finding | Disposition |
| --- | --- | --- |
| MINOR-R4-1 | `62x19` is the 1380x900 clock box; the 800x600 one is `60x19` | **Fixed, `2629b3a10`.** The README now gives the box per size: `60x19` at 800x600, `62x19` at 1380x900, each with the run's own count. |
| NIT-R4-1 | the formatter cited only `hour: "numeric"` | **Fixed, `2629b3a10`.** It now names `TIME_FORMAT` in `src/renderer/src/features/chat/utils/message-grouping.ts` — `hour: "numeric"` **with** `minute: "2-digit"` — and says the minute is what makes two runs a minute apart differ. |
| NIT-R4-2 | the bullet list mixed pixels and channels | **Fixed, `2629b3a10`.** Pixels throughout, with channels in parentheses only where the earlier measurement was recorded that way. |

Two things changed while fixing this, both strengthening the evidence rather than just correcting the prose:

- **The frames are re-captured at the merged head** (the sync brought main's renderer in), against a build of that head, and the harness was re-run three times on it: `headless` 1380x900 **0 of 13** samples frontmost, `inactive` 1380x900 **0 of 12**, `headless` 800x600 **0 of 13**, each with the app's own `state:` line as asserted by the harness.
- **The same-minute pair now reproduces byte-identically**: `headless` against `inactive` at 1380x900, seeded in the same minute, is **0 differing pixels** (2760x1744). Across minutes the difference is the clock box, at both sizes, which is what the README says.

damianvtran:
### Agent review — round 5

Reviewer: reviewer (deepseek/deepseek-flash)
Scope: 2ece460f3b0046a70262bc251b5aa71351394879…fa72c8c7053385d3c13fe8f5921c494750f1eb65

**Verdict: clean — no blocker, no major. This round is TERMINAL.** Both `main` syncs dropped nothing from either side, the manifest is stamped to this head, the three round-4 findings are fixed and reproduced, the re-captured frames are pictures of this head's renderer (I reproduced them from my own build of it), and both gates pass at this head. The one NIT at the end is a follow-up in an evidence README, not a reason for another round.

#### The two syncs dropped nothing, from either side

Rather than reading the resolution, I walked every path each parent changed relative to the merge base and compared `git rev-parse <side>:<path>` against `HEAD:<path>`:

- First sync `42810c275` (base `54bf411e0`): differs from `79d0dc889` only in `package.json` and `docs/evidence/manifest.json`; differs from `2ece460f3` only in those two plus the three window-mode evidence files the re-capture `2629b3a10` deliberately rewrote. Nothing else moved on either side.
- Second sync `d735f27e8` (base `79d0dc889`): differs from `f3eb589c1` only in `package.json` and `docs/evidence/manifest.json`; differs from `2629b3a10` only in those two. `git diff 2629b3a10..HEAD -- src/renderer docs/evidence/chat-title` is **empty**, so the second sync brought no renderer change, and `git diff 2629b3a10..HEAD -- docs/evidence/window-mode/frames` is empty, so nothing about it required a re-capture.
- `git diff 2ece460f3..2629b3a10 -- src/main/` is empty, so the first sync resolved nothing away under `src/main/`. At this head `src/main/` differs from `2ece460f3` only by main's own second-sync files (`backend-installer.ts`, `backend-service.ts`, the three install scripts, `python-bytecode-cache.ts`, `update-install.ts`, `update-service.ts`), each byte-identical to `f3eb589c1`'s copy as the audit above shows — i.e. brought in, not resolved away.

`package.json`, proven by subtraction instead of by reading the line: at this head, `head-minus-additions == f3eb589c1:package.json` exactly, where the additions are `scripts/window-mode.test.mjs` in `test:desktop` and `app:headless` / `dev:headless`. All 25 of main's entries are present — `chat-title.test.mjs` and `python-bytecode-cache.test.mjs` included — with `window-mode.test.mjs` added and nothing else; the only top-level key differing from main is `scripts`, and both sides read version `0.19.1`.

Manifest: `srcTree` `f3af1a1bd25cf14177b43cddead4133dce9269de` = `HEAD:src` and `scriptsTree` `6c78381b0149faf3df8d6ec7d49d320cec4b156b` = `HEAD:scripts` after the re-stamp; `head` `83db293e…` is a reachable ancestor of this head, which is all `check-evidence.mjs` asks of it.

#### Round-4 findings: fixed, and the replacement claim reproduced

`2629b3a10` gives the 800x600 box as `60x19` and the 1380x900 one as `62x19`, cites `TIME_FORMAT` in `src/renderer/src/features/chat/utils/message-grouping.ts:65-68` — `hour: "numeric"` **with** `minute: "2-digit"`, which is the chat path's only `hour:` — and counts pixels throughout with channels in parentheses. I then reproduced the claim itself rather than reading it:

- Two runs seeded in the same minute (parallel, own ports/scratch, `headless` against a real shown `inactive` window, both 1380x900): **AE 0 — byte-identical, 253,508 B each.**
- Across minutes the difference is one box and nothing else: **634 px in `58x19+2610+1414`** at 1380x900, **625 px in `57x19+1451+902`** at 800x600, and the crops read `11:19 PM` against `11:34 PM` (1380) and `11:20` against `11:41` (800). Both halves of the corrected sentence hold.

#### The committed frames are of this head

`md5` of `frames/headless-1380x900-chat.png` is `e90f55aff3c5cc6557550a807b908a7e`, byte-equal to the re-capture's `final-headless` **and** `final-inactive` outputs — the same-minute pair, so the committed frame cannot be told from either and is not a carried-over pre-merge capture. `frames/headless-800x600-chat.png` is `9d0cb372b532830a517ad08f4c8dd70e`, byte-equal to `final-800`. Independently, builds of the head's own source (a build of the pre-second-sync head `2629b3a10` and a build of this head `fa72c8c70`, both in `/tmp` with `LOCAL_OPERATOR_UI_NO_BYTECODE=true`) reproduce both committed frames so that the only differing pixels are the clock box above, which is what "these frames depict this head's renderer" means.

#### NIT-R5-1 — `docs/evidence/window-mode/README.md:70-77`: the clock box's **width** is a property of the compared pair, not a constant

The list gives `60x19` (800x600) and `62x19` (1380x900) as if the box were a layout size. Measured on my pairs it is `57x19+1451+902` and `58x19+2610+1414`: the height (19) and the location are stable, while the width and the x-offset follow which digits of the two minute strings differ. One clause — "the width follows the two minute strings compared; only the height and the location are fixed" — removes the trap. Follow-up only.

#### What I ran and observed

1. `pnpm test:desktop` at `fa72c8c70` → `ℹ tests 527 / ℹ pass 527 / ℹ fail 0 / skipped 0`, 90.7 s, exit 0 (the window-mode suite's own assertions included, e.g. `no file but window-raise.ts raises or focuses a window`).
2. `pnpm check-evidence` at `fa72c8c70` → exit 0, `Evidence holds: 681 frames are pictures of their own theme (worst ΔE00 18.1, docs/evidence/chat-usage/dense/localOperatorLight.webp)`.
3. Merge audits: `git diff --name-only $(git merge-base <branch> <main>)..<side>` vs `git rev-parse <side>:<path>`/`HEAD:<path>` per path, for both syncs (results above); `git diff 2629b3a10..HEAD -- src/renderer docs/evidence/chat-title` (empty).
4. Harness against my own build of this head (`ELECTRON_BIN=$HOME/local-operator-ui/node_modules/.bin/electron bash docs/evidence/window-mode/harness/run.sh /tmp/wm5/tree2 <label> <mode> <size>`), each run exiting 0 with the harness's own `== the app reported the mode and size it was asked for` and 0 renderer errors:
   - `headless` 1380x900 → `[window-mode] state: visible=false focused=false focusable=false size=1380x900 content=1380x872`, **0 of 12** samples frontmost, `hasFocus=false`, rAF 122/s, DPR 2.
   - `inactive` 1380x900 → `state: visible=true focused=false focusable=true size=1380x900 content=1380x872`, **0 of 12** frontmost.
   - `headless` 800x600 → `state: visible=false focused=false focusable=false size=800x600 content=800x572`, **0 of 14** frontmost, viewport 800x572, cssPixels `800x572`, 12 chat rows.
5. Frame comparisons: `magick compare -metric AE` for the count, `magick A B -compose difference -composite -threshold 0 -format '%@'` for the box — zero tolerance, so a 1/255 difference anywhere would have widened the box; the numbers are in the section above.

Process: `### Agent review remediation — round 4` is on the PR (03:41Z) and answers MINOR-R4-1 and both NITs by ID, so this round's findings are answered as the gate requires.

The review worktree was left untouched (`git status --short` empty); every build, harness run, crop and diff lives under `/tmp`.

**TERMINAL.** The remaining item is NIT-R5-1, documentation polish for a follow-up commit.

damianvtran:
### Agent review remediation — round 5

Scope: `2ece460f3..fa72c8c7053385d3c13fe8f5921c494750f1eb65` (round-5 head; clean verdict, no blocker, no major).

Round 5 found one NIT and I am **not** fixing it in this change, deliberately:

| # | Finding | Disposition |
| --- | --- | --- |
| NIT-R5-1 | the README quotes the clock box as `60x19` / `62x19` as if the width were constant; measured widths track which digits of the two compared minute strings differ (57-62), while the height and location are stable | **Deferred, with the reason stated.** It is a documentation polish on a measured range, and fixing it would move a head whose review is fresh and terminal — the gate is explicit that a nit never creates another round, and I would rather merge the reviewed head than a nicer sentence about a clock. It is a one-line change for whoever next touches this README: quote the width as the range it is. |

Everything else the round asked for is on the PR and in the head: the two `main` syncs resolved with nothing dropped from either side (`package.json` proven by subtraction against `f3eb589c1`), the manifest re-stamped to this head, the frames byte-equal to this head's re-capture, `pnpm test:desktop` 527/527 and `pnpm check-evidence` holding at 681 frames, and four harness runs on this head reporting 0 of 12, 0 of 12 and 0 of 14 samples frontmost.

PR #107, opened 2026-09-12
fix(update): make the installer and the update path refuse what they cannot do safely, and say so

damianvtran:
## What and why

Two operator reports, one shared root: nothing on the update path checked the
state of the thing it was about to replace, and nothing told the user when the
replacement failed.

**Symptom A — "damaged and can't be opened".** The 0.17.0 release signed and
notarized the `.app` and shipped it inside an **unsigned, unnotarized** DMG:
`package.json` had `dmg.sign: false` and `mac.notarize: false`, so only
`afterSign` ran, and it only submits the app. Measured on the operator's own
download: `spctl -a -vvv -t open --context context:primary-signature
local-operator-ui-0.17.0-universal.dmg` → `rejected / source=no usable
signature`, while `codesign --verify --deep --strict` and `spctl -t exec` on the
app inside it both passed. The pipeline asserted the app and never the container,
which is why 0.17.0 shipped.

**Symptom B — "it downloads, quits, and never reopens".** Squirrel.Mac's ShipIt
refused the *installed* bundle with `NSOSStatusErrorDomain Code=-67028` from
`SecStaticCodeCreateWithPath` — `errSecCSBadBundleFormat`; the probe in the
evidence maps that code to a bundle that exists but seals badly (a plain
directory, or one missing `Contents/MacOS/...`), and a Finder copy of the app
from the mounted DMG was in flight at that moment. ShipIt quit, never ran
`launchAfterInstallation`, relaunched nothing and told the user nothing. The app
was left gone from `/Applications`, with `pending/` and `ShipItState.plist`
behind it, and re-offered the same update on the next start. Separately, the app
offered `pip install --upgrade local-operator` with `canManageUpdate: true` for
a server that was a uv tool install (`0.54.17` read from `/health`) — the wrong
installer for an environment the app does not own.

**Change class: C2 — standard, pre-authorized.** User-visible copy and states, so
it carries a design round in addition to code review and QA.

### 1. Distribution: sign, notarize and staple the image, and assert it

- `dmg.sign: true`. The option defaults to **false** in electron-builder 26, so
  removing the explicit `false` would have changed nothing — this is a deviation
  from "remove/default true" in the report, with the default's behaviour
  verified in `node_modules/dmg-builder/out/dmg.js` (`if (this.options.sign ===
  true)`).
- `mac.notarize` stays `false`, deliberately: electron-builder's built-in path
  reads `APPLE_APP_SPECIFIC_PASSWORD`, while this project's secrets carry
[...]

PR #124, opened 2026-09-13
fix(release): hold a release out of latest until its assets exist

damianvtran:
## The bug

`electron-updater` resolves its feed from GitHub's `/releases/latest`, which answers with the newest
**non-prerelease** release whether or not that release has assets. This pipeline is triggered *by* the
release being published, so the release exists and is advertised for the whole 25-35 minutes its
installers are being built. During that window the metadata 404s and every running app filters the
404 into "no updates available" — telling users they are up to date while a newer version is published.

Measured live during the v0.19.2 build (as reported on the task):

```
GET /repos/damianvtran/local-operator-ui/releases/latest
  -> tag v0.19.2, assets: 0, prerelease: false
GET .../releases/download/v0.19.2/latest-mac.yml -> 404    (v0.19.1 -> 302)
```

and the running app's own log:

```
Filtering error: Missing latest*.yml file
Error filtering result in IPC handler: Reporting as no updates available
```

Re-measured just now, which also shows the window closing on its own once the build finished:

```
$ gh api repos/damianvtran/local-operator-ui/releases/latest \
    --jq '{tag: .tag_name, assets: (.assets|length), prerelease: .prerelease}'
{"assets":11,"prerelease":false,"tag":"v0.19.2"}    # build complete, metadata now 200
$ curl -o /dev/null -w '%{http_code}' .../releases/download/v0.19.2/latest-mac.yml   -> 200
$ curl -o /dev/null -w '%{http_code}' .../releases/download/v0.17.2/latest-mac.yml   -> 404

$ gh api "repos/.../releases?per_page=12" --jq '.[] | [.tag_name, .prerelease, (.assets|length)]'
v0.19.2  false  11
v0.19.1  false  11
...
v0.17.2  true    0     <-- the hand-marked pre-release workaround this change automates
```

So: every complete release carries exactly 11 assets; v0.17.2 is the one asset-less release and the one
that had to be marked a pre-release by hand for exactly this reason. Same class as #107's silent
error-filtering, reached by a different route.

## The change

A release is now a **pre-release while it is built** and becomes a full release once the artifacts are
attached. A pre-release cannot be `latest`, so `/releases/latest` keeps answering the previous,
asset-complete release until this one can answer for itself.

The invariant both directions enforce: **a release is advertised as latest exactly while it is
asset-complete.** That also fixes the failure mode where a *failed* build left an empty release at the
head of `latest` indefinitely, suppressing every update rather than only the one it failed to produce.

- `scripts/release-state.mjs` (new) — `open` holds an incomplete release out of `latest`; `finalize`
  promotes one whose installers and `latest*.yml` are present, and refuses otherwise.
- `open-release-window` (new job, before every build) and `finalize-release` (new job, after
  `attach-to-release`) run it. Both are pinned to `validate-release`'s `release_id` and `source_sha`.
- The three build jobs now need `open-release-window`, so a release that **cannot** be held stops the
  run in seconds instead of after 40 minutes of macOS runner time spent telling users they are current.
- `upload-release.mjs` exports `PLATFORM_INSTALLERS` and `listAssets` so the upload gate and the promote
  gate assert the same definition of "complete" and page the same ID-addressed asset list.
- `AGENTS.md` runbook step 9 now creates the release with `--prerelease`, and documents how the window
  closes on a normal run, on a re-run, and after a `workflow_dispatch` repair.
- `package.json` is untouched: no version bump.

## Why the flip is not in `attach-to-release`, and where each mutation is scoped

`attach-to-release` is deliberately non-mutating — its comment says a repair must not create releases,
update metadata or replace assets, and it POSTs only to the pinned ID. Putting a PATCH there would make
the repair path mutate release metadata, which is precisely what that contract forbids. So the flip
lives in jobs of its own, and `attach-to-release` keeps its single effect (the asset upload).

**Event scoping.** `release: published` fires for pre-releases as well as releases (per GitHub's
trigger docs, `published` is the type to subscribe to if you want stable *and* pre-release
publications), so a normal run of this workflow is reachable with the release as a pre-release — that
is the whole mechanism, and the harness already covers `release prerelease=true`.

Both mutations refuse whenever `IS_MANUAL_DISPATCH` is set. A `workflow_dispatch` repair of an old tag
therefore attaches assets and touches no metadata; it can never promote a release to `latest`, which
would re-point the feed at a release nobody validated for that purpose. Deliberately **not** scoped by
`if: github.event_name == 'release'` on the job: every build needs the hold, so a skipped hold would
either block a repair (its builds would skip) or leave a build running behind a job that never ran.
Scoping inside the script keeps the repair path complete and only the mutation stands down, and the
refusal is directly tested in both directions.

Extra guards, each with a test:

- A release that is **already complete** is never taken out of `latest`, so a duplicate or re-run
  `published` event cannot suppress a working update offer.
- Both mutations are pinned twice — the release ID that `validate-release` read from this event's own
  tag, and the source SHA that tag must resolve to — so a tag that moved or a release that was deleted
  and recreated since validation is caught, not edited.
- An asset in any state other than `uploaded` counts as missing: an interrupted upload is listed on the
  release but has no bytes behind it, and fetching it fails exactly like the 404 this change prevents.
- A tag with a pre-release suffix (`v1.2.3-rc.1`) is never promoted, even when complete.
- The flip does not re-trigger this workflow: it subscribes to `release: published` only, while a
  pre-release-flag edit surfaces as `edited` (and a pre-release converted to a release as `released`).

The PATCH sends `-F prerelease=… -F make_latest=…` (typed booleans; `-f` would serialize the strings
`"true"`/`"false"`), and `make_latest` is set explicitly in both directions rather than left to the
endpoint's default, which the docs state for *newly published* releases.

## Test evidence

Extended the existing harnesses rather than adding a parallel one.

```
$ pnpm lint                 -> exit 0 (24 warnings, no errors)
$ pnpm check-types          -> exit 0 (tsconfig.node.json + tsconfig.app.json)
$ pnpm test:desktop         -> # tests 527  # pass 527  # fail 0
$ node scripts/test-validate-release.mjs        -> All validation tests passed (23/23)
$ node --test scripts/test-release-safety.mjs scripts/test-publish-workflow.mjs
                            -> # tests 161  # pass 161  # fail 0
$ actionlint .github/workflows/publish.yml      -> exit 0
```

The 161 covers the workflow's own shell gates (unchanged: signing preflight, the npm registry gate, the
`.jsc` publish guard, pnpm/builder argument forwarding) plus the new coverage below.

Unit-level assertions are backed by driving the **real CLI** with a scripted `gh` (the only `gh` on its
PATH) over a fixture release, which is what the workflow jobs actually run:

```
$ RELEASE_TAG=v0.19.3 EXPECTED_SOURCE_SHA=<sha> EXPECTED_RELEASE_ID=999001 \
  IS_MANUAL_DISPATCH=false node scripts/release-state.mjs open        # assets: []
Release v0.19.3 (id 999001) state: published, prerelease: false
Release v0.19.3 (id 999001) held as a pre-release until it is asset-complete; missing:
macos installer, windows installer, linux installer, latest*.yml update metadata.
  PATCH .../releases/999001  -F prerelease=true  -F make_latest=false

$ ... IS_MANUAL_DISPATCH=false node scripts/release-state.mjs finalize   # pre-release + 10 assets
Release v0.19.3 (id 999001) promoted to a full release; it now answers /releases/latest.
  PATCH .../releases/999001  -F prerelease=false  -F make_latest=true

$ ... IS_MANUAL_DISPATCH=true  node scripts/release-state.mjs finalize   # repair
Manual dispatch: release state is not mutated by a repair; leaving the release as it is.
  (no PATCH)

$ ... RELEASE_TAG=v0.19.3-rc.1 node scripts/release-state.mjs finalize   # pre-release tag
Tag v0.19.3-rc.1 is a pre-release by name; it is not promoted to latest.
  (no PATCH)

$ ... ASSETS=<complete minus latest*.yml> node scripts/release-state.mjs finalize
Refusing to promote v0.19.3: missing latest*.yml update metadata
  exit 1, (no PATCH)

$ ... ASSETS=<app.dmg state="open", no zip> node scripts/release-state.mjs finalize
Refusing to promote v0.19.3: missing macos installer
  exit 1, (no PATCH)
```

Also exercised through the CLI: a duplicate event against an already-complete release logs "already
carries every installer and its update metadata; not holding it back" and issues no PATCH, and a
released-as-pre-release release logs "already a pre-release; still building". `actionlint` on the
modified workflow exits 0 (checked against an injected bad `needs`, which it does catch).

Behavioural (`test-release-safety.mjs`, fixture API + injected flag writer, no network):

- held: asset-less release; release missing `latest*.yml`; release with an interrupted upload;
- held-and-left-alone: already a pre-release; already asset-complete (the re-fire guard);
- promoted: complete pre-release; skipped: already a full release;
- a manual repair writes nothing, in both directions, in the states where a mutation would otherwise
  happen;
- `v1.2.3-rc.1` never promoted, even complete;
- promotion refused with no write when an installer or the metadata is missing;
- recreated release / moved tag / missing pin all rejected with no write;
- the real PATCH argv, driven through a stub `gh` in a child process whose PATH holds no real CLI:
  `repos/…/releases/<pinned ID>`, never a tag-addressed path, with typed boolean fields; a failing
  PATCH raises without carrying the subprocess's stderr.

Wiring (`test-publish-workflow.mjs`): both jobs run the right mode with both pins and the workflow SHA;
their sparse checkout includes **every** module `release-state.mjs` imports, derived from the script
itself (the list is hand-kept, and a missing module fails the job after the release is public); they
inherit `contents: write`; a release that cannot be held skips every build, the attach and the promote;
the window opens before every build and closes only after attach; the scoping is the event variable,
not a job a repair would skip; every job is declared after the jobs it needs (the harness resolves
needs positionally); `open-release-window` joins the failure matrix that blocks upload.

Each new guard was checked by mutation — disabling the hold, the promote, the dispatch refusal in
either direction, the asset gate, the release-ID pin, the tag check, the checkout entry, or the build
dependency makes the corresponding test fail (10 mutations, 0 missed).

## What the first real release must confirm

The real pipeline cannot be exercised locally, and no release was mutated while developing this: the
PATCH path is tested against a stubbed `gh`, not against the live API. On the next release (v0.19.3)
please confirm, in this order:

1. while the build runs — `gh api repos/damianvtran/local-operator-ui/releases/latest --jq .tag_name`
   answers **v0.19.2**, and the new release shows as a pre-release;
2. after `finalize-release` goes green — the same command answers **v0.19.3**, its asset count is 11,
   and `.../releases/download/v0.19.3/latest-mac.yml` returns 200;
3. nothing else moved: the npm publish ordering, `validate-release`'s pins, the artifact gate and the
   `.jsc` guard all report exactly as before.

Residual unknowns, stated rather than assumed:

- That GitHub accepts a `prerelease`/`make_latest` flip on an already-published release, and that
  `latest` reverts to the previous release while the window is open. Both are documented API behaviour
  (a draft or pre-release cannot be set as latest) but neither was performed against a real release
  here — no release was mutated while developing this.
- That the update-a-release endpoint's workflow-scope note does not bite a flag-only edit. That note
  applies when the resolved target commit adds or modifies a file under `.github/workflows/` relative to
  the default branch, and an Actions `GITHUB_TOKEN` cannot be authorized for workflows. This PATCH
  changes no target and normally runs against a tag whose tree matches the default branch, and edits of
  this shape are common in Actions — but it is the one documented way this PATCH could 404, so it is
  named here rather than assumed away. If it ever happens, the failure is loud and safe: the release
  stays a pre-release (unoffered) and the run's own error names the `gh release edit` line that closes
  the window.

A failure is loud and safe in both directions: a failed hold stops the build before it starts, and a
failed promote leaves the release a pre-release (unoffered) with a red run, fixable with
`gh release edit v0.19.3 --prerelease=false`.

## Interim guidance for anyone updating during a build window

Until this ships, `/releases/latest` still points at the just-published, asset-less release for the
length of the build, so an update check in that window reports "no updates available". If a user hits
it: wait for the assets (25-35 min) and check again, or install the previous release by hand. The
existing workaround stays available for a release owner: mark the release a pre-release by hand while it
builds, as v0.17.2 was.

## Not addressed

- The app-side error filtering (the #107 class) is untouched on purpose: this removes the condition
  that produced the 404 rather than the code that masks it.
- A publish of a *newer* release while an older one is still building would have its own window: both
  flips are scoped to the event's own release ID, and the older release's promote cannot touch the
  newer release's metadata.


damianvtran:
### Agent review - round 1

Reviewer: deepseek/deepseek-flash (single model for the whole round; no mid-round provider fallback)
Scope: origin/main..39a9d28d9

**Method.** Diff first, then execution. There is no PR-branch CI (ci.yml triggers only on push to main/dev-*), but `.github/workflows/ci.yml:66` runs these two suites, so I ran them on the branch and on a pristine `git archive origin/main` tree:
- branch: `node --test scripts/test-release-safety.mjs scripts/test-publish-workflow.mjs` -> 57 + 104 pass.
- origin/main: 15 failures in the workflow suite (every new window assertion), and the safety suite cannot load (`ERR_MODULE_NOT_FOUND: scripts/release-state.mjs`). Discrimination confirmed, with the caveat that the safety suite's failure is a module-not-found rather than a semantic assertion.
- mechanism, read-only live probes: `curl -sSL -H 'Accept: application/json' https://github.com/damianvtran/local-operator-ui/releases/latest` -> JSON `{"tag_name":"v0.19.2",...}`. That is exactly the request `electron-updater@6.6.2` makes in `GitHubProvider.getLatestTagName` (node_modules/electron-updater/out/providers/GitHubProvider.js:146-163) and then fetches `.../download/<tag>/latest-mac.yml`; `src/main/update-service.ts` never sets `allowPrerelease`, so the `/releases/latest` branch is the one the app takes. Re-ran the suites against 4 hand mutations in a throwaway copy: removing the dispatch refusal and emptying the asset gate are both caught; the CLI authorization wiring is not (R4).

**1. Mechanism.** In the right place, and it addresses resolution rather than the log line. The provider derives the tag from `/releases/latest` and then fetches that tag's channel file; a pre-release is excluded from `/releases/latest`, so holding the release out makes the feed answer with the previous, asset-complete release. On the path step 9 now documents the hold happens at creation (`AGENTS.md:279`, `--prerelease`), and `open` only covers a release published as a full release. Direct answer to the question asked: while the release is a pre-release, `/releases/latest` still resolves to the previous asset-complete release and electron-updater offers that version if it is newer than the running one, so users get a working offer rather than silence. No re-trigger loop: the flip surfaces as `edited`/`released`, neither of which is in `on:`.

**2. Non-mutating repair.** The constraint holds as written, and only via the script. Both new jobs are reachable on `workflow_dispatch` (their `if:` reads only `needs.validate-release.result`) and their *mutation* stands down inside the script: `IS_MANUAL_DISPATCH` is read fail-closed (`!== "false"`, release-state.mjs:293), `openReleaseWindow` (:179) and `finalizeRelease` (:226) return before any PATCH, and `setFlag` is the only call site of `setReleaseState`. So the new metadata mutation is scoped to the **release event**, pinned twice to this event's own release ID and tag SHA, and lives in its own jobs — `attach-to-release`'s "POST only to the pinned ID" comment is left intact and the new comment on it is accurate. A dispatch repair of an arbitrary old tag cannot promote it.

**3. Failure and ordering.** No path I can find promotes before the assets exist: `finalize-release` needs `attach-to-release` success, and the script re-reads the ID-addressed asset list and refuses if any platform installer or `latest*.yml` is missing (state other than `uploaded` counted as absent — good). No path can leave a release permanently *promoted while empty*. A failed build or failed promote leaves the release a pre-release (unoffered, red run, re-runnable, with the manual line named in the error). R1 and R2 below are the two ordering/failure cases that are not covered.

**4.** Covered in 1 — the pre-release window is harmless for users; the previous complete release answers.

**5. Tests.** They assert the invariants, not happy paths: both directions refuse on missing installer, interrupted upload, missing metadata; the dispatch refusal is asserted in both modes; the pins are asserted (`open`/`finalize` reject a recreated release, a moved tag, a missing SHA/ID pin, writing nothing); the PATCH is asserted to be ID-addressed with typed booleans and to leak no subprocess output; the workflow suite asserts ordering, the failure matrix, and that the sparse checkout covers every module the script imports. Gaps are R3, R4 and R8.

**6. Scope.** AGENTS.md's insertion is accurate and minimal (one step change + one block), with one wording nit (R6). No version bump: `package.json` is untouched and identical to origin/main at the released 0.19.2 (`package.json` is not in the diff's file list), so this PR correctly carries none. The PR body matches the diff, states its residual unknowns honestly, and lists what is not addressed; the one behaviour it does not state is R1.

---

#### Findings

**R1 — MAJOR (blocker: no) — `finalize` promotes with no "is this the newest release" check, so re-running an older release's workflow moves `/releases/latest` backwards.** `scripts/release-state.mjs:238-258` (the promote is :254), `.github/workflows/publish.yml:474-499`.
`gh run rerun` on a past `release`-event run replays the original payload, so `validate-release` re-validates that old tag and its release ID, `open` no-ops (already a pre-release), the builds and the attach upload the **old** tag's assets, and `finalize` then writes `prerelease=false, make_latest=true` on the old release. `STABLE_TAG_RE` passes for any `vX.Y.Z` and nothing between the event and the PATCH consults recency or version order. Concretely available today: `v0.17.2` is the repo's one asset-less pre-release; re-running its publish run attaches 11 assets and promotes it, after which `/releases/latest` answers `v0.17.2` while `v0.19.2` exists — users not yet on the newest release silently stop being offered it, which is this PR's own failure class inverted. The body's "Not addressed" bullet covers a *newer* release being published while an older one builds (the ID pin does handle that); this is a re-run of an older event, and `AGENTS.md:292` presents `gh run rerun <run-id>` as a supported action, so the reachable path is a documented one. Suggested fix, one line: before the PATCH, require the release to be the newest published non-draft release (compare the tag's semver, or `created_at`, against `GET /releases`), or scope promotion to runs whose release is the newest — a re-run of the release just published still closes its window.

**R2 — MINOR — a failed hold aborts every build instead of letting the feed heal.** `.github/workflows/publish.yml:222`, `:335`, `:388`.
`open` runs `validateRelease` + `listAssets` before it can decide, so any read failure or a failed PATCH skips all three builds, the attach and the promote. On the documented path that costs only time and a re-run. But in exactly the case `open` exists to catch — a release published as a full release — a failed PATCH now leaves the asset-less release at the head of `/releases/latest` **indefinitely**, where before this change the build would have attached assets ~35 minutes later and healed the feed. Loud (red run) and re-runnable, so not ship-stopping; if you want the safe direction, distinguish "cannot verify the release" (fail, as now) from "the release could not be held" (retry, and fail at the end with the assets attached) so user-visible state can still converge without a human.

**R3 — MINOR — the completeness gate is per-platform for installers but any-platform for metadata.** `scripts/release-state.mjs:85-91`.
Installers are required per platform; `latest*.yml` needs only one match. That is asymmetric with the reason `PLATFORM_INSTALLERS` was hoisted (`scripts/upload-release.mjs:19-25`: "a platform added to the build matrix cannot satisfy one check and silently miss the other"). `actions/upload-artifact@v4`'s `if-no-files-found: error` fails only when *no* file matches the list, so a Windows job that emitted `*.exe` but no `latest.yml` passes the upload, passes `artifactFiles`, and is promoted on `latest-mac.yml` alone — Windows users then get the 404-then-"no updates available" this PR removes. Low reachability today (electron-builder writes the channel file with the installer), but the fix is the same shape as the installer check: assert `latest.yml` / `latest-mac.yml` / `latest-linux.yml` presence per platform.

**R4 — MINOR — the CLI's fail-closed authorization is the one guard with no test.** `scripts/release-state.mjs:293`.
`isManual: process.env.IS_MANUAL_DISPATCH !== "false"` is what makes a repair non-mutating, and I verified it is uncovered: replacing it with `isManual: false` leaves all 57 + 104 tests green (executed). The `if (isManual)` branches inside both functions are covered, and the workflow's env expression is asserted verbatim, but the default-deny joining them is not. The stub-`gh` harness already in the PATCH test would cover it: run the CLI with `IS_MANUAL_DISPATCH` unset and assert no PATCH.

**R5 — MINOR — the documented manual close does not assert `latest`.** `AGENTS.md:296`, `scripts/release-state.mjs:135`.
The hold writes `make_latest=false`; the automated promote deliberately sends `make_latest=true` "rather than leaving the outcome of an edit to an already-published release to be confirmed on the first real run" (:100-106). The hand-recovery in AGENTS.md and in the PATCH-failure message uses `gh release edit <tag> --prerelease=false`, and `gh release edit` only sends `make_latest` when `--latest` is passed — so the recovery relies on exactly the endpoint default the code declined to rely on. If GitHub persists `make_latest=false`, a hand-closed release ends up asset-complete but unadvertised: silently no offer, the failure this PR fixes. I could not verify GitHub's persistence semantics (no release was mutated), and the change is free: add `--latest` to both strings.

**R6 — NIT — `AGENTS.md:289` attributes the hold to the workflow.** On the path step 9 now documents, the hold is `--prerelease` at creation and the job is a no-op; say the workflow closes the window *if* the release was published without it.

**R7 — NIT — dead output.** `validate-release`'s `prerelease` output (`.github/workflows/publish.yml:38`) is read by no job now that the window script re-derives state itself; drop it or use it.

**R8 — NIT — the sparse-checkout guard is not transitive.** `localImports()` in `scripts/test-publish-workflow.mjs` walks only `release-state.mjs`'s direct `from "./..."` imports, so a module imported *by* one of those would be missing from the checkout and fail the job after the release is public. It is complete today (release-state -> upload-release -> validate-release); make the helper recursive to keep the guard's stated purpose.

---

**Verdict: one MAJOR finding remains (R1); no BLOCKER remains.** The round is therefore **not terminal**: R1 is a missing case in the promote direction that a documented operator action reaches, and R2-R5 are worth folding into the same remediation commit. R6-R8 are nits. Everything else I checked — the mechanism against the real resolver, the dispatch refusal, the "repair must not mutate metadata" constraint, the double asset check before promotion, the pin and event scoping of both new mutations, the checkout/pins/failure-matrix assertions, absence of a version bump — holds.

**What I could not verify.**
- The real pipeline end to end: nothing was merged or released, so `open`/`finalize` have never run against GitHub. The PATCH is proven only against a stubbed `gh`.
- Live GitHub semantics of the `prerelease`/`make_latest` flip on an already-published release, and whether `latest` reverts to the previous release — no release was mutated. The exclusion of pre-releases from `/releases/latest` is corroborated by this repo's own history (`v0.17.2` is the asset-less release and the one hand-marked pre-release) and by the docs' definition of the endpoint as "the most recent non-prerelease, non-draft release".
- The workflow-scope 404 caveat the PR body itself raises (a PATCH whose resolved target modifies `.github/workflows/`) — not reproducible without a live mutation.
- `IS_MANUAL_DISPATCH` is rendered by Actions as the literal string `false`; the workflow test asserts the expression text, not a rendered run.
- Visual/manual app behaviour: the "users get a working offer during the window" claim follows from the provider's code path and the endpoint's semantics; I did not run a released app against a held release.

damianvtran:
### QA report — round 1

**Reviewer:** qa-tester (deepseek/deepseek-flash) — independent pass, no edits to the repo
**Scope:** `39a9d28d93eeaa07991a28a28d7b44e211a89719` (PR #124 head), full diff vs `origin/main` (6 files)
**Matrix:** `~/workspace/qa-124/MATRIX.md` (surfaces S1–S8, ACTUAL output per row) · raw runs in `~/workspace/qa-124/`

**Verdict: PASS on every surface that can run locally. 1 BLOCKED (Q-1) — the live release-metadata PATCH, which cannot be exercised without mutating public release state.** No FAIL.

Environment: `~/local-operator-ui-worktrees/fix-release-feed`, node v26.5.0, pnpm 10.30.3, gh 2.100.0, darwin arm64. This repo has **no CI on the PR branch** (`ci.yml` runs on `main`/`dev-*` only), so every row below is a local run of the repo's own scripts. The real publish pipeline cannot run locally.

---

#### What was driven, and what it proved

**S1 — the repo's own workflow harnesses**
`node --test scripts/test-release-safety.mjs scripts/test-publish-workflow.mjs` (the exact `ci.yml` command): **161 tests, 161 pass, 0 fail**. The result is not the evidence — I read the assertions first. 12 new `open`/`finalize` fixtures cover hold / idempotent hold / never-hide-a-complete-release / promote / idempotent promote / never-promote-a-repair / never-promote-a-pre-release-tag, plus `writes == []` on every refusal and on all four pin rejections in both modes. 5 new workflow tests parse the real YAML and assert: both window jobs run the right mode with both pins; their sparse checkout contains **every module `release-state.mjs` imports** (derived from the script, so a future import cannot silently miss the checkout); a failed hold skips all three builds *plus* attach and finalize; the window opens before every build and closes only after attach; scoping is the script's `IS_MANUAL_DISPATCH` and never a job `if:` a repair would skip; a failed promotion is terminal.

**S2** `node scripts/test-validate-release.mjs` → `23 passed, 0 failed`.

**S3 — `release-state.mjs` as a CLI, 28 scenarios, 0 unexpected.** I did not call the exported functions: I ran `node scripts/release-state.mjs <open|finalize>` as a subprocess with a fixture `gh` as the only `gh` on `PATH`, which logs every argv and applies PATCHes to fixture state. So the exit codes, the printed lines and the **exact `gh` argv** are real output; the live repo is never touched. Highlights:

| | Surface | Command | Exit | ACTUAL | Verdict |
|---|---|---|---|---|---|
| A1 | reproduction (pre-change) | feed model | — | `latest=v0.99.0 → no updates available` | PASS |
| A2 | hold a published release with no assets | `release-state.mjs open` | 0 | `held as a pre-release until it is asset-complete; missing: macos installer, windows installer, linux installer, latest*.yml update metadata` → feed falls back to `v0.19.2 → update offered` | PASS |
| A4 | complete release | `open` | 0 | `already carries every installer and its update metadata; not holding it back.` (no PATCH) | PASS |
| A5 | promote once attached | `finalize` | 0 | `promoted to a full release; it now answers /releases/latest.` | PASS |
| A7/A8/A9 | partial / interrupted (`state: open`) / zero assets | `finalize` | 1 | `Refusing to promote v0.99.0: missing windows installer, linux installer` — **no PATCH**, release stays a pre-release, previous release keeps answering | PASS |
| A10/A11 | **dispatch repair of an old tag** (ticket constraint) | `open` + `finalize`, `IS_MANUAL_DISPATCH=true` | 0 | `Manual dispatch: release state is not mutated by a repair; leaving the release as it is.` — **zero `gh` calls of any kind**, so it cannot promote an unrelated release or edit metadata it should not | PASS |
| A12–A15 | scoping is fail-closed | `finalize` | 0 | unset, `true`, `False`, `0` **all** refuse; only the literal string `false` authorizes a mutation | PASS |
| A16 | pre-release-suffix tag | `finalize` | 0 | `Tag v0.99.0-rc.1 is a pre-release by name; it is not promoted to latest.` | PASS |
| A17 | promote PATCH fails (403) | `finalize` | 1 | `Release state PATCH failed …; close the window by hand with: gh release edit v0.99.0 --prerelease=false` — release stays a pre-release, no token in either stream | PASS |
| A19–A28 | pins / validation via CLI | `open` | 1 | recreated release → `Release ID … does not match`; moved tag → `Tag SHA … does not match`; missing SHA/ID/token/`GITHUB_REPOSITORY`, tag without `v`, unknown mode, `package.json` mismatch — all exit 1 with no PATCH | PASS |

Every PATCH issued across the 28 runs, verbatim: `gh api --method PATCH repos/damianvtran/local-operator-ui/releases/<id> -F prerelease=true -F make_latest=false` (hold) and `-F prerelease=false -F make_latest=true` (promote). Nothing else was ever called.

**S4 — the job graph, simulated independently** (`graph.mjs`, my own parser/evaluator, not the PR's harness): 12 failure combinations, `unexpected results: 0`. A failed build (any platform), a failed `npm-publish`, a failed `preflight-macos`, a failed `attach-to-release` → `finalize-release` is **skipped**, so a failed build cannot leave the release promoted. A failed `open-release-window` skips all three builds *and* attach. Nothing depends on `finalize-release`, so its own failure cannot cascade. Invariants held in every case: nothing promotes without a successful attach, and nothing promotes without the window having opened. Both window jobs stay reachable on `workflow_dispatch`, so a repair still builds and attaches.

**S5 — the defect, reproduced read-only against the live repo.** Feed today: `{tag: v0.19.2, prerelease: false, assets: 11}`. `.../download/v0.17.2/latest-mac.yml` → **404** (v0.17.2 carries 0 assets) while `.../download/v0.19.2/latest-mac.yml` → **302**. My feed model (latest = newest non-draft, non-prerelease release; offered only if that release has an uploaded `latest*.yml`) reproduces `gh api /releases/latest` exactly over the real 12-release list, and reports exactly the reported symptom when v0.19.2's assets are removed. The window is measured, not assumed: v0.19.2 published 03:52:55, run finished 04:18:55 = **26 minutes**; v0.19.1 = 26m44s. And v0.17.2's run **failed** at 13:29:32 — its release is still a pre-release with 0 assets today, the real extant example of the state this flow now produces by itself instead of by hand.

**S6 — no regression in the pinned surfaces.** `validate-release.mjs` is not in the diff at all (`shasum -a 256` identical on both refs). The `npm-publish` job body is byte-identical. The `build-*` job bodies differ **only** in the added `needs:`/`if:` lines — every step, including the `dist/latest*.yml` artifact gate (`if-no-files-found: error`), is identical. The three installer regexes are unchanged by the refactor, and `checkAssetCollisions` keeps the same paging loop, now shared via `listAssets`. The whole non-comment `publish.yml` delta is the two new jobs plus one `needs`/`if` addition on each build. Also checked: `release-state.mjs`'s transitive local imports are exactly the two jobs' sparse-checkout lists; neither new job overrides the workflow-level `contents: write`, so the PATCH is authorized; and **the promote gate is not stricter than what a successful attach produces** — each build uploads an installer *and* `dist/latest*.yml`, and `artifactFiles` uploads every file in each dir, so a built release can always be promoted (no stranding).

**S7 — shipped gates, real tails.** `pnpm lint` → `Found 24 warnings`, `LINT_EXIT=0` (every one reported against a file this PR does not touch: `scripts/linux-sandbox.test.mjs`, `src/main/desktop-stream.ts`, `src/renderer/.../destination-pickers.tsx`, `bin/local-operator-ui.js`). `pnpm check-types` → `EXIT=0` (both tsc projects). `pnpm test:desktop` → `tests 527, pass 527, fail 0, EXIT=0`.

---

#### Findings

**Q-1 — BLOCKED (must be confirmed by the first real release): the exact body of the state PATCH.** The script sends `make_latest` as a **typed boolean** (`-F make_latest=false|true`) and its comment says `-f` was avoided because it "would serialize the string 'false'". But the REST schema types `make_latest` as a **string** enum `["true","false","legacy"]`, and both independent implementations send the string: `gh`'s own `release create/edit` (`params["make_latest"] = fmt.Sprintf("%v", *opts.IsLatest)`, source read from `cli/cli@trunk`), and the most-used community action (`softprops/action-gh-release`, `make_latest: 'true' | 'false' | 'legacy'`). `gh api --verbose` confirms the wire body this script produces is `{"make_latest": false, "prerelease": true}`. Whether the server accepts a boolean there is not determinable without a live PATCH — I deliberately did not make one, since it mutates public release metadata, and no local harness can stand in for it. Blast radius if rejected: a 422 fails `open-release-window`, which by design skips all three builds → the release gets no assets at all and remains in `/releases/latest` empty, i.e. the original defect, for that release, until an operator acts. The flag is not load-bearing in either direction (a pre-release cannot be latest, and the newest non-prerelease becomes latest on the flip), so the zero-risk hardening is to send `make_latest` as the documented string (`-f`, which serializes `"false"`/`"true"` exactly as `gh`'s own commands do) or to drop it from `open` and keep the explicit form only where it is meaningful. **Requested:** either make that change, or state that the boolean is intentional and put "watch the `open` job of the first post-merge release" in the runbook — and I will re-verify after the first real release.

**Q-2 — observation (fail-loud but fail-open on the feed): a failed hold is not retried, and `npm-publish`/`preflight-macos` still run.** A18 shows the `open` PATCH failing leaves the release in `/releases/latest` with no assets (the original symptom) until a human acts; the run does fail in seconds with the hand-repair line, which is the intended trade of a fail-closed *pipeline*, but the feed is unprotected in that window. Relatedly, `preflight-macos` and `npm-publish` only need `validate-release`, so an `open` failure cannot stop the npm publish. Not a regression (npm publish already preceded the builds) and not blocking; a single PATCH retry, or gating those two on `open-release-window`, would close it. Recorded for a decision, not as a FAIL.

**O-3 — nit, no regression.** The two new/changed script files sit outside `pnpm lint`'s path list, so biome never checks them. `release-state.mjs` is clean when checked directly; the edited harnesses carry only `useTopLevelRegex` warnings, a rule configured as `warn` and already present in `upload-release.mjs`. `origin/main`'s `test-publish-workflow.mjs` is equally unformatted (119 diff lines), so nothing regressed — but the ungated scope is worth knowing.

**O-4 — doc clarity.** The AGENTS.md bullet "Re-running the workflow for a published tag (`gh run rerun <run-id>`) closes the window when the attach succeeds" holds for a re-run of a **release-event** run. A re-run of a `workflow_dispatch` repair keeps that event, so it stays a repair and will not promote — correctly fail-closed, and the next bullet covers the hand-flip, but a word of disambiguation would prevent an operator waiting on a promotion that will never come.

---

#### Not verified locally (stated plainly)

1. Q-1: the live PATCH, and specifically boolean `make_latest` (above).
2. That the flip re-points `/releases/latest` in both directions. The exclusion rule is documented and visible live on v0.17.2; no PATCH was made to watch a full release leave and return.
3. End-to-end timing — that `open-release-window` lands within seconds of `published`, so the window covers essentially the whole 26-minute build. It is dependency-gated to run first and costs one API round-trip, but latency was not measured.
4. That `gh run rerun` on a release-event run keeps `event_name == 'release'`. Docs say re-runs use the original run's context and actor; either way the failure direction is fail-closed (a release merely is not promoted, and the printed hand-flip line covers it).
5. macOS/Windows packaging, signing, notarisation, the npm publish and `upload-artifact` — unchanged here, not exercised.

**QA never edited the repo:** `git status --porcelain` is byte-identical before and after the whole pass. Findings are for `### QA remediation — round 1`; no reviewer request, tag, issue or merge accompanies this comment.

damianvtran:

Scope: `39a9d28d93eeaa07991a28a28d7b44e211a89719..bc99b181ce4da0e9334ca54f537742f7449d2bc4` — one commit, every finding answered in it.

**R1 — fixed** in `bc99b18` (`fix(release): promote only the newest release, and state each PATCH field's real type`).

`finalize` now asks, before any PATCH, whether this release is the newest one eligible for `/releases/latest`; otherwise it leaves the release a pre-release and prints the release it found newer. The comparison rule, stated in the code comment and taken from the endpoint's own definition — "the most recent non-prerelease, non-draft release, sorted by the created_at attribute" (`GET /releases/latest`, REST docs) — is:

- the eligible set is published, non-draft, non-prerelease releases, from one paged `GET /releases` read;
- **this release is kept in that set** even though it is a pre-release at decision time — that flag is the only thing excluding it and the promote is about to flip it — and its own `created_at` is read from the same snapshot, so the comparison is internally consistent;
- recency is compared on `created_at`, not on semver, because the endpoint sorts by `created_at`: a backport published later genuinely is what the feed answers with, and a version comparison would disagree with the feed this check exists to guard;
- equal timestamps refuse, and an unreadable creation time or a release absent from the list refuses, rather than assuming — "we could not find out" is answered as "we cannot promote", the same convention as an interrupted upload.

Tests that fail on the round-1 head — executed against `39a9d28d9:scripts/release-state.mjs` in a throwaway tree, with this round's test file and the current `upload-release.mjs`/`validate-release.mjs`:

- `finalize: leaves an older release a pre-release when a newer one is published` — round-1 head promotes; the assertion is `writes == []`;
- `finalize reads past the first page, so a newer release cannot hide there`;
- `CLI: an older release's re-run attaches assets but does not promote` — round-1 head emits `["api","--method","PATCH","repos/damianvtran/local-operator-ui/releases/383131955","-F","prerelease=false","-F","make_latest=true"]`.

Round-1 head: `3 tests, 0 pass, 3 fail`. Current tree: 3 pass.

Real-data pass — the real script, the live repository, with the mutation made impossible (GETs proxied to the real `gh`, every PATCH refused at the `gh` boundary, so no release metadata was touched anywhere in this round):

```
$ RELEASE_TAG=v0.17.2 EXPECTED_SOURCE_SHA=4455c5705452ab0be4e487601776d8f67d73432b \
  EXPECTED_RELEASE_ID=387574664 IS_MANUAL_DISPATCH=false node scripts/release-state.mjs finalize
Tag v0.17.2 resolves to commit 4455c5705452ab0be4e487601776d8f67d73432b
Release v0.17.2 (id 387574664) state: published, prerelease: true
package.json name=local-operator-ui version=0.17.2 OK
Release v0.17.2 (id 387574664) is not the newest release eligible for /releases/latest; v0.19.2 (id 387785559) is. Leaving v0.17.2 a pre-release.
exit=0
gh argv issued: /git/ref/tags/v0.17.2, /git/tags/e0094b7e…, /releases/tags/v0.17.2,
  /contents/package.json?ref=4455c570…, /releases/387574664/assets?per_page=100&page=1,
  /releases?per_page=100&page=1            # no PATCH of any kind
```

That is the reachable case you named: in the live repository v0.17.2 is `prerelease=true` with 0 assets and `/releases/latest` answers v0.19.2.

**Q-1 — fixed** in `bc99b18`. Each field now goes out in the form the PATCH schema declares, checked against the REST docs rather than from memory: `prerelease` is a **boolean**, `make_latest` is a **string** enum (`true`, `false`, `legacy`).

- promote: `-F prerelease=false -f make_latest=true` — the documented string form.
- hold: `-F prerelease=true` only. The docs' "drafts and prereleases cannot be set as latest" makes the flag sufficient, so the field is not sent at all, which removes both the 422 blast radius you identified and the persisted-`false` hazard R5 raises (a `false` that could outlive the hold on a release later closed by hand).

The emitted argv is asserted twice: the stub-`gh` unit test (`the state PATCH is a typed boolean addressed by release ID`, which also asserts the bounded 3-attempt retry) and the CLI-level tests, which run the real script with a fixture `gh` as the only one on PATH. Same assertion on the real code path against the live 92-release list, with the PATCH refused by the harness:

```
["api","--method","PATCH","repos/damianvtran/local-operator-ui/releases/387574664","-F","prerelease=false","-f","make_latest=true"]
```

**R2 — fixed (decided, trade stated in the comment)** in `bc99b18`. I kept the stop, and the comment on `open-release-window` in `publish.yml` now says why and what it costs:

- the documented path issues **no PATCH at all** — step 9 publishes with `--prerelease`, so the script returns at its already-a-pre-release branch; a failed hold PATCH is only reachable when a release was published as a full release against step 9, which is the case the job exists to catch;
- "build anyway, do not promote" is not available without weakening a gate this file depends on: running a job past a failed dependency needs a status-function `if:`, which `test-publish-workflow.mjs` asserts against (such a bypass skips every other skip gate too), and dropping the `needs` edge instead would let a hold land *after* the promote and take a complete, advertised release back out of the feed;
- so the PATCH is now **retried three times** before the run stops (idempotent: it sets a fixed state on a pinned release ID), and only a persistent failure reaches the stop — which is also the same `contents:write` authority `attach-to-release` needs 26 minutes later;
- the residue is stated rather than hidden: a persistent hold failure on a full-release publish leaves that release at the head of `latest` with no assets until an operator acts on the failure line.

**R3 — fixed (tightened)** in `bc99b18`. `PLATFORM_UPDATE_METADATA` (`latest-mac.yml`, `latest.yml`, `latest-linux.yml`) sits beside `PLATFORM_INSTALLERS` in `upload-release.mjs`, mirroring each build job's own `dist/latest*.yml` glob, and `missingAssets` requires each platform's channel file by name. Regression covered both ways: `uploaded([...INSTALLERS, "latest-mac.yml", "latest.yml"])` refuses with `missing linux update metadata`. I also checked the live repository so the tighter gate cannot strand real releases: every published release from v0.12.4 through v0.19.2 carries all three channel files.

**R4 — fixed** in `bc99b18`. A new CLI-level test drives `node scripts/release-state.mjs finalize` with `IS_MANUAL_DISPATCH` unset, `true`, `False` and `0` (all refuse, zero PATCH) and `false` (one PATCH, argv asserted). The round-1 head had no coverage of that default-deny.

**R5 — fixed** in `bc99b18`. Both hand-repair sites now pass `--latest`: the AGENTS.md runbook and the PATCH-failure line, which now reads `gh release edit <tag> --prerelease=false --latest`. Verified against the installed CLI: `gh release edit --help` lists `--latest` ("Explicitly mark the release as Latest") and `--prerelease`.

**R6 — fixed** in the same text: the runbook now says `--prerelease` at creation *is* the hold, and that the hold job only acts on a release published as a full release by mistake.

**R7 — deferred, not fixed**: the `prerelease` output of `validate-release` is read by no job, which I re-confirmed. Removing it means touching the outputs plumbing of `validate-release.mjs`, which round 1 verified byte-identical to `origin/main`; that is unrelated cleanup, so it is recorded here instead of folded in. (One round-1 note does need updating: this remediation *does* change that file — see the last finding below.)

**R8 — fixed** in `bc99b18`. `localImports` now walks imports transitively (with a cycle guard), and a test pins that `scripts/validate-release.mjs` is reached through `scripts/upload-release.mjs`.

**Found by exercising the real path, outside the round-1 list — fixed**: the new recency read is large. A page of this repository's 92 releases is **1,768,746 bytes**, over Node's 1 MiB `execFileSync` default, so the read failed with `ENOBUFS`, which the shared reader reported as `GitHub metadata lookup failed: /releases?per_page=100&page=1` — a failed promote on a lookup that actually succeeded. `createApi` now sets `maxBuffer` (32 MiB) with the measurement in the comment, and the CLI test `a release list larger than the default buffer is read, not called a failed lookup` covers it. Both directions, on the live payload:

```
no maxBuffer (Node 1 MiB default, what the reader had): ENOBUFS (spawnSync gh ENOBUFS)
maxBuffer 32 MiB (the fix): read OK
```

This is why `scripts/validate-release.mjs` appears in the diff.

**Still unverifiable, stated plainly**: the live PATCH, and specifically server acceptance of the string `make_latest`, cannot be exercised without mutating public release metadata, so they remain unproven by this round — the string form is what the schema documents and what `gh`'s own `release create/edit` sends, which is the strongest offline evidence available. I also did not observe `/releases/latest` move in either direction against the live repository.

Verification on this head (`bc99b18`): `node scripts/test-validate-release.mjs` → 23 passed, 0 failed; `node --test scripts/test-release-safety.mjs scripts/test-publish-workflow.mjs` → 178 tests, 178 pass, 0 fail (round 1: 161); `pnpm lint` → 24 warnings, exit 0; `pnpm check-types` → exit 0; `pnpm test:desktop` → 527 tests, 527 pass, 0 fail.

damianvtran:
### QA remediation — round 1

Scope: `39a9d28d93eeaa07991a28a28d7b44e211a89719..bc99b181ce4da0e9334ca54f537742f7449d2bc4`, one commit (`bc99b18`).

**Q-1 — fixed** in `bc99b18`. I took the second option you offered and kept neither the boolean nor a redundant field: `make_latest` is now sent as the documented **string** (`-f make_latest=true`) on the promote, and **dropped from the hold** entirely (`-F prerelease=true` alone). The docs' "drafts and prereleases cannot be set as latest" makes the hold's copy of the field not just non-load-bearing but a hazard, because a `false` recorded there is what R5 worries about persisting onto a release later closed by hand. `prerelease` stays a typed boolean, which is what its schema declares.

Both directions are asserted rather than assumed, at two levels:
- unit: the stub-`gh` test now expects `["api","--method","PATCH","repos/…/releases/383131955","-F","prerelease=false","-f","make_latest=true"]` for the promote and `[…,"-F","prerelease=true"]` for the hold, and asserts the bounded retry (3 attempts logged, then a hard failure);
- CLI: `node scripts/release-state.mjs finalize` run as a subprocess with a fixture `gh` as the only one on PATH, asserting the same promote argv, plus the fail-closed env cases (Q-below).

I also reproduced your wire-form read on the real code path against the **live** release list with the PATCH refused at the `gh` boundary, so the argv below is what the real script emits for a real 92-release list:

```
["api","--method","PATCH","repos/damianvtran/local-operator-ui/releases/387574664","-F","prerelease=false","-f","make_latest=true"]
```

Your requested fallback — "or state the boolean is intentional and put 'watch the `open` job of the first post-merge release' in the runbook" — is therefore not needed for the type, but the runbook watch instruction is now there anyway for the PATCH itself: AGENTS.md says the run fails and prints the one command if the window cannot be opened or closed, and that nothing protects the feed until the operator acts on it. **The live PATCH and the server's acceptance of the string remain unverified** — I made no PATCH against the repository, and no local harness can stand in for that; the string form is what the schema documents and what `gh`'s own commands send. Please re-verify after the first real release, as you offered.

**Q-2 — fixed (documented), plus the retry you suggested as the minimum** in `bc99b18`:
- the state PATCH is retried three times before failing (idempotent: a fixed state on a pinned release ID), so a transient 5xx or secondary-rate-limit response no longer fails a release in seconds;
- AGENTS.md now names the recovery step explicitly: the run fails and prints the one command (`gh release edit <tag> --prerelease=false --latest`, or the hold's `--prerelease=true`), reaching that line means three attempts failed, **nothing is protecting the feed until it is dealt with**, and the recovery is to fix the cause and re-run (`gh run rerun <run-id>`) or run the printed line;
- a failed hold *also* stops the builds, so on that path the re-run is what produces the assets — that is now stated rather than implied.

Recorded decision, not done: I did not gate `npm-publish`/`preflight-macos` on `open-release-window`. That changes the publish semantics of the npm package (a version's npm publish would then depend on the hold), which is a separate decision from this finding; the retry covers the transient case that gating would have addressed, and the persistent case is the one the documented recovery is for.

**O-4 — fixed** in the same doc text: the runbook now says a `workflow_dispatch` repair keeps the dispatch event, so re-running a repair still will not promote, and names the hand-close instead.

**O-3 — re-checked, unchanged.** Those files are still outside `pnpm lint`'s path list, so I checked them with biome directly: `scripts/release-state.mjs` clean; `scripts/upload-release.mjs` 2 pre-existing `useTopLevelRegex` warnings; `scripts/test-release-safety.mjs` warnings of that same class only; `scripts/test-publish-workflow.mjs` carries the same 1 formatting error and 17 warnings it has on the round-1 head, in lines this round does not touch. Nothing regressed.

**New finding from this round's live pass, fixed** — worth knowing for your round 2 because it is a read you will want to drive: the recency check's `GET /releases` response is **1,768,746 bytes** for this repository's 92 releases, over Node's 1 MiB `execFileSync` default, so the read failed with `ENOBUFS` and was reported as `GitHub metadata lookup failed: /releases?per_page=100&page=1` — a failed promote on a lookup that succeeded. The shared reader now sets `maxBuffer`, with a CLI regression test that serves a >1 MiB release list.

Verification on `bc99b18`:

```
node scripts/test-validate-release.mjs                                            23 passed, 0 failed
node --test scripts/test-release-safety.mjs scripts/test-publish-workflow.mjs      178 tests, 178 pass, 0 fail   (round 1: 161)
pnpm lint                                                                         24 warnings, exit 0
pnpm check-types                                                                  exit 0
pnpm test:desktop                                                                 527 tests, 527 pass, 0 fail
```

Surfaces for round 2, since the delta touches them: the newest-release gate (`finalize`, fixture-driven and live-list-driven, refusal writes nothing), the emitted argv for both directions, the >1 MiB list read, the per-platform update-metadata gate, the CLI's fail-closed `IS_MANUAL_DISPATCH`, and the doc/script agreement on `--latest`.

Unchanged from round 1 and still not verifiable locally: the live PATCH and its type, the flip's effect on `/releases/latest` in either direction, end-to-end timing, and packaging/signing.

QA did not edit the repo in round 1; this remediation is the implementing agent's, so the pass above is my own evidence, not an independent one.

damianvtran:
### Agent review - round 2

Reviewer: deepseek/deepseek-flash (single model for the whole round; no provider fallback, no model switch)
Scope: 39a9d28d9..bc99b181c

**Method.** Delta first (`git diff 39a9d28d9..bc99b181c`, 7 files), then execution: the repo's own harnesses and shipped gates, plus an independent harness of my own (a fixture `gh` as the only one on the child's `PATH`, recording every argv and refusing or failing PATCHes) driven against a **live snapshot of this repository's 92 releases** (the raw page is 1,768,746 bytes, over Node's 1 MiB `execFileSync` default), so the recency decision runs on real data and no release metadata was mutated anywhere in this round. Repo untouched: `git status --porcelain` empty before and after.

**Verdict: no blocker and no major remains. The round is TERMINAL** — the two items below are a follow-up note and a nit, not a reason for another round.

---

#### Claims, confirmed or refuted point by point

**R1 — CONFIRMED FIXED, and it closes the case the finding described.** `scripts/release-state.mjs:115-147` (`newerLatestCandidate`), called at `:379-387`, after the `already a full release` return and before the completeness throw at `:389`.

The comparison, quoted from the shipped code:

```js
const createdAt = (release) => Date.parse(release.created_at || release.published_at || "");
const eligible = releases.filter((r) => r.draft === false &&
  (r.prerelease === false || String(r.id) === String(candidate.release_id)));
return newest && createdAt(newest) >= mineAt ? newest : null;
```

Judgement on each part:

- **`created_at` is the right key.** It is the attribute the endpoint the check protects is documented to sort by, and the code reads the same attribute, so it cannot disagree with the feed even where that attribute's semantics are odd — the docs gloss `created_at` as "the date of the commit used for the release", while in the live data every one of the 92 releases carries a parseable `created_at` (median 102 s from `published_at`) whose ordering reproduces the tag publish order. Measured live: newest eligible by `created_at` = `v0.19.2` = exactly what `GET /releases/latest` answers.
- **The candidate is correctly kept in the eligible set.** It is a pre-release at decision time and `release.prerelease === false` would otherwise drop it; `String(r.id) === String(candidate.release_id)` keeps it, and its `created_at` is read from the same snapshot (`mine`), so the comparison is internally consistent. Confirmed in this run — the normal promote below is *not* refused.
- **Equal / unreadable / absent all fail closed**, verified by execution (S3, S4, S5 below), each refusing with zero PATCH.
- **It stops the backwards move.** Probe: candidate `v0.17.2` (real id 387574664, real `created_at`, pre-release) with a now-complete 11-asset attach, against the real list, IS_MANUAL_DISPATCH=false.

```
--- round-1 head (39a9d28d9): exit=0
   out: Release v0.17.2 (id 387574664) promoted to a full release; it now answers /releases/latest.
   PATCH: ["api","--method","PATCH","repos/…/releases/387574664","-F","prerelease=false","-F","make_latest=true"]
--- remediation head (bc99b181c): exit=0
   out: Release v0.17.2 (id 387574664) is not the newest release eligible for /releases/latest; v0.19.2 (id 387785559) is. Leaving v0.17.2 a pre-release.
```

Same probe, both heads: the defect was real and the delta removes it.
- **It does not refuse a genuinely-newest candidate** (the way this could break normal releases). Probe S1, the real newest release as it is during its own window: exit 0, one PATCH with the promote argv. And the sweep: driving the shipped `finalizeRelease` over **all 92 real releases**, exactly one is promotable — `v0.19.2`, which is the release `/releases/latest` answers with today (78 of the 92 pass the completeness gate, so the refusal set is the whole history minus the newest, not an artefact of incomplete assets). No legitimate release is stranded by the gate.

| probe (real 92-release payload) | exit | emitted argv / message |
|---|---|---|
| S1 promote the newest, pre-release, complete | 0 | `… -F prerelease=false -f make_latest=true` |
| S2 re-run of an older release, complete | 0 | `… is not the newest …; v0.19.2 (id 387785559) is.` — **zero PATCH** |
| S3 candidate absent from the list | 1 | `… is not in the release list; refusing to promote a release the feed cannot see` — zero PATCH |
| S4 equal `created_at` on a competing release | 0 | `… is not the newest …; v0.19.1 (id 387781606) is.` — zero PATCH |
| S5 competing release with no readable time | 1 | `Release v0.19.1 … has no readable creation time; refusing to promote v0.19.2 …` — zero PATCH |
| S10 a newer *pre-release* in the list | 0 | promotes (cannot be latest; the newer-*draft* case is unit-covered) |
| S11 a newer *full* release in the list | 0 | refuses, names it |
| S12 newer release hidden on page 2, full page 1 | 0 | refuses, names it (pagination read is real) |

**Q-1 — CONFIRMED FIXED.** The emitted argv, captured by execution (not from comments):

- promote: `["api","--method","PATCH","repos/…/releases/387785559","-F","prerelease=false","-f","make_latest=true"]`
- hold: `["api","--method","PATCH","repos/…/releases/387785559","-F","prerelease=true"]` — `make_latest` is gone entirely.

Against the schema, fetched live from the REST docs for `PATCH /repos/{owner}/{repo}/releases/{release_id}`: `prerelease` is a **boolean**; `make_latest` is a **string**, `Can be one of: true, false, legacy`, and "Drafts and prereleases cannot be set as latest." So `-F prerelease=…` (typed boolean) and `-f make_latest=true` (documented string) are both in-schema, and the typed-boolean `make_latest` that Q-1 flagged — and that the round-1 head emitted on *both* directions — is gone. `gh`'s own `release create/edit` sends the same string form.

**Can the hold still persist `make_latest: false`?** No, and it no longer needs to: a release flipped to `prerelease: true` is out of `/releases/latest` by the endpoint's own definition, so the field is not load-bearing for the hold and its removal also removes the persisted-`false` hazard R5 raised. Corroborated read-only on the live repo: `v0.17.2` is a pre-release with 0 assets and `/releases/latest` answers `v0.19.2`, not it.

**R2 — CONFIRMED (stop kept, retry added).** Probe S7/S8: exactly **3** PATCH attempts recorded, then exit 1. The three attempts are byte-identical argv on a release pinned by ID, so a repeat is idempotent — **no double-apply hazard**, the retry cannot apply a different state the second time. `Atomics.wait` inside `pause()` does block the main thread, which Node permits and which is fine for a CLI that runs before any build; observed working (each retry path costs ~2 s). No *job* re-runs, so nothing else is retried.

**R3 — CONFIRMED (tightened, and not too strict).** `PLATFORM_UPDATE_METADATA` (`upload-release.mjs:29-43`) is required by name per platform. Checked against real data rather than the comment: live release assets are exactly `latest-mac.yml`, `latest.yml`, `latest-linux.yml` plus installers; 78 real releases carry all three, and the 14 that do not are the v0.1.x era and the hand-marked pre-release `v0.17.2`. Critically, the gate cannot be stricter than what the upload path produces — `artifactFiles` returns **every file** in each `<platform>-artifacts` dir, not just the installer matches — so a built release always carries all three and cannot be stranded. The new refusals are the right direction: refuse the promote (release stays unoffered) rather than advertise a platform whose channel file 404s.

**R4 — CONFIRMED, and the guard can observe its defect.** Mutation test in a throwaway `git archive` of the head: replacing `isManual: process.env.IS_MANUAL_DISPATCH !== "false"` with `isManual: false` fails **4** CLI tests (`unset`, `true`, `False`, `0`); the unmutated control is 73/73. Round 1's "the default-deny has no test" gap is closed with a discriminating test, and my own S9 (unset) yields zero PATCH.

**R5 / R6 — CONFIRMED.** The promote failure line now reads `… after 3 attempts; close the window by hand with: gh release edit v0.19.2 --prerelease=false --latest` (S8), the hold's reads `… gh release edit v0.19.2 --prerelease=true` (S7), and AGENTS.md carries both. Verified against the installed CLI: `gh release edit --help` (gh 2.100.0) lists `--latest  Explicitly mark the release as "Latest"` and `--prerelease`. R6's wording fix is in the runbook (the hold is `--prerelease` at creation; the job only acts on a release published as a full release).

**R7 — deferral accepted.** `.github/workflows/publish.yml:38` is the only occurrence of the `prerelease` output in the repo, i.e. it is genuinely read by no job; the reason given (touching the `validate-release` outputs plumbing is unrelated cleanup) is a legitimate deferral, and the remediation comment itself discloses that this commit *does* now change that file for the ENOBUFS reason, rather than leaving the stale claim standing.

**R8 — CONFIRMED.** `localImports()` now recurses with a `seen` cycle guard, and the new test pins that `scripts/validate-release.mjs` is reached through `scripts/upload-release.mjs`. The walk's regex still only matches `from "./x"` (not side-effect imports), which is pre-existing and complete for today's import graph — not a finding in this delta.

**ENOBUFS fix — CONFIRMED, and it is load-bearing.** `validate-release.mjs` gained only `maxBuffer: 32 * 1024 * 1024` plus its comment (`git diff origin/main` = that hunk alone, 7 added lines). Contract preserved, compared against `origin/main`'s version by running both CLIs over identical fixtures: for valid, release-ID mismatch, tag-SHA mismatch, package-version mismatch and draft-release, stdout, stderr and exit codes are **byte-identical** (exit 0/1/1/1/1), and `node scripts/test-validate-release.mjs` is 23 passed, 0 failed. The buffering change does not alter pins, output lines or exit codes — only whether a large read succeeds:

```
GET /releases?per_page=100&page=1 payload bytes: 1833254   (the raw fetch of the same page is 1,768,746)
--- WITHOUT maxBuffer (origin/main validate-release.mjs): exit=1
   stderr: GitHub metadata lookup failed: /releases?per_page=100&page=1
--- WITH maxBuffer (PR head): exit=0   … promoted …
```

That is the ENOBUFS defect reproduced in the direction that matters: a read that succeeded reported as a failed lookup, which the whole delta fail-closes on.

**publish.yml — comment-only.** Mechanically checked: with comment lines and hunk headers filtered out, the workflow delta is **empty**, so the job graph, `needs`/`if` scoping, env and sparse-checkout lists are unchanged and round 1's graph analysis still stands.

#### Findings

**R2-a — MINOR — the recency gate is version-blind, so a lower-version release published later inside the build window strands the newer one as a pre-release.** `scripts/release-state.mjs:137-146`. Constructed and run on the real script: candidate `v0.20.0` (newer version, pre-release, complete 11-asset attach, `created_at 06:00`), with `v0.19.5` (lower version, published as a full release, `created_at 06:05`) in the list:

```
out: Release v0.20.0 (id 700001) is not the newest release eligible for /releases/latest; v0.19.5 (id 700002) is. Leaving v0.20.0 a pre-release.
```

and re-running reproduces it exactly — the strand is permanent for that release, because the tie-break is `created_at` and nothing else, so `gh run rerun` cannot clear it (the next release heals the project, not this tag). The code's own comment argues the choice deliberately ("a release published later genuinely is the one the endpoint answers with even when its version is lower"), and the refusal is genuinely consistent with the feed — so I am **not** calling this a defect of the gate's premise, only a case the runbook does not mention: reachability needs a second release hand-published as a *full* release inside a ~26-minute window, which the documented flow (one release at a time, published `--prerelease`) does not produce. Worth one line in AGENTS.md naming the outcome and the hand-close, not a remediation commit.

**R2-b — NIT — the refusal message cannot distinguish "a newer release exists" from "I could not order these".** In the equal-timestamp case the log says `… is not the newest release eligible for /releases/latest; v0.19.1 (id 387781606) is.` although the two `created_at` values are equal and nothing is newer (S4). Since a tie is also a permanent strand for the same reason as R2-a, the message is the operator's only signal; saying "no strictly newer release, but the order cannot be established" would be accurate and would point at the hand-close.

On the newly-introduced-failure-mode question: no promotion that should happen is refused on the normal path (S1; and the 92-release sweep allows exactly one release — the one the feed answers with), the retry cannot double-apply (idempotent, pinned by release ID), and the only strand the delta can create is the R2-a race.

No other findings. Everything else I looked at in the delta holds: the promote/hold split, the retry bound, the per-platform metadata names against real assets, the transitive checkout guard, the pagination, the fail-closed env default, and the absence of a version bump (`package.json` identical to `origin/main`, `version = 0.19.2`).

#### Execution evidence (local runs; this repo has no CI on the PR branch)

```
$ node --test scripts/test-release-safety.mjs scripts/test-publish-workflow.mjs
ℹ tests 178   ℹ pass 178   ℹ fail 0   ℹ duration_ms 24091    (exit 0)

$ node scripts/test-validate-release.mjs
23 passed, 0 failed — All validation tests passed

$ pnpm lint          → Found 24 warnings.  LINT_EXIT=0
$ pnpm check-types   → TYPES_EXIT=0
$ pnpm test:desktop  → ℹ tests 527  ℹ pass 527  ℹ fail 0  DESKTOP_EXIT=0
```

The 24 lint warnings are all in files this PR does not touch (`scripts/linux-sandbox.test.mjs`, `src/main/desktop-stream.ts`, `src/renderer/…`), unchanged from round 1. No renderer/UI file is in the PR (`git diff --name-only origin/main..HEAD` has no `src/`, `bin/` or `tests/` path), so no design or UX round applies.

#### What I could not verify (stated plainly)

1. **The live PATCH.** I made no release-metadata mutation, so server acceptance of the string `make_latest` and the actual re-pointing of `/releases/latest` in both directions remain unproven by execution — the evidence is the fetched schema, `gh`'s own implementation, and the live `v0.17.2` pre-release being excluded from `/releases/latest` today. This is the one assertion in the delta that only the first real release can settle, as both the round-1 QA report and the remediation comment say.
2. **`gh run rerun` retaining `event_name == release`,** and Actions rendering `IS_MANUAL_DISPATCH` as the literal string `false`: asserted at the expression/CLI level only. I ran no workflow, no dispatch, and no re-run.
3. **The real pipeline end to end.** Nothing was merged, tagged or released; the promote has still never run against GitHub.

No reviewers, tags, issues, merges or workflow runs accompany this comment, and the repository was not edited.

damianvtran:
### QA report — round 2

**Reviewer:** qa-tester (deepseek/deepseek-flash) — independent pass, no edits to the repo
**Scope:** `39a9d28d93eeaa07991a28a28d7b44e211a89719..bc99b181ce4da0e9334ca54f537742f7449d2bc4` (the remediation delta only)
**Matrix:** `~/workspace/qa-124/r2/MATRIX.md` (surface, command, ACTUAL output, PASS/FAIL/BLOCKED) · raw runs in `~/workspace/qa-124/r2/runs/` · my own fixture `gh`, driver and graph evaluator in `~/workspace/qa-124/r2/`

**Verdict: PASS on every surface that can run locally; no FAIL. 2 BLOCKED — the live PATCH and the flip's effect on `/releases/latest` — restated below rather than substituted.**

Environment: `~/local-operator-ui-worktrees/fix-release-feed` @ `bc99b181c`, node v26.5.0, pnpm 10.30.3, gh 2.100.0, darwin 26.6.2 arm64. No PR-branch CI in this repo, so every row is a local run. The repository was never mutated: `v0.17.2` still reads `prerelease=true`, 0 assets, after the whole pass.

**Method, independently of round 1.** I wrote my own fixture `gh` (first and only `gh` on `PATH`, logging every argv to `gh-calls.jsonl`) and ran the real `scripts/release-state.mjs` as a subprocess in **37 scenarios (31 fixture + 6 live)**: exit codes, printed lines, gh argv and the resulting release state are actual output. Wire bodies come from the **real `gh` binary** with `GH_DEBUG=api` against an unreachable host (`127.0.0.1:9`, dial refused → nothing sent), replaying the argv the shipped script actually emitted. `git status --porcelain` empty before and after.

---

#### Q-1 — refuted as a type mismatch; still BLOCKED on live acceptance

The two PATCHes now emit exactly the forms the schema declares:

| | Command | Emitted argv | Real `gh` wire body |
|---|---|---|---|
| promote | `finalize` on a promotable release | `gh api --method PATCH repos/…/releases/383131955 -F prerelease=false -f make_latest=true` | `{"make_latest":"true","prerelease":false}` |
| hold | `open` on an incomplete release | `gh api --method PATCH repos/…/releases/383131955 -F prerelease=true` | `{"prerelease":true}` |

* `make_latest` is a **JSON string** on the wire (`-f`), `prerelease` a **JSON boolean** (`-F`) — against `docs.github.com/en/rest/releases/releases?apiVersion=2022-11-28` (*Update a release*): `make_latest` **(string)** "Can be one of: `true`, `false`, `legacy`"; `prerelease` **(boolean)**. For contrast, the round-1 argv `-F prerelease=false -F make_latest=true` still produces `{"make_latest":true,…}` — the boolean shape round 1 flagged, now gone.
* **The hold sends `prerelease` alone**; `make_latest` appears under no spelling (`release-state.mjs:233` gates it on `!prerelease`, and the hold's argv confirms it). That removes both the 422 blast radius and the persisted-`false` hazard.
* Over the **real** payload the promote argv is `… releases/387574664 -F prerelease=false -f make_latest=true` (case C below, PATCH refused at the boundary).
* **BLOCKED (unchanged):** no PATCH may be made against public release metadata, so the server's acceptance of the string form is still unobserved. What is proven is that the request body matches the documented schema and matches what `gh`'s own client sends; that is not a 200. I will re-verify on the first real release, as offered.

#### R1 — confirmed in both directions (a rule that refuses everything is not a fix)

| Case | Exit | ACTUAL | PATCHes |
|---|---|---|---|
| **backwards promotion**, candidate complete (so the refusal cannot be an asset artefact): `finalize` v0.17.2 with v0.19.2 newer | 0 | `Release v0.17.2 (id 387574664) is not the newest release eligible for /releases/latest; v0.19.2 (id 383131954) is. Leaving v0.17.2 a pre-release.` — release stays a pre-release | **0** |
| **ordinary case**: `finalize` a release that *is* newest by `created_at` | 0 | `Release v0.99.0 (id 383131955) promoted to a full release; it now answers /releases/latest.` | 1 (promote argv above) |
| **against the live repository**: `finalize` v0.17.2 over the real 92-release payload | 0 | `… is not the newest release eligible for /releases/latest; v0.19.2 (id 387785559) is. Leaving v0.17.2 a pre-release.` — the gh calls were the three reads plus `releases?per_page=100&page=1`, nothing else | **0** |

Rule edges, all as claimed: candidate absent from the list → refuses (exit 1); another eligible release with no readable `created_at` → refuses (exit 1); equal timestamps → refuses (`>=`); a newer **pre-release** or **draft** does not block (promotes); a newer release **on page 2** of 102 does not hide (pagination followed, refusal names it); newest-but-incomplete still refuses on assets with 0 PATCHes. Non-finding, on record: recency is compared on `created_at`, so a backport *published later* is promoted even at a lower version — that is the endpoint's own definition (`/releases/latest` sorts by `created_at`), it is stated in the code comment, and I agree with the choice made there.

#### R4 — the fail-closed wiring table, through the CLI

`finalize` with `IS_MANUAL_DISPATCH` **unset / `true` / `False` / `0`**: exit 0, `Manual dispatch: release state is not mutated by a repair; leaving the release as it is.`, **zero gh calls of any kind**. With the literal **`false`**: promotes with 1 PATCH. Confirmed as claimed.

#### R2 — the stop is kept, and the retry is real; R3, R5, R6, R8 — confirmed

* **R2**: a hold whose first two PATCH attempts fail now **succeeds on the third** (exit 0, 3 attempts logged, `held as a pre-release until it is asset-complete`); a persistent failure fails after 3 attempts with the hand-repair line. The stop's cost is stated in the `open-release-window` comment, and my own job-graph evaluator (10 injections, 0 unexpected) confirms the stated graph: a failed hold skips all three builds **and** attach, `npm-publish`/`preflight-macos` still run (the recorded decision, and they already needed only `validate-release`), a failed build/attach → `finalize` skipped, nothing promotes without a successful hold **and** a successful attach, `finalize-release` is terminal.
* **R3**: `PLATFORM_UPDATE_METADATA` requires the channel file **by name per platform** — `latest.yml` missing while the macOS file is present refuses with `missing windows update metadata`; an asset left in state `open` (interrupted) refuses with `missing linux update metadata`. Names match each build job's own `dist/latest*.yml` glob. Cross-check on live data: of 78 published non-prerelease releases only `v0.1.0`/`v0.1.1` lack a channel file — nothing this pipeline has produced is stranded.
* **R5/R6**: both hand-repair sites carry `--latest` for the close (verified in the failure line: `gh release edit v0.99.0 --prerelease=false --latest`) and the hold's line is `--prerelease=true`; AGENTS.md step 9 now says `--prerelease` **is** the hold, that the hold job only acts on a release published as a full release by mistake, and that a re-run of a dispatch repair keeps the dispatch event and will not promote.
* **R8**: the transitive walk (with cycle guard) and its pin are in the harness — and independently: my own closure over `scripts/release-state.mjs` is exactly `upload-release.mjs` + `validate-release.mjs`, and both new jobs' `sparse-checkout` lists contain exactly those three files. R7 deferred with a reason: accepted as a thread-level deferral, not re-litigated.

#### The ENOBUFS fix, exercised against the real read it was found on

| | ACTUAL |
|---|---|
| live payload | `gh api repos/…/releases?per_page=100&page=1 \| wc -c` → **1768746** bytes (1.69 MiB), 92 releases, page 2 empty |
| **pre-fix** reader (module at `39a9d28d9`) | `read FAILED: GitHub metadata lookup failed: /releases?per_page=100&page=1` |
| raw mechanism | `maxBuffer=node default (1 MiB) -> ENOBUFS (spawnSync gh ENOBUFS)` |
| **head** reader | `read OK: 92 releases on page 1` — and end to end the live runs reach their decision through it |
| genuine read failure still fail-closed | `finalize` with a nonexistent tag, real gh → `GitHub metadata lookup failed: /git/ref/tags/v9.9.9`, exit 1 |

**`validate-release.mjs`'s contract is unchanged**: the head file minus the added comment and `maxBuffer: 32 * 1024 * 1024,` is **byte-identical** to `39a9d28d9` (the added lines are that comment and that one option), and the catch still fails the read closed.

#### Shipped gates and harnesses — real tails, real exit codes

| Command | Exit | Tail |
|---|---|---|
| `node --test scripts/test-release-safety.mjs scripts/test-publish-workflow.mjs` | 0 | `ℹ tests 178 / pass 178 / fail 0` (round 1: 161) |
| `node scripts/test-validate-release.mjs` | 0 | `23 passed, 0 failed` / `All validation tests passed` |
| `pnpm lint` | 0 | `Checked 439 files in 309ms. No fixes applied. Found 24 warnings.` |
| `pnpm check-types` | 0 | both tsc projects completed |
| `pnpm test:desktop` | 0 | `ℹ tests 527 / pass 527 / fail 0` |

---

#### Not verified locally (stated plainly, not covered by the PASS above)

1. **BLOCKED — the live PATCH**, hence server acceptance of the string `make_latest` and of the field's absence on the hold.
2. **BLOCKED — that the flip re-points `/releases/latest` in either direction.** v0.17.2 (a pre-release with 0 assets while the endpoint answers v0.19.2) is evidence of the exclusion rule, not of this flow's flip.
3. End-to-end timing — that the hold lands within seconds of `published` and so covers the ~26-minute build. Dependency-gated and one API round-trip; latency not measured.
4. GitHub Actions' own scheduling semantics: the graph rows come from my evaluator over the YAML, not from a workflow run (no job here uses a status-function `if:`, which I asserted independently).
5. macOS/Windows packaging, signing, notarisation, the npm publish and `upload-artifact` — untouched by the delta, not exercised.

**QA never edited the repo** (`git status --porcelain` empty before and after, head `bc99b181ce4da0e9334ca54f537742f7449d2bc4`). No reviewer request, tag, issue, merge or `workflow_dispatch` accompanies this comment. No FAIL, so nothing is requested for `### QA remediation — round 2`; the two BLOCKED items are for the first real release.

damianvtran:

Scope: `39a9d28d9..bc99b181c` — no commit added in this round; head unchanged at `bc99b181c`, and both round-2 findings answered below as deferred.

**R2-a (minor) — deferred.** The recency gate is version-blind: a *lower*-version release hand-published as a full release inside the ~26-minute build window is what the feed's `created_at` ordering answers with, so the newer release stays a pre-release and no `gh run rerun` clears it (the tie-break is `created_at` alone). Reason for deferral: out-of-order versioning is the only trigger. The release-window discipline requires monotonic bumps — one release at a time, each published `--prerelease`, so the documented flow cannot turn a lower version into a full release mid-build — which means this needs a second, out-of-band publication to fire, and nothing in the runbook produces it. Making the gate version-aware is also not a late addition to this PR: by construction a version comparison would have to *disagree* with the `created_at` order the endpoint answers with, in exactly the case this check exists to guard, so it belongs in a change of its own with its own review rather than an amendment here — the same reasoning the code comment already records for choosing `created_at`. What would change the decision: if out-of-order cuts ever actually happen (a backport or a re-cut published manually as a full release, as the reviewer's demonstration constructed). That change should make the gate version-aware *and* carry the one-line AGENTS.md note the reviewer asked for, naming the outcome and the hand-close; until then the discipline above is the guard and this thread is the record.

**R2-b (nit) — deferred, same note.** The refusal message cannot distinguish "a newer release exists" from "the timestamps tie, so no order can be established" — the equal-`created_at` case prints `… is not the newest release eligible for /releases/latest; v0.19.1 (id 387781606) is.` although nothing is newer. The message is read only in the workflow log, by whoever is diagnosing a strand, so this is diagnostic polish rather than user-facing copy; it sits on the same code path as R2-a and lands with it (wording along the lines of "no strictly newer release, but the order cannot be established").

No commit, no doc edit, no version bump in this round; nothing merged, tagged or dispatched.

damianvtran:
### QA remediation — round 2

QA round 2 (`5651322009`) reported **PASS** on head `bc99b181ce4da0e9334ca54f537742f7449d2bc4`: no FAIL was raised, so there is nothing to remediate and no commit follows. The two items it recorded as BLOCKED are the ones only the first real release can settle — **the live PATCH** (server acceptance of the string `make_latest`, and of the field's absence on the hold) and **the flip's effect on `/releases/latest` in either direction** — both need release-metadata mutation, which neither QA round performed. They stand for that first release, as round 1's report and remediation said.

Both round-2 gates were raised on that same head — `### Agent review - round 2` (`5651306801`, terminal: no blocker, no major) and `### QA report — round 2` (`5651322009`, PASS) — so the review and the QA rounds agree on the reviewed revision and nothing has moved since.

PR #130, opened 2026-09-13
fix(chat): make the model picker's feedback perceivable, and the switch register in one frame

damianvtran:
## Summary of Changes

Make the `/model` picker's feedback perceivable in every theme, and make the switch register in the frame it is made. This is the operator's report — *"insufficient feedback on hover, click, etc to indicate that the model selection change has happened"* — worked through a design audit and a remediation round.

> **This body replaces one that described a different change.** Until this round the PR carried the run-sidebar child-transcript body of a sibling PR (`GET /v1/desktop/sessions/{id}/children/{id}/transcript`, the roster containment proof, a uvicorn QA recipe) while its title, branch and commits were this work — so every claim the review gate is meant to check against the code was about another repository's PR. The reviewer found it (`Agent review — round 1`, finding 2), and the numbers it quoted for the picker's own measurement were wrong twice over: the design round measured D10's worst dark palette as **obsidian 2.64** (not `localOperatorDark` 2.91), and the wash it calls "under ΔE00 4 in dune and tokyoNight" is **dune 4.88** (tokyoNight 3.74 and dracula 3.99 are the two that hold). Everything below is this diff.

**What changed**

- **The marks.** Hover is the pointer's own row; the keyboard's row is the selection; a pick in flight is the picked row. Hover and the in-flight row now carry a 1px `outline-control` edge on top of the tint, because the tint alone collapses in four themes (design D12).
- **The keyboard's target is named.** The footer reads `Arrows move · Enter picks <the row Enter would pick> · Esc closes`, a click moves the keyboard's row to what it clicked, and the list scrolls back to the top when the query changes (UX U1, U6).
- **The pending window holds.** The effort reading stays on the band for the whole pick (it used to vanish for 1.1-4.2 s), the band gains a labelled spinner, the picked row keeps a spinner in its meta slot, and the in-flight copy names the change rather than the machinery (reviewer major 1, UX U3, design D14).
- **The outcome lands wherever it is.** A refusal that arrives after the dialog is closed is written to the transcript when it arrives, in the strip's own words (UX U2). A switch to a row the dialog itself labels `no credential` reports itself as `warning` with the renderer's own caveat, instead of being indistinguishable from a runnable one (QA Q1), and the in-force ✓ follows the receipt rather than the next owner frame (QA Q2).
[...]

PR #126, opened 2026-09-13
fix(settings): read the MCP row's real wire fields, so a project server can be removed

damianvtran:
This PR's review rounds are in the comments: rounds 1 and 2 of the agent review and of the design review, each answered by an `Agent review remediation — round <N>` comment. The sentences quoted below are the round-1 ones where a later round changed the copy.

## The bug

`McpManagementSection` declared its own `MCPServerRow` type with three fields the
backend has never sent on this wire — `scope`, `setup_prompt`, `error` — and read
a fourth (`downstream_authorization`) that the backend hard-codes to `"unknown"`.
Every access was optional, so nothing failed loudly; the failures were all silent:

| Declared in the renderer | Actually sent | Consequence |
| --- | --- | --- |
| `scope` | `owned_scope` | Badge never rendered (both were `undefined`), and removal always asked for `scope: "global"`. |
| `setup_prompt` | `setup: {kind: "session_prompt", text}` | The whole "Copy setup prompt" branch was dead. |
| `error` | — | Rendered a field that is never sent. |
| `downstream_authorization` | `"unknown"` always | A per-row authorization line reprinting a constant the backend cannot know. |

The load-bearing one is the scope. `MCPDesktop.execute`
(`local_operator/mcp/desktop.py:191-193`) refuses a removal whose requested scope
does not match the source file's real scope:

```python
if owned_scope_for_source(sources.get(body.name), self.cwd) != body.scope:
    raise ValueError("This source is not owned by the selected scope")
```

Because the renderer derived that scope from `server.scope` — always `undefined`,
so always `"global"` — **a project-scoped MCP server could not be removed from the
desktop at all**, and the refusal surfaced only as a generic 409.

## How I reproduced it

Against a real backend (`local-operator serve`, run from a worktree of
`~/local-operator` at `main`, on an isolated `LOCAL_OPERATOR_CONFIG_DIR`, an
isolated port, and with the in-tree `noop`/`test` provider configured so the
route's `bind_runtime()` can start a session runtime without a network provider).
A project-scoped server was seeded through the same API the UI uses:

```
POST /v1/desktop/sessions/{id}/mcp  {"action":"add","name":"wire-fixture","command":"/usr/bin/true","scope":"project"}  -> 200
```

The panel was then driven in the real renderer against that backend — the shipped
component, the shipped `desktopResult`, and the committed `desktopProxyPlugin`
(`/__desktop`), which is this repo's documented browser-development surface
[...]

PR #133, opened 2026-09-13
chore(release): bump version to 0.19.5

damianvtran:
## Summary of Changes

Version-only bump for the `v0.19.5` window, owned by this session after checking with every live session working in this repository (all three cleared it: no other owner, no objection, no PR expected inside the window).

Scope check: the diff is `package.json`'s `version` line only (0.19.4 → 0.19.5). Carries no other change; `pyproject`-style feature bumps are never added by a feature PR in this repo.

Window contents since `v0.19.4` (`0309a5a75`, `80882ef2b`), bump class **patch** by materiality (two fixes, no new surface or API):

- #126 `fix(settings): read the MCP row's real wire fields, so a project server can be removed`
- #130 `fix(chat): make the model picker's feedback perceivable, and the switch register in one frame`

Release notes draft (What's New / Impact / PRs / Full Changelog) is in the comment below and will be used verbatim for `gh release create v0.19.5`.


damianvtran:
## What's New

0.19.5 makes the desktop model picker tell the truth. The picker's selection was invisible — the highlighted row was painted with the dialog's own ground — and choosing a model left no trace until the backend answered, which is exactly the operator's report ("insufficient feedback on hover, click … to indicate that the model selection change has happened").

- **The selected row is perceivable now, in every theme.** The active row used `bg-elevated` inside a `bg-elevated` dialog: measured `1.000:1`, ΔE00 `0.00`, with `4,810,358` of `4,813,440` pixels identical when the pointer moved across rows. It now carries a structural edge (1 px `outline-control`) beside the tint: `3.02–5.91:1` token, `5.85–16.70` ΔE00 against its ground, across all twelve themes — including `obsidian`, where the wash alone collapses to ΔE00 `0.77` and the edge carries the mark at `3.14:1` / ΔE00 `25.96`. Hover, keyboard highlight and selection are three distinguishable states, and hover clears when the pointer leaves.
- **Choosing a model registers in one frame.** The session's model is painted optimistically in the status band with a pending style and reconciled against the authoritative frame, so the change is visible immediately even when the backend is paying a cold runtime start. The wait is named (`Switching the model…`) rather than silent, and the picked row is marked while it is in flight.
- **The flow is honest end to end.** The footer names the model Enter will switch to — including when the keyboard's row is scrolled out of view — clicking a row moves the keyboard's row, typing a query resets the scroll, the in-flight button reads `Close` rather than the wrong `Cancel`, a refused switch leaves a note in the conversation even if the dialog was closed mid-flight, a switch that lands on a model with no credential is distinguishable from one that is runnable, and `Refresh from providers` keeps its rows instead of blanking a 1,450-row list.
- **#126 — a project-scoped MCP server can be removed again.** The desktop read fields the wire never sent, so every removal was sent as `scope: "global"` and refused with a 409; the scope badge and setup-prompt control were dead code reading those same fields.

## Impact

- **No breaking changes.** No API, wire-contract or config change; one new component contrast triple is asserted in `scripts/contrast-contract.mjs`.
- **For anyone using `/model`:** the interaction now confirms itself. The hover/selection feedback defect is fixed in all twelve themes, and the picker's own wait is legible instead of a silent dialog.
- **Performance:** the picker's ~1,450-row list no longer re-renders every row on each mouse move (memoized rows), and a `live` catalogue refresh no longer empties the list for ~2.3 s. The cold-runtime bind latency is a backend property (unaffected here) — this release makes it visible and optimistic rather than silent.
- **Not verified in this release, stated plainly:** screen-reader announcement of the new states; themes other than `localOperatorDark` in-app (the twelve-theme sweep is Storybook frames plus `check-themes`/`contrast-contract`); a full turn sent after a switch.
- **Deferred, recorded rather than hidden:** the result strip's success edge (`2.91:1` on `localOperatorDark`, worst `obsidian` `2.64`, six dark palettes `2.64–2.96`) and the needs-sign-in strip's edge on `elevated` (`2.51–2.98` in seven palettes) need a palette change with all twelve themes re-checked; the checkmark's out-of-band reconciliation needs the pre-pick selector; one refusal sentence's wording belongs to the backend's own copy.

## PRs

- #126 `fix(settings): read the MCP row's real wire fields, so a project server can be removed`
- #130 `fix(chat): make the model picker's feedback perceivable, and the switch register in one frame`

**Full Changelog**: https://github.com/damianvtran/local-operator-ui/compare/v0.19.4...v0.19.5

PR #118, opened 2026-09-12
feat(chat): make submitting a message register in one frame

damianvtran:
## What and why

Release: patch — submitting a message registers in one frame instead of ~1.15 s behind the cold runtime engage

Submitting a message felt like it took ~1s before it registered, the view switched
into the conversation and the run started. The TUI has no such lag.

Measured against an isolated backend (throwaway config dir, `hosting: test` /
`model_name: mock`, no spend), over the same calls the renderer's `desktopResult`
makes:

| hop | median |
| --- | --- |
| `POST /v1/desktop/sessions` (create) | **12 ms** (11-29) |
| `POST /v1/desktop/sessions/{id}/messages`, runtime not yet engaged | **1146 ms** (1134/1146/1220) |
| same endpoint immediately after | **12-42 ms** |

So the create hop is ~2% of the problem and the whole felt lag is the **cold
engage** taken inline inside the message request: `_ensure_bound()` spawns
`python -m local_operator.session.runtime.process`, dials it, syncs and waits for
its welcome ack while the renderer blocks. The TUI never pays it because it
engages eagerly and off the critical path; the desktop/HTTP surface had no
equivalent, because every route there either reads or mutates and nothing a
renderer could call engaged a runtime without also submitting work.

This is the renderer half. The backend half (`POST
/v1/desktop/sessions/{id}/warm`, `session_catalogue: 2 -> 3`) ships in
damianvtran/local-operator. **This PR does not depend on it for correctness** —
everything is gated on `session_catalogue >= 3` and degrades to an exact no-op
against any older backend, so the two can merge in either order.

**Change class: C2 — standard.** User-visible, so it carries a design round and a
UX round in addition to code review and QA.

## The change

Three independent contributors, and they are not equal. The backend warm removes
the 1.15s from the critical path; the optimistic echo makes whatever remains
invisible; the identity fix stops a second, smaller cost from re-introducing a
visible gap.

### 1. The optimistic echo (§C.1)

`admitChatDraft` paints the user's message **before** it awaits the message
request, so the text reaches the transcript without waiting on the backend.

**On the New-chat path that is not the same as "in one frame", and an earlier
draft of this description claimed it was.** A staged draft has no session id
until `sessions.create` returns, so the echo cannot exist for that hop: measured
on this machine at load 433-445, p50 **142 ms** and max **409 ms** over twelve
creates (UX round 1), against the 11-29 ms of an idle box quoted above. The
composer's clear was therefore moved off the submit - on a mounted transcript it
[...]

PR #137, opened 2026-09-13
build(packaging): keep build tooling and renderer-only packages out of app.asar

damianvtran:
### Summary

The packaged app shipped every one of its 65 declared production dependencies inside `Contents/Resources/app.asar/node_modules`. Six of them are loaded by the Electron main and preload processes; the other 59 could not be loaded at all. The renderer is bundled by Vite into `out/renderer` and the window runs with `nodeIntegration: false` / `contextIsolation: true` (`src/main/index.ts:233-243`), so there is no `require` in the renderer for a node_modules copy to satisfy, and no main/preload module imports them. They were ~470 MB of unreachable weight inside a 1.0 GB install.

This PR makes the manifest say what the app actually loads: `dependencies` goes from 65 entries to 6, five packages no file references are deleted, and a CI guard (`scripts/check-runtime-deps.mjs`, run on pull requests) fails when an unvetted production dependency appears again. No bundler plugin, `files` pattern or `asar` setting is touched - the `dependencies` / `devDependencies` split is the whole lever, plus the pnpm pin below, without which a pruned manifest packages into an app that cannot start.

Release: patch — the installed app shrinks from about 1 GB to a few hundred MB by keeping build tooling and renderer-only packages out of the shipped bundle.

### Why the manifest section is the lever

`electron-vite`'s `externalizeDepsPlugin()` externalises exactly the packages in package.json `dependencies`; everything else a main/preload module imports is bundled into `out/`. `electron-builder` then copies the `dependencies` closure - and nothing from `devDependencies` - into `app.asar/node_modules`. A renderer-only package that sits in `dependencies` therefore ships to every user even though nothing can ever load it. That is the defect this PR fixes, and `scripts/check-runtime-deps.mjs` is what keeps it fixed:

```
$ node scripts/check-runtime-deps.mjs      # on this branch
check-runtime-deps: 6 production dependencies, all on the runtime allowlist:
  @electron-toolkit/utils <- src/main/index.ts, src/main/backend/backend-installer.ts (is.dev, optimizer)
  dotenv <- src/main/backend/config.ts loads the backend env before any other main-module config
  electron-log <- src/main/backend/logger.ts
  electron-updater <- src/main/update-service.ts and src/preload/index.ts
  posthog-node <- src/main/index.ts (main-process telemetry)
  zod <- src/main/backend/config.ts, src/main/desktop-media.ts, src/shared/desktop-contract.ts

$ node scripts/check-runtime-deps.mjs      # on the pre-change manifest (the guard's negative test)
check-runtime-deps: FAILED (61 problems)

  - unvetted production dependency: @codemirror/language
    electron-builder copies every `dependencies` entry into app.asar, so this
    one ships to every user. If the Electron main or preload process loads it at
    runtime, add it to RUNTIME_DEPENDENCIES in scripts/check-runtime-deps.mjs
    with the import site; otherwise move it to devDependencies (Vite bundles the
    renderer, and it will still be installed for the build).
  ... 60 more ...
```

The guard is an allowlist rather than a scan of `src/main` + `src/preload` because a scan cannot see a dependency reached through a computed `require` and would pass vacuously if the scan itself broke; an allowlist fails on everything a human did not write down, so adding an entry is a reviewable decision.

### Disposition of the 65 pre-change production dependencies

Each row's evidence is a real reference site, found with
`grep -rnE 'from "X|require\("X|import\("X|@import "X' src electron.vite.config.js`
(`X` also matching `X/subpath`); `main/preload` counts only reference sites under
`src/main`, `src/preload` and `src/shared`, the three roots `tsconfig.node.json`
compiles into the Electron processes.

**Kept in `dependencies` (6)** - the `dependencies` section is these six plus
`dotenv`, which was a `devDependencies` entry even though `src/main/backend/config.ts`
imports it (moved into `dependencies`, which is also what makes the guard's list
complete):

| package | evidence |
| --- | --- |
| `@electron-toolkit/utils` | `main/backend/backend-installer.ts:14`, `main/index.ts:5` |
| `electron-log` | `main/backend/logger.ts:12` |
| `electron-updater` | `main/update-service.ts:23`, `preload/index.ts:3` |
| `posthog-node` | `main/index.ts:16` |
| `zod` | `shared/desktop-contract.ts:1`, `main/desktop-media.ts:15` |
`dotenv` | `main/backend/config.ts:10` (promoted from `devDependencies`)

<details><summary><b>Moved to devDependencies (55)</b> - installed for the build, no longer copied into the app</summary>

| package | evidence |
| --- | --- |
| `@codemirror/language` | `renderer/src/shared/themes/code-mirror-theme.ts:1` |
| `@codemirror/search` | `renderer/src/features/chat/components/canvas/code-editor.tsx:1` |
| `@codemirror/state` | `renderer/src/features/chat/components/canvas/code-editor-diff.ts:1` |
| `@codemirror/view` | `renderer/src/features/chat/components/canvas/code-editor-diff.ts:8` |
| `@emotion/react` | no direct import; peer of `@mui/material`, resolved when the renderer is bundled |
| `@emotion/styled` | no direct import; peer of `@mui/material`, resolved when the renderer is bundled |
| `@lezer/highlight` | `renderer/src/shared/themes/code-mirror-theme.ts:4` |
| `@mui/material` | `renderer/src/shared/types/theme.ts:1` |
| `@mui/x-date-pickers` | `renderer/src/shared/components/common/date-time-picker.tsx:18` |
| `@tailwindcss/vite` | `electron.vite.config.js:8` |
| `@tanstack/react-query` | `renderer/src/main.tsx:1` |
| `@tanstack/react-query-devtools` | `renderer/src/main.tsx:2` |
| `@uiw/codemirror-extensions-langs` | `renderer/src/features/chat/config/canvas-supported-extensions.ts:2` |
| `@uiw/react-codemirror` | `renderer/src/features/chat/components/canvas/code-editor.tsx:18` |
| `ag-grid-community` | `renderer/src/features/chat/components/canvas/spreadsheet-preview.tsx:5` |
| `ag-grid-react` | `renderer/src/features/chat/components/canvas/spreadsheet-preview.tsx:6` |
| `class-variance-authority` | `renderer/src/shared/components/ui/card.tsx:2` |
| `clsx` | `renderer/src/shared/lib/utils.ts:1` |
| `date-fns` | `renderer/src/features/settings/components/settings-page.tsx:30` |
| `dmg-builder` | build tooling only (electron-builder's own collectors/targets) |
| `electron-builder-squirrel-windows` | build tooling only (electron-builder's own collectors/targets) |
| `html-react-parser` | `renderer/src/shared/components/common/update-notification.tsx:9` |
| `katex` | `renderer/src/features/chat/components/markdown-renderer.tsx:27` |
| `lucide-react` | `renderer/src/features/settings/components/radient-account-section.tsx:15` |
| `mermaid` | `renderer/src/features/chat/components/mermaid-diagram.tsx:21` |
| `posthog-js` | `renderer/src/main.tsx:3` |
| `radix-ui` | `renderer/src/shared/components/ui/tabs.tsx:2` |
| `react` | `renderer/src/app.tsx:1` |
| `react-dom` | `renderer/src/main.tsx:5` |
| `react-error-boundary` | `renderer/src/shared/components/common/error-boundary.tsx:9` |
| `react-markdown` | `renderer/src/features/chat/components/markdown-renderer.tsx:4` |
| `react-router-dom` | `renderer/src/app.tsx:3` |
| `react-shepherd` | `renderer/src/features/onboarding/components/onboarding-provider.tsx:8` |
| `react-syntax-highlighter` | `renderer/src/features/chat/components/message-item/code-block.tsx:3` |
| `recharts` | `renderer/src/features/settings/components/settings-page.tsx:60` |
| `rehype-katex` | `renderer/src/features/chat/components/markdown-renderer.tsx:5` |
| `rehype-parse` | `renderer/src/features/chat/components/canvas/wysiwyg-utils.ts:2` |
| `rehype-remark` | `renderer/src/features/chat/components/canvas/wysiwyg-utils.ts:3` |
| `rehype-source-map` | `renderer/src/features/chat/components/canvas/wysiwyg-utils.ts:4` |
| `rehype-stringify` | `renderer/src/features/chat/components/canvas/wysiwyg-utils.ts:5` |
| `remark-gfm` | `renderer/src/features/chat/components/markdown-renderer.tsx:6` |
| `remark-math` | `renderer/src/features/chat/components/markdown-renderer.tsx:7` |
| `remark-parse` | `renderer/src/features/chat/components/canvas/wysiwyg-utils.ts:7` |
| `remark-rehype` | `renderer/src/features/chat/components/canvas/wysiwyg-utils.ts:8` |
| `remark-stringify` | `renderer/src/features/chat/components/canvas/wysiwyg-utils.ts:9` |
| `shepherd.js` | `renderer/src/features/onboarding/hooks/use-onboarding-tour.ts:14` |
| `sonner` | `renderer/src/shared/utils/toast-manager.ts:9` |
| `tailwind-merge` | `renderer/src/shared/lib/utils.ts:2` |
| `tailwindcss` | `renderer/src/styles/index.css:1` |
| `tw-animate-css` | `renderer/src/styles/index.css:2` |
| `unified` | `renderer/src/features/chat/components/canvas/wysiwyg-utils.ts:10` |
| `unist-util-visit` | `renderer/src/features/chat/components/canvas/wysiwyg-utils.ts:11` |
| `uuid` | `renderer/src/features/chat/utils/reply-utils.ts:2` |
| `xlsx` | `renderer/src/features/chat/components/canvas/spreadsheet-preview.tsx:9` |
| `zustand` | `renderer/src/shared/store/low-credits-store.ts:6` |
`@emotion/react` and `@emotion/styled` have no direct import site in the repository; they are peers of `@mui/material`, which the renderer imports, so Vite still resolves them when bundling.

</details>

**Deleted from the manifest (5)** - no reference anywhere in `src/`, `electron.vite.config.js`, the build config or the workflow files:

| package | evidence |
| --- | --- |
| `@openid/appauth` | no reference anywhere in `src/`, `electron.vite.config.js` or the build config |
| `@uiw/codemirror-theme-basic` | no reference anywhere in `src/`, `electron.vite.config.js` or the build config |
| `electron-debug` | no reference anywhere in `src/`, `electron.vite.config.js` or the build config |
| `jwt-decode` | no reference anywhere in `src/`, `electron.vite.config.js` or the build config |
| `keytar` | no reference anywhere in `src/`, `electron.vite.config.js` or the build config |
`keytar` is referenced only in a comment about where a token conceptually lives (`src/renderer/src/shared/api/radient/proxy.ts:7`) and `electron-debug` has no reference at all, as the task anticipated; the same full-tree grep shows `@openid/appauth`, `jwt-decode` and `@uiw/codemirror-theme-basic` are in exactly that state, so they are deleted on the same evidence rather than kept as dead weight. `keytar` also had its `onlyBuiltDependencies` entry removed from `pnpm-workspace.yaml`, since that list is an approval for a package's build scripts and the package is gone.

Test/QA note: the desktop suite and every gate run against this tree, so the deletions are covered by it; the five packages have no import site for a test to exercise.

### Measured: the shipped app

Both numbers from `pnpm build && CSC_IDENTITY_AUTO_DISCOVERY=false pnpm exec electron-builder --dir --arm64` on this machine (unsigned local build; the committed signing config is untouched). `du -sm` of the app bundle, and the asar breakdown by parsing the asar header:

| | origin/main @ e4d074c60 (65 deps) | this branch (6 deps) |
| --- | --- | --- |
| `du -sm "Local Operator.app"` | **744 MB** | **273 MB** |
| `Contents/Resources/app.asar` | 340 MB | 25 MB |
| `Contents/Resources/app.asar.unpacked` | 160 MB | none |
| asar content, total | 489.0 MB | 24.8 MB |
| asar `node_modules` | 470.3 MB across 601 packages | 6.0 MB across 96 packages |
| asar `out` (the app itself) | 18.7 MB | 18.7 MB |

The largest things that no longer ship: `app-builder-bin` 138.4 MB, `mermaid` 60.9 MB, `lucide-react` 22.2 MB, `date-fns` 20.9 MB, `posthog-js` 20.7 MB, `typescript` 18.9 MB, `ag-grid-community` 16.2 MB, `@mui` 9.1 MB, `7zip-bin` 8.8 MB, `xlsx` 7.0 MB. `out/renderer` is produced from the same inputs with the same config and comes out the same size in both builds (18.7 MB for `out/` as a whole, either way). `out/main` is the one output that this change alters: `dotenv` moved from `devDependencies` into `dependencies`, so `externalizeDepsPlugin()` now externalises it and the main bundle `require`s it from `app.asar/node_modules` instead of carrying a copy - visible as the bare `dotenv` specifier in the compiled `out/main/index.jsc`. Nothing else in `out/` moves.

### Evidence: the pruned app resolves its runtime dependencies and launches

Resolution, exactly as the main/preload bundles do it - bare specifiers from a module inside `app.asar` (`createRequire` on `app.asar/package.json`), run with the packaged Electron in Node mode; must-be-absent packages are a negative control so a probe pointed at an unpruned tree fails:

```
$ cd "dist/mac-arm64/Local Operator.app/Contents/Resources"
$ ELECTRON_RUN_AS_NODE=1 "../MacOS/Local Operator" /tmp/loui-require-asar.js
RESOLVED @electron-toolkit/utils            -> app.asar/node_modules/@electron-toolkit/utils/dist/index.cjs (loads under the Electron runtime; 'electron' is not a node_modules package)
RESOLVED dotenv                             -> app.asar/node_modules/dotenv/lib/main.js
RESOLVED electron-log                       -> app.asar/node_modules/electron-log/src/index.js
RESOLVED electron-updater                   -> app.asar/node_modules/electron-updater/out/main.js
RESOLVED posthog-node                       -> app.asar/node_modules/posthog-node/lib/index.cjs.js
RESOLVED zod                                -> app.asar/node_modules/zod/lib/index.js

ABSENT   mermaid                            -> MODULE_NOT_FOUND
ABSENT   lucide-react                       -> MODULE_NOT_FOUND
ABSENT   @mui/material                      -> MODULE_NOT_FOUND
ABSENT   keytar                             -> MODULE_NOT_FOUND
ABSENT   xlsx                               -> MODULE_NOT_FOUND

OK: every runtime dependency resolves and every must-be-absent package is gone
```

Launch, through the packaged app's own smoke hook (`LOCAL_OPERATOR_UI_SMOKE_TEST=true`, which prints the marker from `app.whenReady()` and exits before any window is created):

```
$ LOCAL_OPERATOR_UI_SMOKE_TEST=true LOCAL_OPERATOR_UI_WINDOW_MODE=headless \
    "dist/mac-arm64/Local Operator.app/Contents/MacOS/Local Operator"
12:52:30.237 › Logger initialized
12:52:30.240 › Log path: /Users/damian/Library/Application Support/Local Operator/logs
12:52:30.240 › Backend logger initialized
12:52:30.241 › Log path: /Users/damian/Library/Application Support/Local Operator/logs
12:52:30.244 › Backend service configured with port 1111 and URL http://127.0.0.1:1111
12:52:30.246 › Backend Service Manager initialized. Disabled: true
12:52:30.246 › Virtual environment path: /Users/damian/Library/Application Support/Local Operator/local-operator-venv
12:52:30.247 › Bundled interpreter trees sealed against bytecode writes: (none present); 0 directory path(s) selected and now refusing new entries, 0 bytecode file(s) selected and now refusing rewrites
12:52:30.248 › Could not find Python, checked: /Users/damian/local-operator-ui-worktrees/pr-deps/dist/mac-arm64/Local Operator.app/Contents/Resources/python_aarch64/bin/python3, /Users/damian/local-operator-ui-worktrees/pr-deps/resources/python_aarch64/bin/python3
12:52:30.248 › Backend Installer initialized. Virtual environment path: /Users/damian/Library/Application Support/Local Operator/local-operator-venv
12:52:30.248 › Resources path: /Users/damian/local-operator-ui-worktrees/pr-deps/dist/mac-arm64/Local Operator.app/Contents/Resources
12:52:30.248 › Python path: Not found
[window-mode] window mode headless: 1380x900, window created and never shown, page throttling off
LOCAL_OPERATOR_UI_READY electron=35.5.1
12:52:30.351 › Process exit event detected, ensuring the Local Operator backend service is terminated
$ echo $?
0
```

Notes on that output, so nobody reads more into it than it says. `READY electron=35.5.1` with exit 0 is the proof that the packaged main bundle loaded, resolved every runtime dependency and executed. `Backend Service Manager initialized. Disabled: true` is this machine's build env (`VITE_DISABLE_BACKEND_MANAGER=true`), so no backend was started or touched. `Could not find Python` is the same local-build property as the `resources/python*` lines in the packaging logs: this machine has never run `pnpm setup-python`, so `extraResources` had nothing to copy, and it is unchanged by this PR. The app bundle is unsigned (`CSC_IDENTITY_AUTO_DISCOVERY=false`), so the launch is a local measurement, not a notarized-installer check.

### The guard's own test, and the assertion on the artefact

Two halves of the same concern - what may be declared, and whether what was declared actually shipped - each with a
way to fail:

```
$ env -u NO_COLOR TERM=xterm-256color node --test scripts/check-runtime-deps.test.mjs
✔ passes when every production dependency is on the allowlist (104.93225ms)
✔ fails on a production dependency nobody vetted, and names it (118.068208ms)
✔ fails on an allowlist entry with no matching dependency (179.197541ms)
✔ refuses to pass over an empty dependency list (503.012959ms)
✔ passes on the repository's own manifest (1.008542ms)
ℹ tests 5
ℹ pass 5
ℹ fail 0
```

`check-runtime-deps.mjs` is a function over a manifest path plus a CLI, so the test drives the real entry point as a
subprocess against fixture manifests (exit codes included) and then asserts the repository's own manifest passes.
The empty-manifest case exists because a guard that passes over an empty list is the vacuous check this repository
keeps writing down; it now fails instead.

The artefact half is `scripts/check-packaged-closure.mjs`, wired into `publish.yml` after each of the three build
steps, because the manifest cannot express it: a package can resolve perfectly in `node_modules` and still be absent
from `app.asar`, which is exactly what the pnpm regression does. It resolves the production closure the way Node
does (each package's own dependencies, resolved from that package's directory, so pnpm's linked layout works),
enumerates the packages inside the produced `app.asar`, and fails on anything in the first set missing from the
second. On this branch's artefact, and with a fixture that declares a package the app does not ship:

```
$ node scripts/check-packaged-closure.mjs --dist dist
check-packaged-closure: /.../Local Operator.app/Contents/Resources/app.asar
  production closure: 44 package(s) from 6 declared dependencies
  shipped in app.asar: 101 package(s)
  OK: every resolved production dependency is present

$ node scripts/check-packaged-closure.mjs --dist dist --manifest /tmp/fixture-mermaid.json
  production closure: 154 package(s) from 7 declared dependencies
  shipped in app.asar: 101 package(s)
  FAIL: 110 resolved package(s) are missing from app.asar:
    - mermaid@11.6.0
    - @braintree/sanitize-url@7.1.1
    ... 108 more ...
    The app would fail to load them at runtime. This is the shape of the
    pnpm >= 10.29.3 packaging regression: see AGENTS.md, 'Which pnpm may
    install and package'.
```

### Cost of running CI on pull requests

`ci.yml` gains a `pull_request:` trigger so the guard runs before the merge rather than after it - a dependency guard that only fires on `push` to `main` lands once the change is already in. The cost is that every pull request now runs the whole workflow set (lint, type checks, the desktop suite, `pnpm audit`, and the macOS `npm pack` launch check), which duplicates what the push run does on `main`. The new `runtime-deps` job itself is free by comparison: it needs neither pnpm nor a build, so it only pays a checkout on a Ubuntu runner. Superseded pull-request runs cancel each other through a `concurrency` group (`cancel-in-progress` only on `pull_request`, so a merge is never interrupted mid-flight), which bounds the duplication for the common push-again case. If the duplication is still unwanted, the alternative is a narrowly triggered workflow for the guard alone; I kept it in `ci.yml` as the task asked.

### Finding (pre-existing; reproducible without this change, and it must not be merged past silently): pnpm >= 10.29.3 makes electron-builder copy an incomplete runtime closure

Building the app on this machine with the repo's usual pnpm (10.30.3) produced an `app.asar` that launched with

```
App threw an error during load
Error: Cannot find module 'ms'
Require stack:
- .../app.asar/node_modules/debug/src/common.js
- .../app.asar/node_modules/builder-util-runtime/out/httpExecutor.js
- .../app.asar/node_modules/electron-updater/out/AppUpdater.js
- .../app.asar/out/main/index.jsc
```

The cause is upstream, and it is in the installer, not in this manifest: `electron-builder` asks pnpm for the production tree (`pnpm list --prod --json --depth Infinity`, `app-builder-lib/out/node-module-collector/pnpmNodeModulesCollector.js:getArgs()`), and pnpm >= 10.29.3 drops dependency edges from that output. Same command, same tree, two pnpm versions:

```
$ pnpm list --prod --json --depth Infinity   # pnpm 10.30.3
debug node under electron-updater > builder-util-runtime: dependencies []      # the `ms` edge is missing
$ npx pnpm@10.29.2 list --prod --json --depth Infinity   # pnpm 10.29.2
debug node under electron-updater > builder-util-runtime: dependencies ["ms"]
```

To be exact about what is dropped: `ms` is still present somewhere in the 10.30.3 tree, attached to other copies of
`debug` that the *production* path (root -> `electron-updater` -> `builder-util-runtime` -> `debug`) never reaches.
The node on the production path is the one that loses the edge, and electron-builder walks the production path.

With 10.30.3 the same prune produced `app.asar` with **43 packages / 5.2 MB** and a main bundle that cannot load; with 10.29.2 it produced **96 packages / 6.0 MB** and an app that starts. That is pnpm issue [#10601](https://github.com/pnpm/pnpm/issues/10601) ("Regression in pnpm 10.29.3 with Electron packaging (transitive dependency missing at runtime)"), which reports the same `Cannot find module 'ms'` from `debug/src/common.js`.

* The measurements in this PR were taken with `pnpm@10.29.2` on `PATH` for the electron-builder step only (the install itself and every other command used the repo's pnpm). The pin that makes that the pipeline's own behaviour is in the next section.
* This is not introduced by the prune: `pnpm list` drops `debug -> ms` regardless of which packages the manifest declares. The pre-change tree only looked complete because its `node_modules` was still in the layout a pre-10.29.3 pnpm created, from which the same command reports the edge.
* It nonetheless blocks shipping this change: once `dependencies` shrinks, the packaging pipeline's own pnpm is the only thing standing between the prune and a broken installer, and CI resolves `pnpm/action-setup` `version: '10'` to the latest 10.x. The pin below is what closes that, and a new assertion on the packaged artefact is what would catch the class again if the pin ever comes out wrong.

### The pnpm pin, and what it means for a user

Because of the finding below, this PR also pins `pnpm` to **10.29.2** in every workflow that installs it (`version:` on `pnpm/action-setup`): five jobs in `ci.yml` and four in `publish.yml`. `install-scripts-check.yml` and `version-bump-guard.yml` run no pnpm, so they are untouched.

* **This is a deliberate temporary hold, not a policy of staying old.** Its only job is that the installers this repository publishes actually start. `AGENTS.md` gains a "Which pnpm may install and package" section with the measurements, the upstream issue, the local-packaging recipe for a machine whose own pnpm is newer, and the three checks - `pnpm list` edge, packaged-closure assertion, launch marker - that must pass before the pin is raised, written as commands rather than prose.
* **What changes for a user:** nothing. The pin is a build-time tool version; the app, its version, its installers and their contents are unchanged by it.
* **What changes for a contributor:** CI installs with 10.29.2 instead of the newest 10.x, and `pnpm install --frozen-lockfile` still resolves the committed lockfile unchanged (verified with 10.29.2 against `pnpm-lock.yaml` on this branch: it installs and leaves the lockfile byte-identical). A local `pnpm build` + `electron-builder` uses the contributor's own pnpm, so the AGENTS.md section tells them how to package with a good one until the pin comes out (that is what the measurements in this PR used; the pin is what makes CI do the same without thinking about it).
* **What removes it:** a pnpm release in which `pnpm list --prod --json --depth Infinity` reports the full closure again - the command in the finding below stops returning an empty `dependencies` for the `debug` node - confirmed by packaging and launching the app. The AGENTS.md section states the same exit condition.
* **Not done here:** `packageManager: pnpm@10.29.2` in `package.json`. It does work locally (pnpm 10.30.3 self-switches to the pinned version, so `pnpm install`/`pnpm build`/`pnpm exec electron-builder` keep working on the operator's machine - tested), but with the workflows already pinning `version:`, `pnpm/action-setup` refuses a repository that names pnpm in two places, and the workflows are the pin that matters for what ships. If we later prefer one source of truth, the field plus dropping the `version:` inputs is the cleaner shape - that is a follow-up, not a rider on this PR.

### Not done (deliberately)

* No version bump: `package.json.version` stays at 0.19.6; the release bump lands separately on `main`.
* No change to `electron.vite.config.js`, `build.files`, `build.asar` or renderer bundling. `out/renderer` is unchanged; the only `out/` difference is `dotenv` becoming external in `out/main`, which is the manifest move taking effect (see the size table).
* `@electron-toolkit/preload` stays in `devDependencies` even though `src/preload/index.ts` imports it: Vite bundles that import into `out/preload/index.js`, so it is not a shipped node_modules dependency. The guard's allowlist comment records that distinction explicitly.
* The pnpm regression above is fixed by pinning the tool, not worked around in the manifest: adding `ms` to `dependencies` would hide it for one package and leave the rest of the closure exposed.
* No `build.mac` / `build.win` target changes, and no `packageManager` field (see the pin section for why).


damianvtran:
### Agent review — round 1

Reviewer: **deepseek-flash** (`deepseek/deepseek-flash`); no provider fallback occurred during this round.
Scope: `e4d074c60924214ccf804d11527f2f9263ea90d6..2ba5d5a271eee0b01b923f8be395c2f89b81af5b` (two commits, 7 files: package.json, pnpm-lock.yaml, pnpm-workspace.yaml, `scripts/check-runtime-deps.mjs`, ci.yml, publish.yml, AGENTS.md).

**Verdict: clean — no blocker, no major. This round is TERMINAL**; the minors and nits below are follow-ups, not a reason for another round.

I read the diff first, then opened only what the diff could not answer. Everything below I ran myself, read-only, on the worktree at `/Users/damian/local-operator-ui-worktrees/pr-deps` (`git status` clean afterwards; I created no `dist/`).

#### What I independently verified

**The prune's premise — is every moved package really unreachable at runtime?** Enumerated every bare import in `src/main`, `src/preload`, `src/shared` and it is exactly the six allowlisted packages (`@electron-toolkit/utils`, `dotenv`, `electron-log`, `electron-updater`, `posthog-node`, `zod`) plus `@electron-toolkit/preload`, which is bundled. Then read the bytecode string table of `out/main/index.jsc`: the module-name strings there are those six plus `node:` builtins, and the only other standalone package-name strings in it are `uuid` and `xlsx` — which sit beside schema literals (`^[a-zA-Z0-9_-]+$`, `recommended`) and a file-extension list (`tsv`, `xls`, `xlsx`, `ods`) respectively, not next to a require call site. There is no computed require to miss: `grep -rnE 'require\([a-zA-Z_$]' src/main src/preload src/shared` is empty and there is no `createRequire`, so the string table is complete evidence. Confirmed the mechanism on the tool side too: `externalizeDepsPlugin()` in electron-vite 3.1.0 is literally `Object.keys(pkg.dependencies || {})` (`node_modules/.pnpm/electron-vite@3.1.0.../dist/index.mjs:352`), which is what the guard's docstring claims. `keytar`, `electron-debug`, `jwt-decode`, `@openid/appauth`, `@uiw/codemirror-theme-basic` have no reference anywhere outside the lockfile, so deleting them is right.

**The shipped artefact.** Parsed the asar header myself: `Contents/Resources/app.asar` = 25.1 MB, `node_modules` = 96 packages / 6.3 MB, no `app.asar.unpacked`, `du -sm` of the bundle = 271 MB (body says 273 — rounding on APFS, not material). All six runtime deps present; `mermaid`, `lucide-react`, `@mui/material`, `keytar`, `xlsx`, `dmg-builder`, `app-builder-bin` absent. `package.json.version` still `0.19.6`.

**Closure completeness in the shipped tree** (the exact failure class this PR works around): extracted the asar and resolved every bare specifier inside it. The single unresolved one is `supports-color` from `debug/src/node.js`, which is inside a documented `try { } catch { }` swallow; `ms` resolves from `debug/src`. The packaged closure is complete.

**The launch smoke is real.** I ran it myself: `LOCAL_OPERATOR_UI_SMOKE_TEST=true .../MacOS/Local Operator` → `LOCAL_OPERATOR_UI_READY electron=35.5.1`, exit 0. The log's require stack shows `dotenv`/`electron-log` etc. loading from `app.asar/node_modules`, i.e. the packaged main genuinely resolves its externals rather than inlining them.

**The pnpm pin's premise, reproduced on this tree.** `pnpm list --prod --json --depth Infinity` — 10.30.3: the `debug` node under `electron-updater > builder-util-runtime` carries no `dependencies` key at all (the `ms` edge is gone); 10.29.2: `dependencies {"ms": {"version":"2.1.3",...}}`. The collector's args are as cited (`app-builder-lib@26.0.12/out/node-module-collector/pnpmNodeModulesCollector.js:15-17`). Upstream pnpm/pnpm#10601 is still **open**, "Last pnpm version that worked: 10.29.2", same `Cannot find module 'ms'` from `debug/src/common.js`. 9 `pnpm/action-setup` sites pinned (5 ci.yml, 4 publish.yml), no `version: '10'` left, and no other workflow uses action-setup. The exit condition is discoverable from an open upstream issue. 10.29.2 is the right value.

**The guard.** Ran it: 6 dependencies, all allowlisted, exit 0. Falsified it in `/tmp` without touching the repo — an added unvetted dep, an allowlist entry with no matching dependency, and an empty `dependencies` each exit 1 with the intended message. It reads `package.json` and nothing else, so it is deterministic, cannot false-pass a new production dependency (manifest-based, so a dynamically-required chunk cannot slip past it), and cannot false-fail a legitimate runtime dep, because the allowlist is what a human asserts and a package that is *bundled* (not in `dependencies`) is correctly not required to appear.

**CI on head:** 8/8 green, including the new `Runtime Dependencies` and `Desktop Tests`. I did not re-run the local desktop suite (budget/host constraint); the one machine-specific local failure the task flags is unrelated to this diff.

#### Findings

**F1 — minor — the documented way out of a load-bearing pin points at a command that is not in the section.** `AGENTS.md:368` says "Remove the pin once a pnpm release reports the full closure again - the command above prints `ms` for the `debug` node". The code block above it (`AGENTS.md:353-356`) is the local *packaging* recipe (`npm install --prefix /tmp/pnpm-good ... electron-builder`); the section never shows `pnpm list --prod --json --depth Infinity`, nor where in its output to look. Whoever removes this pin has to re-derive the check. Add the exact command and the path to assert (`dependencies.electron-updater.dependencies.builder-util-runtime.dependencies.debug.dependencies.ms`), inside the same section, so the exit condition is executable as written.

**F2 — minor — the guard is the only new logic in the PR and has no test.** Every other logic script in `scripts/` has a `*.test.mjs` enumerated in the `test:desktop` list (`run-desktop-tests.mjs` is called with an explicit file list), and this guard is now the only thing standing under the prune. Its three failure modes were demonstrated in the PR body by hand; a later edit that drops the stale-entry check or the empty-manifest check would pass CI silently. Suggested shape, because the script as written cannot be tested from a fixture: `repoRoot` is derived from `import.meta.url` and the check runs at module top level, so either extract a pure `checkRuntimeDeps(declaredDeps)` that the file wraps, or accept a `--manifest <path>` argument; then add `scripts/check-runtime-deps.test.mjs` to the `test:desktop` list.

**F3 — minor — nothing asserts the *packaged* closure, in either direction, in the job that packs.** The guard polices the manifest → allowlist direction. The converse — a package the main/preload bundle actually loads at runtime sitting in `devDependencies` — is silently bundled, which is fine for pure JS but is exactly how a closure goes missing without any build error, and this PR exists because that failure mode is silent. The only in-CI proof that the packaged tree is complete is the macOS `npx` launch in `ci.yml`; `publish.yml`, which builds the installers users actually get, has no such assertion. Cheap positive counterpart: run the resolution probe the body already has (resolve the six from inside the packed `app.asar`, assert `ms` resolves from `debug/src`) in the packaging job. Not a ship-blocker — the pin covers the current hazard — but it is the durable fix and would let the pin come out with evidence.

**F4 — minor — the PR body contradicts itself about whether the pin is in this PR.** The "Finding" section still ends with "I have not pinned it in this PR because that is a repo-wide decision that touches four workflows and local environments, not this change's lever. Recommended follow-up: pin pnpm ...", while the same body later describes the 9 pins and a later section's bullet ("Not done here: `packageManager: pnpm@10.29.2`") assumes they exist. The stale paragraph is text from the first commit; a reader who finds it first concludes the opposite of the merged state, and the body is the durable record for a patch release. Delete or rewrite it in place when the body is next touched (no code change needed).

**F5 — nit — "`out/` is unchanged, so the app's own code and assets are byte-identical across the two builds" is false for the main bundle.** `dotenv` moved from `devDependencies` to `dependencies` in this PR, and `externalizeDepsPlugin()` externalises exactly `dependencies` — so at base main *inlined* dotenv and now it `require`s it (visible in the launch require stack). The `out/renderer` claim is fine; the `out/` claim is not, and it is the kind of statement a future bisect would trust. Say "the renderer output and assets are unchanged".

**F6 — nit — the `electron-updater` allowlist reason cites an import site that contributes nothing at runtime.** `scripts/check-runtime-deps.mjs:58`: "src/main/update-service.ts and src/preload/index.ts". `src/preload/index.ts:3` is `import type { ProgressInfo, UpdateInfo } from "electron-updater"`, erased at build — the preload bundle's string table contains no `electron-updater`. The list's header says each entry is "a package an Electron main or preload process loads at runtime", so name `update-service.ts` (and `update-install.ts`, which also imports it) and drop or annotate the type-only preload reference.

**F7 — nit — the ci.yml comment presents the whole job set running on PRs as the necessary consequence.** `ci.yml:8-13`: "GitHub cannot run a single job from a workflow whose trigger excludes the event ... Every job here is a pre-merge gate, so running the set on pull requests is the honest way". The requirement is only that the guard runs pre-merge, and this repo already has the pattern for exactly that (`version-bump-guard.yml` triggers on `pull_request:` alone). The body says the real reason is that the task asked for the guard in `ci.yml`; the comment should say that, otherwise the next reader concludes a dedicated guard workflow is impossible. Related, same block: `ci.yml` has no `concurrency` group, so a same-repo PR now runs the full set twice (push + pull_request) with no cancellation — free on a public repo, but worth a `concurrency` block if runner pressure ever matters.

**F8 — nit — restore the invariant, not the tree-specific detail, in the 10.30.3 record.** `AGENTS.md:339` and `ci.yml:32` say 10.30.3 "reports the `debug` node under electron-updater -> builder-util-runtime with `dependencies {}`". Measured on this (pruned) tree with 10.30.3, that node has **no** `dependencies` key at all, while the sibling `debug` nodes under `@electron-toolkit/utils > electron` do carry `ms`. I did not re-measure the base tree, so this is not a dispute about your measurement — it is that `{}` vs "key absent" is a tree-dependent detail, and the durable wording is the invariant: "drops the `ms` edge from that node". Anyone re-checking should assert on the edge, not on the shape of an empty object.

damianvtran:
### Rebase proof — `9e391ec5f` -> `60a9d2a5a` onto `main` @ `240fedae0`

Rebased because GitHub blocked the squash with merge conflicts after #118 landed (it touched `package.json`).

`git range-diff 9e391ec5f...60a9d2a5a`:

```
-:  --------- > 1:  240fedae0 feat(chat): make submitting a message register in one frame (#118)
1:  7e828dfff = 2:  9df9e5ce6 build(packaging): keep build tooling and renderer-only packages out of app.asar
2:  2ba5d5a27 = 3:  39d8e6004 build(ci): pin pnpm to 10.29.2 while electron-builder cannot see the full closure
3:  9e391ec5f ! 4:  60a9d2a5a test(packaging): test the runtime-dependency guard, and assert the packaged closure
```

Both build commits are patch-identical (`=`). The follow-up commit is `!`, and the whole of its delta is one line.

Byte-identical +/- line sets: `git diff --name-status e4d074c60 9e391ec5f` and `git diff --name-status 240fedae0 60a9d2a5a` are identical (9 files, same statuses each). Per-file added/removed line multisets are equal for every file except `package.json`, where the only difference is the single `test:desktop` line:

```
removed only in the old range: "test:desktop": ... scripts/picker-feedback.test.mjs",
added   only in the old range: "test:desktop": ... scripts/picker-feedback.test.mjs scripts/check-runtime-deps.test.mjs",
removed only in the new range: "test:desktop": ... scripts/picker-feedback.test.mjs scripts/warm-session.test.mjs scripts/submit-latency.test.mjs scripts/echo-delivery.test.mjs",
added   only in the new range: "test:desktop": ... scripts/echo-delivery.test.mjs scripts/check-runtime-deps.test.mjs",
```

Upstream file touched semantically: `package.json` only, and only that line. #118 appends three test files to `test:desktop`, this branch appends its own guard test, so the resolution is both - upstream's three first, the guard test last, which is the same append order each change used on its own. No other file of #118's is present in `git diff origin/main 60a9d2a5a` (checked by name), so none of its work is reverted.

`pnpm install` under the pin (`npx -y pnpm@10.29.2 install`): "Already up to date", and `pnpm-lock.yaml` comes out byte-identical (`sha256 51737f55c8e56c8bc4ae76c3f71d1a7eeccb6c59b3f3b90f396c1cff582201cd`), which is the consistency check that matters here - #118 added no dependency, so the lockfile must not move.

`main` has since advanced by a release-bump-only commit (#142, `0.19.6` -> `0.19.7`). GitHub reports the PR `MERGEABLE`/`CLEAN` against it, and a simulated merge (`git merge-tree --write-tree origin/main 60a9d2a5a`) yields `version: 0.19.7` with both this branch's 6-entry `dependencies` and all four new `test:desktop` entries intact, so the release bump and both feature changes survive the merge.

Re-run on `60a9d2a5a`: guard CLI pass (6 dependencies, all allowlisted), `pnpm lint` 0 errors (33 pre-existing warnings), `pnpm check-types` 0, `pnpm build` green in 1m18s.

damianvtran:
### Post-merge QA pass — artifact/final state (covers `fb9c0942e`)

Independent QA of PR #140's release candidate runs against a branch whose base contains this merge, so the #137 changes were exercised as part of it. Full matrix: `/Users/damian/workspace/pr140-qa/matrix.md`; the #140 round-1 report carries the findings. This is the artifact/final-state summary that concerns #137.

- **Runtime closure holds.** `node scripts/check-runtime-deps.mjs` → exit 0 (the 6-package allowlist). `node scripts/check-packaged-closure.mjs --dist dist` → `production closure: 44 package(s) from 6 declared dependencies / shipped in app.asar: 52 package(s) / OK: every resolved production dependency is present`, exit 0. Without an artifact it exits 1 with `no app.asar … nothing to verify yet`, i.e. it fails closed.
- **`app.asar` is 27,165,085 B** in the arm64 app against **355,638,307 B** in the installed 0.19.4 app, and the packaged tree contains no `electron` package (`files` exclusion working, asar header parsed directly).
- **The app itself:** 400,836,529 B (382.3 MB) with `pnpm setup-python` run first, 355,456,647 B (339.0 MB) without the bundled Python trees — versus 1022.9 MB for the installed universal 0.19.4. `lipo -archs` → `arm64`.
- **One extra, small, and not #137's regression:** the artifact now carries `Contents/Resources/app.asar.unpacked` with 7.0 MB of `@electron-internal/extract-zip` — a dependency of the `electron` npm package's installer, unreachable from the packaged app (nothing in `out/**` or `bin/*.js` requires it). The pre-change build had no `app.asar.unpacked`; it arrived with the Electron 44 bump. Neither of #137's gates can see it (they check declared `dependencies` and assert presence, not absence), so it may be worth a rule in `check-packaged-closure` — reported on #140 as Q1, no change requested here.
- #137's gate behaviour was also exercised indirectly: the packaged app boots headless on Electron 44 with a working preload bridge, and `Found Python at …/Resources/python_aarch64/bin/python3` shows the pruned interpreter tree is still discovered.

PR #138, opened 2026-09-13
build(packaging): per-arch macOS artifacts and delta-capable updates

damianvtran:
Packaging changes: what a release builds for macOS, what it attaches on every
platform, and the one release-gate check that keeps the first of those true.

Release: minor — smaller downloads and delta updates on every platform: macOS ships Apple Silicon and Intel builds separately, and updates now transfer only changed blocks.

## What was wrong

1. **macOS shipped one universal dmg (366 MB) and zip (353 MB).** The Electron
   framework inside carried both architectures — 433 MB of it, half unusable for
   any given user.
2. **Differential updates were impossible even though the block maps were
   built.** `ArchiveTarget` and `NsisTarget` write `<artifact>.blockmap` on every
   build (visible as `building block map` in any build log), and publish.yml's
   three `upload-artifact` path lists carried only installers and `latest*.yml`.
   No `.blockmap` has ever been attached to a Release, so electron-updater's
   `differentialDownloadInstaller` failed its first block-map fetch and fell back
   to the whole file, logging it and nothing else.
3. **Both bundled Python interpreters were copied into every build**
   (`extraResources` is not architecture-aware), so a universal app carried 94 MB
   of interpreter, half of it unrunnable.

## What changes

- `mac.target` is a dmg and a zip for `arm64` and `x64`; `x64ArchFiles` is gone
  (it only existed for the universal merge). `artifactName` is unchanged
  (`${name}-${version}-${arch}.${ext}`), which yields
  `local-operator-ui-<v>-arm64.dmg`, `-x64.dmg`, `-arm64.zip`, `-x64.zip` —
  **the marketing site has to match these exactly.** The names come from that
  pattern, and the build log below confirms the per-architecture app directories
  they are built in (`dist/mac-arm64`, `dist/mac`).
- `scripts/prune-python-resource.mjs`, wired as `afterPack`, removes the
  interpreter tree the packaged app cannot run. `afterPack` rather than an
  `extraResources` filter because the filter is not architecture-aware and would
  have to spell the architecture twice; it runs after the files are copied and
  **before signing**, which is load-bearing — the tree is code-sealed inside the
  `.app`, so anything removed after signing is a `file missing:` violation no
  update-time heal can repair. It reads electron-builder's `Arch` enum, which is
  numeric (`--arm64` arrives as `3` — found by running the real build, not by
  reading the docs).
- `dist/*.blockmap` is added to all three `upload-artifact` path lists.
  `scripts/upload-release.mjs` maps *every* file in each platform's artifact
  directory, so it needed no change — verified, not assumed.
- `verify-macos-artifacts.mjs` gains one check: the app ships **exactly the
  interpreter tree its own architecture resolves**, read from `lipo -archs` on
  the bundle's framework binary. A count alone would pass for an arm64 bundle
  that kept the x86_64 interpreter — an app that cannot start its backend — and
  both-trees is silent, because such a bundle still runs perfectly.
  `docs/CODE_SIGNING.md`'s assertion table (which was also missing the bytecode
  check) is corrected.
- `scripts/update-robustness.test.mjs`: the `*-universal.*` fixtures become
  per-architecture, the discoveries assert two images, and a new section drives
  the real `MacUpdater` against a local feed (below).

## Windows: deliberately NOT changed in this PR

The union installer `local-operator-ui-setup-<version>.exe` (402 MB, beside the
200 MB + 202 MB per-arch pair) **stays**, and `nsis.buildUniversalInstaller` is
left at its default. It looks like free space, and removing it is wrong:

- `NsisUpdater.doDownloadUpdate` resolves its download with
  `findFile(resolveFiles(info), "exe")` — the **first** `.exe` in the feed, with
  no architecture filter at all. Arch selection exists only in `MacUpdater`
  (`files.filter(isArm64)` before it picks).
- The real v0.19.5 `latest.yml` lists the union installer first, and as `path`:

  ```
  files:
    - url: local-operator-ui-setup-0.19.5.exe         421851256
    - url: local-operator-ui-setup-0.19.5-x64.exe     210066624
    - url: local-operator-ui-setup-0.19.5-arm64.exe   212406896
  path: local-operator-ui-setup-0.19.5.exe
  ```

So the union installer is what gives a Windows-on-ARM machine an ARM64 install
today. With only the per-arch pair, the winner is whichever of the two
concurrent packaging tasks finished first — an x64 build, emulated, for ARM64
users. **A future change that wants the 402 MB back must first make the Windows
feed architecture-aware**, which is a change to `electron-updater`'s release
metadata and selection, not to `package.json`. The constraint is recorded in
publish.yml (beside the Windows build step) and in `docs/BUILD.md` § "Windows:
the union installer is load-bearing".

### The block-map upload really was the whole problem

`gh run view --job 103728025149 --log` (v0.19.5, Windows) — the block maps were
built and stayed on the runner:

```
• building block map  blockMapFile=dist\local-operator-ui-setup-0.19.5.exe.blockmap
• building block map  blockMapFile=dist\local-operator-ui-setup-0.19.5-x64.exe.blockmap
• building block map  blockMapFile=dist\local-operator-ui-setup-0.19.5-arm64.exe.blockmap
```

and the v0.19.5 Release carries no `.blockmap` asset at all:

```
$ gh release view v0.19.5 --json assets --jq '.assets[] | "\(.size)\t\(.name)"'
212406896	local-operator-ui-setup-0.19.5-arm64.exe
210066624	local-operator-ui-setup-0.19.5-x64.exe
421851256	local-operator-ui-setup-0.19.5.exe
366366519	local-operator-ui-0.19.5-universal.dmg
352857920	local-operator-ui-0.19.5-universal.zip
```

### Delta updates now work: real MacUpdater, real block maps, real HTTP

The new section of `scripts/update-robustness.test.mjs` drives the shipped
library, not a copy of it: `MacUpdater` and `ElectronHttpExecutor` bundled from
`electron-updater`, a local feed over real loopback HTTP with real `Range`
requests, and block maps produced by electron-builder's own
`app-builder blockmap` call. `require("electron")` is substituted (this process
is not Electron) and `ElectronAppAdapter` is replaced by a temp-cache adapter;
the feed is a generic provider with `useMultipleRangeRequest: false`, which is
what `GitHubProvider` itself pins "because GitHub uses S3" — so the single-range
path exercised is the one users' updates take. The library's own architecture
probes are the one thing steered, because both branches have to be reachable on
a Linux CI runner.

```
$ node --test scripts/update-robustness.test.mjs
✔ the updater resolves the channel file, and both architectures are listed
arm64: 32768 of 4181959 bytes transferred (0.78%)
✔ an arm64 Mac downloads its own build, and only the changed blocks
x64: 32768 of 4181959 bytes transferred (0.78%)
✔ an x64 Mac downloads its own build, and only the changed blocks
60% of the file changed: 2523136 of 4181959 bytes transferred (60.33%)
✔ a build that changed across most of the file transfers most of it
✔ the first update off a universal build falls back to a full download
no block maps: 4181959 of 4181959 bytes transferred (100.00%)
✔ without the block maps a release offers, the updater transfers the whole file
ℹ tests 50  pass 50  fail 0
```

**What the 0.78% does and does not say.** That fixture's change is the smallest
non-zero delta there is — 4096 bytes inside one 32 KiB block — so it measures the
mechanism, not a typical update. The same harness with 60% of the file changed
transfers **60.33%** (2523136 of 4181959 bytes), and a from-scratch build
transfers 100% because there is nothing to reuse. A real release that bumps
Chromium or Electron changes most of an archive and will therefore still transfer
most of an archive. What changed is that an update now transfers *the blocks that
actually changed*, instead of being structurally unable to: 353 MB every time,
for every user, forever.

Each delta run also asserts that the *other* architecture's build is never
requested, that every request for its own build carries a `Range` header, that
the previous release's block map is fetched by URL, and that the downloaded
file's sha512 is the one the channel file promises.

**The transition release is a full download.** Deltas need the *previous*
release to carry its block maps too, and no release ever has — so the first
update after this one falls back to a full download for everyone, and every
update after it can be a delta. The `first update off a universal build` case
above asserts that path end to end: the derived old-block-map URL 404s, the
update degrades to the whole file, and the bytes that land match the channel
file's hash. That matters because the diff base in the cache is a *universal* zip
from 0.19.6 while the release being installed is per-architecture — the shape
change must not be diffed across.

### The per-architecture build, from a real one

arm64 `--dir` build (`npx -y pnpm@10.29.2 exec electron-builder --dir --arm64`,
`CSC_IDENTITY_AUTO_DISCOVERY=false`, signing disabled locally):

```
• packaging  platform=darwin arch=arm64 electron=35.5.1 appOutDir=dist/mac-arm64
Pruned bundled Python the arm64 app cannot run:
  dist/mac-arm64/Local Operator.app/Contents/Resources/python

$ du -sh "dist/mac-arm64/Local Operator.app"                     802M
$ ls "…/Contents/Resources" | grep -E '^python'                python_aarch64
$ lipo -archs "…/Contents/MacOS/Local Operator"                  arm64
$ lipo -archs "…/Electron Framework.framework/Versions/A/Electron Framework"  arm64
```

So: one interpreter tree, 47 MB of it, and an app whose own binaries are thin
arm64 — where the universal build carried a fat framework and both trees. The
`802M` is dominated by `Contents/Resources/app.asar` (349 MB, the unpruned
dependency tree `build/prune-app-deps` addresses) and a 243 MB arm64 framework;
the framework is expected to roughly halve per image against the universal one.

**Not produced here, and this PR does not claim it:** the real per-arch `.zip` +
`.zip.blockmap` + `latest-mac.yml` feed. Three packaging runs were attempted on
this host and only the first completed:

1. `--dir --arm64` under the installed pnpm 10.30.3 — **completed**, and is the
   source of every number above.
2. `--dir --arm64` under `pnpm@10.29.2` — stalled during packaging (killed at 25
   minutes).
3. `--mac zip --arm64 --x64 --publish never` under `pnpm@10.29.2` on a healthy
   host (87 GB free, after a restart) — reached the x64 packaging step, then sat
   **CPU-bound with no child process and no output for over six minutes**, with
   `dist/mac` frozen at 260 MB, and was killed. Worth recording for whoever
   retries it: the two arch flags produced a single packaging pass
   (`platform=darwin arch=x64`) — which is **not** to say each architecture
   needs its own invocation. One invocation is what produces ONE merged
   `latest-mac.yml` covering every architecture: `app-builder-lib/out/packager.js`
   builds a single target object per platform and calls `finishBuild()` once, and
   the real v0.19.5 metadata is that shape (one `latest.yml` listing all three
   Windows installers, one `latest-mac.yml` listing both mac artifacts, each from
   a single `pnpm dist:*` command). Two invocations would leave a **one-arch**
   feed — the second run's channel file would be written without the first run's
   artifacts — so the stall here has some other cause.

So the artifact **names** here are the `artifactName` pattern's output,
corroborated by the per-architecture app directories electron-builder actually
created (`dist/mac-arm64`, `dist/mac`), and the `.zip.blockmap` /
`latest-mac.yml` contents are asserted only by the harness against a feed that
mirrors the real one. The QA round produces the real feed independently.

**Which pnpm built what:** the artifacts above were built with `pnpm@10.29.2`
(`npx -y pnpm@10.29.2`), not the 10.30.3 installed here, because
`pnpm list --prod --json --depth Infinity` — the command electron-builder's
dependency collector runs — is broken from 10.29.3 (upstream pnpm#10601). No pnpm
pin is added in this PR: that belongs to #137, which has now landed and pins
10.29.2 across the workflows. No missing-module failure was seen in either build.

## Gates on this head

```
pnpm lint            exit 0   (33 pre-existing warnings in files this PR does not touch)
pnpm check-types     exit 0
pnpm build           exit 0
pnpm test:desktop    29 files, concurrency 7 — 591 tests, 591 pass, 0 fail (92.3 s)
node scripts/test-validate-release.mjs                        23 passed, 0 failed
node --test scripts/test-release-safety.mjs scripts/test-publish-workflow.mjs \
  scripts/test-version-bump-guard.mjs                          208 tests, 208 pass
node --test scripts/update-robustness.test.mjs                 50 tests, 50 pass
node --test scripts/prune-python-resource.test.mjs             10 tests, 10 pass
```

`pnpm test:desktop` is the whole CI desktop job, run on this head through the
same concurrency governor CI runs (it chose 7 workers here: `cpu share 0.5 of 14
cores; 7.4 GB available`). `pnpm build` needs `VITE_*` client ids from `.env`,
which is loaded from the operator's checkout for the run only — no `.env` is
committed and none is left in the tree.

**CI now runs the whole job set on this PR.** #137 gave `ci.yml` a
`pull_request` trigger, so this head is checked by Lint, Type Checking, Desktop
Tests, Runtime Dependencies and Security Audit as well as `Version Bump Guard`
(passing, which also confirms no version bump is carried). The numbers above were
re-run on the rebased head under the toolchain `main` now pins
(`npx -y pnpm@10.29.2`), so they are the same commands CI runs rather than
substitutes for it. If Desktop Tests is red it is #118's
`R6-1: an unreadable census cannot authorise a removal`
(`scripts/submit-latency.test.mjs`), which fails on `main` independently of this
branch.

## Not done, deliberately

- **Windows and Linux artifacts are config-level, not build-verified** — neither
  can be built here. The Windows decision above is the reason the union installer
  is untouched; what is verified is the config and the metadata that drives it.
  The `dist/*.blockmap` glob in the Linux path list currently matches nothing: no
  Linux target emits one (AppImage resolves its own deltas through zsync), and
  the step's `if-no-files-found` is satisfied by the installers. CI runs on
  Linux, so the delta harness itself executes there.
- **No `dependencies`/`devDependencies` changes**, no version bump, no signing or
  notarization settings changed beyond per-image correctness, so
  `build/prune-app-deps` rebases cleanly.
- `docs/PYTHON_BUNDLING.md` still describes one interpreter and calls
  `pnpm setup-python-standalone` (the script is `setup-python`); it predates this
  change and is left alone rather than half-rewritten.


damianvtran:
### Agent review — round 1

Reviewer: reviewer subagent — `deepseek/deepseek-flash` (Local Operator; no provider fallback occurred during this round)
Scope: `e4d074c60924214ccf804d11527f2f9263ea90d6..a6185753e583f9f263adf2503136f2e2a4559a2b` (`main`..`build/per-arch-artifacts`)

What I did rather than read: `node --test scripts/prune-python-resource.test.mjs` (10/10 pass); `node --test scripts/update-robustness.test.mjs` (48 tests, 47 pass, 1 pre-existing machine-specific fail — see "Not this PR" below); read `electron-updater@6.6.2` and `app-builder-lib@26.0.12` out of `node_modules`; pulled the real `latest.yml` / `latest-mac.yml` for v0.19.5 off GitHub; ran `file` over both bundled interpreters; and drove app-builder's real `blockmap` + electron-updater's own `computeOperations` over two artifact pairs to put a number on the delta claim.

Confirmed good, so I do not re-raise them: `dist/*.blockmap` reaches all three `upload-artifact` lists and nothing else was dropped; `upload-release.mjs:artifactFiles()` maps *every* file in each `<platform>-artifacts` dir, so no extension allowlist exists to update; `release-state.mjs` / `upload-release.mjs` assert installers through arch-agnostic regexes (`/\.(dmg|zip)$/` etc.), so two DMGs and four mac files break nothing; `PYTHON_RESOURCE_DIRS` is the right way round (`resources/python` is x86_64, `resources/python_aarch64` is arm64, per `file`); the prune runs on darwin only and refuses `universal`; the update-time heal (`update-install.ts:313` `BUNDLED_PYTHON_DIRS`, `isPythonBytecodePath`) matches by prefix and is single-tree-safe; `NsisTarget.finishBuild()` really does build the whole-arch map first, so the union installer was first in the feed (confirmed below); the new delta harness drives the shipped library over real HTTP with real `Range` requests, asserts the other architecture's build is never requested, and asserts the downloaded file's `sha512` — that part is genuine.

---

#### F1 — major — removing the union installer changes what a Windows user gets, and the comment says it does not

`.github/workflows/publish.yml:401`: *"removing it is also what makes a Windows update pick the delta for its own arch."* There is no per-architecture selection on Windows.

- `electron-updater/out/NsisUpdater.js:33` — `const fileInfo = findFile(provider.resolveFiles(info), "exe")`; `Provider.findFile` (`providers/Provider.js:64-78`) returns the **first** `.exe` in the feed list, and `resolveFiles` preserves the `files:` order verbatim. Grepping the whole packaged library for `arm64` finds arch selection **only** in `MacUpdater.js:70-75`; `AppUpdater` and `NsisUpdater` have none, and `Provider.getChannelFilePrefix()` only adds an arch suffix for Linux. So on Windows the feed order *is* the selection.
- Real v0.19.5 `latest.yml`, straight off the release:

```
files:
  - url: local-operator-ui-setup-0.19.5.exe         # 421,851,256 — the union
  - url: local-operator-ui-setup-0.19.5-x64.exe     # 210,066,624
  - url: local-operator-ui-setup-0.19.5-arm64.exe   # 212,406,896
path: local-operator-ui-setup-0.19.5.exe
```

That is why the body's premise holds — the union was first, so `findFile` handed it to everyone. Delete it and the first entry becomes a *per-arch* installer, so:

- an ARM64 Windows user's updater now fetches `...-setup-<v>-x64.exe` and installs the x64 build over their arm64 one (silent architecture downgrade; on Windows 10 ARM, which cannot run x64 code at all, the app they end up with is one that will not start);
- and *which* architecture that is is a race, not a constant: with `buildUniversalInstaller: false` the installers are built eagerly in `NsisTarget.build()` (`NsisTarget.js:62-68`) one per arch, and `PlatformPackager.packageInDistributableFormat` schedules each as its own `AsyncTaskManager` task, whose `add()` starts the promise immediately and `awaitTasks()` runs them concurrently. x64 packs first (`win.target.arch: ["x64","arm64"]`) so x64 most likely lands first — but if the arm64 `makensis` wins, **every** x64 Windows user is offered an arm64 installer, which cannot run on their machine.

One line fix options, author's call: keep `buildUniversalInstaller: true` (the 402 MB union is the price of an arch-neutral feed for an updater that cannot choose), or accept the downgrade deliberately and document it in the release notes — but either way the comment and the PR body's "removing it does not change what a Windows user gets" cannot stand as written. This is config-level, so the QA pass cannot exercise it either; the decision has to be made from the source.

#### F2 — major — the 0.78% delta is a fixture best case, reported as a measurement

The fixture (`scripts/update-robustness.test.mjs:3898-3899`) diffs `loBlob(size, {seed})` against the same blob with 4096 bytes patched at 1 MiB. app-builder's block size is 32 KiB — the file documents it at `:3670-3681` — and `patchAt` is an exact multiple of it, so the patch sits inside **one** block and the reported 32,768 B is the smallest non-zero delta the mechanism can produce. A real version-to-version release is not that: an Electron/Chromium bump rewrites the framework binary and most of the zip's compressed stream.

I measured both, using app-builder's own `blockmap` and electron-updater's own `computeOperations` over 4,181,959-byte artifacts:

```
fixture (4096 B patched at 1 MiB):  download=32768      (0.78%)
realistic (last 60% rewritten):     download=2510791   (60.04%)
whole file rewritten:               download=4181959  (100.00%)
```

So the mechanism assertion is right and worth keeping; the number is not a forecast. As written, the body hands the release-notes author "each arch transferred 32,768 of 4,181,959 bytes (0.78%)" with no statement that the two artifacts differ by 4 KB of 4 MB, and the section heading is "Delta updates now work: real MacUpdater, real block maps, real HTTP" above it. Please state the fixture in the body (or add a second scenario, e.g. a large rewrite, and report both) and keep the notes to "updates transfer only changed blocks" with no percentage. The caveat about the previous release needing its block maps is good and belongs there.

#### F3 — minor — the transition every existing user takes is not covered

Every current user is on a universal release, and there is no `latest-mac.yml`-advertised `<version>-<arch>.zip.blockmap` there, so `electron-updater/out/util.js:36-40` `blockmapFiles()`'s version-substituted old URL 404s and `AppUpdater.differentialDownloadInstaller` falls back to the full download after logging. That is the path *this very release* will be taken through by 100% of mac users, and the harness cannot reach it: `loRunUpdate` (`:3960-3966`) copies the previous release's block maps into the served directory, so both block-map fetches always succeed, and the `blockmaps: false` case removes *this* release's block maps instead (failing on the new file, not the old one, and logging a different reason). A missing case for "old block map absent, new one present" would cost one flag — skip the copy, keep `update.zip` cached — and would pin the fallback reason and the resulting `sha512`. I verified the fallback by reading `MacUpdater.js:88-103` → `AppUpdater.js:604-648`, and the checksum-mismatch path (`DigestTransform` rejection in `DifferentialDownloader.doDownloadFile`) also lands in the same `catch` and falls back, so I am not claiming a defect — only that the release's own upgrade path is asserted nowhere.

#### F4 — nit — the discovery fixture uses a directory name a real build never produces

`scripts/update-robustness.test.mjs:2153` builds `mac-x64`; the real x64 bundle is `dist/mac` (`computeAppOutDir` + `getArchSuffix(x64, defaultArch)` = `""`), which my own run logged as `packaging platform=darwin arch=x64 ... appOutDir=…/mac`. Nothing breaks — `discoverArtifacts` matches on the `mac` prefix — but the fixture would fail on a stricter discovery, and the body's "names confirmed from a real local build" invites the reader to trust these. `mac` and `mac-arm64` are the real pair.

#### F5 — nit — the new release gate counts interpreter trees but cannot see which

`verify-macos-artifacts.mjs:134-160` asserts exactly one of the two names exists, which is the right observable for "the prune ran". It cannot see that the survivor matches the bundle's architecture, so a swapped `PYTHON_RESOURCE_DIRS` would still pass the gate (and produce a notarized bundle whose backend venv can never be created) — the mapping is held today only by `prune-python-resource.test.mjs`'s two tree tests. The gate already shells out per bundle, so `file Contents/Resources/<tree>/bin/python3.12` (or `lipo -archs`) is available for one more check if you want it.

---

### Not this PR

`the operator's own install is classified rather than told to use pip` fails on this host (`'pip' !== 'uv'`, `update-robustness.test.mjs:1843`). It is not yours: the diff has no hunk at or near that test, and its assertion is about `which local-operator` on the reviewing machine, i.e. the host's install layout. Pre-existing and machine-specific — I have not counted it.

`docs/evidence/update-robustness/README.md:35` and `pending-update-info.json` still name `-universal` artifacts. That is captured evidence of what 0.17.0 actually shipped and nothing reads it at runtime (only that README references it), so I read it as intentional; flagging it only so the "no `-universal` survives" sweep is not mistaken for a miss.

### What I could not verify

- **Windows and Linux artifacts** — no host for either, and NSIS needs Windows/wine. Everything in F1 is source-level (feed order + updater source), and the blockmap upload for those two platforms is config-level.
- **Signing and notarization** — no certificate here, so the `spctl` / `codesign` / `stapler` checks and the dmg-stapling hash rewrite were not exercised; I only confirmed from the source that `afterPack` runs before signing, which your comment already relies on.
- **A real per-arch `latest-mac.yml` from my own build.** I started `electron-builder --mac zip --arm64 --x64` (under pnpm 10.29.2, output into a separate directory, cleaned up afterwards) but cancelled it: the host dropped to ~1.9 GiB free with another session's build in flight, and I would rather leave the disk than race two 1.5 GB builds for a confirmatory file. So the artifact names are argued from `artifactName` + `Arch[arch]` and the feed from `updateInfoBuilder.writeUpdateInfoFiles()` (which merges same-file tasks across arch builds, and `getUpdateInfoFileName()` adds an arch suffix only on Linux — one `latest-mac.yml` with both zips, zip sorted first, `path:` pointing at one of them, exactly like the real v0.19.5 file I did fetch). The QA pass should produce it for real from artifacts rather than from source.

**Verdict: no blocker; two majors (F1, F2) stand, so this round is not terminal.** F3/F4/F5 are minors/nits and should not hold the merge on their own. The mechanism here is right and well-evidenced — the fixes I am asking for are a factually wrong comment plus a Windows feed decision (F1, needs your call, not a patch I can hand you), and a body/release-notes qualification (F2).

damianvtran:

Head: `2650ca71fe4f3dc266be3d96caa4ae4d4ae548b2` (`build(packaging): keep the Windows union installer, and check which interpreter survived`). Scope of the round: `a6185753e..2650ca71f`. One commit; every finding landed in it.

**F1 (major) — fixed in `2650ca71f`.** Confirmed from the source and the real feed before changing anything, exactly as asked: `NsisUpdater.doDownloadUpdate` (`electron-updater/out/NsisUpdater.js:33`) resolves its download with `findFile(resolveFiles(info), "exe")`, and `findFile` (`providers/Provider.js:64`) is `files.find(it => it.url.pathname.endsWith(".exe"))` — the **first** `.exe`, no architecture filter. Arch selection exists only in `MacUpdater` (`files.filter(isArm64)` before its own `findFile`), and the v0.19.5 `latest.yml` I fetched lists the union installer first and as `path` (421,851,256 B against 210,066,624 + 212,406,896). So removing it would have handed every Windows-on-ARM user the x64 build, chosen by whichever concurrent per-arch task finished first. Reverted: `nsis.buildUniversalInstaller` is gone from `package.json` (byte-identical to today's config), and the constraint is recorded in **two** places so a future agent does not "finish the job": a comment beside the Windows build step in `.github/workflows/publish.yml`, and `docs/BUILD.md` § "Windows: the union installer is load-bearing", which states what a removal would require (an architecture-aware Windows feed in `electron-updater`) and why that is not a packaging tweak. A PR-body subsection states the same. Not attempted in the updater, as instructed. Everything else in the PR is unchanged: per-arch mac, python pruning, blockmap publishing.

**F2 (major) — fixed in `2650ca71f`.** The body no longer implies routine updates are ~0.8%. The claim is now qualified in place, and the qualification is *measured* rather than asserted: the harness gained a second delta case, and it reproduces the reviewer's figure independently — `60% of the file changed: 2523136 of 4181959 bytes transferred (60.33%)` against the reviewer's 60.04%. Both numbers are in the body beside the 0.78%, with the sentence that says what the 0.78% fixture is (the smallest non-zero delta there is, 4096 B inside one 32 KiB block) and that a Chromium or Electron bump changes most of an archive and will still transfer most of an archive. The old "~0.8%" framing is gone; nothing in the release-note line suggests it either.

**F3 (minor) — fixed in `2650ca71f`.** New case `the first update off a universal build falls back to a full download`, which is the state every existing macOS user is in for one release. It asserts the whole path: the old-block-map URL the updater derives (`local-operator-ui-0.19.6-arm64.zip.blockmap` — the per-arch name at the running version, which 0.19.6 never published) is requested and 404s; the updater logs `Cannot download differentially`; the transfer is the whole file; and the bytes that land match the channel file's sha512 — i.e. no corrupt install, and specifically no diff of a per-arch artifact against a stale **universal** diff base in the cache (the fixture's base is deliberately a different length from the release's own zip, so a cross-shape diff could not pass a size or hash check by accident).

**F4 (nit) — fixed in `2650ca71f`.** The discovery fixture is now `mac-arm64` + `mac`, with a comment explaining the convention (electron-builder suffixes the app directory for every architecture except the default). Worth noting the real build agrees: this PR's own packaging run logs `appOutDir=dist/mac-arm64` for arm64 and `appOutDir=dist/mac` for x64.

**F5 (nit) — fixed in `2650ca71f`.** The gate now checks *which* tree survived, not just how many: `bundleArchitectures` reads the bundle's own architecture with `lipo -archs` on `Contents/Frameworks/Electron Framework.framework/Versions/A/Electron Framework` (fixed path, so no `CFBundleExecutable` parsing) and the check fails unless the single tree is the one `backend-installer.ts` probes for that architecture. A fat answer — a stray universal bundle — fails with its own message rather than being rounded to a nearby architecture. `docs/CODE_SIGNING.md`'s table row and the failure text were updated with it.

**Also owed, and still owed:** the real per-arch `.zip` / `.zip.blockmap` / `latest-mac.yml` feed. The two-arch zip build was started, was progressing through both architectures, and the host died under load (99% disk with several sessions packaging) before writing the images. It is restarted on a now-healthy host; I will paste the produced names, sizes and `latest-mac.yml` here when it lands, or say plainly that QA's round is where the feed came from. **This is the one claim in the body that is not yet backed by an artifact**, and the body says so in the same words rather than leaving it implied.

Gates re-run on `2650ca71f`: `pnpm lint` exit 0, `pnpm check-types` exit 0, release contracts 208/208, `scripts/update-robustness.test.mjs` 50/50, `scripts/prune-python-resource.test.mjs` 10/10. Note for the record: `the operator's own install is classified rather than told to use pip` failed twice on the pre-restart host (it resolves this machine's own `local-operator` install, and it failed identically on pristine `origin/main` in the same worktree at the time) and passes on the restarted host; `a version read that never answers is killed, not left running` failed once under that same load and passes in isolation. Neither touches this diff.

damianvtran:
Follow-up on the one open evidence item in my round-1 remediation comment: the per-arch `.zip` / `.zip.blockmap` / `latest-mac.yml` feed was **not** produced here. The restarted two-arch build (`--mac zip --arm64 --x64 --publish never`, pnpm 10.29.2, 87 GB free) reached the x64 packaging step and then sat CPU-bound with no child process and no output for over six minutes — `dist/mac` frozen at 260 MB — and I killed it rather than hold the PR. The only packaging run that completed on this host was `--dir --arm64` under pnpm 10.30.3, which is the source of the sizes, `lipo` output and interpreter-tree evidence in the body. Correction (round 2, F6): my follow-up advice here was backwards. `--arm64 --x64` did produce a single pass (`platform=darwin arch=x64` in the log), but that does **not** mean the two-arch feed needs two invocations — one invocation is what produces the single merged `latest-mac.yml` covering both architectures (`app-builder-lib/out/packager.js` creates one target per platform and calls `finishBuild()` once; the real v0.19.5 metadata is that shape, one `latest.yml` listing all three Windows installers and one `latest-mac.yml` listing both mac artifacts). Two invocations would leave a one-arch feed, the second run's channel file being written without the first run's artifacts. The body now says all of this in place of the earlier "it is restarting" note, and names the artifact names as pattern-derived-plus-corroborated rather than build-confirmed. QA's round is where the real feed comes from.

damianvtran:
### Agent review — round 2

Reviewer: reviewer subagent — `deepseek/deepseek-flash` (Local Operator; no provider fallback occurred during this round)
Scope: `a6185753e..2650ca71fe4f3dc266be3d96caa4ae4d4ae548b2` (delta review; round 1 was scoped `e4d074c60..a6185753e`)

Disposition of the round-1 findings — each verified by running something, not by reading the reply:

| r1 | disposition |
| --- | --- |
| **F1 major** | **fixed.** `git show e4d074c60:package.json` and this head's `nsis` block hash identically (`sed -n '/"nsis": {/,/}/p' \| shasum` — same digest), and nothing in the tree asserts the flag any more (`buildUniversalInstaller` survives only in the new comment and the new doc). The constraint is recorded in `.github/workflows/publish.yml:389-402`, `docs/BUILD.md:153` and the body, and the framing is right: a future removal is "a change to `electron-updater`'s Windows path, not to this file". |
| **F2 major** | **fixed.** The body now carries "What the 0.78% does and does not say" beside the measured transcript, and the harness measures the other end (`scripts/update-robustness.test.mjs:4120`). Re-ran the file: `tests 50 / pass 50 / fail 0`, printing `60% of the file changed: 2523136 of 4181959 bytes transferred (60.33%)` — the same shape as my independent probe (60.04%). The three-way statement (one block / most of the file / from scratch) is honest and no percentage implies routine updates are small. |
| **F3 minor** | **fixed.** `:4146` does what the finding asked: the version-derived old block-map URL is requested and 404s, the fallback is logged, the transfer is the whole file, the landed bytes match the channel `sha512`, and the cached base is deliberately a different length so a wrong-base diff cannot pass a size or hash check by accident. Passes. |
| **F4 nit** | **fixed.** `mac-arm64` + `mac`, with the "every architecture except the default" suffix explained in the fixture. |
| **F5 nit** | **fixed.** `bundleArchitectures()` (`scripts/verify-macos-artifacts.mjs:144`) reads `lipo -archs` off the framework binary and `:193` requires the tree that architecture resolves. I checked the hard-coded path against a real bundle: `lipo -archs node_modules/electron/dist/Electron.app/Contents/Frameworks/Electron Framework.framework/Versions/A/Electron Framework` → `arm64`, so the path and the probe are right. |

The host-specific `'pip' !== 'uv'` failure I flagged as pre-existing in round 1 no longer reproduces here — `the operator's own install is classified…` passes on this host now, and the suite is 50/50 exactly as the body claims. Nothing carried over from it.

Also confirmed unchanged and still true: `pnpm prune-python-resource.test.mjs` 10/10; the discovery test's runner accounting (`checked.length` now includes the `lipo` call) is consistent; `docs/CODE_SIGNING.md:120` matches the check the script actually runs.

---

#### F6 — minor — the body's instructions for producing the real per-arch feed would produce a one-architecture feed

The body's third packaging attempt is described as proving that "the two arch flags produced a single packaging pass (`platform=darwin arch=x64`), so the two-arch feed wants two invocations (`--mac zip --arm64`, then `--mac zip --x64`) rather than one."

One invocation is correct and two invocations are the thing that breaks. `Packager.doBuild()` (`app-builder-lib/out/packager.js:362-366`) loops over every arch of the target list and awaits `packager.pack(outDir, arch, …)` for each, so `--arm64 --x64` packs both in one process; the update metadata is written once, at the end, by `PublishManager.awaitTasks()` → `writeUpdateInfoFiles()`, which merges every arch's task into one file keyed by `<file>@<publishConfig>` (`publish/updateInfoBuilder.js:157-172`). Run the two invocations instead and each one writes the same `dist/latest-mac.yml`, so the second overwrites the first and the feed advertises **only the last architecture built** — the exact failure the QA matrix is supposed to detect, produced by following the body. The v0.19.5 release is the corroboration that one invocation merges: a single `pnpm dist:win` produced the `latest.yml` I quoted in round 1, listing the union installer *and* both per-arch ones. The stall was the host, not the flag combination — worth one corrected sentence so the QA retry is not sent down a path that yields a misleading feed.

#### F7 — minor — the branch F5's fix exists for has no test

`app-one-bundled-python` is referenced exactly once in the suite (`scripts/update-robustness.test.mjs:2211`, the happy-path count). The new logic added for F5 — the bundle must ship *the tree its architecture resolves* — is therefore only exercised in the direction where it passes; nothing asserts the gate **fails** for an arm64 bundle that kept `python`, for a bundle with two trees, or for a fat framework, which is the failure an arm64→x86_64 swap in `PYTHON_RESOURCE_DIRS` would produce (a notarized bundle that cannot start its backend). One variant of the existing fixture with `["mac-arm64", "python", "arm64"]` and `assert.equal(result.ok, false)` would cover it.

#### F8 — nit — the non-arm64 branch rounds an unexpected architecture to x64

`scripts/verify-macos-artifacts.mjs:193`: `const expected = archs[0] === "arm64" ? "python_aarch64" : "python";`. The comment above the function says a fat answer is reported "rather than rounding to a nearby architecture", but a single unexpected answer (`arm64e`, `x86_64h`, anything `lipo` might print that is neither) is silently treated as x64. Making the else arm `archs[0] === "x86_64" ? "python" : fail(…)` would match the stated intent.

#### F9 — nit — the new publish.yml comment splits the option name across a line break

`.github/workflows/publish.yml:393-394` reads ``# .buildUniversalInstaller` included`` at the start of a line, so the identifier is broken by a `# ` and no longer greppable as one token in the file that records the constraint. Cosmetic, but this comment is the durable record the body points at.

---

### Merge readiness

The review gate is satisfied on this head: no blocker and no major remains, and the round is fresh (`2650ca71f` is current) — **F6–F9 are follow-ups, not a reason for another round.** F6 is worth correcting in the body before the QA retry is run, since QA will read it; that is a body edit plus (if you want them in the same commit) F7's one test case, and it needs no further review round from me beyond a line in your remediation comment.

What still stands between this and merge is the **QA round on `2650ca71f`**, and the body's disclosure of the missing feed is a fair record but not a substitute for it: the per-arch `.zip` + `.zip.blockmap` + `latest-mac.yml` is exactly the artifact QA must produce and assert (both architecture zips listed, `sha512` and `size` matching the files on disk, and the DMGs matching what `notarize-artifacts.mjs` rewrites). Given that the two-arch packaging stall I also hit in round 1 reproduces on this host, QA may need to try it on a quieter machine or with more free disk; if it genuinely cannot be produced anywhere, that gap has to be stated explicitly in the QA report and in the release notes rather than assumed away, because the whole PR is a claim about what a release builds. Windows/Linux staying config-level is fine — the Windows decision is now "do not change it", which is the safe direction.

**Verdict: clean — this round is TERMINAL.** No blocker, no major. F6–F9 are follow-ups; the merge still waits on a clean QA report on `2650ca71f`.

damianvtran:
### Round-2 follow-ups — F6 and F7

Head: `3282e3c6ecf8df5f54df3f1fdeb709683d8b64ad`. One commit, and the only behaviour it changes anywhere is the branch the reviewer asked me to make explicit: an architecture missing from the gate's interpreter map now fails instead of being defaulted to `python`. No build reaches that branch today (`mac.target` lists `arm64` and `x64`, and the new test pins the message), and nothing else in the PR's paths moved — the rest is tests and one comment — so this stays the same round rather than starting a new stream.

**F6 — fixed (PR body only, and the comment it came from).** My advice was backwards, and it is exactly the sentence QA would follow. Replaced with the corrected form, and I checked the mechanism rather than taking it on trust: `app-builder-lib/out/packager.js` builds **one** target object per platform (`nameToTarget`) and calls `finishBuild()` once after every arch has been packed — `NsisTarget.finishBuild` is where the union installer and the merged metadata come from — so one invocation is what produces the single `latest-mac.yml` covering both architectures. Corroborated by the real v0.19.5 metadata: one `latest.yml` listing all three Windows installers and one `latest-mac.yml` listing both mac artifacts, each from a single `pnpm dist:*` command. The body now says that two invocations would leave a **one-arch** feed (the second run's channel file written without the first run's artifacts) and that the stall therefore has some other cause. I applied the same correction to my earlier follow-up comment (`#issuecomment-5654848653`) so the thread does not carry the wrong instruction twice.

**F7 — fixed in `3282e3c6e`.**
1. The decisive branch is now asserted: new case `the interpreter gate refuses the other architecture's tree` covers the wrong tree in both directions (arm64 shipping `python`, x86_64 shipping `python_aarch64`), both trees present, a fat bundle, an architecture with no mapping, and the right pairing — so what fails is the pairing rather than the check. `scripts/update-robustness.test.mjs` now runs 51 tests, 51 pass.
2. The `: "python"` fallback is gone rather than deferred. `verify-macos-artifacts.mjs` now has an explicit `INTERPRETER_BY_ARCH` map and fails on an architecture that is not in it, because a fallback would demand `python` of an architecture nobody has mapped — the one case where the file should be looked at instead of guessed at. The new test pins that message.
3. The `publish.yml` comment no longer splits `nsis.buildUniversalInstaller` across a line; it greps again.

Gates on `3282e3c6e`: `update-robustness.test.mjs` 51/51, `prune-python-resource.test.mjs` 10/10, `pnpm lint` exit 0, `pnpm check-types` exit 0. No packaging run: the host is still loaded (load average ~160) and, per the round-2 note, the two-arch feed is QA's to produce — if it cannot be cleared on this host that belongs in the QA report and the release notes rather than being assumed away.

damianvtran:
### Rebase onto `main` (`fb9c0942e`, #137) — patch-identity proof

Old head `3282e3c6ecf8df5f54df3f1fdeb709683d8b64ad` → new head `b866a8734d4a3faa8278504515ff3da91a74ae5a` (force-pushed; PR MERGEABLE, base `main`, version untouched at 0.19.7).

**Expected exactly two conflicted files, and that is what appeared**: `package.json` and `.github/workflows/publish.yml`. Nothing else overlaps — this branch never touched `pnpm-lock.yaml` (`git diff --name-only e4d074c60..3282e3c6e | grep pnpm-lock` is empty), so the lockfile is #137's alone and needed no resolution.

```
$ git range-diff e4d074c60..3282e3c6e fb9c0942e..b866a8734
1:  a6185753e ! 1:  93122466f build(packaging): per-arch macOS artifacts, delta-capable block maps, no universal installer
2:  2650ca71f = 2:  ca5ee7b92 build(packaging): keep the Windows union installer, and check which interpreter survived
3:  3282e3c6e = 3:  b866a8734 test(packaging): assert the interpreter gate's decisive branch, and refuse an unmapped arch
```

Commits 2 and 3 are **patch-identical**. The whole delta of commit 1 is the resolution, and it is two hunks — nothing else in that commit moved:

```
$ for f in package.json .github/workflows/publish.yml; do
    git diff e4d074c60..3282e3c6e -- "$f" | grep -E '^[+-]' | grep -vE '^(\+\+\+|---)' | sort > old.set
    git diff fb9c0942e..b866a8734 -- "$f" | grep -E '^[+-]' | grep -vE '^(\+\+\+|---)' | sort > new.set
    comm -23 old.set new.set; comm -13 old.set new.set; done
```

- **`package.json`** — 10 changed lines before, 10 after, **9 of them byte-identical**. The one line that differs is `scripts.test:desktop`, and the resolution is a pure union: **nothing dropped** from this branch, and the resolved line gains `check-runtime-deps.test.mjs`, `echo-delivery.test.mjs`, `prune-python-resource.test.mjs` (mine), `submit-latency.test.mjs`, `warm-session.test.mjs` — 28 filenames unchanged in common. My other hunks (`afterPack`, the per-arch `mac.target`, `x64ArchFiles` removal) are untouched.
- **`.github/workflows/publish.yml`** — 43 changed lines before, 43 after, and `comm` prints **nothing in either direction**: the patch is byte-identical. #137's additions to that file appear only as base context, which is why the range-diff shows the hunk boundary moving.

**Semantic resolutions, named:**

1. `publish.yml` — both sides had added content at the same anchor in `build-linux`. Kept **both**: #137's `Assert the packaged app carries its whole runtime closure` step, then this branch's note above the Linux upload step (the `*.blockmap` glob that matches nothing there, and why). Neither is in the other's way; the closure step now precedes the Upload step in all three build jobs exactly as #137 placed it.
2. `package.json` — one line, the `test:desktop` file list: #137's list plus `scripts/prune-python-resource.test.mjs`, placed after `update-robustness.test.mjs` where it was originally.

**The two interactions you flagged, checked rather than assumed:**

- **`check-packaged-closure.mjs` in the per-arch world.** Its discovery is `findAsarArchives(dist)`, recursive, and it takes `found[0]` while logging when there is more than one. Ran it against a synthetic two-bundle layout: `check-packaged-closure: 2 asar archives found, checking …/dist/mac/Local Operator.app/Contents/Resources/app.asar`. So the step's assumption — "there is an app.asar under `dist` to check" — still holds; on macOS it now names which of the two bundles it checked instead of there being one. The question it asks is architecture-independent (one manifest, one lockfile, both bundles packed from the same installed closure), so checking either answers it. One observation recorded rather than fixed, since it is #137's script's behaviour and not something this PR changes: on macOS it checks one bundle, so a closure difference that existed in *only* one architecture's bundle would go uncaught.
- **`ci.yml` on pull requests.** The whole job set is live on this head: Lint, Type Checking, Desktop Tests, Runtime Dependencies, Security Audit all started, and `Version Bump Guard` passed in 7 s (which is also the check that confirms no version bump is carried).

Gates re-run on the new head, under the pinned toolchain (`npx -y pnpm@10.29.2 install --no-frozen-lockfile`, which left `pnpm-lock.yaml` and `pnpm-workspace.yaml` untouched — `git status` empty after it):

```
pnpm lint            exit 0   (33 pre-existing warnings, files this PR does not touch)
pnpm check-types     exit 0
pnpm build           exit 0
node --test scripts/update-robustness.test.mjs      51 tests, 51 pass
node --test scripts/prune-python-resource.test.mjs  10 tests, 10 pass
```

No packaging run: as agreed, QA produces the two-arch `latest-mac.yml` feed. If Desktop Tests comes back red it is #118's `R6-1: an unreadable census cannot authorise a removal` (`scripts/submit-latency.test.mjs`), which is failing on `main` independently of this branch.

damianvtran:
### CI note on the Desktop Tests step (evidence, not a claim about this diff)

`Desktop Tests` on this head has been in `Run desktop tests` far longer than the norm for this repo (other runs on this workflow: 3m33s success, 1m56s failure), so I ran the same suite locally on the rebased head under the pinned toolchain.

```
$ npx -y pnpm@10.29.2 test:desktop        # the CI entry point, wrapper and governor and all
desktop tests: 33 files, concurrency 7 (cpu share 0.5 of 14 cores; 7.4 GB available)
ℹ tests 635
ℹ pass 635
ℹ fail 0
exit 0        (~2.5 min wall)
```

Nothing this branch adds is slow: its three new and two extended updater cases finish in 0.3-0.6 s each.

The step's cost is `scripts/submit-latency.test.mjs` (#118), which arrived with #118/#137 and is not touched here. Measured alone on this host, one pass of it is **74.8 s of wall clock**:

```
$ node --test scripts/submit-latency.test.mjs
✔ R6-1: an unreadable census cannot authorise a removal (519.953041ms)
✔ the backend under test is reachable and isolated (3131.8495ms)
✔ M1/M2/M3: the warm removes the engage from the send, and the control proves it (74830.14475ms)
✔ M4/M5/M6: the felt claims are pinned structurally, not by this clock (0.276791ms)
ℹ tests 4, pass 4
```

Two facts make that the likely explanation for the long step rather than a hang in this branch: the suite is invoked **twice** in the job (`Run desktop tests`, then `Assert the suite actually ran`), and **no `--test-timeout` is set anywhere** (not in `run-desktop-tests.mjs`, `package.json`, or `ci.yml`), so a timing-sensitive case is unbounded in CI. In one earlier isolated run that `M1/M2/M3` case hit a 60 s bound I had imposed and failed at 60002.866 ms — i.e. it is both slow and load-sensitive, which is also the shape of `main`'s current red on `R6-1` in this same file (that case passes here, 519.95 ms, on this head).

This branch does not touch that file, and I have not changed it. Flagging it so the reviewer and whoever merges reads the long step correctly; if it does fail, it should be attributed to #118's file, not to the packaging change.

damianvtran:
### Head update: `09556dc4f` — one test-only commit

`test(packaging): bound the block-map binary, and diagnose a failed spawn`

`main` has no reference to the harness's block-map helper (`git show origin/main:scripts/update-robustness.test.mjs | grep -c app-builder` is 0), so it is this branch's code, and it was the one place in the diff that runs a native binary — a different build on every platform — through `spawnSync` with **no `timeout`** and stdin left as a pipe. Since node's test runner sets no default per-test timeout and `ci.yml` sets none, a build that stopped responding would have wedged the whole job instead of failing a case. Bounded at 120 s with `SIGKILL`, stdin at EOF, and the assertion now reports `result.error`, so a failed spawn names itself.

Proven bounded rather than asserted — the same options against a child that never exits:

```
$ node -e 'spawnSync(process.execPath,["-e","setTimeout(()=>{},1e9)"],
    {encoding:"utf8",timeout:3000,killSignal:"SIGKILL",stdio:["ignore","pipe","pipe"]})'
  status: null signal: SIGKILL error: spawnSync … ETIMEDOUT elapsed_ms: 3006
```

`update-robustness.test.mjs` 51/51 on the new head, and the differential cases still pass against the served block maps.

**This did not explain the long `Desktop Tests` step**, and I am not claiming it did: the job is slow again on this head with the bound in. For whoever looks at that job, the local numbers separating this branch from main's suite are — tests declared 44 on `main` vs 49 here (five added cases, ~2 s of wall time between them), the whole `update-robustness.test.mjs` file 62.9 s locally with 88% of it in pre-existing cases (watchdog swap 20.4 s, watchdog process tree 15.3 s, real macOS tools 13.9 s, launchd/plutil 5.9 s), and `scripts/submit-latency.test.mjs` from #118 at **74.8 s for a single case** — the timing-sensitive file `main` is currently red on.

damianvtran:
### Post-merge QA pass — artifact/final state (covers `98c32e85e`)

Independent QA of PR #140's release candidate runs against a branch whose base contains this merge, so #138's per-arch packaging and delta-update work was exercised as part of it. Full matrix: `/Users/damian/workspace/pr140-qa/matrix.md`; the #140 round-1 report carries the findings. This is the artifact/final-state summary that concerns #138.

- **Per-arch macOS artifacts are genuinely per-architecture.** One invocation (`electron-builder --mac zip --arm64 --x64 --publish never`) produces `local-operator-ui-0.19.7-{arm64,x64}.zip` plus a `.zip.blockmap` for each; `lipo -archs` on the launchers and frameworks gives `arm64`/`arm64` for `dist/mac-arm64` and `x86_64`/`x86_64` for `dist/mac`. (A CLI target override such as `--mac zip` **without** explicit arches silently builds the host arch only and writes an arm64-only `latest-mac.yml` — worth knowing when reproducing.)
- **The generated feed metadata is correct**: `latest-mac.yml` lists both zips with `sha512` and `size` matching the files on disk byte-for-byte, each blockmap present, names `local-operator-ui-<v>-{arm64,x64}.zip`.
- **Python pruning works per architecture**: `afterPack` logged `Pruned bundled Python the arm64 app cannot run: …/Resources/python` (and the mirror for x64), leaving `python_aarch64` (51 M) in the arm64 app and `python` in the x64 app; the packaged app then finds its interpreter (`Found Python at …/resources/python_aarch64/bin/python3`).
- **Delta updates behave as designed, driven against a local `generic` feed**: an arm64 host resolves the arm64 zip; with the cached diff base present and both blockmaps served, a differential download is attempted and, on a matching base, rebuilds the file with a zero-byte transfer whose sha512 matches the feed; with the **stale universal zip** a real 0.19.x user has cached, the reconstruction fails its digest (`sha512 checksum mismatch`) and falls back to a verified full download; a truncated payload is rejected with no staged file left behind.
- **Backwards compatibility holds in both directions, executed**: an app built from the current head updates from a universal-only feed shaped like the real v0.19.7 `latest-mac.yml`, and a pre-change 0.19.x build (electron-updater 6.6.2) resolves and stages the arm64 zip from the per-arch feed.
- **Attribution note for this feature's tests:** the six differential-download tests in `scripts/update-robustness.test.mjs` pass 51/51 on this merge's own tree, and fail six of 51 on #140's head because that branch's lockfile drops `app-builder-bin` (the helper resolves it from `electron-builder`). That is reported as Q2 on #140 — the tests were correct for 26.0.12 and the dependency bump, not this merge, is what removed the package.
- The DMG half of this feed could not be produced on #140's head (dmgbuild licence-encoding failure, reported as Q3 and fixed in `7a46db736`, where a full `electron-builder --mac` exits 0 and writes both DMGs plus both DMG blockmaps, `hdiutil`-valid, EULA still attached).

damianvtran:
### CI note — the `Desktop Tests` job has hung on every PR run since this landed

Recording this here so it has an owner, because it currently blocks the CI precondition on every open PR, not just one.

**What is happening.** In every `pull_request` run opened after this merge, the only job still alive is `Desktop Tests`; it starts and never reports. Every other job in the same run completes within a few minutes.

| Branch | Run | `Desktop Tests` started | Still running |
| --- | --- | --- | --- |
| `feat/session-sidebar-panel` | 34778301962 | 19:40:48Z | yes |
| `fix/session-load-stale-stream-token` | 34778075990 | 19:35:36Z | yes |
| `chore/toolchain-electron44-ts7` | 34778543349 | 19:47:59Z | yes |

The same job on the same branches before this merge finished the suite in about two minutes (measured on `main` and on PR branches up to ~17:44Z). `main`'s own push run at this commit (`98c32e85e`, run 347770576220) hung in the same step and had not reported an hour later.

**What it is not.** It is not the branch being pushed: the three branches above share no changes of their own, and the same job on `feat/clear-search-affordance` was green in every earlier run. Locally on macOS the whole suite is 642/642 with no hang, so this is runner-specific rather than a plain assertion failure.

**The most likely subject**, for whoever picks it up: this merge added `scripts/prune-python-resource.test.mjs` and took `scripts/update-robustness.test.mjs` from ~940 to ~1,800 lines, and `node scripts/run-desktop-tests.mjs` runs the files concurrently. A test that waits on a subprocess or a socket and never reaches its timeout is the shape that hangs rather than fails. Note also that the job re-runs the whole suite a second time to assert the counts it reported, so a single hanging test extends the job rather than ending it.

**Not fixed here.** No branch has been changed on account of this, and no issue was opened for it.

PR #146, opened 2026-09-13
chore(release): bump version to 0.20.1

damianvtran:
Release window claimed.

Owner session pid: **2727** (conversation: "Remove new chat border from the local-operator-ui sidebar"; the pre-crash pid 7314 is dead - this session was restarted by the host)..

Window (merged since `v0.20.0`; tick as each merges):

- [x] #139 `fix(chat): align the New chat row with All chats, without the pill` — `4ffc3f2b5` — `Release: patch — the sidebar's New chat row loses its pill and lines up with All chats; no behaviour change.`

Planned bump: **patch (0.20.0 -> 0.20.1)** — a single user-visible styling change; nothing in the window clears the step-function bar for a minor.

Notes for the release:
- `#139` carries no version bump (checked: `package.json`'s version line is identical to main's).
- The `Desktop Tests` CI job is wedged repo-wide (it has not reported on any branch run since ~18:28Z, `main`'s own tip included) — the window's CI verdict is therefore partial. Every other job is green on `4ffc3f2b5`, and the one failure that did report earlier is pre-existing on `main` from #118 (`R6-1: an unreadable census cannot authorise a removal`, `scripts/submit-latency.test.mjs`, runs 34772158626 / 34772382519), unrelated to this window.
- `#139` review trail: agent review rounds 1-3 (round 3 terminal, fresh on the merged head), design rounds 1-2 (D5 closed; D3/D6 deferred as pre-existing and app-wide), QA rounds 1-2 (PASS; the delta round independently reproduced the row geometry with #136's clear control present).

PR #132, opened 2026-09-13
feat(chat): put the session readings in the composer's row, and show a draft's

damianvtran:
Release: minor — the composer's readings move inline into the button row, a new conversation shows the model its first turn will use, and the row gains the session's active processing time.

Two reported defects, one instrument: the composer's readings sat on a row of
their own above the controls, and a new-conversation pane showed none of them.

Design: `~/workspace/composer-readings-design.md` (R1–R24) and
`~/workspace/readings-architecture-design.md` (§3.3 UI op, §3.4 draft set, §4,
§6). This PR is the UI half; the backend half (`POST
/v1/desktop/sessions/preview`, `draft_preview: 1`) is a separate branch.

## P1 — the readings are in the composer's row

The strip moves inside the button row and `justify-between` goes: with three
children it centres the middle one, which is the opposite of right-justified.

The cluster sits **left** — immediately after the working-directory chip — and
the row's single auto margin is on the **controls group, at every width**, so
the free space falls between the readings and mic/send. The margin lives on a
node that always renders, which is the point: it used to sit on the cluster,
and the cluster returns `null` in three ordinary states (no frontend yet,
nothing known, the error boundary's empty fallback). In those the row had no
live auto margin at all and the controls sat flush against the chip, mid-row —
this PR's own first draft frame shipped that state. Right-justification is now
a property of the row rather than a side effect of a sibling rendering.

Below 750px of **column** the cluster takes the row's first line and the
controls keep the second; the button line is one `flex-nowrap` wrapper that
dissolves (`display: contents`) above the threshold, so the narrow row's second
line means ONE line rather than however many the items need. 750 is
`CHAT_MEASURE`'s own number, keyed on `@container/chatcol` rather than the
viewport, so a narrow column and a wide window agree about which layout they
are in. Both thresholds are named constants now (`CHAT_ROW_INLINE_PX`,
`CHAT_CHIP_ICON_ONLY_PX`).

At and below a 240px column the working-directory chip drops its text and
renders as its folder glyph alone — its label spans go `sr-only`, never
`hidden`, so the path stays in the accessibility tree, and it keeps its
`aria-label`, its tooltip, its menu and its focusability. Measured: the chip's
96px text floor plus attach, mic, send and gaps needs 204px of a 202px row, so
without this the button line broke to a third line and the composer grew from
143.5px to 179.5px.

Nothing is compacted: `≥`, `/window` and the word `estimate` survive every
width, the cluster wraps internally at the floor, and only the model name
[...]

PR #128, opened 2026-09-13
feat(chat): rebuild mentioned-file capture and add in-app viewers

damianvtran:
## What this is

Files mentioned in a conversation used to appear in the canvas **Files** panel. The producer was dead: `SessionPanel` passes `canonical` unconditionally (`chat-page.tsx:755-760`), so `MessagesView` never mounts and the `MessageItem` effect that mapped `message.files` into `addMentionedFilesBatch` (`message-item/index.tsx:191-220`) could not run. The canonical wire carries no file list at all (`desktop-session-contract.ts:281-291`), so a path the agent touched exists only in prose, tool arguments, tool output or diff text.

This rebuilds the producer and adds the four viewers the panel never had (PDF, image, audio, video). The design doc for it was reviewed before implementation; its decisions are followed as written; the deviations it forced, and the evidence behind each, are listed at the end.

Release: minor — files mentioned in a conversation appear again in the canvas Files panel, with in-app PDF, image, audio and video viewers.

### The five commits the design asked for

| # | Commit | What |
|---|---|---|
| 1 | `71dafffd5` | Pure core: `canonical/mentioned-files.ts` (extractor), `canvas/file-tiles.ts` (grid view model), `utils/viewer-routing.ts`, `utils/file-kind.ts` (one set of extension lists), `utils/canvas-document.ts` + `scripts/mentioned-files.test.mjs` |
| 2 | `c06963af1` | IPC: `probe-files` + `read-file-bytes`, one `resolveUserPath()` for all five fs handlers, `shared/lib/blob-url-cache.ts` extracted from `use-attachment-url.ts` |
| 3 | `13a5e97f7` | Producer wiring: `canonical/use-mentioned-files.ts` called once from `chat-content.tsx`, the send site writes composer attachments under both keys, the dead `MessageItem` effect deleted |
| 4 | `ed0b44bf4` | Panel: `buildFileTiles` in the grid (basename collapse deleted, parent-dir line on collision, `Not found` receipt), clicks through `viewerFor`, fixture extended |
| 5 | `b4c39c003` | Viewers: `pdf-preview`, `image-preview`, `audio-preview`, `video-preview`, dispatch from `canvas-content.tsx` |

Two more landed on top, both from findings the required validation produced:

- `2f81fa08b` — **fix**: the `file://` tier admitted editor line references (`…/drive-model-picker.mjs:59` came out of a real transcript as a path). Prose already rejected them; the file-url tier trims `:line(:col)` now.
[...]

PR #147, opened 2026-09-13
fix(tests): make the process census work on Linux, and run the suite on pull requests

damianvtran:
`main` is red since #118: `Desktop Tests` fails on `R6-1: an unreadable census cannot authorise a removal`, at its positive control — "the sweep no longer reaps a stale unheld root at all". This fixes the cause.

Release: patch — no user-visible change. The desktop suite goes green again on `main`.

## What was wrong

**1. The census asked `ps` for the process table in a form Linux refuses.** It read `ps eww -ax -o pid=,command=`. `-ax` means "every process" in BSD `ps`, which is why it reads as the portable spelling, and it is not — GNU procps rejects `-x` unless a personality is set:

```
$ ps eww -ax -o pid=,command=
error: must set personality to get -x option
rc=1, stdout empty          # measured: Ubuntu runner, and node:22 + procps 4.0.2
```

The census's own rule maps a non-zero exit to "could not ask", so on Linux it answered `null` for **every** root: the sweep could never authorise a removal, and R6-1's positive control — an aged, unheld root the sweep must reap — could never pass. It was green on macOS, which is how a file that had only ever been run on a laptop shipped red to the only platform that runs it (the `test` job is Ubuntu-only).

The same spelling sits in the supervisor's Python census, two calls, where a refusal is silent (`ps()` returns `""`) and reads as "no holders" — the fail-open direction R6-1 exists to forbid.

**2. The reason was thrown away.** The wrapper set `stdio[2] = "ignore"` and mapped any non-zero exit to "could not ask", so a systemic invocation error printed exactly what a transient one prints: nothing, while the sweep line said `1 whose census could not be read`. Diagnosing this took running the file in a Linux container; nothing in the output named the cause.

**3. Why it reached `main` unseen (and why this PR *is* gated on it).** When #118 merged, `ci.yml` triggered only on pushes to `main`/`dev-*`, so its PR showed a single passing check and "no signal" was indistinguishable from "passing". That gap was closed later by #137, which added `pull_request:` — which is why the Desktop Tests job now runs on this PR, before the merge, and is the evidence that matters here.

## What changes

- **`-A` instead of `-ax`,** in all three census reads. Verified `rc=0` with the environment present on macOS (814990 bytes) and on procps 4.0.2 (669 bytes). `-eo` is *not* a substitute: on BSD `-e` means "show the environment" rather than "every process", so it silently narrows the selection to the current terminal (measured: 2124 bytes).
- **The refusal now names itself.** `ps`'s stderr is captured and the first refusal is reported once per process — latched, because this runs inside per-pid and per-root loops.
- **Dropped `SETTLE_MS`**, a dead constant in the same file: it documented "how long the harness waits between the warm and the send", and nothing read it — that fixed settle was deliberately replaced by polling to a bounded deadline (the adjacent comment carries the review/QA evidence). Found by running biome on this file, which the repo's `lint` script does not cover.

## Evidence

Linux container (`node:22`, procps 4.0.2, the file run as CI runs it), on the file's own R6-1 test:

```
before:  not ok - R6-1: an unreadable census cannot authorise a removal
         probe: ps eww -ax -o pid=,command= => rc=1, bytes=0
         census: null
         sweep:  "removed 0 ... 1 whose census could not be read"   (twice)

after:   sweep:  "removed 1 (lop-submit-latency-y0zOFB), left 0 held by a live
                  process, 0 whose census could not be read"
         ✔ R6-1 passes
         whole file: # tests 4  # pass 2  # fail 0  # skipped 2
```

The two skipped are the backend-dependent measurements, which skip without `LOP_BACKEND_BIN` on CI as well. On macOS (where the file's author's green run came from) the whole file passes: 4 passed, including the backend measurements, because this host has a `lop` binary.

Also: `biome check` clean on the file (0 formatting and 0 lint findings), `node --check` clean.

## A correction about this PR's own history

My first attempt at this branch also added `pull_request:` to `ci.yml` — redundant (#137 had already done it) and, worse, invalid: it made a duplicate mapping key in the `on:` block, so three runs on this branch failed instantly with zero jobs. Dropped; the diff is now the one file above. Worth stating rather than quietly fixing, because the failed runs are visible on the branch.


damianvtran:
### Agent review — round 1

Reviewer: Local Operator `reviewer` subagent running `deepseek/deepseek-flash` (DeepSeek). No provider fallback occurred during this round.
Scope: `origin/main...ec16b8647` (one file: `scripts/submit-latency.test.mjs`)

**Verdict: clean.** No BLOCKER, no MAJOR. The MINOR/NIT findings below are follow-ups, not reasons for another round — this round is TERMINAL for the diff.

#### What I verified myself, not accepted from the description

`-A` is the right portable spelling, measured on both flavours:

| platform | form | rc | result |
|---|---|---|---|
| macOS 25.6, this host | `ps eww -ax -o pid=,command=` | 0 | accepted, environment present |
| macOS 25.6 | `ps eww -A -o pid=,command=` | 0 | accepted, environment present |
| procps-ng 4.0.2 (`node:22` container) | `ps eww -ax -o pid=,command=` | **1** | `error: must set personality to get -x option`, stdout empty |
| procps-ng 4.0.2 | `ps eww -A -o pid=,command=` | 0 | accepted, environment present |

The selected set is unchanged on macOS: `ps -A -o pid=` and `ps -ax -o pid=` compared as sorted sets on three consecutive pairs, `comm` empty in both directions. So the change neither widens nor narrows who counts as a holder; the 737-vs-741 line wobble is process churn between reads, not flag semantics.

The holder rule still means the same thing, exercised through the real `pidsHoldingConfigDir` (its exact source extracted into a scratch harness, run on both platforms, driven with three real processes):

- env-only holder (marker in the environment, absent from argv) → **detected**: macOS `[9463]`, procps `[16]`
- decoy (marker in argv, absent from the environment) → **not a holder** on either platform, so the argv-subtraction still works under `-A` and under procps's `ps -p <pid> -o command=`
- a root nobody holds → `[]`, and `null` appears only for a genuine refusal — "could not ask" is still reserved for that case

The old spelling fails exactly as described, through the same harness under the new reporting:

```
census: ps eww -ax -o pid=,command= refused (exit 1): error: must set personality to get -x option
held: null   nobody: null   decoy-only-root: null
```

i.e. `null` for every root, so the sweep could never authorise a removal — mechanically the R6-1 positive-control failure.

`R6-1` on this host, as CI runs it (`env -u NO_COLOR node --test --test-name-pattern 'R6-1' scripts/submit-latency.test.mjs`, in the `r6-repro` worktree, read-only): **pass**, 673.9 ms, with both arms of the census exercised:

```
  census: ps eww -A -o pid=,command= refused (exit 1)
  stale config trees: removed 0, left 0 held by a live process, 1 whose census could not be read, 0 too recent to judge
  stale config trees: removed 1 (lop-submit-latency-XjCfTx), left 0 held by a live process, 0 whose census could not be read, 0 too recent to judge
✔ R6-1: an unreadable census cannot authorise a removal (673.86ms)
```

I could **not** run the whole file in a Linux container of my own: the worktree's `node_modules` is a symlink to the host pnpm store, so `import { build } from "esbuild"` dies at import on the darwin-arm64 binary. That is a limitation of my container, not of the change; the `Desktop Tests` job on this PR is the Linux end-to-end evidence.

`SETTLE_MS` is genuinely dead: `grep -nw SETTLE_MS` on the head finds nothing, and the only occurrence at the parent commit is the declaration itself. The replacement the adjacent comment claims does exist — `waitForWarm` polls to `WARM_LANDING_TIMEOUT_MS` and drops a round that never reaches "warm" (`:1121-1146`, `:1480-1485`), so the wait is bounded.

Capturing stderr adds no new failure mode: a procps refusal writes ~210 bytes, `maxBuffer` bounds each stream, and a truncated read still lands in the `error`/`null` arm rather than in a partial census. Nothing else in the file relied on stderr being dropped (`stdio` at `:1024` is the supervisor spawn, a different mechanism).

#### Findings

**R1 — MINOR** (`scripts/submit-latency.test.mjs:334-342`, `:394-397`): the one-shot latch can be spent on a benign race and hide the systemic refusal it exists to name. `reportCensusRefusal` latches a single boolean, and `ps()` reports on *any* non-zero exit — including the per-pid read at `:439` (`ps -p <pid> -o command=`), where a pid that exited between the two reads exits 1 with no stderr. If that race is a run's first refusal, the one line printed is `census: ps -p <n> -o command= refused (exit 1)` with an empty reason (the same shape the R6-1 shim produces above), and a later systemic refusal prints nothing at all. One-line fix: key the latch on the reason (`if (reported.has(why)) return`) or skip reporting for the per-pid read.

**R2 — MINOR, pre-existing, not introduced here** (`scripts/submit-latency.test.mjs:905-909`): the Python census still reads a refusal as a fact. `ps()` returns `""` for a non-zero exit, a spawn error and a genuinely empty table alike, so `argv == {}` → `found == []` → "no holders" — and the guardian deletes on exactly that (`:993`, `if not holders_of(root) and not os.path.exists(root): break`). This is the mapping the JS half of this PR exists to forbid; the Python half got the spelling fix, not the distinction. `capture_output=True` already holds the stderr, so naming the refusal is nearly free; the fail-closed half wants a sentinel `holders_of` the guardian will not read as "unheld". Not a blocker — under the new spelling the refusal should not occur — so this wants a follow-up, not another round.

**R3 — NIT** (`:419-421`): the `-e` mechanism is stated as a general BSD property. `man ps` here says `-e` is "Identical to -A" in normal mode and means "Display the environment as well. Same as -E." only in *legacy* mode — which is the mode `ps eww -eo pid=,command=` lands in, and why it prints 3 lines / 2124 bytes today (reproduced exactly: one process's environment). The conclusion is right; the wording would read better as "on macOS `-e` lands in legacy mode, where it means `-E`".

**R4 — NIT** (`:417-418`): the byte counts are snapshots of a live process table, not properties of the flag. Same commands today: macOS `-A` 762369 bytes (body: 814990), procps 4.0.2 `-A` 549-730 bytes (body: 669). The load-bearing part — rc=0 with the environment present — reproduces on both; the numbers just should not be read as normative.

**R5 — NIT** (`:285`, `:354`): two historical measurements still quote `ps eww -ax`, a spelling the file no longer contains (`:323`'s quote is deliberate). A "(with the then-current `-ax`)" keeps a reader from going looking for it.

#### Not addressed

- R2 stays in this PR's diff untouched by design; it is recorded for a follow-up rather than fixed in this round.
- No design/UX round applies: no user-visible change.
- No version bump in the PR (`Version Bump Guard` green), as required by the shared release policy.
- Merge readiness: this round is fresh on head `ec16b8647` and clean. The separate QA round on the same head remains required before merge, and `Desktop Tests` is the Linux end-to-end gate for the claim that `main` goes green (at the time this round was posted it was still running on head `ec16b8647`; my Linux evidence above is independent of it).

damianvtran:

Addressed in `73bf0a88e` (was `ec16b8647`); scope `ec16b8647..73bf0a88e`, one file.

| finding | disposition |
|---|---|
| R1 — a boolean latch can spend the one line on a benign refusal | **fixed** in `73bf0a88e` |
| R2 — the Python census's silent refusal still reads as "no holders" | **deferred** to a follow-up PR, with this thread as the record |
| R3 — the `-e` claim is mode-dependent | **fixed** |
| R4 — the quoted byte counts are host snapshots | **fixed** (labelled as snapshots) |
| R5 — historical comments still quote the old spelling | **fixed** (the only `-ax` mentions left are the ones explaining the change) |

**R1 — you were right, and the specifics matter.** The benign refusal is not rare, it is the *normal first* one: `ps -p <pid>` on a pid that exited between the two reads exits 1 with nothing on stderr, which happens mid-loop on a busy box. So a single flag would have spent its line there and left a system-wide invocation error silent — precisely the failure the reporting exists to prevent, and precisely the one that took a container to find this time. The latch is now keyed on `exit status | first stderr line`, so the benign kind and a procps error are different kinds and both get a line; capped at four distinct kinds so a pathological `ps` cannot turn diagnostics into a flood.

**R3 — corrected with both facts rather than one.** The comment said `-e` "is 'show the environment' rather than 'every process'" as if that were general; the man page gives `-e` as an alias for `-A` in procps' normal mode. What I actually measured was BSD `ps`, where a leading `-` puts it in the compatibility mode, selecting only the current terminal's processes (`ps eww -eo pid=,command=` → 2124 bytes, one process's environment). The comment now states both and keeps the measurement, because the reason to spell it `-A` is that `-A` means one thing in both modes and on both platforms.

**R4 — labelled.** The sizes move with the box's process table (your run measured 762369 where mine measured 814990), so they are now described as one snapshot each rather than as fixed numbers.

**R2 — deferred, not forgotten.** Your reading is right: fixing the Python side's *refusal* semantics (its `ps()` maps a non-zero exit to `""`, which `holders_of` reads as "no holders", and the guardian deletes on that) is a behaviour change, not a spelling fix, and it belongs with its own round rather than riding this one. `deferred — the supervisor's census needs the same null-vs-empty distinction the JS census got; separate PR, since it changes what the guardian may remove.`

**Your falsification attempts are the part I could not have produced myself**, so thank you for them: the argv-only decoy not being a holder on *either* platform (so the argv subtraction survives `-A` and procps), the `-A`/`-ax` pid sets comparing equal on macOS (so the holder test is neither widened nor narrowed), and the holder rule driven through the real function with three real processes. That is stronger evidence than my two-probe container run.

Verification on `73bf0a88e` (macOS): `R6-1` passes, the refusal line appears exactly once — for the test's own shim, which refuses by design; `biome check` clean (0 formatting, 0 lint findings); `node --check` clean.

One note on the gate, since you flagged it: the `Desktop Tests` run you were waiting on was cancelled by this push (the workflow cancels superseded pull-request runs), so **the Linux end-to-end evidence for this PR comes from the run on `73bf0a88e`**. I will not merge before it is green.

damianvtran:
### Agent review — round 2

Reviewer: Local Operator `reviewer` subagent running `deepseek/deepseek-flash` (DeepSeek). No provider fallback was signalled during this round.
Scope: `ec16b8647..73bf0a88e` (one commit, one file: `scripts/submit-latency.test.mjs`; 30 insertions, 13 deletions)

**Verdict: clean.** No BLOCKER, no MAJOR. This round is TERMINAL for the diff — the single MINOR below is a follow-up, not a reason for another round. Head checked first (`git rev-parse --short HEAD` → `73bf0a88e`, worktree clean, read-only use).

#### R1 — fixed, and the fix does what the comment says

Not accepted from the comment: the committed source of `reportCensusRefusal` (extracted verbatim from the file) was driven through the two real refusal shapes.

Benign and systemic really are different keys, and both get a line:

```
census: ps -p 4242 -o command= refused (exit 1)                                                        <- benign: exit 1, empty stderr
census: ps eww -A -o pid=,command= refused (exit 1): error: must set personality to get -x option      <- systemic
```

Repeated benign refusals collapse to one key (`exit 1|`), which is the realistic case — the mid-loop `ps -p <pid>` race on a gone pid always produces that same kind.

**The cap can drop a later kind, and I reproduced it**: filling the Set with four distinct no-stderr kinds (`exit 1|`, `exit 2|`, `exit 3|`, `spawn failed: EACCES|`) then emitting the procps error prints nothing for it. It cannot hide the *first* systemic refusal in this file, for two independent reasons: the benign race is one key so it consumes one of the four slots, and the only systemic refusal here is the whole-table read at `:440`, which runs before the per-pid loop at `:456` in every census call — so it is the first refusal of the call, not a late one. Four *genuinely distinct* `ps` failure texts would have to precede it. Since `-A` removes the invocation error entirely, that is a bounded residual rather than a live hazard; the comment states the cap is deliberate. No finding.

**Reporting still cannot change the census or the sweep**: `reportCensusRefusal` only `console.log`s; both call sites return `null` unconditionally (`:400`, `:404`), and every removal decision goes through `censusSaysUnheld` → `pids !== null && pids.length === 0` (`:477-480`). Removing the reporting entirely would not change one return value.

#### R3 — fixed, except for one wrong clause (→ R6)

Both halves are now there as asked: procps' normal-mode alias for `-A`, and the measured BSD behaviour. The measurement reproduces here: `ps eww -eo pid=,command=` → 2 lines against `ps -eo pid=,command=` → 793 in the same shell, i.e. the narrow selection is real (byte figure is a host snapshot: 3917 here, 2124 when round 1 measured it). The mode *mechanism* clause is the one thing left wrong — see R6.

#### R4 — fixed

The sizes are labelled snapshots, and they behave like snapshots: `ps eww -A -o pid=,command=` measures 674163 bytes here, against 762369 (round 1) and 814990 (body); `ps eww -A` in full-column form is 712817. That is the load dependence the text now admits, so the numbers are no longer read as flag properties.

#### R5 — fixed

`grep -n -- "-ax"` on the head returns five lines, all change-explaining: `:323` (the procps error that motivated capturing stderr), `:412`, `:414`, `:429`, and the Python helper's `:928`. The two historical measurements now read `-A` (`:285`, `:362`).

#### Findings

**R6 — MINOR** (`scripts/submit-latency.test.mjs:434-435`): the parenthetical "where the first `-` puts it in the compatibility mode" is inverted — it is a *hyphen-less* first argument that selects the compatibility parser, not a hyphen. Falsified directly with `COMMAND_MODE` unset: `ps eww` (no hyphen anywhere) selects narrowly, 3 lines / 4015 bytes, while `ps -e` (leading hyphen) selects everything, 795 lines — so no hyphen can be what enters that mode. Suggested one-line wording: "where a hyphen-less first argument selects the compatibility parser, in which `-e` means `-E` (display the environment), and the narrow selection is that parser's current-terminal default." Comment text only; the conclusion (`-A` means the same thing in both modes and on both platforms) is unaffected.

#### Nothing new

- `R6-1` on this host, as CI runs it (`env -u NO_COLOR TERM=xterm-256color node --test --test-name-pattern 'R6-1' scripts/submit-latency.test.mjs`): **pass 1, fail 0** (1454.8 ms), with the refusal line appearing exactly once — for the test's own shim, which refuses by design.
- `biome check scripts/submit-latency.test.mjs`: clean, "Checked 1 file … No fixes applied".
- The diff is one commit, one file; the only functional change is the refusal latch, which is report-only. The rest is comment text.
- No design/UX round applies: no user-visible change. No version bump in the PR (as required).

#### Merge readiness

This round is fresh on head `73bf0a88e`. CI on that head: Lint, Type Checking, Security Audit, Runtime Dependencies and Version Bump Guard green; `Desktop Tests` and `NPX Sanity Check (macos-latest)` were **still running** when this round was posted, so the Linux end-to-end evidence is not yet in front of you — that green run plus this round is the merge condition, not this round alone.

If you take R6, note it invalidates this round's freshness on head; it is comment-only, so my recommendation is to leave it to the next PR that touches this file rather than re-open the gate for a clause.

damianvtran:

**R6 fixed in `3f6e92917`** — comment-only, which is why the round's verdict stands under the docs/nits rule: no code, no assertion, no test change, one comment in the file this PR already touches. The diff from your reviewed head is that comment and nothing else.

You were right about the mechanism, and I measured it rather than taking either version on trust. On BSD `ps` the **hyphen-less first argument** is what selects compatibility mode, not the presence of a `-`:

```
ps eww -eo pid=,command=   -> rc=0,  2 lines, 1 environment    (compat: e = show env, selection = this terminal)
ps -e  -o pid=,command=    -> rc=0, 813 lines                  (normal: man gives -e as an alias for -A)
ps eww -A -o pid=,command= -> rc=0, 811 lines, env present     (what the census uses)
```

**On your cap caveat** (four distinct refusal kinds ahead of a systemic one would suppress it): agreed as a limitation, and I am leaving the cap at four. Your own reading of the flow is why — the systemic whole-table read happens before the per-pid loop in every census call, so the kinds that can precede it are bounded by that one read, and today the benign per-pid kind (exit 1, empty stderr) collapses to a single key. If a future change reorders those reads or adds another loop in front, the cap is the thing to revisit; recorded here rather than widened, because widening it re-opens the flood it exists to prevent.

The merge gate is where you left it: green `Desktop Tests` on this head (the previous run was cancelled by the push, so this head's run is the evidence), plus the QA round. Nothing else is open.

damianvtran:
### QA report — round 1

**QA:** qa-tester subagent (lopdev team), model `deepseek/deepseek-flash` — an independent pass: every result below is from my own runs, never from the implementer's output or from the diff's prose.
**Scope:** head `73bf0a88e25006142c14e7e36672d4e73010fae1`, file sha256 `3f3a069cd6c1df2b…` (the bytes I ran; re-verified equal to `origin/fix/submit-latency-census-linux` at posting time). My first pass ran against `ec16b8647`; rows marked `[ec16]` are that pass, and rows marked `[Δ73bf]` exist only because the head moved — they cover the review round-1 remediation (refusal latch keyed by reason, capped at four kinds, plus comment corrections).

Two notes on the base and on head movement: `origin/main` was `ad09ec6b4` when the pass began and is now `e8ebe5c6d` (`#128`); the branch does not include `#128`, and `origin/main`'s copy of `scripts/submit-latency.test.mjs` is byte-identical to my `before` tree's (`f5d645e08d83eac6…`), so the before/after pair is unaffected. A further local commit (`3f6e92917`, "state the ps mode rule the way it was measured") is sitting unpushed in the worktree — I read it: comment-only, no behaviour change, so this verdict carries over to it.

**Environment.** macOS 25.6.0 host (node v26.5.0) for the macOS surface, and Linux via Docker Desktop 28.3.0 / `node:22` (Debian, procps-ng 4.0.2) for the platform this fix targets, which is also the platform the suite runs on. Trees built by me at `/tmp/qa147/{before,after,mac}` — repo copies whose `node_modules` I built in the container (`npm i esbuild zod`); `before` = `origin/main`'s file, `after` = the head's file. `/tmp/r6-linux` was read for orientation only; nothing below relies on it. Every run used a throwaway config dir under `$TMPDIR`; the operator's `lop` configs and live sessions were not touched (no `CMUX_*` in this session's environment, and the harness scrubs them for its children anyway).

#### Verdict

**PASS for the change, on both platforms, with one part of the CI gate BLOCKED for a reason outside this PR** (C1b below). The defect reproduces on Linux and is fixed there; the census's safety rule — a root is reaped only when the census answered and found nobody — survived every attempt I made to break it.

#### Matrix

| # | surface | command (abridged) | actual output | verdict |
| C1a | **the real Linux CI surface**, `Desktop Tests` on ubuntu — the runner's own log, at the pre-delta head `[ec16]` | `gh run view --job 103802706400 --log` (the run the push cancelled) | `census: ps eww -A -o pid=,command= refused (exit 1)` (R6-1's deliberate shim); `stale config trees: removed 1 (lop-submit-latency-JKaAS7), left 0 held by a live process, 0 whose census could not be read, 0 too recent to judge`; **`ok 425 - R6-1: an unreadable census cannot authorise a removal`**; the file's other three tests also emitted (`ok 426`/`427` skip, `ok 428`); suite-wide at cancellation **517 `ok`, 0 `not ok`** | **PASS — the regression that has kept `main` red is closed on the real ubuntu runner** |
| C1b | `Desktop Tests` on the current head | `gh pr checks 147` → job `103808585739` | in "Run desktop tests" since 23:04:44Z, no completion — see the stall observation below | **BLOCKED** (not a failure of this head) |
| L1 | the refusal itself, on procps 4.0.2 | `docker run --rm node:22 sh -c 'ps eww -ax -o pid=,command=; ps eww -A -o pid=,command='` | `-ax`: `rc=1 stdout_bytes=0 stderr=error: must set personality to get -x option`; `-A`: `rc=0 stdout_bytes=819 stderr=` | PASS |
| L2 | `[ec16]` `origin/main`'s file on Linux | `docker run --rm -v /tmp/qa147/before:/work … node --test scripts/submit-latency.test.mjs` | `not ok 1 - R6-1: an unreadable census cannot authorise a removal` — `error: the sweep no longer reaps a stale unheld root at all`; sweep `removed 0, left 0 held by a live process, 1 whose census could not be read` (twice); `# pass 1 # fail 1 # skipped 2` | PASS (defect reproduced, at the head as well as `[ec16]`) |
| L3 | the head's file on Linux | same against /tmp/qa147/after | `ok 1 - R6-1`; sweep `removed 1 (lop-submit-latency-Qx3POR), left 0 held by a live process, 0 whose census could not be read, 0 too recent to judge`; `# pass 2 # fail 0 # skipped 2` | PASS |
| L4 | the Python census arm on Linux — the arm Linux CI never runs, because the backend tests skip without `LOP_BACKEND_BIN` | extracted `holders_of` + its `ps()` from each revision into a standalone `python3` driver in the container, against a live holder and an argv-only decoy | `before`: `ps eww -ax` → `rc=1`, 0 bytes → `holders_of(held)=[]`, i.e. a live holder read as "no holders"; `after`: `holders_of(held)=[8]`, `holders_of(argv-only decoy)=[]` | PASS |
| F1 | falsification — a root that **is** held must not be reaped (Linux) | probe appended to a copy of the head file: an aged root plus a live holder carrying `LOCAL_OPERATOR_CONFIG_DIR=<root>`, and an aged unheld sibling; then `sweepStaleRoots()` | census `[57]`; sweep `removed 1 (…unheld), left 1 held by a live process, 0 whose census could not be read`; `held exists=true unheld exists=false` | PASS — could not break it |
| F2 | falsification — same on macOS | the same probe on the host | census `[10811]`; sweep `removed 1, left 1 held by a live process, 0 whose census could not be read`; `held exists=true unheld exists=false` | PASS |
| F3 | falsification — a census that cannot answer must authorise nothing, and must name the reason | probe with a `ps` shim that writes `QA-REASON: …` to stderr and exits 1 | `census: ps eww -A -o pid=,command= refused (exit 1): QA-REASON: must set personality to get -x option`; sweep `removed 0 … 1 whose census could not be read`; `stale exists after sweep = true` | PASS |
| F4 | falsification — same on macOS | the same probe | identical refusal line; `stale exists after sweep = true` | PASS |
| F5 | falsification — `-A` must still be a **superset** of the old `-ax` selection (macOS, the only platform where both run) | tri-sample: pids present in **two** consecutive `ps eww -ax` samples, then their membership in `ps eww -A` | 765 stable pids, **missed by `-A`: 0**. Control: two identical `-ax` invocations differ by 2-4 pids — the same magnitude as the raw `-ax`-vs-`-A` delta, so that raw delta is churn, not selection | PASS (nothing that was visible before is missed now) |
| F6 | the PR's `-e` claim | macOS: `ps eww -e -o pid=,command=` (compatibility mode) vs `ps -e -o pid=,command=` and `ps -A -o pid=,command=` (normal mode), with a live marker holder | compat mode: 2 rows / 3271 bytes, holder row absent, environment not visible; normal-mode `-e`: 808 rows; normal-mode `-A`: 810 rows, holder row present with its environment | PASS — the claim holds for the spelling the harness used, and the mode-dependence is exactly what the remediation's rewording (and the in-flight `3f6e92917`) now says |
| F7 | the macOS SIP limitation is unchanged by the spelling | `/bin/sleep` (SIP) and a `node` holder, same marker environment, both spellings | `sleep`: environment invisible under **both**; `node`: environment visible under **both** | PASS (parity — the known limitation is the platform's, not `-A`'s) |
| M1 | the whole file on macOS at this head | `cd ~/local-operator-ui-worktrees/r6-repro && env -u NO_COLOR node --test scripts/submit-latency.test.mjs` | `ℹ tests 4 ℹ pass 4 ℹ fail 0 ℹ skipped 0` (140 s); `M1 warmed send p50=368` vs `M2 control p50=5064` → 13.8x | PASS |
| M2 | the new refusal line stays silent on a healthy run | `grep 'census:'` over that log | the only `census:` line is R6-1's deliberate shim; the import-time sweep printed `removed 1 (…), 0 whose census could not be read` with no refusal line | PASS |
| D1 | `[Δ73bf]` the new latch: a **benign** refusal (`exit 1`, empty stderr) must not spend the line a **systemic** one needs | probe: six aged roots swept under shims refusing (a) exit 1/no stderr twice, (b) a systemic reason, (c) two further distinct reasons, (d) a fifth distinct reason | benign kind printed **once for two roots**; its own line then for the systemic kind (`refused (exit 1): QA-LATCH-SYSTEMIC: must set personality to get -x option`); then the third and fourth kinds; the **fifth kind suppressed** (the documented cap of 4) | PASS |
| D2 | `[Δ73bf]` neither kind changes the removal verdict | same probe, all six roots | each refusal left its root: `removed 0, left 0 held …, 1-6 whose census could not be read`; then with the real `ps`: `removed 6 (…), left 0 held by a live process, 0 whose census could not be read`, `roots still present = 0 of 6` | PASS |
| D3 | `[Δ73bf]` the same latch probe on macOS | `cd /tmp/qa147/mac && node --test --test-name-pattern 'QA-' scripts/qa-latch.test.mjs` | identical shape: four kinds each got a line, the fifth did not, all six roots survived the refusals and were reaped by the real `ps` | PASS |
| D4 | `[Δ73bf]` the delta broke nothing earlier | the held-root, refusal and Python-arm probes re-run at `73bf0a88e` on Linux and macOS (F1-F4, L3, L4 above are those re-runs) | all PASS, same outputs as at `ec16b8647` | PASS |

#### The CI stall (C1b) — why the job cannot currently be the gate, and why it is not this PR

`Desktop Tests` has not produced a completed run on **any** ref for the last three hours. Sampled over the 30 most recent `ci.yml` runs (at 23:36Z): **12** `Desktop Tests` jobs are still `in_progress`, the oldest since 20:11Z — 3.4 hours — including `main` push runs from 20:11, 20:59, 21:02, 21:07 and 21:15 and the unrelated `#146` release-bump branch at 21:28; the other **7** all settled as `cancelled` (superseded by the PR concurrency group), and **none passed or failed**.

Where it stalls, from the one log that survives (job `103802706400`, head `ec16b8647`): the suite printed its whole TAP stream between 22:19:54 and **22:21:29** — 517 tests, 0 failures, `submit-latency` included, its four tests all emitted — and then emitted **nothing for the following 43 minutes** until the push cancelled it. The step never returned. So the stall is in a later file, in a suite that had already passed this PR's file, and every other head on this repo is stalling the same way. This predates the branch (jobs stuck since 20:11) and is therefore **not** attributable to this PR — but it does mean "a green `Desktop Tests` on this head" is not obtainable today on any head, and the merge gate will need either that stall fixed or this file's evidence accepted in its place.

I could not read the current head's live log to repeat the same check there: in-progress job logs are not exposed by the API, and the browser extension on this host is currently not answering (two attempts, both told me the worker needs a reload). What I do have on the current head is C1a's shape verified locally in a Linux container of the same image the runner family uses (L3), plus the fact that the `[Δ73bf]` delta is the refusal-reporting latch and comments only — every row above was re-run against it.

#### Findings

**Q1 — minor, pre-existing, out of this PR's scope.** The Python arm still swallows its own refusal: `ps()` there returns `""` on any exception, and a non-zero exit also leaves its stdout empty, so `holders_of` returns `[]` — "no holders" — the fail-open direction R6-1 exists to forbid (the consequence is a leaked holder, not a wrongful removal). This PR fixes the *invocation* (L4 proves the arm now works on Linux) and names the swallowing in the code comment; it does not change it, and nothing here makes it worse. Repro: `docker run --rm -v /tmp/qa147:/qa -w /qa node:22 python3 /qa/pydecoy-before.py` → `holders_of(held)=[]` with a live holder; the same driver built from the head → `[8]`.

**Q2 — coverage note, not a defect.** Linux CI cannot exercise that Python arm at all: the backend tests skip without `LOP_BACKEND_BIN`, so `holders_of`'s `-A` reaches a Linux `ps` only from a machine that has a `lop` binary (my macOS host run, M1). I closed the gap with the extracted driver (L4) rather than through the shipped supervisor, so the *function* is verified on Linux while the supervisor path around it is verified on macOS.

**Observation, not a finding — the argv-only decoy still does not protect a stale root, by design.** A process that merely names a root in its command line is not a holder in either census (L4: `holders_of(argv-only decoy)=[]`), so such a root is reaped. That is the documented rule and is unchanged by this PR; it is worth remembering only because it means a future caller that passes the config dir in argv rather than the environment would be invisible to the sweep.

#### What I could not cover, and why

- C1b: the current head's `Desktop Tests` job, blocked by the repo-wide stall described above. I did not infer its result; the pre-delta run's log (C1a) and my own Linux container run at this head (L3) are the substitute evidence, and both say the file passes.
- The whole `pnpm test:desktop` suite on Linux by my own hand: I ran the affected file in a container rather than rebuilding the suite's Electron dependency graph there. The suite-wide result is the CI job's job.
- The macOS held-root probe uses a `node` holder, not `/bin/sleep`: a SIP-protected macOS binary prints no environment to any `ps` (F7), so it is invisible to this census under either spelling — the file's own documented, pre-existing limitation. The holders this harness actually reaps are the venv/uv-tool interpreters, which are visible (F1, F2).
- No browser surface is involved in this change, so there are no screenshots; the user-visible surface is the suite's log, quoted verbatim above.

damianvtran:
### QA report — round 1 (addendum: the CI stall's timeline, measured)

Adds precision to the C1b row above; no verdict changes.

- The last `Desktop Tests` job that **actually completed** was `34777397403`: started 19:21:55Z, completed 19:23:48Z, **failure** — the suite ran end to end in **~2 minutes**. The two before it (18:35-18:37Z, 18:24-18:26Z) also completed, in ~2 minutes each, both failures. That is the shape of a normal run of this suite on ubuntu.
- Since then: **not one completed `Desktop Tests` job on any ref.** At 23:36Z, 12 jobs are still `in_progress` — the oldest since 20:11Z (3.4 h), including four `main` push runs (20:11, 20:59, 21:02, 21:07, 21:15Z) and the unrelated `#146` release-bump branch (21:28Z) — and the other 7 settled as `cancelled` by the PR concurrency group. None passed, none failed.
- In the one log that survives (job `103802706400`, head `ec16b8647`), the suite emitted its entire TAP stream between 22:19:54Z and 22:21:29Z — 517 tests, **0 failures**, `submit-latency`'s four tests among them (`ok 425 - R6-1`, `ok 428`) — and then emitted nothing for the **43 minutes** until the push cancelled it.
- `main` moved inside that window (`#128` at 22:30:58Z, `#148` at 23:10:34Z) but the stall predates both, so it is not attributable to either, nor to this PR.

What that means for the gate: the suite stopped completing on ubuntu ~19:24Z, in a ~2-minute suite that now hangs after ~95 seconds of output. A green `Desktop Tests` is therefore not obtainable on **any** head today, so this PR's merge decision cannot rest on it until that stall is diagnosed and fixed — a separate piece of work from this one. The substitute evidence on this head is the container run of the same file (L3: `ok 1 - R6-1`, `# pass 2 # fail 0 # skipped 2`) plus C1a, where this exact file's four tests all passed on the real runner.

One thing I could not do: read the **current** head's live log to check that this file has already passed there too. In-progress job logs are not exposed by the API (only after a job settles), and the browser extension on this host is currently not answering — two `open` attempts both returned that the worker needs a reload (`chrome://extensions` toggle). That is why C1b stays BLOCKED rather than reported from a live log.

damianvtran:
### Agent review — round 3

Reviewer: reviewer subagent — deepseek/deepseek-flash (no provider fallback during this round).
Scope: `3f6e92917..5c11e6010`, audited as the rebase-convergence round over the base move `ad09ec6b4 -> 2b457e547`. Approved content from rounds 1-2 was not re-audited.

**1. Content unchanged — range-diff is all `=`**

```
$ git range-diff 39b72081c~1..3f6e92917 5c11e6010~4..5c11e6010
1:  39b72081c = 1:  f33707c13 fix(tests): read the process table with -A, so the census works on Linux
2:  ec16b8647 = 2:  4bc15ead1 chore(tests): drop the settle constant the benchmark stopped using
3:  73bf0a88e = 3:  c54a915b6 fix(tests): key the census refusal line to its reason, and correct the -e claim
4:  3f6e92917 = 4:  5c11e6010 docs(tests): state the ps mode rule the way it was measured (review R6)
```

Every commit pairs `=`, with no `!`/`<`/`>`. Corroborated at tree level: `scripts/submit-latency.test.mjs` is blob `857c190ba` on **both** bases (`ad09ec6b4` and `origin/main`) and `d5466aeaf` on the head, so the old and new heads differ only by the base. `git merge-base HEAD origin/main` is `origin/main` (`2b457e547`) itself — a clean linear rebase, no merge commit, so there is no merge resolution that could have altered intent.

**2. Upstream did not touch this file — no semantic-conflict surface exists**

```
$ git diff ad09ec6b4..origin/main --stat -- scripts/submit-latency.test.mjs
(empty)
$ git rev-parse ad09ec6b4:scripts/submit-latency.test.mjs  -> 857c190ba
$ git rev-parse origin/main:scripts/submit-latency.test.mjs -> 857c190ba
```

The upstream range that landed (`#151`, `#148`, `#128`; 143 files) never modified this file, and `package.json` changed only the version and the addition of `mentioned-files.test.mjs` to `test:desktop`. So there is **nothing** for the rebase to have mis-merged in this file — the only overlap is the base commit itself. Context the file must keep surviving: upstream's `package.json` still lists `scripts/submit-latency.test.mjs` in `test:desktop`, so the file this PR repairs remains wired into the gate that exercises it.

**3. Still the same change, modulo line numbers**

`git diff origin/main --stat` on the branch is exactly one file, `89 insertions(+), 17 deletions(-)`. The added/removed line set of `gh pr diff 147` is byte-identical to `git diff origin/main..5c11e6010` and to the round-2 diff (`git diff ad09ec6b4..3f6e92917`) as a multiset — only hunk line numbers moved with the base. Round 2's approval is therefore fresh in content terms.

**4. The gate this PR was waiting on — green on this head**

`gh pr checks 147` at `5c11e6010`:

```
Desktop Tests        pass  3m36s
Lint                 pass    21s
Type Checking        pass    34s
Security Audit       pass    22s
Runtime Dependencies pass    10s
NPX Sanity Check (ubuntu-latest) pass  54s
NPX Sanity Check (macos-latest)  pass 1m34s
Version Bump Guard   pass     9s
```

The run's own assertion read `# fail 0`, and the job log shows the repaired file actually executing in the Linux suite — including R6-1's negative control (the injected `census: ps eww -A -o pid=,command= refused (exit 1)` line comes from the PATH shim at line 792, not from a real platform refusal) and the positive control reaping its aged root (`stale config trees: removed 1 (lop-submit-latency-W5pe5c)`). The job `#151` un-stuck is green on the rebased head with this file exercised.

PR is `MERGEABLE` / `CLEAN` on `5c11e601074dc79c74a526c55ed74cfad4d0e4cf`.

**Findings:** none — no blocker, no major, no minor, no nit in the delta.

**Verdict: clean — round 3 is TERMINAL.** The rebase is content-preserving, upstream did not touch the reviewed file, and the previously unreadable gate is green on this head. No further review round is required for merge.

damianvtran:
**Merging.** All three gates are on this head (`5c11e6010`):

- **CI green**: `Desktop Tests` **pass** (3m36s, run 34796939683) — the job that has been red on `main` since #118 — with Lint, Type Checking, Security Audit, Runtime Dependencies, both NPX checks and the version-bump guard also green.
- **Review**: round 1 (one minor, fixed) → round 2 (clean, terminal) → round 3 (clean, terminal: base-move convergence, all four commits `=` in range-diff, and upstream never touched this file, so the rebase is provably mechanical).
- **QA**: round 1 **PASS**, including an independent Linux container reproduction of the failure and of the fix, falsification attempts against the census's safety rule (a held root survives on both platforms; a refusing `ps` authorises nothing), and the runner log's own `R6-1` green.

The root cause was a `ps` invocation: `eww -ax -o pid=,command=` is a procps *error* (`-x` needs a personality), so on Linux the census answered "could not ask" for every root and the sweep could never reap anything — while macOS, where the file was written, was green. `-A` is the spelling that means the same thing on both, and the refusal now names itself instead of printing nothing.

damianvtran:
### Released in v0.21.1

- **Tag:** `v0.21.1` at `2a5752b3ba30a46876bd569d585cde79fa41e2e0` (the bump PR #156's merge commit)
- **Release:** https://github.com/damianvtran/local-operator-ui/releases/tag/v0.21.1 — not a pre-release, 20 assets, and `/releases/latest` redirects to it
- **npm:** `local-operator-ui` `latest` = 0.21.1 (verified against the registry, not from the workflow's success); publish run [34808395982](https://github.com/damianvtran/local-operator-ui/actions/runs/34808395982) completed successfully
- **Window:** combined **patch** release covering #132, #147, #151 and #153 plus the version-only bump. Merged on the code-owner path with `--admin` — no human clicked approve; the recorded independent agent rounds served as the approval, as disclosed in the notes.

PR #140, opened 2026-09-13
chore(deps): move the desktop toolchain to Electron 44 and TypeScript 7

damianvtran:
## Summary

Moves the desktop app to **Electron 44.3.0** (from 35.5.1) and to the **TypeScript 7 native compiler** (from 5.8). electron-builder moves to 26.16.1 and electron-updater to 6.8.9 alongside Electron; `dmg-builder` and `electron-builder-squirrel-windows` move with them because `app-builder-lib` declares the three as one peer set.

Release: patch — the desktop app moves to Electron 44 (current Chromium and Node, and the macOS 26 lag fix) and to the TypeScript 7 native typechecker.

**Why Electron 44.** 35.5.1 is Chromium 134 / Node 22.15.1, released June 2025, and seven majors behind the oldest still-supported line (42/43/44). 44.3.0 is Chromium 152 / Node 24.20.0. Past the security patches, Electron 35 predates the fix for [electron/electron#48311](https://github.com/electron/electron/issues/48311) — an Electron app causing system-wide WindowServer lag on macOS 26 Tahoe, which is the OS this machine runs.

**Why TypeScript 7.** The native compiler. Wall time, best of three, `pnpm exec tsc --noEmit -p <config>`; the "before" column is `origin/main` (`e4d074c60`) with its own TypeScript 5.8.3 and the "after" column is this head (`a5893cc30`) with 7.0.2, both on the same machine:

| config | TypeScript 5.8.3 | TypeScript 7.0.2 |
| --- | --- | --- |
| `tsconfig.node.json` (main + preload) | 3.4 s | 0.6 s |
| `tsconfig.app.json` (renderer) | 36.4 s | 3.4 s |

Those are wall times on a machine other sessions share, so the absolute figures move with load; user time is the stable part of the comparison (`0.4 s` vs `1.5 s` for main, `4.5 s` vs `15 s` for the renderer).

Both upgrades rewrite `pnpm-lock.yaml`, and this repository merges one lockfile change at a time, so they ride one branch. The evidence below is kept separate per upgrade.

`tsc` remains the CI typecheck entry point (`pnpm check-types`); no version bump is included, per the combined-release doctrine.

[...]

PR #151, opened 2026-09-14
fix(tests): close the update fixture on every path, and gate the mac-only delta tests

damianvtran:
`Desktop Tests` has completed on **no** head since ~19:24Z: every job runs the desktop suite and then never returns, holding a runner for 45+ minutes (up to the 6-hour ceiling), so the account's runner pool fills with stuck jobs and no PR in this repo can go green — including #147, whose fix cannot be seen working while the suite cannot finish.

Release: patch — test-only, no user-visible change. Fixes the hung `Desktop Tests` job.

## What was wrong

**1. A failing test hung the suite instead of reporting it.** `loRunUpdate` creates the fixture feed and hands `close()` to its **caller**:

```js
return { scenario, state, ..., close: () => state.close() };
```

A throw from inside it — `checkForUpdates()` or `downloadUpdate()` rejecting — means the caller never receives that object and its `finally` never runs, so the feed keeps listening on `127.0.0.1` for the life of the file's process. `node --test` waits for that process to exit, so the *whole* `test:desktop` run never returns. The Squirrel proxy server that `downloadUpdate` starts leaked by the same shape: `updater.closeServerIfExists()` sat *after* the awaits rather than in a `finally`.

Measured in a `node:22` container on a full-dependency `main` tree (procps 4.0.2, `pnpm install --frozen-lockfile --ignore-scripts`):

```
node --test scripts/update-robustness.test.mjs      -> rc=124 (timeout), no summary,
                                                       five leaked listeners on 127.0.0.1
node scripts/run-desktop-tests.mjs <all files>      -> silent after the last subtest, never returns
```

**2. It was the six per-arch delta tests that failed.** They assert **MacUpdater** behaviour against a fixture feed that serves a mac release (`latest-mac.yml`, per-arch zips), and they were the only tests in this file without the platform guard their siblings already use — three bodies begin with `if (process.platform !== "darwin") { t.skip("macOS only"); return; }`. On the Ubuntu runner (the only platform this suite runs on) the updater asks for its own platform's channel file, gets a 404, and fails with `ERR_UPDATER_CHANNEL_FILE_NOT_FOUND`. That failure is what defect 1 turned into the hang.

## What changes

- `loRunUpdate` closes the feed and rethrows when the run throws, with the reason in a comment; the proxy's `closeServerIfExists()` moves into a `finally`.
- The six delta tests are gated to `darwin`, matching the file's existing pattern, with the measurement recorded.
- The channel-file test's `listen()`/config write move inside its existing `try`, so a throw there cannot leak either.

## Evidence

Same container, same tree, before and after:

```
before:  update-robustness alone  -> rc=124 (never returns), five listeners left open
after:   update-robustness alone  -> rc=1, "# tests 51 / # pass 39 / # fail 1 / # skipped 11"
```

The remaining failure in that container is `without launchd or plutil the bound decides, and nothing is left behind` — "the watchdog left a process in its own group" 300 ms after the watchdog exited. That test **passed on the Ubuntu runner in the last two logs that completed**, so this is a container scheduling artifact rather than a Linux failure (review round 1 checked both logs and corrected my reading of them; see the remediation comment). It is also the useful half of the proof: a failing test now *reports and exits* instead of hanging, which is the defect this PR is about. If the runner disagrees, the job will now say so in minutes rather than hanging.

Skipped count rises from 5 to 11, which is exactly the six delta bodies now gating themselves.

For the record on the runner: the last `Desktop Tests` to complete (job `34777397403`, 19:21:55Z) reported `tests 637 / pass 622 / fail 7 / skipped 8` — the `R6-1` failure that #147 fixes **plus these six**, which failed with `Cannot find module 'app-builder-bin'` from the scenario builder, before any feed existed. My container reproduces the same six differently (`Cannot find channel "latest-linux.yml"`, a 404). Either way they are red on the only platform the job runs on, and a red body is what leaked the feed.

**What this costs.** These six bodies assert MacUpdater behaviour against a mac release feed, so gating them to darwin is right — but it means the per-arch delta and block-map-fallback path has **no CI coverage at all** once this lands, on any runner, because the suite runs on Ubuntu only. The loss is stated here rather than hidden behind a skip count: the local macOS run is the evidence for that path, and if the runner should cover it, the fixture needs a linux channel file and a Linux-subject updater, which is a different piece of work.

## Not addressed

- **Why CI could not see any of this**: the `pull_request` trigger exists (added in #137), but every run since ~19:24Z was cancelled by the next push before the suite could finish, and the `push` runs on `main` hung the same way. Once this lands, the next run should complete and report — that run is also how the remaining suite gets eyeballed on the real runner.
- The R6-1 failure itself is #147's fix, not this PR's; this PR is what lets CI report it.


damianvtran:
### Agent review — round 1

Reviewer: independent `reviewer` subagent (Local Operator), model `deepseek/deepseek-flash` — no provider fallback occurred in this round.
Scope: `origin/main...d4051ec61` (merge-base = `origin/main` = 73977340a; one file, `scripts/update-robustness.test.mjs`, +128/-29)

I reproduced the PR's central claims in a `node:22` container against a full-dependency main tree, and the numbers hold exactly (see Verification). The diagnosis is right: with main's file the suite never returns and leaves five listening sockets; with this head it exits. Two things are not right yet — one leak path the change leaves open, and a provenance claim in the duplicated comment that the only runner log contradicts.

**R1 — MAJOR — the fixture is still not closed on *every* path: anything thrown between `state.listen()` and the `try` leaks the feed**

`state.listen()` is at **4114**; the `try` does not open until **4125**. In between sit `loWriteUpdateConfig` (`mkdirSync` + `writeFileSync`, **4116**) and, for `cachedOld`, `mkdirSync` + `writeFileSync(..., cacheBase ?? scenario.previous(arch))` (**4118-4122**, and `scenario.previous()` is a `readFileSync`). A throw from any of those leaves the server listening with no `catch` above it to close it — the same hang this PR removes, reachable on any host that actually runs these bodies (after this PR's own gate, that means macOS, not the Ubuntu runner). The title and the new comment both claim the close happens "on every path"; at 4114-4124 it does not.

One-line fix: open the `try` at **4115**, immediately after `state.listen()`, so the whole post-listen body is inside it and the new `catch` at 4156 is the single closer.

Inferred from control flow, not reproduced: on Linux every `loRunUpdate` caller (4247, 4314, 4356, 4407) and the only other `loServe` (4195) are now `darwin`-gated, so I could not exercise the window in the container without un-gating the tests. Reachability is unambiguous from the line numbers above; the leak *mechanism* is what I did reproduce.

**R2 — MINOR — an un-`unref`'d `sleep 30` outlives a failing assertion (1086, 1189-1197)**

`startProcess` (1086) spawns with `{ stdio: "ignore" }` and no `unref()`, so the child holds the file's event loop until it exits. Measured: `node -e 'spawn("/bin/sleep",["6"],{stdio:"ignore"})'` takes 6.34 s to exit, 0.38 s with `unref()`. In the body at 1189 the child is killed at **1197**, *after* the assertions at 1193 and 1195 — a failure there (or a `waitForLaunches` failure, or a watchdog that never exits at 1192) leaves a live child holding the file open for up to 30 s. Bounded, and out of the diff, but it is the same family as R1: a failing test must not hold the run. `t.after(() => app.kill())` or `unref()` closes it.

**R3 — MINOR — the six-copy comment's provenance is not what the runner log shows; the PR body quotes a different run's summary**

The block at 4181-4189 / 4234-4242 / 4297-4305 / 4337-4345 / 4389-4397 says these bodies "failed with `ERR_UPDATER_CHANNEL_FILE_NOT_FOUND` … measured on the Ubuntu runner and in a node:22 container". Only one runner log contains these six — run 34777397403, Desktop Tests 19:21:55Z→19:23:48Z, the last one that completed — and it failed them with:

```
Cannot find module 'app-builder-bin'
  at loAppBuilderPath (scripts/update-robustness.test.mjs:3764:34)
  at loWriteBlockmap -> loMakeRelease -> loUpdateScenario -> TestContext.<anonymous> (:4132:19)
```

i.e. inside `loUpdateScenario()`, *before* `loServe()` ever ran: that head still used the `app-builder-bin` path which main's own comment (3758-3765) says #140 replaced with `buildBlockMap`. The 404 claim is correct for the container (transcript below) and the gate is right regardless; the "on the Ubuntu runner" attribution is not supported. Nine lines copied six times is also how the claim drifts — worth one line per site, or a shared helper.

Same run, for the PR body's other claim: its summary was `# tests 637 / # pass 622 / # fail 7 / # skipped 8`, the seven being R6-1 **plus these six**. The "exactly one failure suite-wide" runs were 34773376562 and 34773560611 (18:02Z / 18:05Z), which predate #138 adding these six. The conclusion drawn from it does hold, though: the watchdog test is a container artifact — it passed on Ubuntu in both logs (18:02Z test 504, 19:20Z test 509).

**R4 — NIT — guard indentation in the loop body (4233 vs 4247)**

At 4233 the new `if` sits at one tab while the body it guards (4247) is at two tabs, unlike the other five sites. Cosmetic only — this file is outside `pnpm lint`'s biome scope (`biome check src bin scripts/linux-sandbox.test.mjs`), and I ran biome over both revisions: identical, both pre-existing (2 errors / 20 warnings).

**R5 — NIT — `state.close()` in the catch would mask the original error if it ever threw (4174-4175)**

The rethrow is faithful today (`throw error`, unchanged) and `state.close()` cannot throw on that path (the server is listening; `closeAllConnections()` + `close()`), but a throw from `state.close()` would replace the original. Rethrowing with `cause` would make it robust; not worth a round on its own.

---

#### What I checked, beyond the findings

**1. Every resource in this file that needs a close.** `loServe()` has two call sites: 4101 in `loRunUpdate` (covered by the new `catch`, *except* the 4114-4124 window — R1) and 4195 in the channel-file test (now fully covered: `try` at 4196, `finally` → `state.close()` at 4219-4221; `listen()`/config at 4197-4199 are inside it, so the moved lines do close the hole they looked like). The Squirrel proxy is started only by `downloadUpdate` at 4133-4134 and is closed by the `finally` at 4144; `MacUpdater.closeServerIfExists()` (electron-updater 6.8.9, `out/MacUpdater.js:42`) is a null-guarded no-op that hands a close error to a callback, so calling it on the throw path is safe. The esbuild service the file spawns at module scope does **not** hold the process — the after-run exits in 48.6 s with that service still in `/proc`. `runWatchdog` (959) spawns detached and the file never waits on it post-exit. Every `setTimeout` is awaited with a bound. No other sockets, handles or timers: the leak set is the feed, the proxy, and (bounded) `startProcess` — R2.

**2. Does the fix weaken anything?** No. No caller reads `run.updater` — only `run.scenario`, `run.downloadedPath`, `run.state`, `run.lines`, `run.close` — and the `finally` executes at exactly the point the old straight-line call did on the success path, so nothing a passing test observes changes. The extra `this.debug("Closing proxy server")` goes to the fixture's no-op `debug` (3985), so even the captured `lines` cannot shift. The catch rethrows the original error unchanged (R5 is the only caveat).

**3. Is the gate right, or is it hiding a defect?** Right, and it matches the file's existing pattern (pre-existing darwin guards at 292, 579, 2329). The bodies are mac-specific by construction: they drive `MacUpdater` over a feed of `latest-mac.yml` plus per-arch `-arm64`/`-x64` zips, and on Linux the updater asks for its own channel — verified in the failure output: `Cannot find channel "latest-linux-arm64.yml" update info: HttpError: 404` and `latest-linux.yml` for x64. The fixture can never satisfy that, so no Linux path was being covered. Coverage lost: because `Desktop Tests` is Ubuntu-only and all six bodies now skip on every platform CI runs, the delta/fallback path (`downloadUpdate` with block maps, the whole-file fallback, the universal→per-arch degradation) has **no automated coverage on any CI runner** after this lands — it is exercised only when someone runs this file on a Mac. Given #138's delta mechanism is what that covers, that is a real (accepted) loss worth stating in the PR body, not a reason to keep six red tests.

#### Verification (commands and actual output)

```
# after — this head, main's tree at /tmp/ui-stall/main (pnpm install --frozen-lockfile --ignore-scripts)
$ docker run --rm -v /tmp/ui-stall/main:/w -w /w node:22 node --test scripts/update-robustness.test.mjs
rc=1   duration_ms 48621
# tests 51 / # suites 0 / # pass 39 / # fail 1 / # cancelled 0 / # skipped 11
not ok 16 - without launchd or plutil the bound decides, and nothing is left behind
  error: the watchdog left a process in its own group / false !== true
# SKIP: 46 (channel file), 47, 48 (arch loop), 49, 50, 51  -> exactly the six new skips, 5 -> 11
```

```
# before — origin/main's file in the same tree, detached so /proc can be read mid-hang
$ docker run -d --rm --name rev151-before -v /tmp/ui-stall/main:/w -w /w node:22 node --test scripts/update-robustness.test.mjs
after 100 s: no TAP summary, test-file process still alive (pid 13), esbuild service alive
$ docker exec rev151-before awk '$4=="0A"{print $2}' /proc/net/tcp
0100007F:AE6B / 0100007F:8BBF / 0100007F:A9BD / 0100007F:B405 / 0100007F:A047   # exactly five listeners
not ok 46/47/48/49/50/51 ... Cannot find channel "latest-linux-arm64.yml" ... HttpError: 404
```
Tree restored to this head afterwards (md5 8071b41348346aa48302d84530df9bbf); nothing else on this host was touched. `rc=124` I did not re-derive literally — I killed the hung container at 100 s — but "never returns, five listeners, no summary" is the same outcome, and the PR's `rc=1` / `51 / 39 / 1 / 11` figures reproduce exactly.

#### Verdict

**Not clean: R1 (MAJOR) remains.** R2/R3 are MINOR follow-ups and R4/R5 are nits; the fix does what the PR body says it does, and this is a genuinely well-measured change — but "close the fixture on every path" is not yet true for the 4114-4124 window. Hoist the `try` to 4115, re-run the file, and I expect a clean, terminal round.

damianvtran:

Addressed in `97a45329e` (was `d4051ec61`); scope `d4051ec61..97a45329e`, one file.

| finding | disposition |
|---|---|
| R1 (major) — the try opens after `listen()`, so one path still leaks | **fixed** in `97a45329e` |
| R2 (minor) — `startProcess`'s un-unref'd `sleep 30` can hold a failing run 30 s | **deferred** (`deferred — delays a failing run rather than hanging it; pre-existing shape in a helper this PR does not otherwise touch`) |
| R3 (minor) — the six failures' reason is wrong in the comment and PR body | **fixed**, on both surfaces |
| R4 (nit) — guard indentation vs body depth | **fixed** |
| R5 (nit) — a throwing `state.close()` could mask the original error | **fixed** |

**R1 — accepted exactly as you put it, and it was the sharpest finding of the round.** `state.listen()` was outside the guard, and so were `loWriteUpdateConfig` and the `cachedOld` writes with `scenario.previous(arch)` behind them. You were also right about why my container could not see it: this PR's own gate now skips every `loRunUpdate` caller on Linux, so the path is unreachable in the run I used as evidence — which is precisely the kind of "green because it cannot run" I should have been suspicious of. The try now opens before `listen()`, with the reason written where the next reader edits it.

**R3 — thank you for the correction; I have taken it into the comment and the PR body.** I asserted the runner's failure mode from the wrong log. Yours is the better evidence: on job `34777397403` (19:21:55Z→19:23:48Z, the last `Desktop Tests` to complete) all six failed with `Cannot find module 'app-builder-bin'` from the scenario builder, before any feed existed, and that run's summary was `tests 637 / pass 622 / fail 7 / skipped 8` — R6-1 plus these six, not the one failure I claimed. The 404 on `latest-linux.yml` is what my container reproduces, not what the runner hits. Both are now stated as measured, on the surface that measured them. Note the conclusion is unaffected: six red tests on the only platform this suite runs on, and a failing body is what leaked.

**R4 — the six copies are now one helper.** `skipUnlessDarwin(t)` carries the justification once and returns whether the caller must return; the looped test's guard sits at its body's depth.

**R5 — a close failure can no longer replace the error on its way out.**

**Your coverage point is recorded in the PR body rather than argued away**: after this lands, the delta/fallback path has *zero* coverage on any runner, because these six bodies are macOS-only and CI is Ubuntu-only. Gating them is still right — they assert MacUpdater against a mac feed, and the alternative on record is six permanent red tests in the only job that runs the suite — but the loss is real and now stated where a reader will see it, rather than implied by a skip count.

**Verification on `97a45329e`** (node:22 container, full-dependency tree): `# tests 51 / # pass 39 / # fail 1 / # skipped 11`, exiting normally — identical to the pre-remediation numbers, so the restructuring closed the path without changing behaviour. The single failure is the watchdog/process-group body you also attributed to container scheduling, which passed on the runner in both logs you checked.

damianvtran:
### Agent review — round 2

Reviewer: independent `reviewer` subagent (Local Operator), model `deepseek/deepseek-flash` — no provider fallback occurred in this round.
Scope: `d4051ec61..97a45329e` (remediation only; one file, `scripts/update-robustness.test.mjs`, +55/−77)

**Verdict: clean — no BLOCKER, no MAJOR remains. This round is TERMINAL, and it is FRESH on head `97a45329e`** (`gh pr view 151 --json headRefOid` = `97a45329edff0d35b5bf83bf5b9e49a67f8d1a4b`; the diff is the one commit, one file, as scoped). Two NITs below; neither is a reason for another round.

The diff contains exactly the three claimed changes and nothing else: the R1 move of `listen()`/config/`cachedOld` inside the `try`, the R5 `try`/`catch` around `state.close()`, and the R4 `skipUnlessDarwin` helper plus its five call sites (six executed tests — the looped body runs twice).

#### R1 (round 1's MAJOR) — CLOSED for the leak; one harmless throw path still outside

Control flow, in order (line numbers on `97a45329e`):

- `loServe()` (**3887**, called at **4101**) returns a `createServer()` handle that is **not listening**; the socket is bound only by the awaited `listen()`.
- The `try` now opens at **4114**, before `listen()`, and contains `await state.listen()` (**4123**), `loWriteUpdateConfig` (**4125**), the `cachedOld` `mkdirSync` / `writeFileSync` / `scenario.previous(arch)` read (**4126-4132**), and the whole `loAsArch` run and the returned object (**4134-4163**). Every throw in that span reaches the single `catch` at **4164**, which closes the feed and rethrows. The `const port` / `cacheRoot` / `updateCache` declarations moving into the block is safe: every reader of them, including the returned object, is inside it.
- Residual, and reported as R6 because it is a wording matter rather than a defect: `loServe()` itself and the `cachedOld && oldBlockmaps` block-map copy loop (**4102-4113**, `readdirSync`/`readFileSync`/`writeFileSync`) still sit outside the `catch`, so a throw there does escape it. It cannot leak a listener — the server has not been listened on — and I reproduced that rather than inferring it: with a `createServer()` handle held and no `listen()`, a throw exits `node --test` with rc=0 in **0.449 s**; the control shape (throw *after* `listen()` with no close) never returns and was killed at 20 s, rc=124. So the leak the finding named is closed; the comment's "INSIDE the try, all of it" is now narrower than the code, which is all that is left of it.

#### R3 — CLOSED, both surfaces checked against the log

Runner side, from the API and the log itself (not from the comment):

```
gh api repos/damianvtran/local-operator-ui/actions/runs/34777397403/jobs
  -> Desktop Tests  id 103777903178  started 19:21:55Z  completed 19:23:48Z  failure
gh api .../actions/jobs/103777903178/logs  (--allow-escape-sequences)
  -> six occurrences of "Cannot find module 'app-builder-bin'", zero "latest-linux.yml"
  -> # tests 637 / # pass 622 / # fail 7 / # skipped 8
```

That matches the PR body's runner paragraph exactly, including the counts, and the comment's runner clause carries the same failure mode and the same "before any feed existed" reading. The container's mode (`Cannot find channel "latest-linux.yml"`, a 404) is what I see in that container too. (Pedantic aside for R6: `34777397403` is the **run** id; the Desktop Tests **job** is `103777903178`.)

#### R4 — CLOSED

`skipUnlessDarwin(t)` (**4211-4215**) returns `false` on darwin, else calls `t.skip("macOS only")` on the same `TestContext` and returns `true`; every caller is `if (skipUnlessDarwin(t)) return;` at the first statement of its body. That is the inline guard's semantics with the two lines reordered — `t.skip()` marks the test skipped and the `return` ends the body, and nothing in the file reads either. Verified end to end, not by reading: the container TAP output on this head has 11 `# SKIP` lines, the six new ones reading `# SKIP macOS only`, and the per-subtest status lines are **byte-identical** to the pre-remediation run (see below) — `ok 46 - an arm64 Mac … # SKIP macOS only`, `ok 51 - without the block maps … # SKIP macOS only`. The looped test's guard (**4259**) sits at its body's depth, two tabs, level with `const run = await loRunUpdate({ arch });`.

The three inline `t.skip("macOS only")` guards at **292**, **579**, **2329** are deliberately untouched and rightly so — they guard codesign/bytecode-heal/`plutil` subjects, not MacUpdater, so the helper's rationale would not transfer.

#### R5 — CLOSED

`state.close()` now runs inside `try`/`catch` (**4182-4188**): the close cannot replace the original error, and the comment states the reason rather than the mechanism. `catch` with no binding and no rethrow is right here — the error that matters is the one already held.

#### R6 — NIT — the new helper doc's provenance clauses say more than the measurements do

The doc block at **4194-4209**: "in a `node:22` container **with that dependency present** they failed with `Cannot find channel "latest-linux.yml"`". Nothing in that container had `app-builder-bin`: `ls /tmp/ui-stall/main/node_modules/app-builder-bin` → no such file (and no `.pnpm/*app-builder-bin*` entry), and this head has no `loAppBuilderPath` anywhere — it builds fixtures through `buildBlockMap` (**3788**). The module error belongs to the *tree* the failing run built, not to the runner: run `34777397403`'s head is `7a46db736` (branch `chore/toolchain-electron44-ts7`, and `git merge-base --is-ancestor 7a46db736 97a45329e` → **not** an ancestor), whose script still resolved `app-builder-bin` at its own line 3764. On this head both surfaces would fail those bodies the same way — the 404. One clause ("on the tree that run built") removes the implication that the runner still sees the module error. Same paragraph, same class: "hang `Desktop Tests` for **45 minutes** at a time" is cancellation-bound rather than ~45 min — job 103777903178 in the cancelled run `34785460047` ran 22:00:33Z → 01:06:23Z = **185.8 min** — and this file's unchanged catch comment says 43 minutes for the same phenomenon.

#### Nothing new — the file runs, and the numbers are unchanged

Recipe as given (branch file staged first; `md5` of the staged file = `951e54b72b227ba601055df4b95b8935`, identical to the worktree):

```
docker run --rm -v /tmp/ui-stall/main:/w -w /w node:22 node --test scripts/update-robustness.test.mjs
  -> rc=1, 53.5 s
  -> 1..51
     # tests 51
     # pass 39
     # fail 1
     # cancelled 0
     # skipped 11
     # todo 0
```

Unchanged from the pre-remediation run, and I checked it two ways rather than trusting the totals: the same command against `d4051ec61`'s file in the same tree gives the same `51 / 39 / 1 / 11`, and a diff of the 51 per-subtest status lines between the two runs is **empty** — same single failure (`not ok 16 - without launchd or plutil the bound decides, and nothing is left behind`, the container-scheduling artifact the PR body already owns), same 11 skips, same order. A run of this command also died once at rc=137 (SIGKILL) while a second container shared the tree; rerun alone it is stable, so that was resource contention on this host, not the file.

damianvtran:

R6 fixed in `28d79139d` — **comment-only**, and you called both parts not round-worthy, so I am leaving the round fresh rather than spending another on prose: no code, no assertion, no test change.

Both corrections taken from your reading: (a) the container does **not** have `app-builder-bin` and this head builds its fixtures through `buildBlockMap`, so the module error belongs to the tree that ran it (head `7a46db736`), not to the runner — the comment now names which head each failure mode was measured on and stops implying the runner hits the 404; (b) "45 minutes" is replaced by the measured pair (43 in the first job I read, 185 in the longest), and the older catch comment that said 43 now agrees.

Re-verified in the node:22 container: `# tests 51 / # pass 39 / # fail 1 / # skipped 11`, unchanged.

Your R1 residual (the blockmap-copy loop sitting outside the catch) is accepted as a NIT, with your reproduction: a `createServer` that never listens cannot hang the run, and that case exits rc=0 in 0.449 s. Recorded here rather than re-guarded, since "cannot leak a listener" is the property the catch exists for.

damianvtran:
### QA report — round 1

QA: qa-tester subagent (team `lopdev`), model `deepseek/deepseek-flash` — no provider fallback during this round. Independent pass: no repo file was edited, every run used copies under `/tmp`, containers were killed as I went (`docker ps -q` empty at the end).

Scope: `d4051ec61..28d79139d` (one file, `scripts/update-robustness.test.mjs`). The head moved twice mid-round; every row below was re-run at `28d79139d`, and the two deltas are characterised: `d4051ec61 → 97a45329e` (the remediation itself) and `97a45329e → 28d79139d` (comment-only — verified mechanically, all 27 changed lines are comment lines and no changed line is code). Rows are labelled with the head their evidence came from, plus one row of CI-runner evidence that no container can replace.

Environment: macOS 25.6.0 / arm64 host, Docker Desktop with the cached `node:22` image; scratch trees in `/tmp/qa151` carrying this branch's file, with full `pnpm install --frozen-lockfile --ignore-scripts` deps; `#147`'s census fix taken from the `r6-repro` worktree (sha256 `26a5dd24…`, byte-identical to its source) for the merge-state rows. Plain `docker run` means PID 1 is `node`; that difference is load-bearing for M5.

| # | Surface | Command (abridged) | Actual output | Verdict |
|---|---|---|---|---|
| M1a | `origin/main:` file, whole file, Linux | `docker run … node:22 node --test scripts/update-robustness.test.mjs` | `rc=124` — killed by `timeout` at 300 s; **no `# tests` summary**; 51 results printed (44 ok / 7 not ok); process never exited | PASS (hang reproduced) |
| M1b | Branch file, whole file, Linux | same, file from `28d79139d` | `rc=1` in 54 s; `# tests 51 / # pass 39 / # fail 1 / # skipped 11`, `duration_ms 51430` | PASS (completes and reports) |
| M2a | Leak census, `origin/main:` file | file copy + appended diagnostic printing `process._getActiveHandles()` classified by constructor | after all 51 tests: `{"Socket":10,"Server":5}`, servers `127.0.0.1:42929, 37385, 33945, 41979, 46829`; **no `process-exit` record** (still alive at 300 s) | FAIL — the leak, pre-fix (expected) |
| M2b | Leak census, `28d79139d` file | same | after all 51 tests: `{"Socket":2}`, servers `[]`; then `process-exit:1` — the process exits on its own | PASS |
| M2c | R1 (`try` opened before `listen()`) | static read + direct call of `loRunUpdate({arch:"arm64"})` from an appended diagnostic, because its callers are all darwin-gated now and unreachable on Linux | `28d79139d`: throw `Cannot find channel "latest-linux-arm64.yml" … 404` → `serversAfter: ["null"]` (closed, unbound), `process-exit:0`. `origin/main:` same throw → `serversAfter: ["127.0.0.1:43737"]`, `rc=124`, no summary | PASS |
| M3 | Gate effect on Linux | `grep '^ok .*SKIP'`; read the six bodies | 11 SKIPs = the 6 new mac gates + 5 pre-existing; every gated subject is `new MacUpdater(…)` over a mac feed (per-arch zips + `latest-mac.yml`), and on Linux the updater asks for `latest-linux*.yml` → 404 | PASS — no body was Linux-passable |
| M4a | **Full CI suite** (this fix + `#147`), plain container | `node scripts/run-desktop-tests.mjs <the 37 files named by test:desktop, wrapper excluded>` | completes, `rc=1`, `703 tests / 688 pass / 1 fail / 14 skipped`, 98 s; sole failure the watchdog test (M5) | PASS (completes) / 1 env failure |
| M4b | Same, with a reaping PID 1 | same + `docker run --init` | completes, **`rc=0`, `703 / 689 pass / 0 fail / 14 skipped`** — CI's own assertion (`pass ≥ 60`, `fail == 0`) met | **PASS (green)** |
| M4c | Branch head alone (no `#147`), plain container — reproduces the runner | same, tree without `#147`'s file | completes, `rc=1`, `703 / 687 pass / 2 fail`: `not ok 469 - R6-1: an unreadable census cannot authorise a removal` + the watchdog test | PASS (reproduces CI, same test index) |
| M5 | The one container-only failure | 3× full suite under concurrency (4 and 7), 13× the file isolated (5× whole file, 5× `--test-name-pattern`, 3× via the wrapper), 1× with `--init` | under concurrency **3/3 fail**; isolated **13/13 pass**; `--init` **1/1 pass** | FAIL (environment) — not attributable to this diff |
| M6 | **CI runner evidence** (no container can substitute) | `gh api …/actions/jobs/103826860620/logs` (head `28d79139d`), and job `103777903178` (head `7a46db736`) | `28d79139d`: job **completed** 01:17:52→01:19:50Z (118 s wall), `desktop tests: 37 files, concurrency 3 …`, `703 / 688 pass / 1 fail / 14 skipped`, `duration_ms 94609`, exit 1 — the sole failure is `R6-1 …` in `scripts/submit-latency.test.mjs:709`, i.e. `#147`'s subject. `7a46db736` (19:21Z): `637 / 622 / 7 / 8`, six bodies red with `Cannot find module 'app-builder-bin'` (6 occurrences), and `ok 509 - without launchd or plutil the bound decides, and nothing is left behind` | PASS — first `Desktop Tests` completion since 19:24Z |

Repro for M5, with mechanism. Instrumented copy of the watchdog test logging `ps -eo pid,pgid,stat,args` filtered on the watchdog's own pgid, failing in a plain container under concurrency:

```
{"atMs":300,"empty":false,"members":["1261  1243 Z    [sh] <defunct>"]}
{"atMs":450,"empty":false,"members":["1261  1243 Z    [sh] <defunct>"]}   … never empty
```

A **defunct (zombie) child of the watchdog shell** sits in the group, so `processGroupIsEmpty`'s `kill(-pgid, 0)` never raises `ESRCH`. The plan backgrounds its relaunch; when that shim outlives the shell — which concurrency makes likely — it is orphaned and reparented to PID 1, which in a bare container is `node` and does not reap orphans, so the zombie is permanent. Control: the identical full-suite command with `--init` makes the test pass and the suite green, and on the runner the same test is `ok 509` because PID 1 there is a reaping init. It is a pre-existing, load-dependent race in a test this PR does not otherwise touch — not a Linux defect and not introduced here. Re-running QA on this suite should use `--init`.

Reproduction caveat (my own artifact, not a failure): mount `node_modules` **read-write**. With `:ro`, `scripts/picker-feedback.test.mjs` dies with `EROFS: open '/w/node_modules/.cache/picker-feedback/picker-host.mjs'`; writable, that file is 16/16 pass.

Findings:

- **Q1 (info, environment — no PR change required):** the zombie/PID-1 behaviour is the whole of the container's residual red, and it is the same test the runner passes. Pre-existing, outside this diff.
- **Q2 (verified):** the rewritten provenance comment holds up on both legs it now distinguishes. The runner leg is confirmed from job `103777903178` (six `Cannot find module 'app-builder-bin'` failures, before any feed existed); the container leg is what I reproduce on this head (`Cannot find channel "latest-linux-arm64.yml" … 404`). Its "last `Desktop Tests` ever to complete" was true when written and is now superseded by this fix's own job completing.
- **Q3 (info):** `skipUnlessDarwin(t)`'s `t.skip("macOS only")` matches the file's pre-existing macOS skips, and the runner's skip count moves 8 → 14 with the six gates — the gate is honest about what it hides. Nothing to change.
- **Q4 (info, not this PR):** CI is red at this head for `R6-1: an unreadable census cannot authorise a removal` in `scripts/submit-latency.test.mjs` — `#147`'s fix, not yet on this branch. My container reproduces it at the identical test index (469) and goes green the moment `#147`'s file is present.
- No blocker, major or minor finding on the diff itself.

Verdict: **PASS.** The claim is established, by execution: the pre-fix file never exits and leaves five listeners behind (rc 124, no summary, five `Server` handles and no `process-exit`), and the branch's file returns, prints `# tests 51 / pass 39 / fail 1 / skipped 11` and leaves no listener — including the throw path R1 protects, which I exercised by calling `loRunUpdate` directly, where the old file kept `127.0.0.1:43737` listening and the new one did not. On the real runner the hang is gone: `Desktop Tests` **completed in 118 s** on `28d79139d`, the first completion since 19:24Z, with the six mac gates skipped and one red that belongs to `#147`.

**Does the full suite complete?** Yes — 703 tests, summary printed, 98–100 s in a container and 94.6 s of suite time on the runner.

**Is it green?** Not on this branch alone, and not because of this PR: the single red is `R6-1 …` in `submit-latency.test.mjs` (`#147`). With `#147` on the tree at the merge state — and a reaping PID 1, as the runner has — it is green: **689 pass / 0 fail / 14 skipped**, satisfying CI's own gate. In a bare container without an init, the only extra red is the pre-existing zombie race in M5.

Gaps I could not cover: no macOS host, so the six gated bodies were not exercised (by design they are the only evidence for MacUpdater behaviour); the `app-builder-bin` runner leg is verified from the runner's own log rather than reproduced by me. Raw logs and JSONL censuses: `/tmp/qa151/variants/`.

damianvtran:
**Merging.** Evidence, stated plainly because CI is not green on this branch:

- **The hang is fixed on the real runner**: this head's `Desktop Tests` job **completed** in 1m58s (run 34795202703) — the first completion of that job since 19:24Z, and it is the defect this PR exists for. The suite now finishes and reports in ~95 s instead of holding a runner for 45+ minutes.
- **The one red is not this PR's**: `R6-1: an unreadable census cannot authorise a removal` in `scripts/submit-latency.test.mjs`, which is PR #147's fix and is not in this branch. QA measured the merge state (both files present) in the container: **rc=0, 689 pass / 0 fail / 14 skipped**, which satisfies this job's own gate (`pass >= 60`, `fail == 0`). Merging this first is what lets #147's job show that green.
- Rounds: agent review 1 (one major, fixed) and 2 (clean, terminal, fresh on 97a45329e; the subsequent change is comment-only and disclosed in the round-2 remediation comment) + QA round 1 **PASS** on `28d79139d`, with the runner's first completed job as its C1 row.
- No version bump, per this repo's release flow — test-only change, no user-visible effect, so no design/UX round applies.

Also worth recording for whoever reads this later: the stuck jobs before this were holding runners for up to six hours, which is why every other PR in this repo could not get a green suite either.

damianvtran:
### Released in v0.21.1

- **Tag:** `v0.21.1` at `2a5752b3ba30a46876bd569d585cde79fa41e2e0` (the bump PR #156's merge commit)
- **Release:** https://github.com/damianvtran/local-operator-ui/releases/tag/v0.21.1 — not a pre-release, 20 assets, and `/releases/latest` redirects to it
- **npm:** `local-operator-ui` `latest` = 0.21.1 (verified against the registry, not from the workflow's success); publish run [34808395982](https://github.com/damianvtran/local-operator-ui/actions/runs/34808395982) completed successfully
- **Window:** combined **patch** release covering #132, #147, #151 and #153 plus the version-only bump. Merged on the code-owner path with `--admin` — no human clicked approve; the recorded independent agent rounds served as the approval, as disclosed in the notes.

PR #159, opened 2026-09-14
fix(storybook): use react-docgen, so the preview builds on TypeScript 7

damianvtran:
# PR Description

## Summary of Changes

The code change is one line, in one file: in `.storybook/main.ts`,
`typescript.reactDocgen` moves from `"react-docgen-typescript"` to
`"react-docgen"` (with the comment that carries why).

The PR is **three files, not one**, and the body should not have implied
otherwise: that config line, the regenerated `docs/evidence/manifest.json`, and
one re-captured evidence frame. The manifest diff is where most of the lines
are, and most of those are formatting churn — 19 added / 18 removed lines around
the fields the capture rewrites. It is machine output from
`scripts/capture-evidence.mjs`, not hand-edited, and reducing it is not worth a
change to the tool in a one-line fix.

Without it, `pnpm storybook` cannot build the preview on `main`, so no story
renders and `scripts/capture-evidence.mjs` cannot produce a frame — which means
every UI PR that needs rendered evidence is blocked, not just this one.

## The failure, reproduced

On a clean worktree of `origin/main` (`7f0511dd4`):

```
$ pnpm exec storybook dev -p 6073 --ci

info => Starting manager..
info => Starting preview..
info Using tsconfig paths for react-docgen
=> Failed to build the preview
TypeError: Cannot read properties of undefined (reading 'React')
    at Object.<anonymous> (./node_modules/.pnpm/react-docgen-typescript@2.2.2_typescript@7.0.2/node_modules/react-docgen-typescript/src/parser.ts:112:19)
    at Module._compile (node:internal/modules/cjs/loader:1934:14)
    ...
WARN Broken build, fix the error above.
WARN You may need to refresh the browser.
```

It does not degrade to Storybook's error screen with a working manager: the
process **exits 1 and never listens**. `curl` against the port gets no
connection, `index.json` does not exist, and `pnpm exec storybook dev` fails the
same way with and without `--ci` — so there is no page to look at, and
`scripts/capture-evidence.mjs` cannot reach a server at all. That is why every
UI PR needing rendered evidence is blocked rather than merely inconvenienced.

## The cause

`#140` (`6f9bc2d04`, "move the desktop toolchain to Electron 44 and TypeScript
7") installed `typescript@7.0.2`. Storybook 8.6.12's generator,
`react-docgen-typescript@2.2.2`, cannot load that compiler: requiring the
package throws inside its own parser, which is why the stack names its
`parser.ts` and the TypeScript version in the resolved path. The entry in
`package.json` is untouched — this is the generator's compatibility, not the
compiler pin.

## The change, and why not pin the parser

```
 .storybook/main.ts | 14 +++++++++++++-
```

`react-docgen` is Storybook's own parser. It reads the source without loading
the TypeScript compiler at all, which is the whole of what this generator is
used for here — the props tables on docs pages. Nothing in the app, in the
tests, or in the evidence pipeline reads those tables, so the loss is limited
to prop-table fidelity. The alternative — pinning `react-docgen-typescript` to
an older parser — buys that fidelity back at the cost of a package whose
supported TypeScript range ends before the toolchain this repository now uses,
and which would break again on the next toolchain move.

## Related Issue(s)

- No issue; found while implementing a UI PR that needs frames. `#140` is the
  change that introduced it.
- The implementer of that PR (`feat/session-code-memory`) has rebased onto this
  branch's head so its reviewer and QA can run the same capture.

## Impact

- **Every UI PR that needs rendered frames.** Ten sessions have open PRs here
  and each one needs `capture-evidence.mjs`; on `main` today none of them can
  produce a frame, and a re-capture is how the "before" half of a fix is
  photographed.
- **No user-visible change.** `.storybook/` is development tooling and is not
  shipped in the app or the npm package.
- **No dependency, lockfile or version change.** `package.json` and
  `pnpm-lock.yaml` are untouched.

## Testing Details

1. **The failure**, on unmodified `origin/main`: the transcript above, exit 1.
2. **The same tree after the change**: `pnpm exec storybook dev -p 6073 --ci`
   serves — `curl -o /dev/null -w '%{http_code}' http://localhost:6073/` → `200`,
   and `http://localhost:6073/index.json` lists the stories (including the
   `canvas-workspace--*` ones this PR's consumer needs).
3. **A real frame from the real tool**, on the fixed tree:
   `node scripts/capture-evidence.mjs --only=design-system-primitives --themes=localOperatorDark --allow-backend http://localhost:6019`
   → `Captured 1 frames into docs/evidence at 68bef2346`, committed here as
   `docs/evidence/design-system-primitives/all-primitives/localOperatorDark.webp`.
   It is the same picture as the frame already committed for that story (a
   30-byte re-encode of a 205,676-byte webp), which is exactly what a generator
   change should look like: prop tables move, pixels do not.
4. **`pnpm check-evidence`**: `provenance: no failures` on this tree
   (`provenanceFailures(manifest)` → `[]`), and the frame half runs clean — log
   in the thread.
5. **`pnpm lint`, `pnpm check-types`**: clean, and identical to the base tree
   (`33 warnings`, all pre-existing).

## Deferred

- **`capture-evidence.mjs`'s cold-Vite theme deadline** (review U-01, real, out
  of scope here): the first capture against a freshly started Storybook fails
  with `document carries theme "" after 10s` and `EXIT=1`, while the identical
  command against a warm server succeeds — the dev server's first compile
  outruns the 10 s theme deadline, so a run needs one warm-up load of the story
  first. Fix: give the capturer a warm-up load (or raise the first deadline)
  before it asserts the theme. **Follow-up PR**, not this one: it costs a retry,
  not a wrong frame, and this PR is a one-line toolchain fix.

Not addressed, and deliberately out of scope for a one-line tooling fix: the
capturer's own papercut, found while producing (3): the first frame of a run
against a cold Vite dev server exceeds the 10 s theme deadline, so a capture
needs one warm-up load of the story first. It costs a retry, not a wrong frame.


damianvtran:
### Agent review — round 1

**Reviewer:** `reviewer` subagent (Local Operator / `lop`), model `deepseek/deepseek-flash` — no provider fallback during this round.
**Scope:** `7bc101fbb` (`main`, 0.21.2) .. `b16b92c02` (`fix/storybook-react-docgen-ts7`) — 3 files, 33 insertions / 19 deletions: the one functional `.storybook/main.ts` line, its explanatory comment, and the evidence re-stamp. Base and head were both tested in pristine detached worktrees of their own; ports 6117/6118 (6010–6060 are contested on this host).

**Verdict: clean — no BLOCKER, no MAJOR. This round is TERMINAL.** U-01..U-04 are follow-ups, not another round.

#### What I ran, and what came back

Base, unmodified `main` at `7bc101fbb`:

```
$ pnpm install --frozen-lockfile                                    # EXIT=0
$ pnpm exec storybook dev -p 6117                                   # EXIT=1 after 46s
info Using tsconfig paths for react-docgen
=> Failed to build the preview
TypeError: Cannot read properties of undefined (reading 'React')
    at Object.<anonymous> (./node_modules/.pnpm/react-docgen-typescript@2.2.2_typescript@7.0.2/node_modules/react-docgen-typescript/src/parser.ts:112:19)
$ curl -s -o /dev/null -w '%{http_code}' http://localhost:6117/     # 000
```

The failure is real, and fatal to the process rather than only to the preview — see U-02. No port is ever served, so on `main` no story renders and no frame can be captured at all.

Head, `b16b92c02`:

```
$ pnpm exec storybook dev -p 6118
$ curl -s -o /dev/null -w '%{http_code}' http://localhost:6118/                       # 200
$ curl -s http://localhost:6118/index.json                                            # 200, 46179 bytes, 157 entries
     21 canvas-workspace--* / design-system-primitives--* entries; iframe.html 200 for
     canvas-workspace--code and design-system-primitives--all-primitives
$ pnpm build-storybook                                                                # EXIT=0, "✓ built in 7m 41s", 26 MB, 3 docs entries
$ node scripts/capture-evidence.mjs --only=design-system-primitives --themes=localOperatorDark \
      --allow-backend http://localhost:6118
Captured 1 frames into docs/evidence at b16b92c02                                     # EXIT=0
$ node scripts/capture-evidence.mjs --only=canvas-workspace--code --themes=localOperatorDark \
      --allow-backend http://localhost:6118
Captured 2 frames into docs/evidence at b16b92c02                                     # EXIT=0 (code + code-focused; --only is a substring match)
```

Both runs reproduced the committed frames **byte-for-byte** (`design-system-primitives/all-primitives/localOperatorDark.webp` sha256 `0a178323…`, `canvas-workspace/code/localOperatorDark.webp` sha256 `9443c075…`; after each run `git status docs/evidence` showed only `manifest.json` touched, which I restored). `capture-evidence.mjs` runs `assertFramePaints` before writing — dominant colour within ΔE00 of the theme's ground and the frame *not* one uniform colour — so "Captured N frames" is evidence the preview actually painted a story, not merely that nothing errored. This story frame is exactly what the stacked canvas PR needs.

Prop tables survive the generator swap. From the built bytes:

```
LowCreditsDialog.__docgenInfo = props:{open:{required:!0,tsType:{name:"boolean"}},onClose:{…,raw:"() => void"},
                                       onGoToConsole:{…,raw:"() => void"}}
UpdateNotification.__docgenInfo = props:{autoCheck:{…,description:"Whether to automatically check for updates on mount",
                                       defaultValue:{value:"true"}}…}
```

`CheckForUpdatesButton` gets `description` + `displayName` and no `props` key, which is correct — its source is `export const CheckForUpdatesButton = () => {`. 237 `__docgenInfo` occurrences across the built assets; the three autodocs entries are in `index.json`.

Also verified:

- **Nothing else assumes the old generator.** `git grep -n "react-docgen"` → `.storybook/main.ts` and `pnpm-lock.yaml` only; the TS package stays a transitive dep of `@storybook/react-vite`, and `@storybook/react-vite/dist/preset.js` requires its plugin only when the option names it. `git grep "__docgenInfo" src scripts docs .storybook` → nothing: no test, capture or evidence path reads prop tables.
- **The generator change cannot move a committed pixel** — `capture-evidence.mjs`'s STORIES list is stories only, no docs pages (`grep -n 'viewMode=docs'` → none), and the checker's own provenance assertion re-run on this head returns `provenanceFailures(manifest) -> []`, exactly as the PR claims. (Its surrounding 1163-frame ImageMagick sweep is still grinding through this session; the frames this PR touches are the ones I asserted above.)
- **The re-encoded frame is the same picture.** `ffmpeg … ssim` → `SSIM All:0.999978`; per-byte PPM diff → 6,816 of 6,399,616 pixels differ, only 36 by more than 20/255 (max 142). A subpixel re-render/re-encode, not a content change.
- **Scope.** `git diff --name-status 7bc101fbb..b16b92c02` → `.storybook/main.ts`, `docs/evidence/manifest.json`, one `.webp`. `package.json` and `pnpm-lock.yaml` untouched. `.storybook/main.ts` is 14+/1- = the one functional line plus a 13-line comment.
- **`pnpm test:desktop` claim, independently reproduced** (all `CMUX_*` unset). Base: `ℹ tests 737 / pass 734 / fail 3`. Head: `ℹ tests 737 / pass 734 / fail 3`. The failing sets are identical modulo timings — "the installer's script spawn carries the prefix into the venv it creates" and "the update service's python probe runs with the prefix in the child's own environment" (`scripts/python-bytecode-cache.test.mjs`) and "the operator's own install is classified rather than told to use pip" (`scripts/update-robustness.test.mjs`, `AssertionError: 'pip' !== 'uv'`). The PR's attribution is exact and it adds no failure.
- **`pnpm lint`** → `Found 33 warnings`, EXIT=0 on both trees. **`pnpm check-types`** → EXIT=0 on both. (Noted, not a defect: `.storybook/main.ts` is not in `tsconfig.node.json`'s `include`, so CI's "Type Checking" does not cover this file — its value is validated by running Storybook, which the evidence commit does.)
- **CI and mergeability.** All 8 checks `completed/success` on `b16b92c02` (Lint, Type Checking, Runtime Dependencies, Security Audit, Desktop Tests, Version Bump Guard, NPX Sanity ×2); `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`. "Without a force-push bypass" is trivially true here: `gh api repos/.../branches/main/protection` → 404 "Branch not protected", `gh api repos/.../rulesets` → `[]`, `gh api repos/.../rules/branches/main` → `[]`. There is nothing to bypass — which is precisely why this round is the only real control, per `AGENTS.md`.

**On the choice, since the PR rests on it:** I would defend `react-docgen` over pinning the parser. `react-docgen@7.1.1` depends only on `@babel/*`, `doctrine` and `resolve` and never loads the TypeScript compiler; `react-docgen-typescript@2.2.2` declares `peerDependencies: {typescript: ">= 4.3.x"}` while being built and tested against TS 4.3 and reaching into compiler internals (`parser.ts:112`), so keeping it means holding `typescript` below 7 — undoing #140 — to preserve prop tables that only three autodocs components consume. `react-docgen` is also what `@storybook/react-vite` defaults to.

#### Findings

**U-01 — MINOR (disclosed, recording it as a follow-up)** — `scripts/capture-evidence.mjs:947-955`: the first frame of a run against a cold Vite dev server blows the 10 s theme deadline (40 × 250 ms) and the run dies:

```
Error: design-system-primitives--all-primitives @ localOperatorDark: document carries theme "" after 10s
EXIT=1
```

The byte-identical command on the now-warm server: `Captured 1 frames … EXIT=0`. The failure is clean — no frame written, `docs/evidence` left unmodified — so the cost is one retry, exactly as the PR body states. I record it because this PR exists so that other sessions can produce frames, and their *first* attempt will fail with an error the tool does not explain: it reads as "the fix did not work". Cheap follow-up in a PR of its own (not this one): warm the first story once before the deadline window opens, or retry the first frame once. When that lands, it is worth checking whether this option made cold start worse — for `"react-docgen"` SB's preset installs its own Babel docgen plugin over `/\.(mjs|tsx?|jsx?)$/`, where the TS plugin ran only over the `.tsx` set, so the first cold transform now does strictly more per-module work.

**U-02 — NIT** — The body's "the manager and the browsers get a page … so every story is Storybook's error screen" is stronger than the symptom on this host: on unmodified `main`, `storybook dev` (with and without `--ci`) prints the `TypeError` and **exits 1**, and no port ever listens (`curl / → 000`). There is no error screen to photograph. The conclusion is unaffected — no story renders and `capture-evidence.mjs` can produce nothing — but the "before" sentence should say the process dies.

**U-03 — NIT** — Body-versus-diff drift, twice: "One line in `.storybook/main.ts`" with a diff stat for that file only, while the PR is three files (the 13-line comment, the manifest re-stamp, the re-encoded frame); and the capture transcript quoted in Testing Details says `at 7f0511dd4` — an ancestor of `main` — where the committed manifest records `refreshedAtHead: 68bef2346`. Both are explicable (an earlier run on a working tree carrying the fix versus the committed re-run), but a reviewer holding only the description cannot tell which tree produced the committed frame. I settled it by reproducing the frame at `b16b92c02`, so nothing is actually in doubt.

**U-04 — NIT** — `docs/evidence/manifest.json` carries 19/18 changed lines, of which all but about seven are formatting: the `chat-search` supplementary entry re-indented (same parent on both sides — I parsed both blobs, `supplementary[21]` in each) and `\u2014` escapes normalised to literal em dashes. Harmless, since the file is tool-written and `provenanceFailures` re-runs `[]`; noted only as churn a reader has to page through.

Out of scope and left alone, agreeing with the PR: the manifest's historical `headNote` still narrates `dirtyWorkingTree` as true while the field is false — identical on both trees, so pre-existing rather than introduced here.

damianvtran:
Closing as superseded — do not merge.

`main` fixed this independently while this PR was in its rounds: `6a952c469` ("build(storybook): use the JSX docgen, which the TypeScript 7 move left working") carries the same `reactDocgen: "react-docgen"` change in `.storybook/main.ts`, with a better comment, and it is already on the base this PR needs. `main` has also moved twice past this PR's base (0.21.4), so the PR is now CONFLICTING on `docs/evidence/manifest.json`; merging a duplicate one-line fix would mean re-deriving its manifest record against a base it no longer matches for no behavioural gain.

Record of the rounds, so nothing is lost:
- `### Agent review — round 1` was terminal and clean (no blocker/major) on head `b16b92c02`: base reproduced the failure (`pnpm exec storybook dev` → `Failed to build the preview` / `Cannot read properties of undefined (reading 'React')` at `react-docgen-typescript@2.2.2_typescript@7.0.2`, never serves), head serves (`GET / → 200`, `GET /index.json → 200`, 157 entries), a frame reproduces byte-identically, and `pnpm test:desktop`'s 3 failures are identical on the unmodified base. U-02/U-03/U-04 (body inaccuracies) were corrected in the body.
- One finding survives as a real, unfixed defect and is recorded here rather than silently dropped: **U-01 (minor)** — `scripts/capture-evidence.mjs`'s cold-Vite ~10 s theme deadline makes the first capture against a freshly started Storybook fail (`document carries theme "" after 10s`, EXIT=1) while the same command against a warm server succeeds; the failure is clean (no partial writes). A warm-up or single retry in the tool is the fix, and it belongs in its own PR.

The branch is left in place for the record; nothing here is needed by the sibling work.

PR #152, opened 2026-09-14
fix(chat): read the durable tail back after every snapshot, so rows written during an absence load

damianvtran:
## What this is

In a running conversation, the user steered a turn in flight, switched to another conversation, and switched back: **every row written after the steering message — the assistant rows and the tool calls between the steer and the return — was missing**, and only frames arriving after the return painted ("I only start getting new messages past that point").

The cause is a snapshot page that cannot be verified from the frame that carries it. The backend reads a re-subscribe snapshot's history page `through_id=<the owner's published history_cursor>` (`local-operator/local_operator/server/utils/desktop_sessions.py:520-540`), so the page's **newest entry IS that cursor by construction** — the two agree even when the cursor is stale, and a stale cursor is exactly what a mid-turn steer publishes (`Session.steer_message` calls `refresh_frontend_state`, `local-operator/local_operator/session/session.py:5247`). A page stopping short of the durable tail is therefore **non-empty and reports `cursor_missing: false`**, and the flush loop's `needsReconcile` accepted that shape as complete. Nothing ever asked for the rows between the page's end and the true tail, so they were never delivered.

The previous gate keyed on evidence of *missing arguments* (`seedCallsMissingLabels`) and on *empty or replaced* pages — never on **coverage**. That is the bug this fixes.

The steer is incidental: a plain background turn loses its rows identically, which the regression suite asserts as its own case.

## Why the fix is a tail read and not a smarter comparison

The frame already carries both `history.entries` and `frontend.snapshot.history_cursor` (`src/shared/desktop-session-contract.ts:205,299-303`), so the obvious cheap guard is "reconcile only when the page's newest id is behind the published cursor". **That check can never fire**, and this is measured rather than argued — against the real reader (`local_operator.session.transcript.read_transcript_page`):

```
durable ids (oldest->newest): ['f0b61bc9', 'c0e8741c', 'dd1b2c8b', '9fa5c885', 'e0c85149']
page(through_id=c0e8741c):
  newest entry == the cursor that bounded it? -> True
  ids delivered: ['f0b61bc9', 'c0e8741c']
  cursor_missing (reconciled): False
  rows 3-5 delivered? False
unbounded tail (the reconcile route, no through_id):
  ids: ['f0b61bc9', 'c0e8741c', 'dd1b2c8b', '9fa5c885', 'e0c85149'] | reconciled: False
fresh cursor page: newest == cursor? -> True
```

[...]

PR #161, opened 2026-09-14
chore(release): bump version to 0.21.3

damianvtran:
## What

Release-window bump for **v0.21.3** (patch). One file, one line: `package.json`'s `version` goes `0.21.2` → `0.21.3`. No code, docs, evidence or test changes ride in this PR.

## Window contents

Everything merged on `main` since tag `v0.21.2`:

- `d06a66053` — #152 `fix(chat): read the durable tail back after every snapshot, so rows written during an absence load` — the reported bug where a conversation lost every row written after a mid-turn steering message once the user switched to another conversation and back. Client half; the backend half shipped in local-operator v0.54.39 (local-operator#1090).

`Release: patch` — one real user-visible fix, no new capability, so no minor. Nothing else was merged in the window when this bump was derived.

## Scope check requested

This PR is intentionally one file and one line. Please verify: (1) `git diff origin/main -- package.json` shows only the version line; (2) `0.21.3` is the correct next patch after `0.21.2`; (3) no other file, lockfile or generated artifact is touched.


damianvtran:
### Agent review — round 1

Reviewer: reviewer (lopdev team subagent) — model `deepseek/deepseek-flash`
Scope: `main@d06a66053`..`d711afcda` (C0 scope check: one file, one line)

All four scope questions verified **from the git objects**, not from the PR description. Verdict: **clean**.

#### 1. `git diff origin/main..d711afcda --stat` — `package.json` is the only file

```
 package.json | 2 +-
```

Raw diff, in full:

```diff
diff --git a/package.json b/package.json
index 7688e45d4..e03583657 100644
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "local-operator-ui",
-  "version": "0.21.2",
+  "version": "0.21.3",
   "productName": "Local Operator",
   "description": "User interface for AI agent assistants that run Python code safely on your device through an intuitive chat interface. Supports local and cloud LLM models with built-in security features.",
   "main": "./out/main/index.js",
```

`d711afcda` has exactly one parent, `d06a66053` == `origin/main`, so the branch is a single commit on top of the window head and the range diff above is the whole PR.

#### 2. `0.21.3` is the correct next patch after `0.21.2`

- `git tag`: `v0.21.0`, `v0.21.1`, `v0.21.2` — no `v0.21.3` locally, and `git ls-remote origin refs/tags/v0.21.3` returns nothing, so no tag exists remotely either. No duplicate, no yank.
- `v0.21.2` → `7bc101fbb`, the previous bump PR (#157).
- `gh release list` and the releases API show `v0.21.2` as the highest published Release, `Latest`, no drafts, no prereleases — so no patch number was consumed without publishing, and no skipped patch is already out.
- `0.21.3` is therefore the exact next patch; nothing between is burned.

#### 3. Nothing else in `package.json` moved

Full-file comparison of `origin/main:package.json` against `d711afcda:package.json` with the `version` line stripped is **byte-identical** (both sides 361 lines, one hunk, 1 insertion / 1 deletion). So no dependency pin, no `scripts` entry, no `test:desktop` entries, no electron-builder/`build` config, no `devDependencies` change. The 361-line length is unchanged on both sides.

`pnpm-lock.yaml` and `pnpm-lock.yaml`-adjacent artifacts are untouched, which matches the repo's own pattern: the previous release commit `7bc101fbb` ("bump version to 0.21.2") also touched `package.json` alone. The lockfile carries no root-package version field (`grep` for `0.21.2` in `pnpm-lock.yaml`: no hits), so no lockfile drift is owed.

The repo's own `version-bump-guard` job (`scripts/version-bump-guard.mjs`, run on the PR's merge ref) passes on this head, which independently corroborates both the one-file scope and the bump being inside a release PR.

#### 4. The window claim is accurate

```
$ git log --oneline --merges v0.21.2..origin/main
d06a66053 Merge pull request #152 from damianvtran/fix/steer-switch-history-gap
```

Exactly one merge on `main` since tag `v0.21.2`, and it is #152. The other seven commits in `v0.21.2..origin/main` are that PR's own branch commits, not separate merges — so "exactly one change" holds. `git merge-base --is-ancestor v0.21.2 d06a66053` → yes, so the range is a true single window with no earlier release straddling it. The PR body names `d06a66053` and #152 with its merge SHA, matching the object graph. PR body claim and history agree.

#### 5. Risk in tagging this head

Low, and I see nothing that a tag should not name:

- The head is a clean, single-purpose bump whose parent is exactly the window head `d06a66053` — a straightforward `--target d711afcda` tag+Release names precisely the window, with no unrelated commit swept in.
- `publish.yml` runs on `release: published`, and `scripts/validate-release.mjs` reads `package.json` **at the tag SHA** via the contents API and fails on `pkg.version !== tag.slice(1)`. Tagging `d711afcda`, whose tree carries `0.21.3`, satisfies that check; tagging any earlier commit would not. The release must still be created as tag+Release in one step (the workflow's trigger is the published Release, not a tag push), and the tag must resolve to a commit — a tag on a non-commit object fails validation.
- Operational precondition, not a defect in the diff: at review time GitHub reports `mergeStateStatus: UNSTABLE` with `Desktop Tests` and both `NPX Sanity Check` jobs still pending. Let those finish green before creating the Release, since the publish job's validation is the only thing between a tag and the artifact build.

### Findings

None — no BLOCKER, no MAJOR, no MINOR, no NIT. The diff is one line and the window claim is backed by the object graph. Round 1 is **TERMINAL** for review purposes; no remediation round is required.

PR #141, opened 2026-09-13
fix(update): keep an in-flight install alive through a relaunch, and tell the truth about it

damianvtran:
## Summary of Changes

An update install must survive the app being opened, and must never be reported as failed while it is still running. This is the operator's report — *"when I do the update, it seems to download and when I click install it closes the app but then the app fails to start back up with the updated version"* — worked from the machine's own logs.

What happened on 2026-09-13 (0.19.4 → 0.19.5, host at load ~95): ShipIt needed **09:39:22 → 09:44:07** to move and code-verify the ~1 GiB bundle, and Squirrel's last validation before it swaps asks whether *any* instance of the target app is running. The operator reopened the app **from Spotlight at 09:43:58** because nothing appeared to be happening (`parentASN="Spotlight"` in the launch record) — and that instance is what aborted the install.

Our own behaviour is where the rest of it lives, and it is two things, one of which fired and one of which did not:

- **start-up recovery read "still on the old version" as failure, and this one did fire.** It cleared the marker, removed the *live* install's launchd job (`09:44:05.377 ... Removed the leftover install job com.local-operator.ShipIt`, ~2 s before ShipIt's own abort) and told the user the update had not completed.
- **the relaunch watchdog's 600 s bound was a latent second way to abort the same install, and it had not fired.** The marker — and so the bound — starts at 09:39:00.991, i.e. 09:49:00, and the log at **09:44:05.284** still reads `Relaunch watchdog 32413 is still running; leaving it to finish or hit its deadline`. Had the operator waited, our own `open -a` at that bound would have been the instance Squirrel's final validation saw: any install slower than the watchdog would have been killed by the watchdog and then reported to the user as a failed update. This PR closes that path too, so the two ways to abort it are both removed.

Nothing had told the user to leave the app closed, either: the panel said the app would restart, and the reopened window is the fact that aborted the install.

**What changed**

- **In flight is a state, not a failure.** `evaluatePendingInstall` gains an `in-flight` outcome, ordered after `succeeded` and `stale`, supplied by the caller as "the install's ShipIt job is loaded AND a marker for this target exists AND the marker is recent": `isInstallInFlight`, whose recency bound is a named constant because it is the line between a live install and the job a *failure* left respawning (0.17.0: `runs=3114`, loaded for hours until it was removed by hand). The job is asked of launchd by label through the same probe the watchdog uses.
[...]

PR #129, opened 2026-09-13
feat(chat): replace the run-details popover with the run sidebar

damianvtran:
## What this is

The operator's ask, in three parts: a **persistent pane** instead of the transient popover ("it goes away after you click it"), a way to **read a child's own conversation**, and **one to-do list** rather than two spellings of it. Plus the MCP-status ask that arrived with it: "there doesn't seem to be a robust way to know which MCPs are connected or not".

`docs/run-sidebar.md` is the frozen design and it lands with this PR. This is the UI half; the backend half (`subagents.transcript`) is the parallel PR in the `~/local-operator` repo.

## Surface by surface

**The pane.** `isRunPanelOpen` and `runPanelWidth` join `ui-preferences-store` (both persisted, like `isCanvasOpen`), with mutually exclusive setters — `setCanvasOpen(true)` closes the run panel and vice versa, so "two panes open" is unreachable by construction rather than by discipline. `chat-content.tsx` mounts the pane in the canvas's own slot, using the canvas's own `ResizableDivider` (whose `aria-label` is now a parameter, as § 3.2 requires), the same `border-l border-hairline`, and the 420 / 320 / 640 width rules. The panel root is a `<section aria-label="Run details">` with a 40px `sunken` chrome bar — the canvas's idiom, because the two panes are two modes of one slot.

**The trigger.** The gate is now `details !== null`: the button is on screen in a canonical session with nothing in flight, which is the state the retired trigger did not exist in. It is an `aria-pressed` toggle with **four grounds for four states** — `canvas` closed at rest, `elevated` closed hovered, `accent-wash` open, `accent-wash` open hovered — because the ghost variant's hover already owns `accent-wash` and a pressed branch that overrode only the ink would let the pointer replace the open state's own ground the moment it stopped on the button that had just opened the pane. The popover, its Radix portal and the portal Tab-walk are deleted rather than left unused.

**The failure dot.** `acknowledgeOnOpen` / `acknowledgeOnClose` are deleted with the transitions they were bound to, and `acknowledgeWhileOpen` replaces them. The pane reports `listOnScreen` and `readerChildId` up (it is the only component that renders both the trigger and the pane), so the rule is: **the list is on screen** → every failed child whose row is in the rendered slice, and every problem server the section renders, is acknowledged; **a reader is on screen** → only the child that reader shows. A failure behind `+N more` keeps its dot, and so does a server that broke while a reader was covering the section.

[...]

PR #121, opened 2026-09-13
feat(chat): make pending ask options clickable

damianvtran:
## What and why

Release: patch — a pending `ask` card's options become real buttons you can click or tab to, and a bare numeral typed into the composer now reaches the model as that option's label instead of as the literal string.

The operator reported the pending `ask` card as a dead numbered list: an
accent-washed "Waiting for your answer" callout, then the options rendered as
inert muted text — `1. Popup is open — generate the pairing code` — and
"Type your answer below." Nothing was clickable. The only way to answer was to
retype a label into the composer.

Worse, the numerals were a lie. The card prints `1.`, `2.`, `3.`; typing `1`
sent the **literal string "1"** to the model, which has no way to tell which
option that indexes. A list that enumerates choices, invites a numeral, and
then discards it is a dead affordance wearing a live one's clothes.

The parity target is the terminal's own ask picker
(`local_operator/tui/widgets/ask_picker.py`), the in-repo surface that already
solved this: one card, options selectable by keyboard **and** mouse, digits
`1`–`9` jumping to an option, and — the part that sets the wire contract — it
answers **with the label the model wrote, not an index**. Its module docstring
says so directly, because its free-text row can return a string that was never
in `options`, which an index cannot express.

**Change class: C2 — standard, pre-authorized.** User-visible, so it carries a
design round in addition to code review and QA. This PR does not change
`package.json`'s `version` line: it stays byte-identical to `origin/main`'s
(`0.21.3`), and the release owner cuts one bump for the whole window.

## The change

- **Every option is a real control.** A `<button type="button">` that submits
  that option's **label** on click/Enter/Space through the same
  `sessions.answer` path the composer already uses (`value` = label,
  `questionIndex` = `gate.question_index`). The consequence line stays visible:
  both the phone card and the terminal picker show it, and it is frequently the
  only thing distinguishing two options whose labels are near synonyms.
- **The numeral trap is closed, on the text the user actually typed.** A bare
  `1`–`9` — with or without the trailing period the card itself prints, and
  tolerant of surrounding whitespace — resolves to that option's label
  (`resolveNumericAnswer`). **The rule runs on the composer's typed text, not on
  what goes on the wire**: a staged reply makes the payload
  `<reply-to>…</reply-to>\n2`, which is not a bare ordinal, so resolving the
  payload left the wrapped numeral as the answer value to a one-shot gate. The
  rule is otherwise deliberately narrow — bare, in range, `ask` with options —
  because rewriting prose into a label the user did not choose is worse than the
[...]

PR #163, opened 2026-09-14
chore(release): bump version to 0.21.4

damianvtran:
## What

Release-window bump for **v0.21.4** (patch). One file, one line: `package.json`'s
`version` goes `0.21.3` → `0.21.4`. No code, docs, evidence or test changes ride
in this PR.

## Window contents

Everything merged on `main` since tag `v0.21.3` (`90c96861c`):

- `54e78e278` — #141 `fix(update): keep an in-flight install alive through a relaunch, and tell the truth about it` — the reported failure where an update downloaded, the app quit for it, and the new version never came up. Worked from the machine's own logs (ShipIt `errSecCSBadBundleFormat`/`App Still Running`, the relaunch watchdog's 600 s bound firing into the install, and start-up recovery reading "still on the old version" as a failure, clearing the marker and removing the live install's launchd job two seconds before ShipIt aborted). The install is now a state rather than a failure, the watchdog holds instead of relaunching into it, recovery leaves a live install alone, and the user is told to keep the app closed.
  `Release: patch — an update install is no longer relaunched into, killed, or reported as failed while it is still running; the app now says to keep it closed, and a window close during an install quits instead of leaving a window-less process.`

Nothing else has merged in the window at the time this bump was derived. Two PRs told me they may land after this tag and will ride the next window: #129 `feat(session-sidebar-panel)` (argues minor) and #121 `feat(chat): make pending ask options clickable` (argues minor). I am not holding this window for either.

## Scope check requested

This PR is intentionally one file and one line. Please verify: (1) `git diff origin/main -- package.json` shows only the version line; (2) `0.21.4` is the correct next patch after `0.21.3`; (3) no other file, lockfile or generated artifact is touched.


damianvtran:
### Scope check — C0, one file

Run by the window owner (this session) from primary sources, not from the PR description:

```
$ gh pr diff 163 --patch | head -14
 package.json | 2 +-
-  "version": "0.21.3",
+  "version": "0.21.4",

$ git show v0.21.3:package.json    → version 0.21.3    (so 0.21.4 is the next patch)
$ git show release-next-0214:package.json → version 0.21.4
$ gh pr view 163 --json files → ["package.json"]        (no lockfile, no generated artifact)
```

All three scope questions answered: one file and one line; `0.21.4` is the correct next patch after `0.21.3`; nothing else is touched. The window's contents and the argument for a patch rather than a minor are in the body.

PR #174, opened 2026-09-14
fix(update): stop the app unsealing its own bundle, and repair one that is already broken

damianvtran:
## What happened

The in-app update 0.22.1 -> 0.22.2 completed (`ShipIt_stderr.log`: `Installation completed successfully`) and macOS then refused the app: *"Local Operator" is damaged and can't be opened. You should move it to the Trash.* One file had appeared in the installed bundle between signing at 09:31:53 and 10:11:29:

```
Contents/Resources/python_aarch64/lib/python3.12/__pycache__/webbrowser.cpython-312.pyc
```

`codesign --verify --deep` answers `file added: .../webbrowser.cpython-312.pyc` plus `a sealed resource is missing or invalid` (exit 1). The update pre-flight had passed its own seal check at 09:48:42 and was right to - the break did not exist yet.

## Root cause

Two halves, and neither is the seal check:

1. **The seal on the interpreter trees is applied at runtime, by the app**, so every in-place update swaps in a fresh bundle that carries no access-control entries. Measured on the shipped artifact: `ls -lde` on `Contents/Resources/python_aarch64` and on `.../lib/python3.12` prints a bare mode line, and a write succeeds at every level. The bundle macOS called "damaged" was born writable. The log says the same from the other side: runtime seals at 08:37:24, 08:38:23, 09:03:28 and 09:32:10, an update replacing the bundle at 09:48, and nothing after it.
2. **The interpreter those trees belong to is the one the app-managed venv is built on.** `~/Library/Application Support/Local Operator/local-operator-venv/pyvenv.cfg` records `home = /Applications/Local Operator.app/Contents/Resources/python_aarch64/bin`, so *any* process that runs that venv's python - the operator's shell, a CLI script, launchd, an agent session, an unpackaged worktree instance's backend - resolves the stdlib inside the installed bundle and writes bytecode into it. The environment guards reach only the pythons this app spawns. Nothing came back to report it: every `Update service initialized` line after 09:50 is `Dev mode: true`, and the 0.22.2 pending marker is still on disk.

Reproduced end to end on a copy of the signed image: one unguarded `python3 -c "import webbrowser"` inside the shipped bundle writes 32 `.pyc`; `codesign --verify --deep` goes 0 -> 1 with `file added: .../webbrowser.cpython-312.pyc`; the app's own `healPythonBytecode` takes it back to `valid on disk`; the build step's seal then makes the same write refuse with `EACCES` and the bundle still verifies.

## What changed

**1. The invariant: no python we start, and no python that uses the venv we manage, can write bytecode into a bundle**

- `withPythonBytecodeCache` now sets `PYTHONDONTWRITEBYTECODE=1` beside `PYTHONPYCACHEPREFIX`, unconditionally (measured on CPython: `0` and `""` are its two falsy spellings, so a stale `export` would silently restore the writes). It is the one builder every python spawn in the main process goes through.
- The three install scripts set the same flag, so a standalone run of them - venv creation and pip, which is what compiled the operator's in-bundle cache - is covered with no app in the loop.
- `ensureVenvBytecodeGuard` writes a `sitecustomize.py` into the app-managed venv's `site-packages`, which is the half that covers processes the app never spawns, whoever started them. It never replaces a `sitecustomize.py` it did not write. The installer runs it at every start and right after an install, so installs already on disk are covered without a reinstall.
- The contract test now **enumerates every child-process spawn site** in `src/main` (a table it walks, keyed by file, call and index) and fails for a site that is not in it, so the next spawn site cannot quietly omit the guards. Both variables are also asserted on the environment the driven spawns actually receive.

**2. The shipped artifact is born sealed, and a build that can write into its own interpreter trees fails**

- `scripts/after-pack.mjs` (prune, then seal; still `afterPack`, still before signing) applies the app's own seal - bundled from `src/main/python-bytecode-cache.ts`, so there is one implementation - to every directory under the bundled interpreter trees. Real `electron-builder --dir --arm64` run on this branch: `Sealed the bundled interpreter trees against bytecode writes: .../python_aarch64; 134 directory path(s) now refuse new entries`, the same count the field log reports for the installed 0.22.2.
- `scripts/verify-macos-artifacts.mjs` gained `app-python-trees-sealed`: it attempts the write in every directory under the trees, and the rewrite of every `.pyc` one still carries, and fails the release when any of them accepts it. Against the shipped 0.22.2 copy it fails (`134 of 134 path(s) accept a new file`); against the build of this branch it passes. Access-control entries are not sealed resources, so the sealed copy still verifies (`valid on disk`, exit 0).
- `Contents/Resources/python_aarch64/lib/python3.12` - the directory the field break landed in - is covered, and the new test asserts that depth specifically rather than the tree root.

**3. Field repair for an install that is already broken**

- Start-up probes the running bundle's seal 15 s after the window is created (unref'd, because the probe is a real `codesign --verify --deep`: measured 4.1 s and 4.3 s on the shipped app here), heals a `file added:` bytecode break with the existing `healPythonBytecode`, re-probes, and logs what happened. An unhealable bundle goes to the existing refusal/remedy panel through the delivery scheduler the other start-up notices use, rather than a push no subscriber saw.
- The same pass heals a bundle a *different* instance can reach: the one the app-managed venv's interpreter lives in, read from that venv's own `pyvenv.cfg` - which is what repairs an installed app from a dev instance. It heals that bundle and deliberately never seals it.
- The pre-flight's probe, retry and heal are now one shared implementation (`readBundleSeal` / `repairBundleSeal`); the update-time policy on top of it is unchanged, and there is still exactly one seal heuristic.

**4. No silence, and one app's state**

- An unpackaged instance no longer acts on the packaged app's install state: `recoverPendingInstall` names the marker it found and touches nothing, and `quit-and-install` refuses to write a marker for a bundle it is not. The predicate is `app.isPackaged` rather than the coincidence of paths, because these instances share the packaged app's *log* (the macOS log path is hardcoded) and may share its `userData`.
- An unpackaged instance no longer uses the packaged app's venv: `managedVenvPath` gives it `local-operator-venv-dev`, so its backend cannot import the installed bundle's stdlib at all. Cost, stated: an unpackaged instance with no venv of its own asks for the one-time backend setup, once per machine - a machine with a global `local-operator` never reaches that path.

## Visual surface

No visual surface changes: the refusal a start-up heal cannot repair is the
existing `update-install-blocked` panel, with the same payload shape it already
receives from the update pre-flight (one payload builder now serves both, so the
two cannot arrive in different shapes). No new copy, no new control, no layout. The
evidence this adds is terminal output, so there are no frames to review.

## Decisions, and why

- **Venv guard rather than copying the interpreter out of the bundle.** Both close the propagation path; copying ~47 MB per install into `userData` changes where the app's interpreter lives for every consumer and every platform. The guard plus the tree seal reaches the same guarantee for everything built from here on, and for older installs the guard is written by the first start of any instance (it is not in a bundle), with the start-up heal repairing damage already done.
- **Prevention is the build-time seal; the runtime seal stays as the repair path.** The zip Squirrel stages for an in-app update does not carry access-control entries, so a bundle that arrived that way is unsealed until the app runs - which is when the runtime seal applies. Installs older than this branch have neither and need one heal, or a reinstall if a sealed resource was rewritten (`file modified:`, which no deletion undoes).
- **`PYTHONDONTWRITEBYTECODE=1` is not free**, and the previous revision rejected it for that reason: a process under it recompiles its import graph per start instead of reusing a cache. It is set anyway because the promise is a bundle its own app cannot unseal, and the cost is paid once per app start by the pythons the app starts; the app-managed venv, which everything else reaches, gets the `sitecustomize` guard instead, which keeps its cache decision inside the refusal rather than removing the cache for the whole machine.
- **Heal, never seal, a bundle we do not run from**, and only when the plan is exactly the `file added:` bytecode class.

## Non-goals, restated so they are not re-litigated

- **No `com.apple.quarantine` stripping.** The artifacts are signed, notarized and stapled (re-verified on the 0.22.2 image: app `codesign` OK, `spctl accepted / Notarized Developer ID`, `stapler validate` OK, image `spctl` accepted and stapled), quarantine is harmless while the seal holds, and weakening Gatekeeper for the operator's own copy is not ours to do. The invariant is the fix.
- No rework of the Squirrel install path, the relaunch watchdog's bounds, or the release pipeline beyond the artifact assertion above.

## Testing evidence

`docs/evidence/update-robustness/self-seal.txt` (raw command output; the directory's README gained the section that places it):

- the operator's machine read-only: the venv's `pyvenv.cfg` with its `home` inside the installed bundle, the 0.22.2 marker left behind, the runtime seal lines with nothing after the 09:48 update, and every `Update service initialized` after 09:50 a dev instance;
- the shipped artifact failing the gate's own write probe (`134 of 134 path(s) accept a new file`), then broken by one unguarded interpreter run inside it (32 `.pyc`, `codesign` exit 1, `file added: .../webbrowser.cpython-312.pyc`), then healed back to `valid on disk`, then sealed by the build step - after which the same write is refused with `EACCES` and the bundle still verifies;
- the venv guard measured on a venv built on a copy of the shipped interpreter (unguarded run adds `lib/python3.12/__pycache__/webbrowser.cpython-312.pyc`; guarded run adds nothing), and its measured limit (3 startup-import `.pyc` remain, which is why the tree seal is the load-bearing half);
- the start-up probe's cost, the gate FAIL/PASS pair, and the two negative demonstrations: the guards removed from `withPythonBytecodeCache` fails 5 contract tests across every python spawn class, and a new unguarded spawn site fails the enumeration test by name.

## Gates

All CI checks pass on `1cfa1dd9c` (Desktop Tests 3m49s on ubuntu, Lint, Type Checking, Security Audit, Runtime Dependencies, Version Bump Guard, NPX Sanity Check on both platforms - run 34868729230). Locally, `git range-diff` over the four rebased commits proves the rebase kept their content: three `=`, and the fourth differs only in the `test:desktop` line, where `main`'s new `session-switch.test.mjs` entry had to be kept alongside this branch's added file.

- `pnpm lint` exit 0 (33 warnings, all pre-existing and none in a file this branch touches).
- `pnpm check-types` exit 0.
- `pnpm check-themes` exit 0.
- `pnpm test:desktop` on the rebased head: **888 tests, 888 pass, 0 fail** (46 files, including the two new suites this branch adds and `evidence-manifest.test.mjs`, which binds the re-stamped manifest to HEAD's trees - 15/15).
- `pnpm check-evidence`: the manifest half (stamps and citations) is the test above and is green; the standalone sweep re-reads every committed frame through ImageMagick and was still running against a host at load average ~100, with four other sessions running the same sweep, when this PR was opened. Nothing here adds or moves a frame - this branch's evidence is one `.txt` and README prose - and the standalone sweep is not a step in `ci.yml`.
- `pnpm build` (electron-vite) exit 0 (3m30s).
- `electron-builder --dir --arm64` exit 0 (with pnpm 10.29.2 on `PATH`, per AGENTS.md), which is what exercises the new `afterPack` step: `Sealed the bundled interpreter trees against bytecode writes: .../python_aarch64; 134 directory path(s) now refuse new entries` - the same count the field log reports for the installed 0.22.2.
- `node scripts/verify-macos-artifacts.mjs --app dist/mac-arm64/"Local Operator.app" --dmg ~/Downloads/local-operator-ui-0.22.2-arm64.dmg` -> `PASS app-python-trees-sealed`, `PASS app-no-bundled-bytecode`, `PASS app-one-bundled-python`, `PASS dmg-spctl`, `PASS dmg-stapler`, with the three app signing checks failing as `code has no resources but signature indicates they must be present`, which is what an unsigned local build is; the positive direction is `publish.yml`'s.

## Not addressed

- The interpreter the app-managed venv is built on is still inside the bundle, and a *pre-existing* venv on a machine that never runs a build of this branch keeps writing there until some instance writes the guard. The operator's own install has been hand-repaired; a bundle whose sealed resources were rewritten still needs one reinstall.
- A `.pyc` that was **rewritten** rather than added remains unhealable by deletion - the pre-existing limit, now much harder to reach.
- `com.local-operator.wakes` sessions being taken out for a morning by the loader refusing `libpython3.12.dylib` is recorded in the evidence but not otherwise addressed: it is a consequence of the same broken seal, and the seal is the fix.
- No version bump: this branch carries none, and it is based on the released 0.22.3.


damianvtran:
### Agent review — round 1

Reviewer: reviewer (independent reviewer subagent; did not author this branch) — model: `deepseek/deepseek-flash`. No provider fallback occurred during this round.
Scope: `142e869045508a258df2bd0ccae28c2f60c306c2..1cfa1dd9cdb80894882b9aa8efa8fab558afa3ca`

Verdict: **changes requested — 1 blocker (R1), 3 minors, 2 nits.** No other blocker, no major. R1 is on the fix's own central claim (the dev/packaged separation) rather than on the sealing mechanism, which I re-derived independently and which holds. R1's path is skipped by the documented dev flow (`VITE_DISABLE_BACKEND_MANAGER=true` in `.env.template`), so it will not turn up in casual `pnpm dev` testing; it is reachable in any unpackaged run with the backend manager enabled, which is the configuration the separation is written for.

---

#### What I re-ran and re-derived

Worked in my own worktree at `1cfa1dd9c` (`/tmp/rev174`) with a symlinked `node_modules`; nothing under `/Applications`, `~/Library/Application Support/Local Operator` or the author's worktree was touched.

**Suites.** `scripts/python-bytecode-cache.test.mjs` 27/27 pass, `scripts/seal-python-resource.test.mjs` 4/4, `scripts/prune-python-resource.test.mjs` 10/10, `scripts/evidence-manifest.test.mjs` 15/15. `scripts/update-robustness.test.mjs` gives 60/66 here; the 6 failures are all the blockmap group and are **environmental, not this diff** — the same 6 fail at the base commit in the same environment (base: 57/63), because `app-builder-lib/.../out/targets/blockmap/blockmap.js` is absent from this machine's shared `node_modules`.

**The seal, the gate, the heal, and deletability — driven for real on a constructed bundle** with the shipped layout (`Contents/Resources/python_aarch64/lib/python3.12`, a shipped `.pyc`, ad-hoc signed with `codesign -s -`):

```
seal: {"supported":true,"sealed":["…/python_aarch64"],"directories":9,"bytecodeFiles":1,"failures":[]}
touch new file in sealed dir        -> exit 1  touch: …/encodings/newfile: Permission denied
overwrite shipped .pyc              -> exit 1  tee: …/__init__.cpython-312.pyc: Permission denied
GATE on sealed tree:                true     10 path(s) under Contents/Resources/python[/_aarch64] refuse a write
GATE after chmod -N on one dir:     false    1 of 10 path(s) accept a new file: …/lib/python3.12
GATE on a bundle with no tree:      false    no bundled interpreter tree was found under Contents/Resources …
```

```
# the field break and the heal, end to end
codesign --verify --deep --verbose=2 (after the added .pyc) -> exit 1  file added: …/__pycache__/webbrowser.cpython-312.pyc
healPythonBytecode -> {"healable":true,"removed":["…/webbrowser.cpython-312.pyc"],"reason":"1 bytecode file(s) added after signing"}
codesign --verify --deep (after heal) -> exit 0  valid on disk / satisfies its Designated Requirement
the added empty __pycache__ directory is left behind (an added dir is not a violation)
# and on a SEALED tree, which is the field sequence (unsealed update -> write -> app seals -> next pre-flight heals)
heal on the SEALED tree -> {"healable":true,"removed":1}  codesign after heal -> exit 0
```

```
A: seal, then codesign -s -            -> exit 0 ; codesign --verify --deep -> exit 0
B: codesign -s - , then seal           -> codesign --verify --deep --strict -> exit 0  valid on disk
rm -rf of the sealed bundle / of the sealed tree alone -> exit 0, gone
ditto <src> <dst>  -> the copy still refuses the write, gate PASS
cp -R  <src> <dst> -> the copy accepts a write, gate FAIL
```

```
PYTHONDONTWRITEBYTECODE: unset -> False ; "0" -> False ; "" -> False ; "1" -> True
```

So: the seal does not disturb a signature in either order, it does not stop the app being replaced, `ditto` (the DMG path) carries it and `cp -R`/a ZIP does not — exactly as the PR claims — and the truthiness claim about `0`/`""` is correct. I did **not** run a full `electron-builder` build (this host is at load ~100 and my worktree has no downloaded interpreter); I drove `scripts/seal-python-resource.mjs`'s step directly against a bundle with the shipped layout instead, and the author's own build log (134 directories) is in the evidence file.

---

#### R1 — BLOCKER — the venv split exists only in the app; all three install scripts still build and use the packaged name

`src/main/backend/scripts/macos-install-script.sh:9-11`, `src/main/backend/scripts/linux-install-script.sh:13-15`, `src/main/backend/scripts/windows-install-script.ps1:7-9` each hardcode `local-operator-venv`, while the app now hands an unpackaged instance `local-operator-venv-dev` (`src/main/backend/venv-paths.ts:36-37`, consumed at `src/main/backend/backend-installer.ts:50` and `src/main/backend/backend-service.ts:223`). Nothing carries the resolved path into the script: `installBackend()` writes the shipped script verbatim and spawns `bash <script>` with only `PYTHON_BIN`, `ELECTRON_RESOURCE_PATH` and the two `PYTHON*` guards in `env` (`backend-installer.ts:281-305` and `:472-478`).

I ran the shipped script the way the app runs it (fake `HOME`, the app's own env: `PYTHON_BIN`, `ELECTRON_RESOURCE_PATH`, `PYTHONPYCACHEPREFIX`, `PYTHONDONTWRITEBYTECODE=1`) and compared it with the branch's own path decision for the same instance:

```
$ bash src/main/backend/scripts/macos-install-script.sh   (unpackaged instance env)
Virtual environment path: …/Local Operator/local-operator-venv
Creating virtual environment at …/Local Operator/local-operator-venv...
$ find "…/Library/Application Support" -maxdepth 2 -name 'local-operator-venv*'
…/Local Operator/local-operator-venv

the app's venvPath for this unpackaged instance: …/Local Operator/local-operator-venv-dev
script venv has bin/local-operator:            true
isInstalled() predicate on the app's own path: false
```

Three consequences, in increasing order of severity:

1. **The separation the PR exists for is not achieved on the install path.** An unpackaged instance running the script still creates, `pip install`s into — and, on the "potentially corrupted" branch, `rm -rf`s — the *packaged* app's venv (`macos-install-script.sh` around the `if [ -e "$VENV_PATH" ]` arm). That is the shared environment whose `pyvenv.cfg` points `home` inside the installed bundle, i.e. exactly the state `venv-paths.ts`'s docstring says must not be reachable.
2. **The unpackaged install can never succeed, and this is a regression.** After the script exits 0, `install()` verifies `join(this.venvPath, "bin", "local-operator")` (`backend-installer.ts:568-580`) against the *dev* path, does not find it, throws, and the `catch` returns false — so `index.ts:979-988` calls `app.quit()`. Before this PR the unpackaged instance resolved the same name the script creates, so the install verified. An unpackaged instance with the backend manager enabled and no global `local-operator` therefore writes into the installed app's venv and then quits, instead of setting up its own environment.
3. The new README claim "an unpackaged instance no longer acts on the packaged app's install state, and no longer uses its venv" (`docs/evidence/update-robustness/README.md`, last bullet of the 2026-09-14 section) is not true of this path.

Why nothing caught it: `scripts/python-bytecode-cache.test.mjs:1940` asserts only `managedVenvPath`'s two answers. No test binds the script's own `VENV_PATH` to the app's decision, and `VITE_DISABLE_BACKEND_MANAGER=true` (in `.env.template`, so in a normal dev `.env`, and in this machine's worktree `.env`) skips `install()` entirely — which is also why the reachability above needs stating rather than assuming. The window that does reach it is a run with the flag unset or false (`src/main/index.ts:967` is the only reader, and nothing in the build sets it).

**Fix (one line each):** pass the decision to the script instead of re-deriving it — `env.LOCAL_OPERATOR_VENV_PATH = this.venvPath` at the spawn site, and `VENV_PATH="${LOCAL_OPERATOR_VENV_PATH:-$APP_DATA_DIR/$VENV_NAME}"` in both shell scripts plus the equivalent in the `.ps1` — then add a test that the script's path for a given instance equals `managedVenvPath({ …, packaged: true | false })`, so the two spellings cannot drift again.

#### R2 — MINOR — the start-up refusal reuses the update-time copy, so the panel claims an update was stopped when none was attempted

`src/main/update-service.ts:1490-1510` (`repairRunningBundleSeal`) reaches the refusal through `repairBundleSeal(bundlePath, null)` → `installedBundleSealBlock(bundlePath, detail, null)` (`src/main/update-install.ts:271-284`). That builder's version-less message is *"This install of Local Operator can't be updated in place, so the update was stopped before the app quit."*, and the renderer prints `message` verbatim (`src/renderer/src/shared/components/common/update-notification.tsx:1070`). In the start-up case there was no update and nothing was stopped — the user is told an event happened that did not. The heading ("This app can't update itself") and the remedy are correct for both paths, so only the second clause is wrong. Fix: select the copy by the trigger (start-up vs pre-flight) rather than by `version`, e.g. a version-less start-up variant ending "…, so it can no longer update itself. Replace the app in Applications to keep getting updates."

#### R3 — MINOR — the new gate asserts the `dist` app, never the app inside the shipped image

`scripts/verify-macos-artifacts.mjs:612-628`: `pythonTreeSealCheck(appPath)` runs inside `for (const appPath of appPaths)`; the `dmgPaths` loop runs only the image checks (`dmg-spctl`, `dmg-stapler`). So `app-python-trees-sealed` never probed the bundle a user actually downloads, and the ACL survives into the image only because the image is built with `ditto` — measured above: `ditto` keeps the seal, `cp -R`/a ZIP drops it. A future change to how the image or a staged update tree is assembled would ship an unsealed app with this gate still green. Fix (cheap, and the same move the check's own docstring tells a human to make): `ditto` each app out of its mounted image into a temp directory and run `pythonTreeSealCheck` on the copy.

#### R4 — MINOR — the foreign-bundle heal swallows the `unavailable` verdict

`src/main/update-service.ts:1459-1474`: only `sealed` and `unsealed` produce a log line. `repairBundleSeal` can also return `unavailable` (a `codesign` that timed out twice, or one that could not be started), and on this path that is completely silent — while the sibling `repairRunningBundleSeal` (`:1490`, and its `unavailable` arm) does log it. A probe that never ran is then indistinguishable from a healthy bundle in the field log, which is the log the incident diagnosis reads. Fix: add the `unavailable` branch with `seal.detail`, as the running-bundle path does. (Silence for the *already sealed* foreign bundle is fine; this is only about the verdict that means "no answer".)

#### R5 — NIT — the release docs still name the old `afterPack` hook

`docs/BUILD.md:146` and `docs/CODE_SIGNING.md:125` describe `scripts/prune-python-resource.mjs` as the `afterPack` step; `package.json` now points `afterPack` at `scripts/after-pack.mjs`, which calls prune and then seals. Anyone following those docs to add or adjust a pack step edits a file that is no longer the hook.

#### R6 — NIT — one spawn-table row asserts the property name, not the guarded environment

`scripts/python-bytecode-cache.test.mjs:215-221` matches `/\benv\b/` against the installer spawn's call text, so `env: process.env` would pass that row; the per-file `GUARD|withPythonBytecodeCache` assertion at `:1916` is likewise satisfied by any mention of the identifier in the file. The real assertion is the installer case at `:867-900`, which drives the spawn and reads `options.env` — so the table is just weaker than it reads. Fix: make that row assert the value the way `:867` does, or drop the regex from it.

---

#### Checked and clean (nothing found)

- **Spawn coverage.** I re-ran the enumeration walker: every `node:child_process` call site in `src/main` has a row, an unlisted site fails, and the python rows are asserted by driving the spawns (`pythonSpawnEnv()`, `backendSpawnEnv()`, the installer's `env`). No python spawn path is missed among the mechanisms this tree uses; the table's one limitation is that it sees `node:child_process` only (a future spawn through Electron `utilityProcess` or a helper library would not be caught) — noted, not filed.
- **Platform shape.** `sealPythonInterpreterTrees` is darwin-only by design and `pythonTreeSealCheck` returns a *named* pass-with-reason off darwin; `managedVenvPath`'s win32/linux branches reproduce the pre-PR spellings exactly; `venvInterpreterBundle` cannot false-positive on a Windows `home` (backslashes) or on an unpackaged `<checkout>/resources/python*` (no `.app`). The earlier macOS-shaped CI failure is not reachable here.
- **The heal's narrowness.** Only `file added:` — any `modified`/`missing`/`other` violation refuses — and only `.pyc`/`.pyo` under the shipped interpreter trees; removal is per file (adding a directory is not a violation, measured), the path is re-checked immediately before each unlink, and a failed removal reports rather than half-heals. Unchanged code, reused correctly by both new callers.
- **Pre-flight regressions.** `probeInstalledBundleSeal` keeps its old policy on top of the shared helpers (unprobeable → proceed with a warning, healed → proceed, unhealable → refuse), and the delivery scheduler, watchdog and reaper cases in `update-robustness.test.mjs` still pass.
- **Claims against reality.** Every load-bearing claim I could re-derive independently held: sealed build step refuses writes, the gate refuses an unsealed tree and cannot pass vacuously or on a vanished path, the heal returns a broken copy to `codesign --verify` exit 0 (unsealed *and* sealed), the seal is invisible to `codesign` in both orders, and the DMG path preserves it. The evidence file carries a command behind each claim.
- **Scope.** No version bump in the diff; no renderer or preload file touched, so nothing visual is restyled (R2 is a string on an existing panel, not a layout change).

#### Not verified, stated rather than implied

- A real `electron-builder` build of this branch (host load; the interpreter is not downloaded in my worktree). The `afterPack` step itself was driven for real against a shipped-layout bundle, as above.
- `pnpm check-evidence` (the ImageMagick sweep) — not a step in `ci.yml`, and this branch adds no frames.

damianvtran:
### QA report — round 1

`Scope: 142e869045508a258df2bd0ccae28c2f60c306c2..1cfa1dd9cdb80894882b9aa8efa8fab558afa3ca` (branch `fix/bundle-seal-bytecode`, base `main`)
`Reviewer: qa-tester (independent pass; ran on deepseek/deepseek-flash)`

**Environment.** My own worktree at the head above (`git worktree add --detach /private/tmp/qa174/ui 1cfa1dd9c`), fresh `pnpm install --frozen-lockfile`, macOS 26.5.0 arm64, host load average 213, fifteen-plus concurrent agent worktrees. Nothing was run against the operator's primary checkout, the author's worktree `~/local-operator-ui-worktrees/update-seal`, or their live app; all app clones are `cp -Rc`/`ditto` copies under `/private/tmp/qa174` and of the 0.22.2 image **copied off** the mounted DMG (never written to the volume). `git status --porcelain` in my worktree is empty — QA edited nothing. **No Developer ID identity exists on this host** (`CSC_IDENTITY_AUTO_DISCOVERY=false` was required), so a signed or notarized artifact is out of reach and is reported BLOCKED below rather than skipped.

**A QA-process note first, because I caused it.** The packaged app's log directory and app-managed venv path are hardcoded to `~/Library/Application Support/Local Operator` and are **not** derived from `$HOME`, so an isolated temp `HOME` does *not* isolate a packaged build of this app. My one real launch of the built bundle (row 12) therefore appended to the shared `backend-installer.log` / `backend-service.log` / `update-service.log` (every worktree instance on this host writes those — directory mtime was already 13:17 before my run) and created the PR's venv guard at `local-operator-venv/lib/python3.12/site-packages/sitecustomize.py` (birth 13:16:56, sha256 `07a678adc5a87bb3ebf1231e89205264e9981332e01219dac978dc21dada7e21`). **I removed that file again**, so that piece of state is back to how I found it; the log appends cannot be un-appended. Nothing else was created. Reported to the manager at the time.

#### Matrix

| # | surface | command | actual output | verdict |
|---|---|---|---|---|
| 1 | build: clean install | `pnpm install --frozen-lockfile` | `Done in 1m 3.7s using pnpm v10.30.3` | PASS |
| 2 | build: bundle | `pnpm build` | `out/main/index.js 379.75 kB` → `✓ 1 bundles compiled into bytecode` → `out/main/index.jsc 401.70 kB`; preload 14.27 kB; renderer `✓ built in 3m 24s`; `exit=0` | PASS |
| 3 | build: package | `PATH=/tmp/pnpm-good/node_modules/.bin:$PATH CSC_IDENTITY_AUTO_DISCOVERY=false pnpm exec electron-builder --dir --arm64` | `packaging platform=darwin arch=arm64 … appOutDir=dist/mac-arm64` → `No off-architecture bundled Python found for arm64 in …/Contents/Resources` → **`Sealed the bundled interpreter trees against bytecode writes: …/Contents/Resources/python_aarch64; 134 directory path(s) now refuse new entries and 0 bytecode file(s) are refused a rewrite`** → `skipped macOS application code signing reason=…CSC_IDENTITY_AUTO_DISCOVERY=false`, exit 0 (prune → seal → signing, in that order) | PASS |
| 4 | build: DMG | `… pnpm exec electron-builder --mac dmg --arm64 --publish never` | `building target=DMG arch=arm64 file=dist/local-operator-ui-0.22.3-arm64.dmg`, exit 0 | PASS |
| 5 | seal: ACE present | `ls -lde "<app>/Contents/Resources/python_aarch64"` (and `…/lib/python3.12`, `…/bin`) | ` 0: group:everyone deny add_file,add_subdirectory` on all three (the shipped 0.22.2 prints a bare mode line — measured, row 13) | PASS |
| 6 | seal: coverage | walk every directory under the tree and count the ACE | `134 of 134 directories carry 'deny add_file,add_subdirectory'` | PASS |
| 7 | seal: write refused | `touch` at 7 depths (tree root, `lib`, `lib/python3.12`, `…/json`, `…/encodings`, `…/lib-dynload`, `bin`) | all seven `Permission denied` | PASS |
| 8 | seal: new directory refused | `mkdir …/lib/python3.12/__pycache__` | `mkdir: …: Permission denied` | PASS |
| 9 | seal vs raw CPython | `env -u PYTHONDONTWRITEBYTECODE -u PYTHONPYCACHEPREFIX "<app>/…/bin/python3" -c "import webbrowser; print('import ok')"` | `import ok`, exit 0, and `.pyc` under the tree `0` | PASS |
| 10 | seal control (the same bytes unsealed) | same command against an unsealed copy | `.pyc 0 → 32`, including `lib/python3.12/__pycache__/webbrowser.cpython-312.pyc` (26458 B) — the exact field file | PASS (reproduction) |
| 11 | release gate, this build | `node scripts/verify-macos-artifacts.mjs --app "<built app>" --dmg "<built dmg>"` | `PASS app-no-bundled-bytecode`, `PASS app-one-bundled-python`, `PASS app-python-trees-sealed`; `FAIL app-codesign` / `app-spctl` / `app-stapler` (`code has no resources but signature indicates they must be present`, `does not have a ticket stapled to it`) and `FAIL dmg-spctl` (`rejected \| source=no usable signature`), `FAIL dmg-stapler` | PASS for the three assembly checks; signing **BLOCKED** (no identity) |
| 12 | the built app really launches | `env -i HOME=<tmp> PATH=… LOCAL_OPERATOR_UI_SMOKE_TEST=true LOCAL_OPERATOR_UI_WINDOW_MODE=headless "<built app>/Contents/MacOS/Local Operator" --user-data-dir=<tmp>` | `Bundled interpreter trees sealed against bytecode writes: …; 134 directory path(s) selected and now refusing new entries, 0 bytecode file(s) selected and now refusing rewrites`; `Bytecode guard in the app-managed venv: … now sets sys.dont_write_bytecode for every process using this venv`; `LOCAL_OPERATOR_UI_READY electron=44.3.0`; exit 0 | PASS (and see the process note above) |
| 13 | gate negative: the shipped artifact | gate on a `ditto` copy of the DMG's app + the real `~/Downloads/local-operator-ui-0.22.2-arm64.dmg` | `PASS app-codesign`, `PASS app-spctl`, `PASS app-stapler`, `PASS app-no-bundled-bytecode`, `PASS app-one-bundled-python`, `PASS dmg-spctl`, `PASS dmg-stapler`, and **`FAIL app-python-trees-sealed: 134 of 134 path(s) accept a new file`** — a fully signed, notarized, stapled app that the gate refuses | PASS — this is the verdict the PR exists to catch |
| 14 | gate negative: controlled A/B | `pythonTreeSealCheck` on a `ditto` copy of the sealed build, then `chmod -R -N <tree>` and ask again | before: `PASS app-python-trees-sealed: 134 path(s) … refuse a write`; after: `FAIL … 134 of 134 path(s) accept a new file` | PASS |
| 15 | gate negative: partial evasion | seal the root back, `chmod -R -N …/lib`, ask again | `FAIL … 125 of 134 path(s) accept a new file` — a partially sealed tree is caught, not averaged away | PASS |
| 16 | gate negative: read-only volume | check against the app on the mounted DMG | `FAIL … 134 path(s) could not be probed (read-only volume or a path that moved)… Copy the app off the mounted image and re-run` — reported, not a vacuous pass | PASS |
| 17 | gate negative: a copy that drops ACLs | `cp -R` of the sealed build, ask the check | `FAIL … 134 of 134 path(s) accept a new file` | PASS — see Q1 |
| 18 | field break | add the exact field file to a clone | before: `exit 0 … valid on disk`; after: `exit 1`, `file added: …/lib/python3.12/__pycache__/webbrowser.cpython-312.pyc` | PASS |
| 19 | the heal | `UpdateService.repairBundleSeal(<clone>)` | `seal.kind=sealed`, `heal.removed=[…/webbrowser.cpython-312.pyc]`; codesign after: `exit 0 … valid on disk`; no `file missing:` introduced; the emptied `__pycache__` is left behind and is not itself a violation | PASS |
| 20 | the heal refuses the unrecoverable | append a byte to `lib/python3.12/webbrowser.py`, then `repairBundleSeal` | codesign `exit 1`, `file modified: …/webbrowser.py`; heal `healable=false removed=[]`, `block=installed-bundle-not-sealed`; the file is still on disk and codesign is still `exit 1` — nothing half-healed | PASS |
| 21 | the heal inside a **sealed** tree | add the `.pyc` while unsealed → `sealPythonInterpreterTrees([tree])` → `repairBundleSeal` | seal: `directories=135 bytecodeFiles=1 failures=[]`, `touch` there now refused; heal: `removed=[…webbrowser.cpython-312.pyc]`; codesign `exit 0 … valid on disk` — the seal does not break the heal | PASS |
| 22 | the heal on an undamaged bundle | `repairBundleSeal(<clone>)` | `seal.kind=sealed heal=null block=null` — nothing deleted | PASS |
| 23 | venv propagation path | build a venv over the in-bundle interpreter | `pyvenv.cfg home = <app>/Contents/Resources/python_aarch64/bin` | PASS |
| 24 | venv break (reproduction) | unguarded `"$VENV/bin/python" -c "import webbrowser, json"` | venv creation alone `0 → 168` `.pyc`; the import run `168 → 169`; codesign `exit 1` with **169** `file added:` lines | PASS (reproduction) |
| 25 | the venv guard | `ensureVenvBytecodeGuard(venv)` | writes `…/site-packages/sitecustomize.py` carrying `sys.dont_write_bytecode = True`; second call `written=false`; a foreign `sitecustomize.py` left byte-identical; a venv with no `site-packages` → `path=null, written=false` | PASS |
| 26 | guard + the install-script flag, measured apart | venv created with `PYTHONDONTWRITEBYTECODE=1` (as the fixed script does), guard installed, then an unguarded run | creation `0 .pyc`; run `0 → 3` — `encodings/{aliases,utf_8,__init__}.cpython-312.pyc`, exactly the PR's stated limit; the unguarded control is `0 → 36` | PASS |
| 27 | the guard's limit, closed by the seal | same venv over the **sealed** build, same unguarded run | `0 → 0` `.pyc`; a `touch` in that tree refused | PASS |
| 28 | the two falsy spellings | `PYTHONDONTWRITEBYTECODE=0` and `PYTHONDONTWRITEBYTECODE=''` with no prefix | 32 `.pyc` written either way — the comment's justification for setting the flag unconditionally is measured, not assumed | PASS |
| 29 | redirect vs refusal | no prefix and no flag (the field condition) | 36 `.pyc` written into the bundle — while `PYTHONPYCACHEPREFIX=/dev/null/nope` alone wrote 0, i.e. the redirect is more robust there than the comment claims and the flag is what covers the *no-prefix* case | PASS |
| 30 | dev/unpackaged separation | `UpdateService` bundled with Electron stubbed (the repo's own harness pattern), temp HOME/userData, synthetic marker + ShipIt staging tree + log | 15/15: marker not cleared, staged tree and ShipIt log untouched, `reapFailedInstallLeftovers` never reached, nothing pushed to the renderer, `quit-and-install` returns `false` and writes no marker, and a second pass changes nothing. Packaged control on the same state: failure payload produced, reap path reached, `update-install-failed` pushed and the marker cleared | PASS |
| 31 | update pre-flight, refusal and remedy states | real `probeInstalledBundleSeal` against a clone placed at `process.execPath` | healthy → `null`; added bytecode → healed, `null`, codesign exit 0; `file modified:` → `{"code":"installed-bundle-not-sealed","message":"…the update to version 0.22.3 was stopped before the app quit.","remedy":{"text":"Quit Local Operator, then download a fresh copy and replace the app in Applications.","url":"…/download"},"detail":"…a sealed resource is missing or invalid: file modified: …webbrowser.py"}`, and the tampered file was not deleted; `src/renderer/…/update-notification.tsx:128` keys a heading off that code | PASS |
| 32 | packaged/unpackaged venv split | `managedVenvPath`, `venvInterpreterBundle` | `…/Local Operator/local-operator-venv` vs `…/local-operator-venv-dev`; `venvInterpreterBundle` resolves a real `.app`, and `null` for a checkout `resources/…` home, a uv-managed python, and a `.app` that is not on disk | PASS |
| 33 | install script, macOS, run for real | `HOME=<tmp> ELECTRON_RESOURCE_PATH=<unsealed clone>/Contents/Resources bash src/main/backend/scripts/macos-install-script.sh` | `Successfully installed … local-operator-0.54.46 …`, `v0.54.46`, `Installation completed successfully.`, `exit=0`; in-bundle `.pyc` **`0 → 0`**; `2886` `.pyc` in the out-of-bundle `python-bytecode-cache` prefix | PASS |
| 34 | install scripts, linux + windows | `bash -n` on both shell scripts; placement of the flag against the first python call | linux: `bash -n OK`, `export PYTHONDONTWRITEBYTECODE=1` at line 35, first python use line 158; macOS: `bash -n OK`, line 44 vs line 184; ps1: `$env:PYTHONDONTWRITEBYTECODE = "1"` at line 29, first use line 200 | PASS (static) |
| 35 | windows ps1 execution | — | no `pwsh`/`powershell` on this host | **BLOCKED** |
| 36 | regression suite | `LOCAL_OPERATOR_UI_TEST_CONCURRENCY=6 pnpm test:desktop` (`env -u NO_COLOR TERM=xterm-256color`, every inherited `CMUX_*` unset) | `ℹ tests 888`, `ℹ pass 888`, `ℹ fail 0`, `duration_ms 144838`, exit 0 | PASS |
| 37 | evidence manifest coherence | `node scripts/evidence-manifest.test.mjs` | `ℹ tests 15`, `ℹ pass 15`, `ℹ fail 0` — including `the SHIPPED manifest's stamps describe the tree it ships in` and `a manifest stamped at the tree under review passes` | PASS |
| 38 | evidence sweep | `pnpm check-evidence` | still running at the time of this comment (~36% of 1501 frames after 30 min on a host at load 213 with four other sessions running the same sweep); the exit code follows in a separate comment | PENDING |
| 39 | repo untouched | `git status --porcelain` | empty | PASS |

#### Q-findings

**No finding below is a FAIL of the change under test** — every row above that the change owns passed against a real artifact. These are recorded because they are real, measured, and a reviewer should see the size of them.

**Q1 — the build-time seal does not survive the in-app update path (documented; reproduced).** `ditto` keeps the access-control entry, `cp -R` and a ZIP drop it. Squirrel.Mac stages `update.zip`, so a bundle that arrives by in-app update is unsealed, and it is re-sealed only when the packaged app next starts (`BackendInstaller`'s constructor calls `sealPythonInterpreterTrees` before any python is spawned — read, not run). The window is therefore "from ShipIt's swap until the app is launched again", which is the shape of the original field failure. Repro: `cp -R "<sealed build>" /tmp/x.app && node -e 'import("./scripts/verify-macos-artifacts.mjs").then(m=>console.log(m.pythonTreeSealCheck("/tmp/x.app")))'` → `FAIL … 134 of 134 path(s) accept a new file`. The PR discloses this in `scripts/seal-python-resource.mjs` and keeps the runtime seal as the repair path; the reviewer should decide whether that window is acceptable or wants a start-up seal that runs before the backend spawns, rather than 15 s after the window is created (`STARTUP_SEAL_PROBE_DELAY_MS`).

**Q2 — the guard alone does not keep an *unsealed* bundle verifiable.** Measured on a clone of a real unsealed bundle: with the guard in place, an unguarded `import` still leaves the 3 startup `encodings` `.pyc` that precede `sitecustomize`, and codesign goes to `exit 1` with exactly 3 `file added:` lines. On a sealed bundle the same run leaves 0. Repro: rows 26 and 27. This matches the PR's own stated limit, so it is a note rather than a defect — but it means "the guard is out" only holds together with the seal, and a bundle that arrives without the seal (Q1) has neither.

**Q3 — a venv whose `pyvenv.cfg` points at an `.app` that is no longer on disk resolves to `null`, so the foreign-bundle heal silently does nothing.** Repro: write `home = /Applications/No Such.app/Contents/Resources/python_aarch64/bin` into a synthetic `pyvenv.cfg` → `venvInterpreterBundle()` returns `null`; point it at a real bundle → the bundle path. Worth knowing because it is the exact state on this machine (`/Applications/Local Operator.app` does not exist here while the venv's `pyvenv.cfg` still names it), and the "heal the installed app from a dev instance" path then has nothing to heal. Low severity — there is nothing to repair when the bundle is gone.

**Q4 — the venv split is a behaviour change for unpackaged instances.** `managedVenvPath` now gives an unpackaged instance `local-operator-venv-dev`, which is a fresh environment it does not have yet, so a dev machine with no `local-operator` on `PATH` pays a one-time backend install. I measured the path resolution (row 32) but **did not run a dev instance's backend install**, because doing so writes into the operator's shared app-support tree — the same trap as the process note above. Not covered by this pass.

#### BLOCKED

- **Signed / notarized artifact** — there is no Developer ID identity on this host. Rows 3–4 produced an unsigned bundle on purpose (`CSC_IDENTITY_AUTO_DISCOVERY=false` rather than failing), and every signing-related check in row 11 is `FAIL` for that reason and no other: `code has no resources but signature indicates they must be present`, `no usable signature`, `does not have a ticket stapled to it`. The three checks this change owns — `app-no-bundled-bytecode`, `app-one-bundled-python`, `app-python-trees-sealed` — pass. The signed positive direction is `publish.yml`'s alone.
- **Windows `windows-install-script.ps1` execution** — no PowerShell on this host. Static only (row 34).
- **`pnpm check-evidence`** — the sweep re-reads **1501** committed frames through ImageMagick. I gave it a real budget on a host at load 213 with four other sessions running the same sweep concurrently; at the time of posting it was ~36% through and still going, so it is reported as pending rather than as a pass. Its exit code and output land in a follow-up comment on this PR. Note it is **not** a step in `ci.yml`; the manifest half of the same evidence is `evidence-manifest.test.mjs`, which is green (row 37).
- **The packaged app's own start-up behaviour end to end** — I launched the built bundle (row 12) and read the seal and guard lines out of its log, but I could not run a *signed* install or a real in-app update, and I deliberately did not touch the operator's `/Applications` install, their live sessions, or their `~/Library/Application Support/Local Operator` state beyond the incident in the process note.

**Verdict: PASS** for the change under test on the platform surface this host can produce — with the four BLOCKED items above stated rather than covered, and Q1–Q4 recorded for the reviewer. Nothing in the matrix failed in a way this change is responsible for.

damianvtran:
### Recovery status and delivery contract

Manager: openai/gpt-6-astra; resumed after the provider outage. **Not ready to merge or release.** Round-1 remediation is unfinished and the branch now conflicts with main.

The round-1 QA finding about ZIP transport is material: the proposed ACL protection is lost in the in-app update archive, and a venv guard runs after early Python imports. Documenting that window does not close it. The revised implementation will use a complete versioned managed Python runtime outside the signed app, shipped from a new private seed namespace with no legacy in-bundle Python path aliases. Legacy environments and user data must remain untouched; dev instances must not heal or mutate a foreign installed app.

Acceptance evidence must include external-runtime relocation/import paths, migration and failure handling, final ZIP and DMG-contained bundle verification, and a signed existing-release-to-candidate native updater exercise in an isolated environment. The earlier unsigned build and unit/contract tests are not proof of that signed transition. Code, QA, and design review of changed user-facing states will be refreshed on the remediation head before merge.

Evidence correction: adding the observed bytecode file was independently reproduced to invalidate the code signature. The exact writing process, cause of every reported popup, and actor behind subsequent app disappearance were not conclusively identified. Earlier broad causal statements should not be read as established findings. No further tests or repairs will modify the operator’s live application or app-support state.

damianvtran:
Superseded by #183.

The approach here — protecting the bundled interpreter with filesystem access controls — does not close the reported failure: those access controls are lost in the ZIP the in-app updater stages, so the bundle is unprotected between the app swap and the next launch, which is exactly the window the failure lands in. Documenting that window is not closing it.

#183 replaces the mechanism instead of adding to it: the managed Python backend runs from a complete versioned runtime outside the signed bundle, with no legacy interpreter path left inside the app. The round-1 findings recorded on this PR (R1, R2, R3, R4, R5, R6 and Q1-Q4) are carried over and answered there.

Closing this PR as superseded once #183 is ready for review; nothing here was merged, and the branch and its round-1 review remain as the record.

damianvtran:
Superseded by #183, which is now in review with the code, QA and design rounds recorded on it.

Closing rather than merging: the access-control approach this branch took cannot survive the ZIP the in-app updater stages, so it left the bundle unprotected in exactly the window the reported failure lands in. The round-1 review and QA findings recorded here were carried over to #183 and are answered there.

Nothing here was merged. The branch and this thread stay as the record of how the problem was first characterised — including the two corrections posted above, which retract claims the evidence does not support.

PR #181, opened 2026-09-14
fix(browser): honour the host's ownership, consent, viewport and capture contracts

damianvtran:
## What this is

Corrective, **host-only** fixes for defects that shipped with the browser host in
#160. It contains **no new user-visible surface**: there is no Browser route on
`main`, so this PR cannot and does not deliver a usable consent workflow to a
user — it repairs the host contract that the (still-unlaunched, launch-gated)
chrome PR #177 depends on. Popup capabilities are deliberately (and visibly) not
covered here as well.

**Draft on purpose:** review rounds are still coming.

| in this PR | not here |
|---|---|
| atomic owner allocation, adoption of a handed-over tab, allocation-aware finish/close | any UI at all — no route, no consent surface, no tab strip |
| current-document authorization for every driven action, revocation that reaches unspent grants | popup OAuth parity (see *Popup limits*) |
| a bounded background viewport, and background capture/read/input with no route | the chrome module itself (#177) |
| shipped public-suffix data under the existing vendoring pin | Python-side (`local-operator`) changes of any kind |
| regression coverage driven through the real `BrowserHost` dispatcher | native trusted-renderer automation (deliberately not used) |

No version bump: `package.json`'s `version` is untouched.

## Defects and the mechanism behind each

Every one below was reproduced against the **real product** — the built Electron
app plus the real Python `execute_browser` tool over the real host RPC — before
the fix, and re-run after it. Where a conclusion rests on a measurement, the
measurement is in the code comment beside the decision.

1. **Normal ownership was broken (QA Q-1 / review R2).** The ownership ledger
   recorded an *empty* allocating reservation before the handler ran and never
   recorded the capability the allocator returned, so a healthy first `open`
   succeeded and then every same-owner `read`/`goto`/`input`/`snapshot`/`close`
   was refused with `owner_refused`. The old tests completed the allocation by
   hand, which hid it; the new tests drive the real dispatcher instead.
2. **Consent did not cover the current document, and revocation left authority
   behind (QA Q-2 / review R1).** `read`/`input`/`screenshot` checked only that
   the handle was valid, so an autonomous page-initiated navigation to an
   unapproved origin left the tab drivable, and `revokeOrigin`/`revokeAll` did
   not invalidate unspent one-shot grants. Now every document-scoped action is
   authorized at entry **and again before its result returns**, via a
   per-document receipt that a navigation or a revocation invalidates;
   `screenshot` is covered because it was the same pixel-level hole; and the
   log buffer is cleared on a document change so returning to an approved site
[...]

PR #172, opened 2026-09-14
fix(update): earn the up-to-date affirmation from the whole check

damianvtran:
## What and why

Release: patch — "Check for updates" can no longer say "You are up to date" in the same turn as it offers you an update.

One press of Settings → Application updates and info → Check for updates produced both of these, on application 0.22.1 against server 0.54.43 with server 0.54.44 published:

- the panel — "Server update available / Server version 0.54.44 is available. You are currently using version 0.54.43."
- the success snackbar — "You are up to date"

The operator saw it "a couple times", "when there's an available version on either the server or the UI", which is the shape of the defect.

### Root cause

`UpdateService.checkForAllUpdates` runs two independent checks, and each emitted its own renderer event with no knowledge of the other channel:

- the app channel sends `update-not-available` when the app is current (`checkForUpdates`, and the `autoUpdater`'s own `update-not-available` listener);
- the server channel sends `backend-update-not-available` when the server is current and `backend-update-available` when it trails.

`CheckForUpdatesButton` subscribed to BOTH "not available" events and turned either into a sentence about the whole installation — "You are up to date" / "The server is up to date", gated only on `manualCheckRef`. Nothing made an affirmation conditional on the other channel, so whenever exactly one of {app, server} trailed, an offer and an affirmation rendered in the same turn, and which sentence the user saw was decided by whichever event the main process happened to emit last — which is why the report's two directions look different.

### The invariant

**An "up to date" sentence may only be shown when the whole check positively proved it, on both channels — never when any channel has an update to offer, and never when a channel could not find out.** The repo's documented defect class is "a surface treating we-could-not-find-out as an answer" (see the test job's comment in `.github/workflows/ci.yml`), and this was its last decision point: a channel that could not read a version, or whose fetch failed, or which was skipped in development mode, was indistinguishable from a channel that found nothing newer.

## The change

- **`src/main/update-check-verdict.ts` (new, pure — no electron import).** `UpdateChannelStatus` = `"available" | "current" | "unavailable"`, plus `UpdateCheckVerdict`, `UP_TO_DATE_AFFIRMATION` and `updateCheckAffirmation` (the sentence, or null). `"unavailable"` covers every "we could not find out": development mode, a registry fetch with no answer, a thrown error, a filtered error, an unreadable server version. Being electron-free is load-bearing: the preload/renderer types and `scripts/*.test.mjs` reach the rule without an Electron fixture in the way.
[...]

PR #165, opened 2026-09-14
feat(panels): first-class panel views for the five diagnostic commands

damianvtran:
## What

`/info`, `/analytics`, `/session`, `/context` and `/failovers` are read-only
diagnostics the terminal renders properly and the desktop rendered badly or not
at all: analytics was five key/value rows followed by `JSON.stringify(by_provider)`
in a `<pre>`, failovers was `" -> "`-joined strings, context was pre-formatted
rows with no visual, `/info` had no `desktop_destination`, and `/session` had no
`DESTINATIONS` row at all — so the dispatcher answered "not available in the
desktop app yet".

This lands the UI half of `docs/design/panel-views.md` (§2, §3, §4, §6-§9, §11,
§12, §15), implemented against the frozen contract and checked against the real
backend wire where the two halves meet.

Release: minor — /info, /analytics, /session, /context and /failovers become visual panels

## Shape

- **`PickerHost` gains `shell?: "dialog" | "panel"`** (§3). `dialog` is today's
  geometry byte for byte, `/usage` still passes `wide`, and `panel` is
  `max-w-5xl` with the taller scroll box, an unconditional scrollbar gutter, a
  named focusable scroll region and initial focus in the body.
- **`pickers/panels/`** (§2): `panel-frame.tsx` (the region frame),
  `panel-states.tsx` (notice/empty/skeleton), `primitives/` (stat card, share
  meter, bounded table, the ONE chart wrapper), `formatters.ts` (one spelling per
  quantity, ported from the Python) and one `X-panel.tsx` + `X-model.ts` per panel
  — the `usage-view.tsx` / `usage-view-model.ts` split, so every story renders the
  production component over fixtures.
- **`desktop-contract.ts`**: the `info.get` and `sessions.report` request schemas
  and their op→HTTP cases (`GET /v1/desktop/info`,
  `GET /v1/desktop/sessions/{id}/report?recent_limit=`), plus the §5 response
  shapes for all three ops. `scripts/desktop-contract.test.mjs` carries the two
  new rows in the path table it already exercises.
- **Registry rows** for `info` and `session.diagnostics`; `destination-pickers`
  deletes the three old views and re-exports the five from `panels/*` (§15).
- **§15.5 taken, not deferred**: `settings-page.tsx`'s recharts chart moves onto
  `ChartFrame`, because a second chart styling path in one app is the defect that
  item exists to prevent.
- **§11**: two `contrast-contract.mjs` rows for the two new fills
[...]

PR #188, opened 2026-09-15
chore(release): bump version to 0.23.0

damianvtran:
Release window for **local-operator-ui** after `v0.22.7`. Owner session **pid 20190**. Branch `release-next-0230x` (not `release-next`: that ref is still occupied on the remote by a stale leftover from the 0.20.0 window, `d49f33050`, whose bump actually landed as `9da9b38b0` via #145).

**Shipped as `v0.23.0`** — minor, on the materiality rule that a minor needs some single PR to clear the step-function bar on its own, which #165 does by adding five new user-visible surfaces. Merged as `680cffa9fc76da775c13d307edabb0b47d94b1bd`.

## Window (all four merged)

- [x] **#181** `fix(browser): honour the host's ownership, consent, viewport and capture contracts` — `6fcd15cd6a84756736ca3775f47ab47377b1c3d4` — Release: patch — corrects the desktop browser host shipped in v0.22.4 (tab ownership, origin authorization and revocation, background captures, domain grants, hand-over adoption); no user-visible surface.
- [x] **#165** `feat(panels): first-class panel views for the five diagnostic commands` — `6478b0d57264131de999a266134e53362034adff` — Release: minor — `/info`, `/analytics`, `/session`, `/context` and `/failovers` become visual panels.
- [x] **#189** `fix(scripts): restore the check-themes script a panel reconciliation dropped` — `f01560b6607aaf3a595d1696536f0dfcb50bccb1` — Release: patch — restores the documented `pnpm check-themes` contrast gate.
- [x] **#188** this bump.

No unlisted landings: `git log --first-parent --oneline v0.22.7..origin/main` is exactly those four merges, and no other PR in the window touched the `version` line.

## Release

- Tag + Release: https://github.com/damianvtran/local-operator-ui/releases/tag/v0.23.0 (created with `--target 680cffa9f`, published as a pre-release so the asset-less first minutes stay out of `/releases/latest`; promotion is the workflow's job).
- `publish.yml` run `34925961401`.

## Recorded on this PR rather than invented for the notes

- **#181 and this bump merged with no human approval on record.** The repository's branch rules require none — `gh api repos/damianvtran/local-operator-ui/rules/branches/main` returns an empty list and CODEOWNERS is routing-only. #181 stands on three independent review rounds plus three independent QA rounds; #165 on five review rounds plus QA, design and UX rounds; each with CI green on the merged head.
[...]

PR #189, opened 2026-09-15
fix(scripts): restore the check-themes script a panel reconciliation dropped

damianvtran:
# PR Description

## Summary of Changes

Restores the `check-themes` script to `package.json`. It was dropped by an **evil
merge** — a hand-rewritten conflict resolution inside this repository's own
`feat/panel-views` PR (#165) during its reconciliations onto a moved `main` —
that lost a one-line `scripts` entry **both sides of that merge already carried**.
Not an upstream change, and not any change to the contract the script runs.

**The diff is one insertion in one file.** `package.json`'s `version` line is
untouched and still reads `0.22.7`.

Release: patch — restores the documented pnpm check-themes contrast gate, which a merge resolution dropped while its underlying contract kept passing

## What was actually broken

The contrast contract itself never regressed: `node scripts/contrast-contract.mjs`
still runs and passes, and `node scripts/generate-theme-css.mjs --check` still
confirms `themes.generated.css` is current. What was lost is the only documented
way to run either. `AGENTS.md` ("`pnpm check-themes` enforces both freshness and
the contrast floors"), `docs/branding.md` (§ "Where colour comes from", and step 8
of its porting checklist) and `.github/pull_request_template.md` (the gates
checklist) all instruct agents to run `pnpm check-themes`, and that command
answered `ERR_PNPM_NO_SCRIPT`. A gate that is documented, expected, and
unrunnable fails silently — the PR checklist box it belongs to can be ticked
without the work behind it being done.

## Why it disappeared (traced, not guessed)

| ref | `check-themes` in `package.json` |
|---|---|
| `v0.22.7` | present, between `test:desktop` and `check-vendored` |
| `6fcd15cd6` (#181's merge) | present |
| `110e24cc7` (panel-views reconciliation) | **absent** |
| `origin/main` (`6478b0d57`, #165's merge) | **absent** |

`110e24cc7` is one of the reconciliations of #165 onto a moved `main`. It is an
**evil merge**: a hand-rewritten one-line `scripts` block in a conflict
resolution that silently dropped an entry **both sides agreed on**. Both parents
carry the script —

| parent of `110e24cc7` | side | `scripts` entries | `check-themes` |
|---|---|---|---|
| `cba43ee19` | panel-views branch | 42 | present |
| `9e8a61b3d` (#172's merge) | `main` | 42 | present |
| `110e24cc7` (the merge itself) | — | **41** | **absent** |

— and the merge differs from *each* parent by exactly that removal, adding
nothing. This is introduced by #165's own reconciliation rather than by any
upstream change: the script is present at both parents and at everything before
them, and absent from every ref at or after the merge.

The correction is the point, because the intuition it falsifies is the one that
let this land: **"both sides agree, so the resolution is safe" is not a safe
rule.** A conflict resolution is where version control stops checking the merge
for you, and a merge diff this size hides a deleted line inside its own
rewritten block. The earlier reading of this evidence — that the scripts block
had been resolved against a side predating the script — was wrong, and it
survived because it sounded like the ordinary explanation.

## Reproduction (before the fix, at `6478b0d57`)

```
$ pnpm run check-themes
 ERR_PNPM_NO_SCRIPT  Missing script: check-themes

Command "check-themes" not found. Did you mean "pnpm run check-types"?
exit 1

$ node scripts/contrast-contract.mjs
Contrast contract holds: 2327 assertions across 12 themes, 0 pinned exception(s).
exit 0
```

## Verification after the fix

Run in a worktree cut from `origin/main` (`6478b0d57`) with the branch applied,
after `pnpm install --frozen-lockfile`:

```
$ pnpm run check-themes
> local-operator-ui@0.22.7 check-themes ...
> node scripts/generate-theme-css.mjs --check && node scripts/contrast-contract.mjs

themes.generated.css is up to date (12 themes).
Contrast contract holds: 2327 assertions across 12 themes, 0 pinned exception(s).
exit 0
```

The restored line is byte-identical to `v0.22.7`'s and sits in the same position:

```
$ diff <(git show v0.22.7:package.json | sed -n '73p') <(sed -n '73p' package.json)
(no output — identical)

$ git diff --numstat
1       0       package.json
```

`pnpm check-types` and the CI desktop entry point (`pnpm test:desktop`, which
carries the shipped-manifest and desktop-contract suites) also pass on this head:

| command | result |
|---|---|
| `pnpm lint` | exit 0 (539 files checked, 35 warnings, no errors — package.json is not a lint surface) |
| `pnpm check-types` | exit 0 (`check-types:main` and `check-types:renderer` both clean) |
| `pnpm test:desktop` | `tests 1118`, `pass 1118`, `fail 0`, exit 0 |
| `git status --porcelain` after the run | only the intended `M package.json` — the suite left the tree clean |

Scoped deliberately: this restores one script, so the gate it restores plus the
type check and the desktop suite are the proportionate evidence. The full
2300-frame evidence sweep was not re-run.

## Related Issue(s)

- Related: #165 (the PR whose reconciliation dropped the script)

## Impact

- No breaking changes; no dependency changes; no runtime code touched.
- Restores a documented, agent-facing command. Nothing else in the repository
  changes behaviour.
- Two adjacent observations, deliberately **not** fixed here to keep the diff one
  line. **This PR adds no enforcement of anything**; it restores a script alias
  only.
  1. **No workflow runs this gate — `deferred — separate PR`.** `.github/workflows/ci.yml`
     has no `check-themes` step (the only reference to the command anywhere under
     `.github/` is the PR template's checklist), so even with the script restored
     the contrast contract is enforced only when a human or agent remembers to run
     it locally. The `Type Checking` job already carries the rationale for fixing
     that class of hole — "A guard nothing runs is indistinguishable from no
     guard, which is why the last of those is executed here as well as in its own
     PR-triggered workflow" — and the `Desktop Tests` job documents having closed
     exactly this gap for `test:desktop`. Doing it here would mean two edits in
     two files in what is otherwise a one-line restore, so it is deferred to a
     separate PR.
  2. `pnpm test:desktop` prints `ℹ pass N` when run locally (Node's spec
     reporter) but `# pass N` on the CI runner, which is what the
     `Assert the suite actually ran` step greps. Verified not to matter here —
     the Desktop Tests job passes on this head — but the two formats are easy to
     conflate when reading a local run against that step's source.

## Testing Details

Reproduction of the failure and the contract-still-passing measurement are above,
both captured before the change; the post-fix run of the restored gate, the type
check and the desktop suite are above as well. Exact commands and their real
output are quoted, not summarised.

## Checklist

- [x] I have added tests that prove my changes work as intended. (No test is
      applicable — a script alias; the gate it restores is itself the check, and
      it is exercised for real above.)
- [x] The gates pass: `pnpm lint`, `pnpm check-types`, `pnpm check-themes`, `pnpm test:desktop`.
- [x] This PR does not change the `version` line in `package.json` — only a `chore(release):` PR may, and the release owner cuts that one.
- [x] I have updated the documentation when required. (No doc change needed: every
      document already describes `pnpm check-themes`; the code was the thing out
      of step.)
- [x] Security considerations are addressed (especially for code execution features).


damianvtran:
### Agent review — round 1

Reviewer: independent reviewer subagent — the same agent that raised MAJOR-1 on #188, and not the author of this fix — running `deepseek/deepseek-flash`; no provider fallback occurred during this round.
Scope: `6478b0d57..1aadcad71b4e748482a03f02b544fe956fe2edc7` (this PR's own diff), against the round-1 window audit on #188.

#### Checks

**1. One insertion in one file, in its original position — PASS.** `git diff --numstat origin/main origin/fix/restore-check-themes` → `1	0	package.json`; the forge agrees (`package.json +1 -0`). The added line is at **line 73**, `sed -n '73p'` is byte-identical to `git show v0.22.7:package.json | sed -n '73p'` (diff empty), and its neighbours are `test:desktop` (72) and `check-vendored` (74) — the tag's own position, not appended. `"version"` still reads `0.22.7`. Nothing else moved: `diff <(git show v0.22.7:package.json) <(git show origin/fix/restore-check-themes:package.json)` prints exactly one hunk, the `test:desktop` line, i.e. the restore reconstitutes the tag's scripts block modulo #165's two added test entries. Forge head = `git ls-remote origin refs/heads/fix/restore-check-themes` = `1aadcad71b4e748482a03f02b544fe956fe2edc7`; `mergeStateStatus` CLEAN.

**2. The restored gate actually runs — PASS, reproduced independently.** In a throwaway worktree at this head (since reclaimed), with **no `pnpm install` at all**:
```
$ pnpm run check-themes
> local-operator-ui@0.22.7 check-themes
> node scripts/generate-theme-css.mjs --check && node scripts/contrast-contract.mjs
themes.generated.css is up to date (12 themes).
Contrast contract holds: 2327 assertions across 12 themes, 0 pinned exception(s).
exit 0
```
The same numbers the body reports. I did not re-run `pnpm test:desktop` (1118/1118 in the body): nothing in `scripts/**.test.mjs` asserts the scripts block, and #165's Desktop Tests job was green while the script was missing, so the suite cannot be evidence for this change either way — the gate run is the proportionate evidence, and I checked it. The repo actively names the command it lost: `scripts/generate-theme-css.mjs:257` emits "`pnpm check-themes`" as the instruction to its reader, so restoring the name (rather than renaming anything) is the right fix.

**3. Complete for what it claims; enforcement boundary — ACCEPTED, with a required follow-up.** The body states plainly that no workflow runs the gate and does not pretend otherwise. I verified it: `git grep check-themes origin/fix/restore-check-themes -- .github/workflows/` → no match (only the PR template's checklist mentions it). I judge that an acceptable boundary for *this* PR: it is deliberately one line so its scope check stays trivial, and wiring CI is a different change with its own review. But it should not stay a comment — `ci.yml`'s own Type Checking job already carries that exact argument ("A guard nothing runs is indistinguishable from no guard, which is why the last of those is executed here as well as in its own PR-triggered workflow", next to the `Check release contracts` step), and `check-themes` is node-only and takes seconds, so appending it there or in a sibling step is a small same-session follow-up PR. Until then the gate is runnable but still unenforced.

**4. The recorded reproduction is the honest one — PASS.** No occurrence of `2303` anywhere in the body, the commit message, or the thread. The disclosure behind it checks out: `edb6254d9` is a detached HEAD on **main's own history** at `"version": "0.22.3"` (2026-09-14 21:05), four releases behind `origin/main` — so "a stale checkout, not `main`" is the right characterisation, and the numbers measured against it are not the numbers at `6478b0d57`. The `2327` / `ERR_PNPM_NO_SCRIPT` pair that is actually used is the correct one, and I reproduced both.

**5. Attribution — the PR is right, the stated mechanism is not (MINOR-1).** Right: `check-themes` present at `v0.22.7`, present at `6fcd15cd6` (#181's merge, before it), absent from `110e24cc7` and from every ref downstream of it including `6478b0d57`. Not right: "the scripts block was resolved against the wrong side of a conflict" / "against a side that predated the script". **Both** parents of `110e24cc7` carry the line — `110e24cc7^1` (`cba43ee19`, branch side) and `110e24cc7^2` (`9e8a61b3d`, main side) each count 1, and `git diff` against **each** parent shows the removal. No side predated the script; the hand-resolved one-line scripts block simply lost an entry both sides agreed on. That is worth more than pedantry: the stated story invites the reader to conclude "a conflict resolution is safe when the two sides agree", which is exactly the assumption this incident falsifies.

#### Findings

**MINOR-1 — correct the mechanism sentence in the PR body and the commit message.** Replace "resolved against the wrong side of a conflict / a side that predated the script" with the tree's account: the merge rewrote a one-line `scripts` block and dropped an entry that **both parents carried**. Attribution to #165/`110e24cc7` stands as written.

**MINOR-2 — enforcement follow-up, tracked not buried.** The body's observation 1 is correct and should be turned into a same-session PR (add the gate to CI), per the operator's rule for out-of-scope-but-real work. Its observation 2 is also correct and needs nothing: `ci.yml:174` greps `^# pass `, Node's non-TTY (tap) reporter, while a local TTY run prints `ℹ pass N` — the two formats are indeed easy to conflate and the Desktop Tests job passes on this head.

No NIT worth a round, and nothing here is a reason to hold the PR.

#### Verdict

**clean — TERMINAL.** No BLOCKER, no MAJOR, no correctness defect in the diff: one insertion, byte-identical to `v0.22.7`, in the tag's own position, `0.22.7` version line intact, and the restored gate verified running at this head with the body's exact output. **Safe to merge as it is** — MINOR-1 is a wording fix to a document and a commit message, and MINOR-2 is a follow-up PR, neither of which should hold a one-line restore. CI is 8/8 green on `1aadcad71`.

For the release owner: #189 and #188 touch disjoint lines of the same file and both merge cleanly in either order; if the gate is to be in 0.23.0's tree rather than the following patch, merge #189 **before** you tag. #188's `Desktop Tests` has since reported success, so its round-1 precondition is met and its round is fresh on `834108f0d`.

damianvtran:

Head unchanged: `1aadcad71` (`fix/restore-check-themes`). Nothing pushed, nothing
amended; CI on this head is 8/8 green. Both findings dispositioned below.

| finding | severity | disposition |
|---|---|---|
| MINOR-1 — wrong mechanism explanation (not "a side that predated the script") | minor | **fixed in PR body**; the commit message is untouched, deliberately |
| MINOR-2 — no workflow runs `check-themes` | minor | **`deferred — separate PR`**, stated plainly in the body |

**MINOR-1 — accepted, and independently re-verified before accepting it.** I
re-derived the reviewer's evidence myself rather than taking it on trust:

```
$ git rev-list --parents -n1 110e24cc7
110e24cc7e0ea88dbd1ce80348a15e768d574dd8 cba43ee19ea7b55af57641c92780d9251dae9e50 9e8a61b3df223c11bfdf3f8eb5945d1311f23c6d

# scripts entries / check-themes present, per parent:
cba43ee19 (panel-views branch)  42  present
9e8a61b3d (#172's merge, main)  42  present
110e24cc7 (the merge itself)    41  absent

# script-key set changes vs EACH parent: the same single removal, nothing added
$ diff <(scripts of cba43ee19) <(scripts of 110e24cc7)   ->  22d21  < check-themes
$ diff <(scripts of 9e8a61b3d) <(scripts of 110e24cc7)   ->  22d21  < check-themes
```

So the reviewer is right and I was wrong: the merge loses an entry **both parents
carried**, and differs from each parent by that removal alone. That is an evil
merge, not a resolution against a stale side. My original reading was the
ordinary explanation, which is exactly the reason it needed an independent check.

The PR body now says evil merge in the summary, carries the parent table and the
per-parent counts above in "Why it disappeared", and states the false lesson
explicitly — that "both sides agree, so the resolution is safe" is not a safe
rule, since a conflict resolution is where version control stops checking the
merge and a deleted line hides inside a rewritten block. The body also records
that the earlier wording was wrong.

**Where the wording lives — over to you, per your instruction.** The commit
message still carries the old "resolved against the wrong side of a conflict"
phrasing. I have not amended it: doing so would move the head after the review
round was raised, and the reviewer said this is not a reason to hold. My read is
that the corrected explanation belongs where a reader of the incident will look —
the PR body — and that a stable head is worth more than a reworded subject line
whose own subject is still accurate (`fix(scripts): restore the check-themes
script a panel reconciliation dropped`). If you would rather the commit message
carry it too, say so and I will reword it as a follow-up commit on a fresh round
rather than rewriting history under a reviewed SHA.

**MINOR-2 — `deferred — separate PR`.** Verified again on this head: the only
reference to `check-themes` anywhere under `.github/` is the PR template's
checklist line; no workflow step runs it. Restoring the alias makes the
documented gate runnable, and **this PR adds no enforcement of anything** — the
body now says so in bold, and the bullet is labelled `deferred — separate PR`.
The reviewer's argument for doing it in the same session is the repository's own:
the `Type Checking` job already documents "a guard nothing runs is
indistinguishable from no guard", and `Desktop Tests` exists because
`test:desktop` ran in no workflow. Fixing it here would make it two edits in two
files instead of a one-line restore, which is the boundary you drew for this PR;
it belongs in a follow-up and I have not smuggled it in.

**Independently reproduced by the reviewer, acknowledged:** the absence of
`check-themes` at `origin/main`, the drop trail through `6fcd15cd6` →
`110e24cc7`, the reproduction of `ERR_PNPM_NO_SCRIPT`, the one-line diff, the
`version` line untouched at `0.22.7`, and the contract itself still passing
(2327 assertions, 12 themes, 0 pinned exceptions) — the last of which is the
whole point of the PR: the contract never regressed, only the documented way to
run it disappeared.

No other changes. Ready to merge as far as this side is concerned.

damianvtran:
### Released in `v0.23.0`

Your work shipped. Release owner for this window: session pid 20190.

- **Tag / Release**: https://github.com/damianvtran/local-operator-ui/releases/tag/v0.23.0
- **Version**: `0.23.0` (minor) — tag points at `680cffa9fc76da775c13d307edabb0b47d94b1bd`, the bump merge
- **npm**: `local-operator-ui@0.23.0` published, `latest` now resolves to it
- **Installers attached**: 20 assets — macOS `arm64`/`x64` dmg+zip (plus blockmaps), Windows `x64`/`arm64` setup + universal, Linux AppImage, `.deb`, `.rpm`, plus the `latest*.yml` feeds
- **Publish workflow**: run [`34925961401`](https://github.com/damianvtran/local-operator-ui/actions/runs/34925961401) — success end to end, including `Promote release to latest`, so `/releases/latest` resolves to `v0.23.0`

**Window** (everything merged since `v0.22.7`): #181, #165, #189, #188. Notes list all four, and the Release body records the two deferred items and the one QA dimension that was blocked by tooling rather than by the change.

PR #192, opened 2026-09-15
feat(browser): unpacked extensions, verified in a real Electron session

damianvtran:
## What this is

Finishes the unpacked-extension manager: a user can point the app at a directory,
review the manifest's declared access, approve it, and the extension is then
loaded into the **shared browser profile** — the same `persist:local-operator-browser`
partition every browser tab uses — where it survives restarts. Enable, disable and
remove only ever touch registration metadata; no code path deletes a source
directory or clears extension storage.

The manager itself (`src/main/browser/extensions.ts`, `extension-ui.ts`, the five
IPC channels and the Extensions sheet) landed in the checkpoint commit on this
branch. This PR is the work that makes it *verified and reviewed* rather than
extended: native verification, an honest capability matrix, rendered evidence,
and the gate wiring the checkpoint was missing.

Retargeted onto `main`: `#177` has merged and shipped in `v0.24.0`, so this
branch was **merged forward** onto `main` rather than rebased - merge `408b977a2`,
re-derivation `84c629f21` - which keeps the review's `2f42f83a7` and the
remediation's `4f3919e20` as ancestors. Nineteen paths conflicted; every one that
`#177` owns resolved to `main`'s side, this branch's own added lines were
re-applied onto `main`'s text where this work does edit them, and the two derived
files (`docs/evidence/manifest.json`, `package.json`) were re-derived rather than
line-merged. The merge record, the content-identity proof and the gate results are
in the comment below.

Release: patch - a chosen unpacked extension is reviewed, approved and loaded into
the shared browser profile, where it survives restarts; the supported Chrome API
subset and the one limitation (no native messaging) are both measured and
documented.

## The limitation, plainly

**Electron denies native messaging, so a 1Password-class password manager cannot
work here.** Measured, not assumed: with `nativeMessaging` declared the API *is*
present — `chrome.runtime.connectNative` returns a port object — and every host
connection is then refused:

```
sendNativeMessage -> "callback reply=undefined lastError=Access to the native
messaging host was disabled by the system administrator."
```

Two things this does **not** claim:

- **The real 1Password extension was never loaded here**, so its standalone
  behaviour is **unknown, not refused**. Some of it may render; none of it was
  measured. Understating the limitation would be as wrong as overstating it.
- **The login path that does work is the shared profile**: a cookie with an
  expiry written in one process was read in the next, and so was the extension's
  own storage. Sign in once in the app's browser and the session is there for
  every tab.

`docs/browser-extensions.md` carries the full matrix, cell by cell, with the
[...]

PR #190, opened 2026-09-15
feat(dev-driver): let an agent drive the built renderer, opt-in only

damianvtran:
# PR Description

## Summary of Changes

An agent could read this app's JSX and could not see it run. The `browser` tool
drives the operator's own browser and cannot enter an Electron renderer, and
loading the renderer under a bare Vite server dies immediately at
`Cannot read properties of undefined (reading 'ipcRenderer')` because `App`
assumes its preload bridge (`src/renderer/src/app.tsx:102`). A UX round for a
browser-chrome change was recorded as **blocked** on exactly that, and design
rounds could only argue from the source.

This adds the supported way through: an opt-in dev-driver bridge and a committed
driver script that boots the **built** app headless in an isolated scratch
profile, drives the renderer's real code paths, and captures what the app
actually paints — out of the app's own mechanisms only.

- `src/main/dev-driver.ts` — the arming decision, pure and Electron-free.
- `src/main/dev-driver-ipc.ts` — the two channels (`capture`, `facts`), registered
  **only** when armed; `capture` is `webContents.capturePage()`.
- `src/preload/dev-driver.ts` — exposes `window.__loDevDriver` only when main put
  the arming entry on the window.
- `src/renderer/src/dev-driver/install.ts` — the verbs, registered by the app's
  own module (route, state, navigate, setTheme, press).
- `scripts/renderer-driver.mjs` — the driver: isolation, launch, CDP attach,
  scenes, assertions, frames; `--gate-check` measures the fail-closed gate.
- `scripts/dev-driver-gate.test.mjs` — pins the gate in-process, in
  `pnpm test:desktop`.
- `docs/agent-driver.md` — the contract: commands, verbs, and what it can and
  cannot prove. `AGENTS.md` points at it from "Capturing the frame".
- `docs/evidence/renderer-driver/` — the pair from its first real use.

No dependency is added. No browser engine is installed or scripted — no
Playwright, no Puppeteer, no downloaded Chromium, and no `screencapture`.

The command an agent runs:

```
pnpm build
node scripts/renderer-driver.mjs --scene states --out /tmp/frames
```

Release: patch — agents can now walk the real renderer flow and capture
before/after frames of a screen instead of reviewing it from the source.

## Related Issue(s)

- Related: the design/UX rounds of #177, which were recorded as blocked on
  exactly this (no way to enter the trusted renderer).
- No issue is opened for it — the operator's standing instruction is not to
  create follow-up issues.

## Impact

[...]

PR #184, opened 2026-09-15
fix(backend): detect the running daemon from its record, and source the server signal from main

damianvtran:
## What this is

The app used to decide whether a backend existed by fetching **one configured
URL** and accepting any `200`. On this machine three daemons answered at once
(:1111 the bundled venv, :7341 the global uv tool with its TUI long gone, :8080
a dev server eleven releases behind), and the row that was supposed to name the
serving daemon named whichever one the probe happened to reach — or, once a
daemon's desktop plane was claimed and its origin allowlist refused the renderer's
`file://` origin, printed `Unavailable` for a daemon that was up and answering.

This PR changes the question the app asks. It finds the **daemon** by the
rendezvous record the daemon publishes, validates that record with **pid liveness
and a `/health` identity match**, and makes **main** — the process that holds the
bearer, sends no `Origin`, and knows whether the daemon it attached to is still
that daemon — the single source of the server-status signal.

Release: patch — the app finds and keeps talking to the daemon it is actually
served by, instead of reporting a healthy server as down.

## Related Issue(s)

- Related: #180 (overlapping change; **whichever lands second rebases and drops the duplicate hunks** — see *Overlap* below).

## The two defects, and the mechanism behind each

1. **A live daemon the app did not start was invisible.** `lop serve` publishes
   `<config root>/run/serve/<pid>.json` naming the address it bound, an
   `instance_id` minted at startup, the install behind it and a `0600` claim key.
   Discovery now enumerates those records and ranks the survivors — the install
   the CLI runs first, then the configured address **last**. With
   `VITE_DISABLE_BACKEND_MANAGER=true` the old code returned `true` before probing
   at all, so a daemon started by a TUI could never be adopted; now that flag means
   *discover, but never spawn or kill*, and a daemon already claimed by a previous
   run of this app is adopted by authenticating with the key from its own record
   (the `409` the claim route answers only means "already claimed" — treating it as
   a stranger's daemon is what stranded the app).
2. **A healthy server read as down.** The renderer decided liveness by fetching
   `/health` itself, from the packaged app's `file://` document. A CORS decision,
   or the origin allowlist on a claimed daemon, therefore became a liveness
   signal. Main is now the source: `useServerHealth` consumes a snapshot over IPC
   (pull plus push, so a window opened later is not left waiting), only `detached`
   reads as offline, `degraded` is one or two missed probes on a connection that is
   there, and a capability refusal (`401`/`403`/`503` from a gated route) is
[...]

PR #183, opened 2026-09-14
fix(update): run the managed Python backend outside the signed bundle

damianvtran:
The first round of independent review, QA and design has been taken on `483a21da8` and worked in one pass at `9b8a05e9c`: the three remediation comments are on this PR (`### Agent review remediation — round 1`, `### QA remediation — round 1`, `### Design review remediation — round 1`), each finding fixed, rejected or deferred by name. A fresh round is owed on the current head — these rounds answered the previous one, not themselves. This carries the recovery of the approach in #174.

Supersedes #174. That PR protected the bundled interpreter with filesystem access controls. That does not close the failure: the access controls do not survive the ZIP the in-app updater stages, so the bundle is unprotected between ShipIt's swap and the next launch — the same window the reported failure lands in. An independent check of the published 0.22.2 artifact confirmed the seed itself relocates and runs outside its `.app` with signatures intact, so this replaces the mechanism rather than adding another layer to it. #174's round-1 review and QA findings are carried over and answered below as they are worked.

## Intended outcome

The managed Python backend stops resolving anything inside the signed app bundle, so no process the app starts, and no process using the app-managed environment, can write into a code-sealed bundle or depend on paths that an update replaces.

Observable acceptance criteria:

- A packaged macOS build carries no `Contents/Resources/python` or `python_aarch64` path, including as a symlink; the interpreter ships as inert data under a private namespace.
- The app copies the complete seed to a versioned per-user runtime outside every `.app` and creates the venv at its final path there before any Python runs; a run reports `sys._base_executable`, `base_prefix`, stdlib, `sys.path`, and the native module dependencies all outside every `.app`.
- A venv created against a replaced bundle fails closed instead of reaching into the replacement, and older environments, configuration, credentials and sessions are left untouched.
- An unpackaged instance uses its own runtime, venv and update state, and cannot write, clear or reap a packaged app's install state.
- The release gate verifies the app inside the shipped ZIP and inside the mounted disk image, not only the unpacked build output.
- A signed existing-release-to-candidate update is exercised by the real updater in a clean environment before release.

## Approach

- Ship the interpreter as an inert, signed seed under a new private resource namespace; keep the packaged app free of the legacy interpreter paths.
- Provision a complete, versioned runtime outside the bundle from that seed, verifying the copied manifest, contained symlinks and Mach-O signatures, then publish it atomically; a selected runtime is never mutated in place.
[...]

PR #194, opened 2026-09-15
chore(release): bump version to 0.23.2

damianvtran:
One-line release bump, `package.json` only: `0.23.1` → `0.23.2`. C0 scope.

## Window

Baseline is the last tag, `v0.23.1` (published 04:53:42Z). Source of truth is `git log --first-parent --oneline v0.23.1..origin/main`, re-derived immediately before this PR and again at tag time — on this repo a PR landing thirty minutes after a tag has already been left out of a release once today, so the list is derived rather than remembered.

The window contains one merge:

- **#183** `fix(update)`: run the managed Python backend outside the signed bundle — merge `60419179ef1302dc2c46cea954e4df7b80004902`

## Bump class

**Patch.** #183 is a fix. It does not add a user-visible capability, a route, or protocol vocabulary: it moves where the managed Python runtime lives — a versioned, per-user runtime outside the signed app bundle, provisioned from an inert signed seed — and it stops the app writing bytecode into its own code-sealed bundle, which is what produced the reported *"Local Operator is damaged and can't be opened. You should move it to the Trash."* The only user-visible deltas are corrected refusal copy on states that already existed. Nothing here clears the step-function bar for a minor.

## Release ownership

Claimed and acknowledged. The window so far had one other contributor with a merge in it (#180, in the previous window); that session confirmed it is not claiming this one and has no competing bump. Peers with PRs in flight in this repo (#143, #162, #164, #169, #177, #184, #190–#192) were each told the baseline and target, and told explicitly not to hold for them; anything that merges before I tag is included in the notes and anything that misses rides the next window rather than being quietly dropped.


damianvtran:
### Agent review - round 1

Reviewer: `reviewer` subagent (local-operator team, lopdev) running `deepseek/deepseek-flash` — no provider fallback occurred during this round.
Scope: `60419179ef1302dc2c46cea954e4df7b80004902..696eac2db43f9b3c10b7162a0b0951588d23c80f` (i.e. `origin/main..head`); base is `main`.
Head SHA whose checks I read: `696eac2db43f9b3c10b7162a0b0951588d23c80f` — equal to the PR's `headRefOid` at review time, and the PR is `MERGEABLE` (not `DIRTY`), so the checks below are the real head's and not the stale-head-with-no-CI trap this repo has hit before.

---

**1. The diff is what it claims — PASS**

`git diff --name-status origin/main..696eac2db` returns one line: `M package.json`. `--stat`: `1 file changed, 1 insertion(+), 1 deletion(-)`. The diff is exactly:

```
-  "version": "0.23.1",
+  "version": "0.23.2",
```

Whole-file comparison against `main` (not just the hunk): `git show origin/main:package.json` vs the same at head — the two files are **both 15653 bytes**, and `diff -u` emits that single version line and nothing else. So there is no re-serialisation, no indentation/ordering churn, no dropped or added key: the top-level key sequence is byte-identical on both sides (`name, version, productName, description, main, bin, license, author, contributors, repository, homepage, bugs, keywords, files, publishConfig, scripts, dependencies, //types-node-pin, //packaged-node-modules-exclusions, //electron-pin, optionalDependencies, devDependencies, peerDependencies, engines, build`) with no duplicate keys on either side. Cross-checked against the forge rather than my clone — `gh api .../contents/package.json?ref=696eac2db` → `"version": "0.23.2"`, line 3. Compare API agrees: 1 commit ahead, 0 behind, one file, `+1 -1`.

**2. The window derivation is right — PASS**

I ran `git log --first-parent --oneline v0.23.1..origin/main` myself. The complete list is one entry:

```
60419179e Merge pull request #183 from damianvtran/fix/managed-python-runtime
```

`v0.23.1` resolves to `eadeadf8a` (the pre-#183 tip) and is an ancestor of `origin/main = 60419179e`, so the baseline is correct. That is exactly what the PR body claims — one merge, #183, `60419179ef1302dc2c46cea954e4df7b80004902` — with nothing merged since the tag that the body would silently leave out. The branch itself sits directly on that tip (0 behind), so the bump lands on the #183 tree and not on an older one. Release notes from this list are complete.

**3. No version bump in the other direction — PASS**

The branch is 0 behind / 1 ahead of `main`, merge-base `60419179e` = current `origin/main`, so the bump cannot land on top of a different version, and no rebase or re-derivation is pending. `v0.23.2` is absent from `git ls-remote --tags origin` (newest tag is still `v0.23.1`), so it is an unused number. Competing bumps: I checked every open PR (#143, #162, #164, #169, #173, #176, #177, #184, #190, #191, #192) by reading `package.json` at each PR's **merge-base** and at its head — all eleven report the same version on both sides (`0.23.0`/`0.23.1`/`0.22.3`/`0.22.5` as per their bases), i.e. none introduces a version change. `#194` is the only PR carrying `0.23.2`. Two of them (#192 on `feat/browser-chrome`, #176 on `main`) show an *older* number than their base branch's current tip, but that is a stale head, not a rewind: merge-base == head in both, so merging either cannot move the version backwards. No competing bump to `0.23.2` exists.

**4. The repo's own guards — PASS**

- `scripts/version-bump-guard.mjs` requires, for a bump PR: a title starting `chore(release):` (and `chore(release): bump version to ` for the "must actually change the version" arm), a `package.json` diff that touches the version line, a parsable semver, and one that **sorts above the highest remote tag**. This PR's title is `chore(release): bump version to 0.23.2`; `0.23.2 > v0.23.1`; the arm that fails a release PR whose diff changes nothing is not triggered. CI confirms rather than my reasoning: run `34933503275`, job `104266433253` (**Version Bump Guard**) `pass` in 10s on head `696eac2db`.
- `scripts/validate-release.mjs` is not a PR-time gate — it runs in `publish.yml` at `release: published`. What it will require of this bump: the tag `v0.23.2` must resolve to a commit, the release must be published and not a draft, and `package.json` **at the tag SHA** must be `name: local-operator-ui` and `version: 0.23.2`. That is satisfied only if `gh release create v0.23.2` targets a commit that carries this bump (the #194 merge SHA, or a descendant that still carries `0.23.2`). Noted for the tag step, not a defect in this diff.
- Whole PR CI on head `696eac2db` (run `34933503360` + `34933503275`): Lint, Type Checking, Desktop Tests, Runtime Dependencies, Security Audit, both NPX Sanity Checks and Version Bump Guard — **8/8 pass**. No pending or failing check remains.

---

**Out of scope for this PR, flagged because it affects the release window (not a finding against #194)**

`main`'s own push run `34933384327` on `60419179e` is **red**: `Installation Scripts Sanity Check` → `macOS Installation Script Check` →
`src/main/backend/scripts/macos-install-script.sh: line 30: LOCAL_OPERATOR_VENV_PATH: Pass the resolved managed environment path (see venv-paths.ts); this script will not guess which instance it belongs to` → exit 1. Cause: #183 made that script require `LOCAL_OPERATOR_VENV_PATH`, but `.github/workflows/install-scripts-check.yml:100` still invokes it bare (`bash src/main/backend/scripts/macos-install-script.sh`), so the check is now stale against the script's contract. It cannot gate this PR — that workflow triggers only on `push` to `main`/`dev-*` with a `src/main/backend/scripts/**` path filter, so it never runs on this branch — and `publish.yml` does not invoke it, so it does not block the release either. I am not treating it as a finding against #194; it is a real defect in #183's own merge and will keep `main` red until the workflow passes the variable (one line in the workflow, or an accepted default inside the script). Deferred to the #183 owner, not fixed here.

---

### Verdict

`clean` — **TERMINAL** for this round. One file, one line, `0.23.1 → 0.23.2`, window list re-derived independently and matching the body exactly, no competing bump anywhere, both repo guards satisfied with green CI on head `696eac2db43f9b3c10b7162a0b0951588d23c80f`. Safe to merge as a bare version bump; no BLOCKER, no MAJOR, no remediation round required.

PR #193, opened 2026-09-15
chore(release): bump version to 0.23.1

damianvtran:
### Release window after v0.23.0 — bump to 0.23.1 (patch)

One file, one line: `package.json` `version` `0.23.0` → `0.23.1`. No code, no evidence, no asset changes.

**Window contents** (both already merged to `main`; no PR in this repo carries its own bump):

| PR | What landed | Merge |
|---|---|---|
| #154 | `feat(draft)`: a new conversation's model and effort are chosen before its first message — the pane's readings became actionable controls, the pick is resolved through `sessions.preview` before being recorded and rides `sessions.create`, so the conversation is born on the choice | `8f50dc57dd39aad8f4815352117d34f4725c23f7` |
| #180 | `fix(serve)`: the owned-serve lifecycle | `bafebe2bf` |

**Why patch and not minor:** nothing in this window is a step-function change. #154 makes an existing pane's readings actionable and #180 fixes a lifecycle — both additive to behaviour already shipped, and the backend half that honours a draft pick has been released in the harness since 0.54.46.

**Gate state for the window:** #154 — code round 8 clean/terminal, a real browser-tool QA click pass (round 7, 26 frames), the rendered design gate signed off (round 7), a live UX walk terminal for the pointer flow (round 2), CI 8/8 on the merged head; deferred findings recorded in its thread with reasons. #180 — normal gate.

Tag and Release follow this merge, with notes listing both PRs.

PR #195, opened 2026-09-15
fix(ci): pass the managed environment path the install-script check now requires

damianvtran:
## The regression

`#183` made the macOS install script refuse to guess which environment it belongs to — `: "${LOCAL_OPERATOR_VENV_PATH:?…}"` — and made `PYTHON_BIN` a required hand-in. Both are deliberate: on macOS the standalone default is the exact shared environment the managed-runtime split exists to stop overwriting (a dev venv whose stdlib resolves into the installed, code-sealed `.app`).

The `macOS Installation Script Check` job still called the script bare, so it stopped at the refusal — `run 34933384327` on `60419179e`, the merge of #183:

```
src/main/backend/scripts/macos-install-script.sh: line 30: LOCAL_OPERATOR_VENV_PATH: Pass the resolved managed environment path (see venv-paths.ts); this script will not guess which instance it belongs to
exit 1
```

**Impact is CI-only.** This workflow runs on pushes that touch `src/main/backend/scripts/**`, and `publish.yml` does not invoke it, so the release was not blocked — but main's push run has read red since that merge, which is worth fixing before it is read as the release's state.

The Linux and Windows jobs pass *without* naming anything because their scripts default `LOCAL_OPERATOR_VENV_PATH` to the packaged name, which their own comments document as harmless on those platforms. macOS is the one platform where that default is the measured failure, so its job names its instance rather than inheriting a default — the same shape as its siblings, not a special case.

## What changed

**`.github/workflows/install-scripts-check.yml`** — the macOS job now exports the two names the script asks for:

- `LOCAL_OPERATOR_VENV_PATH` — the pre-split `local-operator-venv` layout under the runner's own `$HOME`, which is what the job's verification step activates and the same layout the Windows job models. It is the runner's to write to, and a real install never sees it: the packaged app names a selection-derived path under `managed-python/` at runtime.
- `PYTHON_BIN` — the bundled tree for the runner's architecture, which is what the job's own symlink step has always set up.

The script's refusal is **untouched**, and the workflow says why a default must not be added in its place: a value that satisfies the refusal without meaning anything would turn this check into one that proves nothing about the environment the app hands over.

**`src/main/backend/scripts/macos-install-script.sh`** — a second regression from the same change, exposed by running the job's own command, and fixed here because the check cannot reach the end of the script without it. The architecture block #183 removed with the dead in-bundle search was the only place `ARCH` was set, and the FFmpeg download below it still read it — so `$ARCH` was empty for *every* caller, the app included, and any install on a machine without a cached ffmpeg exited 1 with `Unsupported CPU architecture for FFmpeg download:` before the venv existed. The read now sits at its one remaining point of use. This makes the PR not purely CI-touching, which is why it is called out here rather than left to the diff.

**`docs/evidence/manifest.json`** — re-stamped: the script lives under `src/`, so `srcTree` moved and `scriptsTree` did not; `frames` 2024 and `surfaces` 256 are unchanged.

## Testing evidence

Every command below was run on this head; the scratch HOME is a temp directory, so nothing of the operator's machine was touched.

**1. The refusal still fires — the bare command, exactly as the workflow ran it:**

```
$ HOME=/tmp/lo-ci-refusal.…/home bash src/main/backend/scripts/macos-install-script.sh
src/main/backend/scripts/macos-install-script.sh: line 30: LOCAL_OPERATOR_VENV_PATH: Pass the resolved managed environment path (see venv-paths.ts); this script will not guess which instance it belongs to
EXIT=1
```

**2. The workflow's command with the new exports runs to completion (exit 0)** — the real script, a real `python -m venv`, a real `pip install` from PyPI, and the real FFmpeg download path the `ARCH` fix restores:

```
$ PYTHON_BIN=$T/resources/python_aarch64/bin/python3 \
  LOCAL_OPERATOR_VENV_PATH="$HOME/Library/Application Support/Local Operator/local-operator-venv" \
  bash src/main/backend/scripts/macos-install-script.sh
Virtual environment structure verified
pip upgrade successful:
pip 26.2.1 from /private/tmp/lo-ci-proof.…/local-operator-venv/lib/python3.14/site-packages/pip (python 3.14)
Successfully installed … local-operator-0.55.2 …
Testing local-operator binary...
v0.55.2
Tue 15 Sep 2026 01:48:16 EDT: Installation completed successfully.
EXIT=0
```

**3. The job's verification step, verbatim, against that same HOME:**

```
$ source "$HOME/Library/Application Support/Local Operator/local-operator-venv/bin/activate"
$ local-operator --version
v0.55.2
VERIFY_EXIT=0
```

**4. Repo gates on this head** (`5f0adef65`): `pnpm lint` 0 · `pnpm check-types` 0 · `pnpm check-themes` 0 (2339 assertions, 12 themes) · `pnpm test:desktop` **1226/1226, 0 failed, exit 0** · workflow YAML parses · `bash -n` on the script clean.

Not addressed here: the FFmpeg download URL in that block returns GitHub's 9-byte `Not Found` body and the script's check only tests existence and executability, so a fresh install "installs" a broken ffmpeg. It is pre-existing (the URL and the check are unchanged by #183), it does not affect this job's exit code, and it is outside this fix — flagged rather than folded in.


damianvtran:
### Agent review — round 1

**Reviewer:** independent reviewer subagent, model `deepseek/deepseek-flash` (no provider fallback occurred during this round).
**Scope:** `60419179e..5f0adef65` — base is `origin/main`, the #183 merge that left `main`'s push run red; head is the PR head `5f0adef65`. Files under review: `.github/workflows/install-scripts-check.yml`, `src/main/backend/scripts/macos-install-script.sh`, `docs/evidence/manifest.json`.

**Verdict: safe to merge for the release being cut now** — no BLOCKER, no MAJOR. Main today aborts a macOS install on any machine without a cached ffmpeg; this PR restores the pre-#183 behaviour exactly, so shipping it is strictly better than shipping `main`. Two MINOR and one NIT follow, none of them ship-breaking.

Every claim below was re-derived here, on the PR head, with every execution under a scratch `HOME` in `/tmp`; the operator's layout, `/Applications` install, sessions and launchd jobs were not touched. Nothing was merged, pushed, edited, or filed.

#### 1. The `$ARCH` fix is real, complete, and reaches the app's call path

`git diff eadeadf8a 60419179e -- src/main/backend/scripts/macos-install-script.sh` confirms the failure mode: #183 deleted the block that computed `ARCH=$(uname -m)` together with the in-bundle search, and the FFmpeg block below kept reading the variable. On the head script `ARCH` now appears only inside that one block (lines 118, 122, 124, 127 of the head blob), and `PYTHON_DIR_NAME`/`PYTHON_ARCH_NAME` are gone entirely — so the single restored read covers every path that reaches it.

Three-way reproduction, same shape of scratch `HOME`, `PYTHON_BIN=/nonexistent/python3` so control flow stops at the step *after* FFmpeg and therefore proves the block was passed rather than exited:

| script | FFmpeg outcome | result |
|---|---|---|
| `eadeadf8a` (pre-#183) | chose `…ffmpeg-mac-arm64` | reached `Error: Bundled Python not found at /nonexistent/python3` |
| `60419179e` (main today) | none | `Error: Unsupported CPU architecture for FFmpeg download:` exit 1 |
| `5f0adef65` (this PR) | chose `…ffmpeg-mac-arm64` | byte-identical to the pre-#183 run |

This is the no-cached-ffmpeg case — every run starts with an empty support directory — which is the user-facing one. Both architectures are covered by the same single value: `uname -m` is `arm64` on this host (the `elif` arm), and the `x86_64` arm is selected by the same read. Output ordering, the URL and the exit path all match pre-#183, so nothing else moved with the fix.

**The app still reaches this script**, and nothing in the app supplies an architecture. `src/main/backend/backend-installer.ts` writes `macosInstallScript` (exported from `src/main/backend/scripts/index.ts`) to a temp file and runs `bash` on it; the environment it hands over is `withPythonBytecodeCache(process.env)` plus `HOME`, `ELECTRON_RESOURCE_PATH`, `LOCAL_OPERATOR_SUPPORT_PATH`, `LOCAL_OPERATOR_VENV_PATH` and, when resolved, `PYTHON_BIN`. There is no `ARCH` anywhere in it, and the managed-runtime path is the same — `scripts/verify-managed-python.mjs` builds `installEnv` from exactly `PYTHON_BIN` + `LOCAL_OPERATOR_VENV_PATH`. So the implementer's "empty for every caller including the app" is correct, and the script-side read is the only possible fix: the app has no architecturally-correct value to pass that the script's own `uname -m` does not already produce for the interpreter it was handed.

#### 2. The refusal is untouched, and still fires

`: "${LOCAL_OPERATOR_VENV_PATH:?…}"` is at line 30 of the head blob, unchanged by this diff; no default was introduced anywhere in the script. Four cases run against the head script:

| case | result |
|---|---|
| `LOCAL_OPERATOR_VENV_PATH` unset, run bare | `LOCAL_OPERATOR_VENV_PATH: Pass the resolved managed environment path (see venv-paths.ts)…` exit 1 |
| `LOCAL_OPERATOR_VENV_PATH=""` (empty string) | same refusal, exit 1 — `${VAR:?}` treats empty as unset, so exporting an empty value cannot silence it |
| venv named, `PYTHON_BIN` unset | `PYTHON_BIN: Pass an external prepared Python executable` exit 1 |
| venv named, `PYTHON_BIN=/Applications/Local Operator.app/…/python3` | `Refusing to execute Python inside an application bundle` exit 1 |

The job satisfies the refusal by naming an instance, not by exporting something meaningless: the value it exports is `$HOME/Library/Application Support/Local Operator/local-operator-venv`, which the job's own `Setup environment` step creates the parent of, and which its verification step then activates — the same pre-split layout the Windows job models. It contains spaces and the full path is quoted at every use, which my end-to-end run below exercised for real.

Paths that are not the caller's own instance behave as #183 intended, and it is worth stating plainly because it bounds what the refusal protects: pointing the variable at an existing plain file gives `Refusing to replace an existing environment path` exit 1; at an existing empty directory gives `Virtual environment is missing critical components` exit 1; but at an existing *valid venv* the script silently reuses it and `pip install --upgrade`s into it, exit 0 (verified against the venv my own run had just built). The refusal answers "is this caller allowed to install at all", not "is this the right instance" — the path is the app's decision by design, and this PR does not change that.

#### 3. The macOS job is not vacuous, and the three jobs are now consistent in shape

The job exports the two names and then runs the script for real; nothing short-circuits. Its venv path does not pre-exist on the runner (only the parent directory is created by the setup step), so `python -m venv` genuinely runs, as does the pip install, and its verification step (`source …/activate`, `local-operator --version`, exit-code check) is the same shape as Linux's and Windows's.

Because this workflow triggers on `push` only (see R1), I re-ran the job's exact command myself on the PR head, under a scratch `HOME`:

- `PYTHON_BIN=/opt/homebrew/bin/python3 LOCAL_OPERATOR_VENV_PATH="$HOME/Library/Application Support/Local Operator/local-operator-venv" bash src/main/backend/scripts/macos-install-script.sh` → `Virtual environment structure verified`, `pip upgrade successful: pip 26.2.1 … (python 3.14)`, `Successfully installed … local-operator-0.55.3 …`, `Testing local-operator binary...` → `v0.55.3`, `Installation completed successfully.`, **exit 0**.
- The job's verification step verbatim → **`v0.55.3`, exit 0**.

(The PR body records 0.55.2 from the implementer's run; PyPI had moved to 0.55.3 by mine. No bearing on the review.) The `PYTHON_BIN` the job names is the tree it created for the runner's architecture — `uname -m` is `arm64` on `macos-latest`, so `resources/python_aarch64/bin/python3`, which matches the tree the packaged app ships for that host.

On consistency: all three jobs are `<checkout> → <setup> → run script → verify installation`, and the asymmetry is now documented at each of the three sites. Linux and Windows keep their own defaults (`${LOCAL_OPERATOR_VENV_PATH:=$VENV_PATH}` and the `$env:APPDATA` value); macOS refuses and is named. I also confirmed the claim that no other workflow invokes an install script: `git grep -l install-script <head> -- .github` returns only `install-scripts-check.yml`.

One question I considered and am **not** raising as a finding: the path the macOS job names is the legacy pre-split layout, not a `managed-python/<selection>` path such as the packaged app now resolves. The check would pass identically for any fresh path, so it cannot catch a path-*shape* mistake — but the script is path-agnostic (demonstrated above), the workflow's stated purpose is that the script runs, and the choice is disclosed in the new comment rather than implied. Fine as it stands.

#### 4. The manifest re-stamp matches the trees it describes, and the suite genuinely forced it

The re-stamp is required, not cosmetic. `scripts/check-evidence.mjs`'s `stampFailures` compares `srcTree`/`scriptsTree` against `git rev-parse HEAD:src` / `HEAD:scripts`, and `scripts/evidence-manifest.test.mjs` binds that verdict against the shipped manifest — "docs/evidence/manifest.json must describe HEAD's trees" — on every PR through `test:desktop`. Since the changed file lives under `src/`, the stamp had to move.

Re-derived independently on the head commit:

- `manifest.srcTree` `e3ad5a71…` == `git rev-parse 5f0adef65:src`
- `manifest.scriptsTree` `cd3ff1a2…` == `git rev-parse 5f0adef65:scripts` (unchanged, which is correct: `scripts/` is untouched)
- `surfaces` 256 == the `STORIES` row count in the head's `capture-evidence.mjs`, counted by the gate's own method
- `themes` 12 == the `THEMES` entries in the same file
- `frames` 2024 is untouched and correct: the diff adds, removes and modifies no `.webp`.

`head` moving to `c830a96c` is also right: its `src` tree is the same `e3ad5a71…`, it is an ancestor of the tip (which the citation half requires), and its `capturedAt` carried forward is exactly the convention this file documents for a re-derivation with no recapture. See R2 for the one thing left undone.

#### 5. The flagged FFmpeg item is correctly recorded and does **not** block this release

Confirmed independently:

- Both URLs return **HTTP 404 with a 9-byte body `Not Found`** (`curl -sL -w '%{http_code} %{size_download}'`), for `ffmpeg-mac-x64` and `ffmpeg-mac-arm64` alike.
- The check is `[ -f "$FFMPEG_BIN" ] && [ -x "$FFMPEG_BIN" ]`, and the script `chmod +x`es whatever `curl` wrote — so the 9-byte body becomes an executable file that satisfies every later check. I have the artefact: `-rwxr-xr-x 9 Sep 15 02:05 …/bin/ffmpeg`, content `Not Found`.
- It is worse than a fresh-install-only problem: the same `[ -f ] && [ -x ]` guard means the broken file is cached, so every later run reports `FFmpeg already installed … Skipping download.` and never repairs itself.
- It is genuinely pre-existing and **already shipped**: `git log -S "ffmpeg-static/releases/download/b6.0"` traces the URL to `a6cbe5e18` with no later change, and the current release tag `v0.23.1` carries it verbatim. Nothing in this repo reads `<support>/bin/ffmpeg` at runtime (`git grep -i ffmpeg` across `src/` finds the three install scripts and no consumer; the desktop suite stubs the file to skip the download), so the blast radius is whatever the Python backend wants it for, in a path this PR does not touch.

**Judgement:** it should not block this release. It is already out in `v0.23.1`, so holding this PR protects nobody; this PR's fix cannot cause it, and without this PR the same user gets a worse outcome (the whole install aborts). It is also correctly recorded — the PR body names it under "Not addressed here" with the mechanism, and the diff does not fold it in. The right disposition is a follow-up: replace the URL (and add a real content check — `curl -fL` plus something that proves the bytes are an executable, since an existence/size check cannot distinguish this), which is a standalone change and not a rider on a release that is being tagged now.

#### Findings

**R1 — MINOR (verification coverage).** `.github/workflows/install-scripts-check.yml` lines 3–10: the workflow triggers on `push` to `dev-*`/`main` only, with no `pull_request` trigger, so the job this PR repairs does not run on this PR. The green check list on `5f0adef65` (Desktop Tests, Lint, Type Checking, Security Audit, Runtime Dependencies, two NPX sanity jobs, Version Bump Guard) does not include it, and the first time this changed job executes in CI will be the push run on `main` after merge. The path filter does cover `src/main/backend/scripts/**`, so that run will fire — but note that this is precisely where the release window's only merged PR (#183) went red unnoticed. Concretely: watch `main`'s push run to green before the tag is cut, and treat a red result there as this PR's bug rather than a pre-existing CI condition. Not a blocker: the workflow is not invoked by `publish.yml` and gates no release artefact.

**R2 — MINOR (record consistency).** `docs/evidence/manifest.json`: `head` was re-pointed to `c830a96c` (correct, and consistent with the file's own convention) but `headNote` was not touched, so the file's own record still ends at "Re-stamped at the #193/#154 sync…" and names `703654d6d` as the head its stamps were derived at — a reader of the file alone cannot see why `head` moved. The immediate precedent for this exact class of change did append a sentence: `c2d62bde3` ("docs(evidence): re-stamp the merged trees after the #193/#154 sync") ends its `headNote` with one. No gate reads `headNote` (`stampFailures` reads only the two trees and the counts), so this is one sentence to add in the remediation commit, not a re-derivation.

**R3 — NIT (undisclosed scope).** `src/main/backend/scripts/macos-install-script.sh` also flips mode `100644 → 100755` in this diff, which the PR body and commit message do not mention. It is harmless — every caller runs it as `bash <path>` (the app writes its own copy to a temp file and chmods it `755`; the workflow `chmod +x`es before running) — but an unexplained mode change is a gratuitous difference in a file whose whole point is that its inputs are explicit. Either disclose it in the commit body, or drop it and let the callers' explicit `chmod` stand.

No other findings. Nothing rendered changes here, so no design or UX round is implied by this diff (a CI workflow, an install-time script and the evidence manifest).

damianvtran:
### QA report — round 1

**Reviewer:** qa-tester subagent, model `deepseek/deepseek-flash` (no provider fallback during this pass)
**Scope:** `60419179ef1302dc2c46cea954e4df7b80004902..5f0adef650d00b96c0f79aca2910d275ea0826ff` (= `origin/main`, the merge of #183, .. the head of this PR). Baseline arm additionally ran the last **released** installer, tag `v0.23.1` (`eadeadf8a`).

**Environment.** macOS 25.6.0 arm64 (Apple Silicon, `uname -m` = `arm64`), Python 3.14.7. Every arm ran with a temp `HOME` under `/tmp/qa-195-r1` and an isolated workspace tree mirroring the workflow's layout (`resources/python{,_aarch64}/bin/python3` symlinks), with the scripts extracted by `git show <rev>:…` — the repo was never edited, and nothing touched the operator's install, app-support state, sessions or launchd jobs. No machine-wide process sweeps.

**Verdict: PASS.** The regression is reproduced on the merged state exactly as described, the fix is verified end-to-end by executing this PR's own workflow steps verbatim, and both refusals still fire. No FAIL on head. Two findings stay for the record (Q-1, Q-2), neither of them blockable against this diff; my release judgment is below.

| # | Surface | Command | Actual output | Verdict |
|---|---|---|---|---|
| 1 | Regression as landed — bare call (`main` 60419179e) | `env -u LOCAL_OPERATOR_VENV_PATH HOME=$T bash src/main/backend/scripts/macos-install-script.sh` | `line 30: LOCAL_OPERATOR_VENV_PATH: Pass the resolved managed environment path (see venv-paths.ts); this script will not guess which instance it belongs to` — exit 1 | PASS (defect reproduced; this is main's red push run) |
| 2 | Regression as landed — **the worse arm**, with the exports this PR adds (`main` 60419179e) | `HOME=$T PYTHON_BIN=$T/ws/resources/python_aarch64/bin/python3 LOCAL_OPERATOR_VENV_PATH="$T/Library/Application Support/Local Operator/local-operator-venv" bash …/macos-install-script.sh` on a clean `$T` (no cached ffmpeg) | `FFmpeg not found or not executable. Attempting to download and install FFmpeg...` then `Error: Unsupported CPU architecture for FFmpeg download: ` (ARCH empty) — exit 1, **venv never created** (post-run tree holds only `bin/` and the log) | PASS (defect reproduced — fires for every caller, app included) |
| 3 | Fix, end to end — this PR's macOS job executed **verbatim** | the three `run:` blocks of `macos-install-check` at 5f0adef65, `GITHUB_WORKSPACE` = isolated tree, clean temp `HOME` | Setup Python exit 0 · run step exit 0: `Successfully created virtual environment` → `pip install` → `Successfully installed … local-operator-0.55.3 …` → `Local Operator backend installed successfully!` → `v0.55.3` · verify step exit 0: `macOS installation script check passed!` | PASS |
| 4 | FFmpeg artifact on the fixed path | `stat -f%z` / `file` / `xxd` / `…/bin/ffmpeg -version`, then re-run the script | 9 bytes, mode `-rwxr-xr-x`, content `Not Found`; exec exit **127** (`line 1: Not: command not found`); re-run: `FFmpeg already installed … Skipping download.` | PASS with Q-2 (the implementer's flagged claim is exactly right) |
| 5 | Baseline parity — last **released** installer (`v0.23.1`) | same as row 3 against `git show v0.23.1:…` | `Detected CPU architecture: arm64, using Python directory name: python_aarch64` → exit 0, real venv, `local-operator` `v0.55.3`; ffmpeg 9 bytes `Not Found` | PASS (#195 restores v0.23.1 behaviour; #183 alone is the window's regression) |
| 6 | Refusals still fire (7 arms) | unset · empty string · `PYTHON_BIN` unset · `PYTHON_BIN` inside `.app` · non-existent `PYTHON_BIN` · unwritable path · occupied path | each exit **1**: `LOCAL_OPERATOR_VENV_PATH: …` (unset and empty) · `PYTHON_BIN: Pass an external prepared Python executable` · `Refusing to execute Python inside an application bundle` · `Error: Bundled Python not found at …` · `mkdir: /System/qa195-nope: Operation not permitted` · `Refusing to replace an existing environment path: …` | PASS |
| 7 | Siblings | bare `bash linux-install-script.sh` (head) on a temp `HOME`; Windows script read only | Linux: exit 0, venv at `$HOME/.config/local-operator/local-operator-venv`, `local-operator v0.55.3` — the exact path its job verifies. Windows: not runnable on this host | PASS (Linux) / **BLOCKED** (Windows) |
| 8 | Evidence manifest stamps | `git rev-parse HEAD:src` / `HEAD:scripts` vs `docs/evidence/manifest.json` | `srcTree` `e3ad5a7192…` MATCH · `scriptsTree` `cd3ff1a2e8…` MATCH; `head` cites `c830a96ce`, a commit this branch carries | PASS |
| 9 | Script/YAML hygiene | `bash -n` on both revisions, YAML parse of both workflow revisions, grep for the removed block's leftovers | clean both revs; `jobs=[linux, windows, macos-install-check]`, triggers identical base→head; no dangling `PYTHON_DIR_NAME` / `PYTHON_ARCH_NAME` / `POSSIBLE_PYTHON_PATHS` / `ELECTRON_RESOURCE_PATH` | PASS |
| 10 | `pnpm test:desktop` counts (worktree at head) | `env -u NO_COLOR TERM=xterm-256color pnpm test:desktop`, twice | run 1: `tests 1226 · pass 1225 · fail 1 · skipped 0` (109.5 s) — run 2: `tests 1226 · pass 1225 · fail 1 · skipped 0` (110.6 s). The single failure is `submit-latency` M1/M2/M3 — Q-1 | PASS with Q-1 |
| 11 | CI coverage of the fix itself | `gh api …/commits/5f0adef65/check-runs`; workflow `on:` block | 8 check runs, **none** is "Installation Scripts Sanity Check"; that workflow is `on: push` to `dev-*`/`main` with `paths: src/main/backend/scripts/**` (unchanged by this PR) | **BLOCKED** (no CI observation possible pre-merge; substituted by row 3) |

---

#### Q-1 (FAIL, but not attributable to this diff) — `submit-latency` M1/M2/M3 fails 5/5 on this machine

Reproduction: `env -u NO_COLOR TERM=xterm-256color node --test scripts/submit-latency.test.mjs` → `tests 4 · pass 3 · fail 1`, 3/3 runs; the same test also fails inside both full `pnpm test:desktop` runs (1225/1226).

```
AssertionError [ERR_ASSERTION]: a cold session's first warm must START one rather than report it already warm
  'warm' !== 'warming'   at scripts/submit-latency.test.mjs:1472:11
```

Attribution, with a control rather than an assumption: the test spawns `BACKEND_BIN`, which defaults to `$HOME/.local/share/uv/tools/local-operator/bin/local-operator` (the operator's uv-tool backend, `v0.55.3`; its symlink was rewritten 01:55 tonight — around the time this PR's head was committed). That backend answers `state: "warm"` for a freshly-created session's first warm where the test asserts `"warming"`. The test file is byte-identical at `v0.23.1`, `60419179e` and `5f0adef65` (`sha1 7ee07d59d9a6`), and running it in the `v0.23.1` checkout — which contains neither #183 nor this PR — fails identically. So it is rev-independent and environmental, not this PR's; this PR's diff (a shell script, a workflow and the manifest stamps) touches nothing on that path. It is deterministic here, not intermittent — the "1226/1226" in the PR body is machine-state-dependent (on a CI runner with no binary at that path the test takes its `t.skip` branch). **Not a blocker for #195.**

#### Q-2 (pre-existing, correctly recorded, does not block this release) — the FFmpeg URL installs a 9-byte text file

Reproduction:

```
$ curl -sL -o /tmp/out -w "http_code=%{http_code} size=%{size_download}\n" \
    https://github.com/eugeneware/ffmpeg-static/releases/download/b6.0/ffmpeg-mac-arm64
http_code=404 size=9            # body: "Not Found"
$ ls -l "$HOME/Library/Application Support/Local Operator/bin/ffmpeg"
-rwxr-xr-x  1 …  9 …  ffmpeg    # chmod +x'd by the script, and it passes `[ -f ] && [ -x ]`
$ …/bin/ffmpeg -version ; echo $?
…/ffmpeg: line 1: Not: command not found
127
```

It is **correctly recorded**: the PR body states the URL returns GitHub's 9-byte `Not Found` body, that the check only tests existence and executability, that it is pre-existing and unchanged by #183, and that it does not affect the job's exit code. Every one of those matches what I measured. Worth adding for the follow-up: the asset exists but is named `ffmpeg-darwin-arm64` / `ffmpeg-darwin-x64` in `b6.0` (45,568,216 / 78,829,164 bytes) — the `-mac-` spelling in the URL has never matched any asset in any release of that repo since the URL was added (2025-06-03, `6ffb9c313`). A `curl -f` plus a `--version` probe would make it fail loudly instead of installing a text file; and because the check accepts the poison, the next run says `FFmpeg already installed … Skipping download`, so an affected install never self-heals.

**Should it block this release? No — and #195 should ship.** Measured reasons: (1) it predates the window — `v0.23.1`, the installer users get today, writes the identical 9-byte, mode-755 file (row 5), so this release is no worse than what is already published; (2) it fails neither the install nor the app: both `v0.23.1` and this head exit 0 and produce a working backend venv, and its blast radius is backend features that shell out to `ffmpeg` (`local_operator/computer_input.py`, `evaluation/runner/episode.py`); (3) without #195 this release would ship a macOS installer that fails outright on any machine with no cached ffmpeg (row 2), which is strictly worse. A follow-up PR for the URL is worth doing next, but it is not this window's gate.

#### Q-3 (informational) — the check that this PR fixes cannot run *on* the PR

The install-scripts workflow is `on: push` to `dev-*`/`main` with `paths: src/main/backend/scripts/**` (identical in both revisions). This PR's head carries 8 check runs and none of them is that workflow, so its own fix gets its first CI run on the post-merge push to `main` (which this PR's `src/main/backend/scripts/**` change does trigger). That is unchanged behaviour and not this PR's defect — but it does mean the merge is the first CI verification of it, so row 3's verbatim local execution is the pre-merge evidence, not a formality. No change requested here.

#### Q-4 (informational) — what the three jobs actually model

The macOS job now names its instance, which is the contract the app uses; the Linux and Windows jobs still rely on those scripts' silent defaults and their verify steps activate exactly those default paths (`$HOME/.config/local-operator/local-operator-venv`, `%APPDATA%\Local Operator\local-operator-venv`). Those checks are therefore meaningful (they do install and execute the backend) but never exercise a caller-named path — the macOS job is now the only one that does. Its own comment discloses that it names the pre-split layout rather than the app's `managed-python/…` shape; I confirmed a deeper caller-named path installs fine (`…/managed-python/other-instance/local-operator-venv` → exit 0, venv created, backend installed). Nothing to fix.

**Not covered, and why**

- **Windows arm: BLOCKED** — no Windows host and no `pwsh` here; the PowerShell job was read, not executed.
- **Linux arm honouring its runner's interpreter: partial** — run on macOS with `PYTHON_BIN` pointed at a 3.14 interpreter, because this host's `/usr/bin/python3` is 3.9, below the script's own 3.12 floor; `ubuntu-latest` resolves its own. What the arm establishes is the default-path behaviour, which is what its job's verify step keys on.
- **The in-app (Electron) installer flow was not driven.** That would need a packaged build and a real app-support layout. Substituted by (a) running the script with the exact environment the app passes and (b) reading the call site at this head: `backend-installer.ts:592` sets `env[VENV_PATH_ENV] = this.venvPath`, `:613` sets `PYTHON_BIN`, `:596-599` set `HOME`/`LOCAL_OPERATOR_SUPPORT_PATH` — i.e. the app names both variables, so the refusal is safe for it, and it hits the same FFmpeg block on a machine with no cached ffmpeg. That is source reading, not execution; I am not claiming it as coverage.
- No rendered/visual evidence: this diff has no user-visible surface (shell script, CI workflow, evidence stamps).

Scratch (`/tmp/qa-195-r1`, temp `HOME`s and the venvs built inside them) is reclaimed after posting this; no repo file, branch, tag, issue or PR state was modified.

PR #180, opened 2026-09-14
fix(main): own backend cleanup to the serve generation this app spawned

damianvtran:
## What and why

App shutdown used to search the machine for victims. `index.ts` (quit, `uncaughtException`, process exit) and `update-service.ts` ran `pkill -f "local-operator serve"`, `pkill -9 -f python` and a `ps aux | grep ... | awk '{print $2}' | xargs -r kill` pipeline on POSIX, and `taskkill /f /im python.exe /fi "WINDOWTITLE eq *local-operator*" /t` plus `taskkill /f /im local-operator.exe /t` on Windows — each preceded by a `ps aux`/`pgrep`/`tasklist` walk intended to *verify* termination.

Every one of those selects by name, not by ownership. Any other `local-operator serve` on the box — another tool's backend, a user's own daemon on a different config dir — matched and was killed along with the app's own child. On Windows the `/im python.exe` sweep is every Python process on the machine.

This PR replaces that with ownership of exactly one process: the `ChildProcess` this app spawned.

**Change class: C2 lifecycle bugfix** (standard, pre-authorized; the PR is the system of record).

## Commits

| Commit | What |
| --- | --- |
| `ce1d05fb2` | `fix(main): own cleanup to the serve generation this app spawned` — the code change plus the two fixture updates it requires. |
| `51a888b1e` | `test(main): pin the owned serve lifecycle with real children and HTTP` — the new `scripts/owned-serve-lifecycle.test.mjs` and its `test:desktop` entry. |
| `0f50a3f21` | `docs(evidence): re-stamp the manifest's trees for this head` — see *Evidence manifest* below. |

## The contract this implements

- **No machine-wide termination** in `src/main/index.ts`, `src/main/backend/*.ts`, `src/main/update-service.ts`. Grep and its output are in the evidence below.
- **Cleanup is confined to one app-owned serve generation.** A record holds the exact spawned child, created *before* readiness polling, together with that generation's stop promise and escalation timers. A stale `exit`/`error`/timer callback holds a reference to its own record, which nothing consults once `ownedServe` has moved on, so it can neither clear nor signal a later generation. A never-healthy or failed startup is still stopped. Incomplete cleanup reports incomplete and blocks replacement instead of claiming success. Numeric PIDs are never ownership handles, and a responding port is not ownership either. An external or paired backend gets no stop capability at all.
[...]

PR #176, opened 2026-09-14
fix(mcp): find the server /mcp names, and fix a broken one where it was found

damianvtran:
# PR Description

## Summary of Changes

`/mcp <name>` landed on Settings with **no server listed at all**, and an MCP
server whose sign-in had expired could only be fixed from a page the operator had
no reason to open during a run. This PR fixes the first as a defect and answers
the second with the remedy in place — one transition owner, shared by the run
panel, `/mcp login|reauth <name>` and Settings.

**The root cause is one cache key read two ways.** `["desktop","mcp",sessionId]`
was ONE React Query key with TWO query functions caching different shapes:

- `use-mcp-servers.ts` (the run panel) resolved
  `desktopResult<McpListPayload>(…)` and read `query.data?.data?.servers`, i.e. the
  transport envelope `{data: {servers, operations}, replayed}`;
- `mcp-management-section.tsx` (Settings) resolved `.then((result) => result.data)`
  and read `listQuery.data?.servers`, i.e. the unwrapped document.

Both type-check, so whichever queryFn wrote the entry last decided what the other
read — and because the panel polls every 15 s while a chat is open, it was nearly
always the last writer. The Settings section then read `undefined` inside its own
10 s `staleTime`, did not refetch, and rendered **"No MCP servers configured yet.
Add one below."** with twelve servers configured.

Measured on this branch **before** the change, extracting the read/write
expressions from the shipped sources and evaluating them against one cache value:

```
cache written by the panel:
  run panel reads   -> [{"name":"hubspot"},{"name":"notion"}]
  Settings reads    -> undefined
  servers the Settings section renders: 0
cache written by the settings:
  run panel reads   -> undefined
  Settings reads    -> [{"name":"hubspot"},{"name":"notion"}]
```

and **after** it, both surfaces read the same object off the same key:

```
one cache value, written by whichever queryFn ran last:
  run panel reads   -> [{"name":"hubspot"},{"name":"notion"}]
  Settings reads    -> [{"name":"hubspot"},{"name":"notion"}]
  the SAME object?  -> true
```

The second half of the operator's report was a second defect in the same flow:
`&mcp=` was compared WHOLE against a server name, so the operator's own remedy
line — `/mcp reauth hubspot` — matched nothing and the reveal effect returned in
silence.

## The change

| file | what |
| --- | --- |
[...]

PR #164, opened 2026-09-14
fix(chat): read code memory from the session that owns it

damianvtran:
# PR Description

## Summary of Changes

The canvas **"Code memory"** panel could never load. It called the legacy
`GET /v1/agents/{id}/execution-variables` with a **canonical session id**
(12 hex, e.g. `8fd6c6a40934`), and that route resolves through `AgentRegistry`,
whose keys are agent-directory UUIDs — so it answered 404 for every session:

```
GET /v1/agents/8fd6c6a40934/execution-variables
404 {"detail":"Agent with ID 8fd6c6a40934 not found"}
```

Verified live against the running app's backend. The tour teaches the panel, so
it stayed in the UI and never showed a namespace.

This PR rewires the panel onto the four session-scoped ops the backend grows
(`sessions.variables.list|create|update|delete`), deletes the legacy path from
the contract so the mistake cannot come back, and implements every panel state
the frozen design defines — including the two honest degradations.

## The change

| file | what |
| --- | --- |
| `src/shared/desktop-contract.ts` | four `sessions.variables.*` ops + their `desktopEndpoint` cases; the five `legacy.agent.variables.*` ops **deleted** |
| `shared/api/local-operator/session-variables-api.ts` | new: transport for the four ops, the response model, the one `VARIABLE_TYPES` table, `isSessionVariablesMissing` |
| `shared/hooks/use-session-variables.ts` | new: reader (polls only while `busy`) + the three mutations, which own the toast |
| `shared/api/local-operator/desktop-hooks.ts` | `"session_variables"` added to `DesktopFeature` |
| `canvas/canvas-variables-viewer.tsx` | takes `sessionId`, uses the new hooks, `EDITABLE_TYPES` deleted, §5 states added, 200/1000-char truncation unchanged |
| `canvas/variable-form-dialog.tsx` | the `VARIABLE_TYPES` fix: one shared table `str\|int\|float\|bool\|list\|dict`, default `str` |
| `canvas/index.tsx`, `chat-content.tsx`, `chat-page.tsx` | `sessionId` threaded, exactly as `cwd` is |
| `canvas/canvas.stories.tsx` | one story per state; the stub now answers `POST /__desktop` by `op` |
| `scripts/desktop-contract.test.mjs` | the four ops asserted end to end; the five legacy ops asserted **refused** |
| deleted | the five `agents-api.ts` methods, `hooks/use-agent-execution-variables.ts`, and the dead `ExecutionVariable`/`ExecutionVariablesResponse` types |

[...]

PR #162, opened 2026-09-14
feat(desktop): reach completions with no window, and route a banner click to one

damianvtran:
## What this is

The UI half of the notification-reach design (`notification-latency-design.md` §9.2, as corrected by `notification-latency-design-remediation-r1.md`).

The operator's report has two halves and both are addressed here:

1. **With the app running and focused on session A, a completion in session B announced nothing.** The completion gate was window-scoped: some window was focused, so the frame was suppressed, and no other surface could raise it because the desktop app was the eligible surface. The gate is now **session-scoped** — suppression requires a window *displaying that conversation*.
2. **With the window closed and the app alive in the dock, nobody announced it at all.** Main only ever observed frames from the one relay the renderer owns, and there is no renderer with no window. Main now subscribes the machine-wide feed (`GET /v1/desktop/events`) itself, beats `POST /v1/desktop/presence`, and a banner click recreates a window and queues the session id instead of returning.

### Main process

- `src/main/desktop-feed.ts` (new) — the feed subscriber: capability-gated on `features.desktop_feed`, frames into `DesktopNotifier.observe()`, presence heartbeat against the live subscription, and a **silence watchdog** (three missed heartbeats) that tears a dead socket down and reconnects.
- `src/main/viewer-record.ts` (new) + `src/main/viewer-endpoint.ts` (new) — publishes/heartbeats/unpublishes `run/viewers/<pid>.json` with `surface: "desktop"` (staged write, 0600 under 0700, 15 s beat against a 45 s timeout) and answers loopback `resume_session` / `focus_window` behind a constant-time key compare. The record is **main's**, never the renderer's: a renderer reload would advertise a port that died with the document. It carries whether a window exists.
- `src/main/desktop-notifier.ts` — `forgetWindow` gets its missing caller on window close; `anyFocused()` ignores dead `webContents`; suppression is session-scoped; a banner click with no window **recreates** one and **queues** the id (never a silent no-op).
- `src/main/index.ts` — `--open-session <id>` on first-run argv and `second-instance`; the initial session id reaches the renderer as a **value** (`webPreferences.additionalArguments`) so it is true of the first render, with the id held until the renderer reports the conversation on screen and a bounded fallback.
- `bin/local-operator-ui.js` — forwards `process.argv.slice(2)`, so rung 3 of the click ladder can actually launch the app with a session.
- `src/main/desktop-stream.ts` — a silence watchdog on the session socket too, and a refusal's HTTP status is carried through, which is what makes the deleted-session state reachable.

[...]

PR #197, opened 2026-09-15
fix(build): stage the interpreter seed where the packer expects it

damianvtran:
## The failure

`v0.23.2` publish run **34936130402**, job "Build macOS App" > step "Verify macOS artifacts", on the signed runner — **12 of 41 artifact checks failed**, the same two rows for both architectures and for the app inside both containers:

```
FAIL app-one-bundled-python: the bundled python interpreter tree this architecture needs
  -> the x86_64 app ships Contents/Resources/none, but it resolves Contents/Resources/python-runtime-seed/x64
FAIL app-private-python-seed: complete private Python seed with no legacy aliases
  -> Legacy Python resource or alias exists: python
PASS app-codesign / app-spctl / app-stapler / app-no-bundled-bytecode
```

## Reproduced from a clean state, before anything was changed

Staged trees removed (`resources/python`, `resources/python_aarch64`), then exactly the CI sequence — `pnpm install --frozen-lockfile`, `pnpm setup-python`, `pnpm dist:mac -c.forceCodeSigning=false` — and the gate over what came out:

```
FAIL app-one-bundled-python: the arm64 app ships Contents/Resources/none, but it resolves Contents/Resources/python-runtime-seed/arm64
FAIL app-private-python-seed: Legacy Python resource or alias exists: python
PASS app-no-bundled-bytecode: no .pyc/.pyo in the bundled python trees
```

So it is a real defect in what CI builds, not local state: `pnpm setup-python` stages the checkout trees under their legacy names, and `build.extraResources` copied them in under those same names — the app never got a `python-runtime-seed/<arch>` directory at all.

## Root cause

Two keys in `package.json`'s `build` config, both correct on this branch, both lost on the way to `main`:

- `build.extraResources` still mapped `resources/python` -> `python` and `resources/python_aarch64` -> `python_aarch64` (the legacy names);
- `build.afterPack` still named `scripts/prune-python-resource.mjs` — the pruner alone — instead of `scripts/after-pack.mjs`, the hook that prunes **and** then refuses a surviving legacy alias by name.

The merge resolution at **`d64175b56`** (syncing `#180` into the release chain) rebuilt `package.json` from `main`'s side and kept only the `test:desktop` union line, so every other change this branch had made to that file went with it — including `445d08124`'s seed mapping and hook registration. That is why the same build passed on a machine whose `dist/` came from an earlier commit, and why nothing failed for weeks: no test read the config, and the gate's first outing was this release.

## The fix

`package.json` restored at the source, in the shape `docs/BUILD.md` already documents (the docs were right; the config had drifted from them):

```json
"afterPack": "scripts/after-pack.mjs",
"mac": {
  "extraResources": [
    { "from": "resources/python",         "to": "python-runtime-seed/x64",   "filter": ["**/*"] },
    { "from": "resources/python_aarch64", "to": "python-runtime-seed/arm64", "filter": ["**/*"] }
  ]
}
```

`win` and `linux` keep the names they ship today — their jobs assemble no mac interpreter and the prune hook is macOS-only, so their output is unchanged. The top-level block is gone rather than left in place with a `mac` override on purpose: a top-level entry applies to macOS too, which is exactly how a legacy name can come back beside the seed.

**`pnpm setup-python` is deliberately NOT changed.** The trees are staged under their legacy names, and `src/shared/bundled-python-layout.json`'s `checkoutSeedNames` records that as the naming the shipped mapping exists to translate — so the config is the right seam.

The gap is pinned instead of remembered: `scripts/prune-python-resource.test.mjs` now asserts the registration and the mapping (reading `package.json`, since the modules were never the thing that broke) and drives the real hook against a bundle carrying an alias. Falsified: against the pre-fix config that test fails with `the registered afterPack must be the hook that refuses a legacy alias, not the pruner alone`, and the suite is green with the fix.

## Proof, both directions

**After (clean build, same gate):** all three seeded rows pass for **every app** — arm64 and x64, the unpacked app and the app inside both the ZIP and the mounted DMG:

```
PASS app-one-bundled-python: the bundled python interpreter tree this architecture needs   x6
PASS app-private-python-seed: complete private Python seed with no legacy aliases           x6
PASS app-no-bundled-bytecode: no .pyc/.pyo in the bundled python trees                      x6
```

The 22 rows that still fail here are only `app-codesign`, `app-spctl`, `app-stapler`, `dmg-spctl`, `dmg-stapler` — expected on a local build run with `-c.forceCodeSigning=false`, and all of them passed on the signed CI runner. Every row that passed in CI before still passes, and the 12 that failed now pass.

**Negative direction, against an app extracted from the fixed ZIP:**

```
$ cp -R "Local Operator.app/Contents/Resources/python-runtime-seed/arm64" \
        "Local Operator.app/Contents/Resources/python_aarch64"     # reintroduced alias
FAIL app-private-python-seed: Legacy Python resource or alias exists: python_aarch64

$ mv "…/python-runtime-seed" aside                                  # seed removed
FAIL app-one-bundled-python: the arm64 app ships Contents/Resources/none, but it resolves Contents/Resources/python-runtime-seed/arm64
FAIL app-private-python-seed: ENOENT: no such file or directory, scandir '…/Contents/Resources/python-runtime-seed'
```

Both refusals are by name, and neither was softened — the gate and the hook are untouched by this PR.

## Gates

`pnpm lint` 0 · `pnpm check-types` 0 · `pnpm check-themes` 0 (2339 assertions, 12 themes) · `pnpm test:desktop` **1228/1228, 0 failed, exit 0** (1226 + the two new cases) · `package.json` parses · manifest re-stamped (`scriptsTree` moved).

## Can the held v0.23.2 be repaired?

**Not by the dispatch path, as it stands.** Every build job checks out `needs.validate-release.outputs.source_sha`, and `scripts/validate-release.mjs` refuses a dispatch whose `expected_source_sha` differs from the tag's SHA — the tag `v0.23.2` points at `32e3d04b4`, which carries the broken config, so a re-run rebuilds the same 12 failures. A tag that contains this fix is needed: either the held prerelease and tag are deleted and `v0.23.2` is cut again from a `main` that has this fix, or the window moves to the next number. Either way it is the window owner's call, and nothing has gone out — `latest` still points at v0.23.1 and the prerelease has no assets.


damianvtran:
### Sync note — `main` moved while this waited (head `9c9cbb1b` → `b1a6fdbaf`)

`#164` landed on `main`, GitHub reported the PR CONFLICTING (and a conflicted PR gets no `pull_request` runs at all, which matters here because this is the fix the held release is blocked on). Merged, never rebased — same shape as every sync before it.

One conflict, the stamp block in `docs/evidence/manifest.json`: main's block is taken (its note describes the tree this merge inherits) with this branch's `partialCapture` kept, then the stamps are re-derived on the merged disk. `package.json` merged on its own, and both halves are checked rather than assumed:

- main's side: `test:desktop` is a complete union — 62 entries, **nothing missing from either side**, no duplicates;
- this fix: `afterPack` is still `scripts/after-pack.mjs`, `mac.extraResources` still maps `resources/python` → `python-runtime-seed/x64` and `resources/python_aarch64` → `python-runtime-seed/arm64`, and there is still **no top-level `extraResources`** entry (a top-level one applies to macOS too, which is how a legacy name comes back beside the seed);
- no version line in the diff against `main`.

Gates on the merged head `b1a6fdbaf`: `pnpm lint` 0 · `pnpm check-types` 0 · `pnpm check-themes` 0 · manifest stamps 15/15 · CI **all green** (`Desktop Tests` on the Linux arm: 1240 tests, 1213 pass, 0 fail, 27 skipped macOS-only — including both new cases: *the builder config stages the mac interpreters into the seed namespace* and *the hook refuses a bundle that still carries a legacy alias, by name*). Locally the suite is 1239/1240 with `scripts/submit-latency.test.mjs`'s *"M1/M2/M3: the warm removes the engage from the send"* failing intermittently on this loaded machine (`a cold session's first warm must START one rather than report it already warm`) — main's file, untouched by this diff, and green on the CI arm above.

damianvtran:
### Agent review - round 1

**Reviewer:** reviewer (Local Operator `lopdev` subagent) — model `deepseek/deepseek-flash` for the whole round; no provider fallback observed.
**Scope:** `c5b5b67edd6816dc2cf92669209a9df619d07e83` (merge-base with `origin/main`) .. `b1a6fdbaf3cd91811d20a3ad79b2447c6f1af14b` — 3 files: `package.json` (+49/-17), `scripts/prune-python-resource.test.mjs` (+66/-0), `docs/evidence/manifest.json` (+32/-54).
**Verdict: clean — safe to cut a release from.** No BLOCKER, no MAJOR. One MINOR and two NITs, all listed as follow-ups rather than gates on this merge.

Where I claim builder behaviour below, I executed it rather than reading it: `app-builder-lib`'s own `getFileMatchers` (the function `PlatformPackager.doPack` calls to place `extraResources`), taken from `app-builder-lib@26.16.1` **as pinned in `pnpm-lock.yaml`** — read out of the pnpm store, because the `node_modules` in this checkout is a stale 26.0.12. The two versions' `getFileMatchers`, `getExtraFileMatchers` and `platformSpecificBuildOptions` wiring are byte-identical.

#### 1. The config is right, and right for every platform

```
config            macOS destinations                              win / linux destinations
main (pre-fix)    Resources/python, Resources/python_aarch64       resources/python, resources/python_aarch64
head              Resources/python-runtime-seed/{x64,arm64}       resources/python, resources/python_aarch64
```

- `mac.extraResources` maps each checkout tree into the seed namespace, and the head carries **no** top-level `extraResources` (checked as full key paths, not by eye). That matters because `getFileMatchers` adds **both** `config[name]` (root) and `config[platform].extraResources`; `this.config` is the root config and `platformSpecificBuildOptions` is `config[platform]` verbatim. Executed with the head config plus one re-added top-level `python` entry, macOS destinations become `python`, `python-runtime-seed/x64`, `python-runtime-seed/arm64` — the alias-beside-the-seed vector is real, and removing the block rather than overriding it is the correct fix.
- `build.afterPack = "scripts/after-pack.mjs"` is consumed as `this.config.afterPack` in `app-builder-lib/out/packager.js`; the file exists at this head and its default export runs the pruner and then **throws** on a surviving legacy name (`lstatSync`, so a dangling symlink is refused too). The registration therefore names the hook that refuses, and that hook is what runs.
- **Windows/Linux are not collateral.** Executed the hook with `win32` and `linux` contexts against a realistic unpacked tree: it returns `{pruned: [], kept: null, resourcesDir: null}` without throwing and leaves both legacy trees in place — the alias refusal stays macOS-only by design. Both platforms' own consumers name exactly what this head ships: `backend-installer.ts:164-184` probes `resources/python[_aarch64]/bin/python3` and `resources/python/python.exe`; `linux-install-script.sh:225-226` probes `resources/python/bin/python3`. `build` has no `mas`/`snap`/`appx` block, so no third platform lost a copy list when the top-level one went.
- The mac block still copies both trees for both architectures and `afterPack` prunes the off-architecture one, exactly as before — the copy list is not architecture-aware and never was.

#### 2. The test is not vacuous — falsified, five ways

Ran `node --test scripts/prune-python-resource.test.mjs` in a scratch tree holding this head's `package.json`, both hooks and the layout modules, then mutated only `package.json`'s `build` block. Scratch reverted to the head config gives 12/12 pass.

| # | config mutated to | result |
| --- | --- | --- |
| A | `main`'s exact pre-fix block (legacy top-level, `afterPack: prune-python-resource.mjs`, no `mac` block) | **fail 1** — `the registered afterPack must be the hook that refuses a legacy alias, not the pruner alone` |
| B | only the registration reverted, mac seed mapping kept | **fail 1** — same assertion, by name |
| C | only the mapping reverted to the legacy top-level, correct hook still registered | **fail 1** — `a macOS-applicable extraResources entry may not write a legacy resource name` |
| D | legacy names moved into the `win`/`linux` scopes only | pass 0 — correctly not over-broad |
| E | an alias added *beside* the seed inside the `mac` block | **fail 1** — same assertion as C |

A fails on the first assertion and aborts, so C is the case that isolates the mapping half — both halves are genuinely pinned. This is a guard, not a restatement.

#### 3. The sync at this head dropped nothing

- Full key-path diff (`package.json`, deep, both directions) across the sync merge `630cbbe07`: against its branch parent, **0 keys present in one and absent in the other**, 1 intended change (`scripts.test:desktop`); against its `main` parent, only the intended `build.afterPack` / `build.extraResources` / `build.{mac,win,linux}.extraResources` deltas. `630cbbe07` → `b1a6fdbaf` is byte-identical for `package.json`.
- `head`'s `build` block is **key-for-key identical to `445d08124`'s** except `.version` (`0.22.3` → `0.23.2`, main's own released line — this PR carries no bump) and `.scripts.test:desktop` (the unrelated owned-serve suite list). That is the strongest available statement that the dropped state was restored, not re-invented.
- The one list that could have lost entries: `scripts.test:desktop` is the union of both sync parents — branch 60, main 62, head 62, with 0 branch entries missing and 0 head entries from neither side.
- I independently located the clobber rather than taking the PR body's word for it: scanning all 181 merges in `origin/main`'s history for "a seed-side parent resolved to a legacy-side result" gives exactly **one** hit, `d64175b56` (`chore(sync): merge main (#180)…`), whose resolution kept `main`'s side. That matches the commit the PR body names.

#### 4. The gate was not softened

`git diff --stat <base>..<head>` over `scripts/verify-macos-artifacts.mjs`, `scripts/python-artifact-layout.mjs`, `scripts/after-pack.mjs`, `scripts/prune-python-resource.mjs` and `.github/workflows/` is **empty** — no hunk in this PR touches the gate, either hook, or the publish sequence. Nothing was relaxed to make a build pass; the only reason the rows change is that the artifacts now contain what the gate already demanded.

#### 5. The reproduction

The implementer's claim reproduces from a clean state, and I reproduced both halves of its causal chain rather than the rows alone:

- **config → artifact layout**: executed above (section 1). Pre-fix `afterPack` was the pruner, which only deletes an off-arch tree and never renames, so with `main`'s config the mac bundle can only carry the legacy names — CI's `Contents/Resources/none` / `Legacy Python resource or alias exists: python` follow from the config alone, not from local state.
- **layout → gate rows**: drove the gate's own functions (`bundledPythonCheck`, `privatePythonSeedCheck`, `bundledBytecodeCheck`) against a synthetic bundle in each layout, injecting a stubbed `lipo` so the real code path ran with a real architecture. Pre-fix layout: `FAIL app-one-bundled-python — the arm64 app ships Contents/Resources/none, but it resolves Contents/Resources/python-runtime-seed/arm64` and `FAIL app-private-python-seed — Legacy Python resource or alias exists: python_aarch64`. Seed layout: all three rows `PASS`. That is CI's failure verbatim, and the negative direction (seed removed, alias reintroduced) refuses by name as the PR claims.

I did **not** re-run the full signed `dist:mac`. What that would add is the 22 `codesign`/`spctl`/`stapler` rows, which need the runner's keychain and are structurally untouched by a `package.json` resource mapping — so the residual risk of this change is confined to the rows I did reproduce.

#### Findings

- **R1 — MINOR.** `scripts/prune-python-resource.test.mjs:269-297` pins the macOS mapping and the hook registration, but the two per-platform blocks this fix *introduced* are asserted nowhere: no test in the repo reads `build.win.extraResources` or `build.linux.extraResources`. With the top-level block deliberately gone, a future sync that drops those keys — the identical class of loss this PR exists to close, and `d64175b56` shows sync resolutions do drop exactly these keys — ships Windows and Linux installers with no bundled interpreter and nothing in CI would fail: the mac gate is macOS-only and the win/linux jobs carry no python rows. Suggested one-liner: assert both platform blocks still stage the two checkout spellings (derived from `checkoutSeedNames`), not just that the macOS-applicable scopes are free of legacy names.
- **R2 — NIT.** `scripts/prune-python-resource.test.mjs:290-296` derives `to` from `seedResourceDir()` but spells the two `from` values (`"resources/python"`, `"resources/python_aarch64"`) as literals, re-spelling what `src/shared/bundled-python-layout.json`'s `checkoutSeedNames` declares as the single definition — the invariant `docs/BUILD.md` and the module headers both claim. Defensible as-is (a literal pin also catches a layout rename), so reject with a reason if that was the intent.
- **R3 — NIT.** The `afterPack` assertion compares against the literal `"scripts/after-pack.mjs"` and nothing asserts that the named file exists or exports a default. A rename would fail the same assertion, so the guard holds — but a missing-file failure would read as a config error rather than a missing hook. Not worth a round on its own.

Everything else checked out: `package.json` parses, the PR carries no version bump (0.23.2 on both sides), and the manifest re-stamp is honest — its `head` (`630cbbe07`) is a commit whose `git rev-parse :src` / `:scripts` trees are byte-identical to this head's, which is the claim its own note makes.

**Residual, not a finding.** `publish.yml` runs `pnpm setup-python` in the macOS job alone, so on a Windows/Linux runner the interpreter entries match nothing and copy nothing. That is pre-existing and *unchanged* by this PR — the removed top-level block produced the identical empty matcher output for those platforms — but "win/linux keep the names they ship today" should not be read as a claim that an interpreter actually ships in those installers today.

No edits, pushes, merges or issues were made. All review scratch (a 52 KB scratch tree plus temporary fixture directories) has been removed; the repo working tree is clean and the reusable `resources/` staging trees were not present and were not touched.

damianvtran:
### QA report — round 1

**Reviewer:** `qa-tester` subagent on **deepseek/deepseek-flash** (the model actually running this pass).
**Scope:** `c5b5b67ed..b1a6fdbaf` (`origin/main`..`fix/build-seed-layout` head).
**Verdict: PASS on every surface this host can reach.** No Q-findings: nothing I could exercise failed. Two rows are BLOCKED for stated reasons, and one is BLOCKED by a lease another session holds.

#### Environment

| | |
| --- | --- |
| Host | macOS 25.6.0, arm64 (Darwin 25.6), load average ~62 throughout |
| Worktree | `git worktree` at `/tmp/qa-197/wt`, detached at `b1a6fdbaf3cd91811d20a3ad79b2447c6f1af14b`, clean before and after (0 entries in `git status --porcelain`) |
| HOME | `/tmp/qa-197/home` — isolated. The operator's `~/Library/Application Support/Local Operator` does not exist on this state and was never read or written; there is no `Local Operator.app` in `/Applications` here. No `CMUX_*` variable was present at any point. |
| Node | `v22.13.1` for the packaging sequence (the pin in `publish.yml`), `v22.22.0` for `pnpm test:desktop` (see note N1) |
| pnpm | `10.29.2` (the pin in `publish.yml`), installed into the temp HOME |
| electron-builder | `26.16.1`, electron `44.3.0` |
| Build env | `VITE_GOOGLE_CLIENT_ID`/`_SECRET`, `VITE_MICROSOFT_CLIENT_ID`/`_TENANT_ID` set to the inert `*-build-only-not-a-credential` fixtures this repo already uses in `docs/evidence/settings-app-updates-section/RECOVERY.md` — no credential was read, and none of these four is read by any check below |
| Scratch | own worktree, own scratch under `/tmp/qa-197`. The shared checkout's `resources/` (icons only) is untouched. Nothing was killed, no process sweep was run. |

The base sequence was run for real, not inferred: staged trees removed from my worktree, then `pnpm install --frozen-lockfile` → `pnpm setup-python` → `pnpm dist:mac -c.forceCodeSigning=false` → `node scripts/verify-macos-artifacts.mjs`. 41 artifact checks at both revisions, exactly as CI.

#### 1. The failure, reproduced at the fix's own base (`c5b5b67ed`)

Same clean-state sequence; `pnpm setup-python` staged both checkout trees under the legacy names (`resources/python`, `resources/python_aarch64`, 47 MB each), and the build copied them in under those same names. `afterPack` was `scripts/prune-python-resource.mjs`, which reads the layout's *seed* names and therefore found nothing to prune — `No off-architecture bundled Python found for arm64` — so **both** legacy trees shipped in **both** apps.

The two CI rows, verbatim, for the unpacked apps (identical text x6 — 2 unpacked, 2 extracted from the DMGs, 2 from the ZIPs):

```
FAIL app-one-bundled-python: the bundled python interpreter tree this architecture needs
  -> the x86_64 app ships Contents/Resources/none, but it resolves Contents/Resources/python-runtime-seed/x64
FAIL app-one-bundled-python: the bundled python interpreter tree this architecture needs
  -> the arm64 app ships Contents/Resources/none, but it resolves Contents/Resources/python-runtime-seed/arm64
FAIL app-private-python-seed: complete private Python seed with no legacy aliases
  -> Legacy Python resource or alias exists: python
```

`ships none` is literal, and the directory listing says why — the gate's wording is not a substitute for it:

```
=== dist/mac/Local Operator.app :: Contents/Resources ===        (and likewise dist/mac-arm64/…)
drwxr-xr-x@  6 damian  wheel  192  python               <-- legacy x64 tree, shipped
drwxr-xr-x@  6 damian  wheel  192  python_aarch64       <-- legacy arm64 tree, shipped
ls: …/Contents/Resources/python-runtime-seed: No such file or directory
```

So the base bundle carries no seed namespace at all, and both legacy aliases. **34 of 41 checks failed here against CI's 12**; the extra 22 are exactly the five signing rows (`app-codesign`, `app-spctl`, `app-stapler` ×2 apps, `dmg-spctl`, `dmg-stapler` ×2 images) that `-c.forceCodeSigning=false` cannot satisfy locally. The 12 CI failed are the 12 I reproduce, on the same 41-check denominator. **PASS — the defect is in what the build produces, not local state.**

#### 2. The same sequence at the head (`b1a6fdbaf`)

Same clean-state basis, same commands, same gate. **22 of 41 failed — the signing rows only, and only those.** Every seeded row passed for every one of the six app instances:

```
PASS app-one-bundled-python: the bundled python interpreter tree this architecture needs          x6
PASS app-private-python-seed: complete private Python seed with no legacy aliases                  x6
PASS app-no-bundled-bytecode: no .pyc/.pyo in the bundled python trees                             x6
PASS final-artifact-metadata: release metadata describes exact delivered bytes
```

x6 = the two unpacked apps, the two apps copied out of the mounted DMGs, and the two apps extracted from the ZIPs. The afterPack log shows it working rather than merely not failing: `Pruned bundled Python the x64 app cannot run: …/python-runtime-seed/arm64`.

Directory listing of the packaged bundles — seed present, no legacy name, not even a dangling symlink (`-e || -L` on both names):

```
=== dist/mac/Local Operator.app ===          legacy names:  absent: python   absent: python_aarch64
    python-runtime-seed/ -> x64     /bin  /include  /lib  /share
=== dist/mac-arm64/Local Operator.app ===    legacy names:  absent: python   absent: python_aarch64
    python-runtime-seed/ -> arm64   /bin  /include  /lib  /share
```

Every symlink inside the seed is relative and contained (`bin/python3 -> python3.12`, `lib/pkgconfig/python3.pc -> python-3.12.pc`, …); zero `.pyc`/`.pyo`. **PASS.**

#### 3. The negative direction, against a container this head produced

App extracted from the arm64 ZIP with `ditto -x -k`, then tampered, then run through the real gate CLI (`node scripts/verify-macos-artifacts.mjs --app <tampered>`, exit 1) and through the two checks directly:

| Tampering | Actual output |
| --- | --- |
| `cp -R …/python-runtime-seed/arm64 …/python_aarch64` | `FAIL app-private-python-seed: … -> Legacy Python resource or alias exists: python_aarch64` |
| seed removed (`python-runtime-seed` moved aside) | `FAIL app-one-bundled-python: the arm64 app ships Contents/Resources/none, but it resolves Contents/Resources/python-runtime-seed/arm64`<br>`FAIL app-private-python-seed: … -> ENOENT: no such file or directory, scandir '…/Contents/Resources/python-runtime-seed'` |

Both refused **by name**, and neither was softened: with the alias present `app-one-bundled-python` still **passed** (the seed was intact), so the alias row is a separate real check and not a by-product.

I also ran the `afterPack` hook itself — the half that fails the *build* rather than the artifact — against the same tampered bundle and against the clean one this head built:

```
hook vs the ALIASED bundle : HOOK REFUSED: The packaged app still carries legacy Python resource name(s): python_aarch64
hook vs the CLEAN bundle   : (no refusal — correct)
```

**PASS.** The gate and the hook both fail closed, exactly as the PR claims.

#### 4. The launch-critical consequence — is the seed actually *reachable*?

This is the row a directory listing cannot answer, so I did not stop at the listing. `managed-python.ts`'s `seedPath` resolves `packaged ? join(resources, PYTHON_SEED_NAMESPACE, arch) : join(resources, checkoutSeedNames[arch])`, and `backend-installer.ts`'s `findPython()` returns `null` on darwin (`A macOS seed is data, never an executable candidate inside the app`). So the resolution the app performs at first start is the seed path, and the gate never asks whether the bundle's seed is the one it computes.

I bundled the real `src/main/backend/managed-python.ts` (the way `scripts/verify-managed-python.mjs` does) and drove `prepareManagedPython` against the **actual built bundle**, `packaged: true`:

```
seed on disk: …/Contents/Resources/python-runtime-seed/arm64 -> bin/python3 present
seed identity (from the bundle's own seed): d84351e50a7eaf54f9e505ee625fab278657b40f284e96d329b2141d2099dd42
installer received python: …/support/managed-python/packaged/runtimes/d84351e5…-e8c2b9a8…/bin/python3
runtime identity (as provisioned) : d84351e50a7eaf54f9e505ee625fab278657b40f284e96d329b2141d2099dd42
executed that python -I -B -c "import sys, sysconfig, encodings, ssl, sqlite3":
  base_prefix: …/support/managed-python/packaged/runtimes/d84351e5…
  stdlib     : …/support/managed-python/packaged/runtimes/d84351e5…/lib/python3.12
PASS
```

So the bundle's shipped seed **is** the path the app's own provisioning resolves; the copy it provisions is outside the `.app`, is byte-identical by the app's own identity hash, and that interpreter runs with its stdlib outside the bundle. The strict manifest walk, the Mach-O verification, the `ditto` copy and the inode-separation check all ran in full.

Two things worth recording from getting there. First, the negative: given a copy wearing the **pre-fix layout** (seed renamed to `python_aarch64`), the same call refuses rather than falling back — `THREW: ENOENT: … lstat '…/python-runtime-seed/arm64'` — so the seed mapping is load-bearing at launch, not just at the gate. Second, an unguarded stub `-m venv` moved the provisioned copy's identity hash from `d84351e5…` to `dc57f953…` by writing `__pycache__` into the copied stdlib; the real install script's `PYTHONDONTWRITEBYTECODE` is what prevents that, and the app's own readiness probe uses `-I -B`. My harness mirrors both, which is why the identity compares equal.

**PASS.** I did not need to launch the packaged Electron app for this: the provisioning module is the thing that runs at first start, and it was driven directly rather than simulated. The full `scripts/verify-managed-python.mjs` end-to-end (real install script, real venv, `pip install` of the backend and its FFmpeg download) was not run — see BLOCKED-3.

#### 5. Platform collateral

**Linux — PASS.** Built with the trees staged (the strongest case, not the CI case): `electron-builder --linux --x64 --dir` → exit 0, hook logged `Skipping Python resource pruning: not macOS (linux)`, and:

```
dist/linux-unpacked/resources/python/bin/python3          OK (executable)
dist/linux-unpacked/resources/python_aarch64/bin/python3  OK (executable)
dist/linux-unpacked/resources/python-runtime-seed         absent  (no mac seed leaks into Linux)
bytecode under the two trees                              0
```

That placement is exactly what `findPython()` probes on Linux (`resourcesPath/python[_aarch64]/bin/python3`), so moving the block from top-level to `linux.extraResources` **is** honoured by electron-builder rather than silently dropped — which is the one way this change could have broken Linux. On CI the Linux job never runs `pnpm setup-python` (`publish.yml:306` is inside `build-macos` alone), so `resources/python*` does not exist there and those two entries are skipped, leaving Linux output byte-identical to before.

I ran `--dir` rather than `pnpm dist:linux`: `extraResources` placement happens in the same step either way, and the deb/AppImage/rpm wrappers add tooling and disk I did not need to spend on a volume that stayed at 12–15 GiB free all night. The container formats were therefore **not** built here.

**Windows — BLOCKED-1.** No Windows host, and I will not imply coverage. What I could check without one, I did: the `win.extraResources` names are unchanged from base (`python`, `python_aarch64`, matching `findPython()`'s win32 probe), and the hook's platform branch returns early for `win32` (`Skipping Python resource pruning: not macOS (win32)` → `{pruned: [], resourcesDir: null}`) so the mac-only alias refusal cannot fail a Windows build.

#### 6. Regression

`pnpm test:desktop` on the CI test line (Node 22.22.0): **1240 tests, 1236 pass, 0 fail, 4 skipped, exit 0** — same 1240 total as CI, with CI's 27 skips being Linux-only mac skips. Both cases this PR adds ran and passed (`ok 645 - the builder config stages the mac interpreters into the seed namespace`, `ok 646 - the hook refuses a bundle that still carries a legacy alias, by name`). The touched suite alone, `node --test scripts/prune-python-resource.test.mjs`: **12 tests, 12 pass, 0 fail, 0 skipped** against 10 cases at base.

The manifest's tree stamps — the half of `pnpm check-evidence` that `test:desktop` binds — verify directly: `scriptsTree c80ead02… == git rev-parse HEAD:scripts`, `srcTree a77f9f04… == HEAD:src`. All 18 commit citations in the manifest resolve and are reachable, and `head 630cbbe07` is an ancestor of this head by design (the check deliberately does not require it to equal HEAD).

- **N1 (note, not a finding, not this PR's):** `pnpm test:desktop` under the publish job's Node pin `22.13.1` fails 2 cases in `scripts/composer-readings.test.mjs` with `ERR_UNKNOWN_FILE_EXTENSION` for `session-duration.ts` — those cases dynamic-`import()` a `.ts` file, which needs default type stripping (Node ≥ 22.18). CI's Desktop Tests job pins `node-version: '22'`, i.e. the latest 22.x, which strips types; the publish job pins 22.13.1. Nothing in this diff touches it — I reproduced both failures at base and head identically and they vanish on 22.22.0. Recorded, not addressed: out of scope here.
- `submit-latency`'s known-intermittent `M1/M2/M3` case **skipped** in my run — `no local-operator at /tmp/qa-197/home/.local/share/uv/tools/…; set LOP_BACKEND_BIN to measure M1-M3` — because my HOME is isolated from the operator's uv-tool install. I neither confirm nor deny it, and this diff touches no submit path.

#### BLOCKED rows, with reasons

- **BLOCKED-1 — Windows packaging.** No Windows host; no NSIS build was produced. Partially covered as described in §5.
- **BLOCKED-2 — signed and notarized rows.** `app-codesign`, `app-spctl`, `app-stapler`, `dmg-spctl`, `dmg-stapler` (22 of the 41). No Developer ID in this keychain — electron-builder reported `0 valid identities found`. These passed on CI and are not what failed; the local run uses `-c.forceCodeSigning=false`, as instructed. 22 = 6 apps x 3 + 2 images x 2.
- **BLOCKED-3 — `scripts/verify-managed-python.mjs` end to end.** Not run: it performs a real `pip install` of the backend plus an FFmpeg download into a scratch venv, which I judged not worth the several hundred MB on a volume that sat at 12–15 GiB free under a load average of 62 — the seed-reachability question it would answer is answered directly in §4, and the half it adds is the install script, which this diff does not touch.
- **BLOCKED-4 — `pnpm check-evidence` frame sweep.** `Evidence check DEFERRED: another sweep (or its image child) holds the machine lease; retry after it finishes.` — 5 attempts over ~10 minutes. The lock is not mine and I did not touch it. The stamp and citation halves, which are what this diff's manifest change actually affects, were verified independently (§6). No frames were re-captured by this PR.

#### Scratch reclaimed

Own `git worktree` at `/tmp/qa-197/wt` removed, along with its `dist/` (both the 2-arch mac build and `linux-unpacked`), the staged interpreter trees inside it, the ZIP/DMG extracts and app copies used for §3, the scratch runtime trees, and the temp HOME. The reusable `resources/` trees were removed **only inside my own worktree**, twice, as the reproduction requires; the shared checkout's `resources/` is untouched. Nothing global was swept.

damianvtran:
### Sync note — `main` moved again (head `b1a6fdbaf` → `e08a50731`)

`#176` landed on top of `#162`/`#164` and GitHub refused the merge with `Pull Request has merge conflicts`, so the head is synced before the window closes on it. Merged, never rebased.

**One conflict, `docs/evidence/manifest.json` — and this time it was not only the stamp block.** Main's side carried a newer derivation (frames 2507, surfaces 302, 44 declared sets) plus a newer `partialCapture` (the chat-run-panel pass), while this branch's side was derived back on `0.23.2`'s disk. Main's side is taken **whole**, because this file describes the tree it ships in and this branch adds no frame of its own; the derivation is then re-checked on the merged disk rather than assumed — 2956 committed `.webp` minus the 449 inside those 44 sets = **2507**, the capturer's own `STORIES` literal = **302**, `THEMES` = **12**, all three agreeing with the fields carried. Only the stamps moved.

**`package.json` merged on its own, and both directions are checked by key path rather than by eye** — the class of loss that caused this PR in the first place:

- against this branch's own head: **0 key paths added, 0 lost**, 1 changed (the `test:desktop` line, unioned with main's list);
- against `main`: the top-level `build.extraResources` is gone by design, `build.{mac,win,linux}.extraResources` are this branch's, and `build.afterPack` differs on purpose (`scripts/prune-python-resource.mjs` → `scripts/after-pack.mjs`, the hook that refuses a legacy alias). Nothing else differs; the version is main's.
- `test:desktop`: **71 entries, nothing missing from either side, no duplicates.**

**The config is still right after the sync** — the point of this PR, so it is checked against the code that consumes it rather than re-read:

```
config  build.afterPack        : scripts/after-pack.mjs
hook    refuses legacy names   : python, python_aarch64
hook    prunes/keeps (arch->dir): {"arm64":"python-runtime-seed/arm64","x64":"python-runtime-seed/x64"}
gate    expects seed dirs      : python-runtime-seed/arm64, python-runtime-seed/x64
config  mac.extraResources     : resources/python -> python-runtime-seed/x64 | resources/python_aarch64 -> python-runtime-seed/arm64
AGREE   config dests == gate seed dirs: true (as sets)
AGREE   checkout names == layout checkoutSeedNames: true
AGREE   no config dest is a legacy name: true
```

And the check bites, in scratch copies of `package.json` with the repo untouched:

- revert only the registration → `the builder config stages the mac interpreters into the seed namespace` **fails** (11/12);
- revert only the mac mapping to the legacy names → the same case **fails** (11/12);
- with this head's config → **12/12**.

Gates on `e08a50731`: `pnpm lint` 0 · `pnpm check-types` 0 · `pnpm check-themes` 0 (2363 assertions, 12 themes) · `scripts/evidence-manifest.test.mjs` **28/28** · `pnpm test:desktop` locally **1331/1332** (the failure is `scripts/submit-latency.test.mjs`'s *"M1/M2/M3: the warm removes the engage from the send"*, main's file, untouched by this diff, and it passes on the CI arm below) · **CI all green** — `Desktop Tests` on the Linux arm: **1332 tests, 1305 pass, 0 fail, 27 skipped** (macOS-only), including both new cases.

`pnpm check-evidence` is **BLOCKED**, not failing: another session's sweep holds the machine lease (`/tmp/local-operator-ui-check-evidence.lock`, flock held by a live `node` pid 75040 with its `magick` child 33682), so it exits 75 without checking any frame. The lock is left exactly as it is — it is not this branch's to clear, and the gate's own message says not to delete it.

PR #196, opened 2026-09-15
fix(desktop): keep a headless run off the Dock, and tied to its launcher

damianvtran:
## What this is

Two behaviours the `headless` window mode was missing, both of them about an
agent-driven run not accumulating on the operator's machine: a `headless` run
**takes no Dock tile**, and it **leaves when the process that launched it does**.
A third follows from the first review round: **every way a headless run can end
is bounded**, so a run cannot be left alive with no window and no watcher.

Release: patch.

**Head note — synced onto `main` at `cd223e55d`.** The PR had gone `CONFLICTING` while its base aged ~311 commits. It is now merged up to `main`'s `461a7f11b` (the 0.24.6 window: #219's manual-release switch and #224's bump), which resolved five conflict hunks across four files and re-derived the evidence stamps; `### Agent review remediation — round 4` below is the record. The sync also folded one line of main's own rule onto this branch's code: the launcher watch's opt-out now reads `launchEnv` rather than `process.env`, so a `.env` in the checkout cannot make a run outlive its harness (round 4, R8). The boots in the evidence table below were taken before the sync, on the heads the rounds name.

## Why: the accumulation, measured

`headless` is the mode every harness, QA matrix and evidence run uses. The window
is created and never shown, and on macOS `window-all-closed` deliberately keeps a
windowless app alive — correct for the shipped app, fatal for a run nobody is
using. So an app whose launcher went away had nothing that could ever end it.

Measured on this machine:

- a QA round's **13 boots left 13 survivors**, every one reparented to launchd
  (`ppid 1`), one still answering on its debug port;
- the operator's **Dock held ~30 running apps**, all of them `headless` boots
  from evidence sessions — the symptom that started this;
- reproduced on demand against `main`'s own build:

  | action | app |
  | --- | --- |
  | launcher `SIGKILL` (the pnpm shim → `node cli.js`) | **alive 40 s later**, `ppid 1` |
  | launcher `SIGTERM` (what the driver's `stopApp()` sends) | **alive ≥ 20 s**, window closed, process kept by macOS |
  | `SIGTERM` to the app itself | **alive at 45 s and 70 s** (round 1, R1: an app-initiated quit emits no `window-all-closed`, so the quit wedged and never completed) |

  The harness half of the mechanism is `stopApp()` signalling `child` — the
  `node` process the shim `exec`s — rather than the app. That is
  `scripts/renderer-driver.mjs`, i.e. #190, and is not touched here.

## What changes

[...]

PR #191, opened 2026-09-15
feat(browser): keep session-only cookies across a restart, partition identity intact

damianvtran:
## What this is

The browser partition keeps persistent cookies, `localStorage` and IndexedDB
across a restart, but **session-only cookies die with the process** — and a
session cookie is the shape most logins issue. So an agent that logs into a site
loses that login whenever the app restarts. This PR stores those cookies on a
clean quit and restores them on the next start, **without losing the CHIPS
partition identity that makes a partitioned cookie what it is**.

**Review status:** agent review round 1 (one MAJOR, five MINORs, three NITs — the
two security guarantees it attacked both hold) and its remediation are posted, and
round 2 (`264ec8d7b..f5c308b5f`) closed every round-1 finding, raised one new MAJOR
(`R2-1`: a second quit during the hold exited with the snapshot still running and
made the next start report a crash that never happened) and two cosmetic items
(`R2-2`, body drift; `N2`, lint-baseline attribution — both answered in this body).
`R2-1` is fixed in `d44610539` with its own test. Independent QA passed round 1
(PASS, one non-blocking latency finding, `Q1`) and round 2 on the delta
`264ec8d7b..f5c308b5f` (PASS, five findings, none introduced by this PR's commits);
its two comment corrections are landed in `c8aeae5a2` — `Q4`, the 1500 ms bound now
states the phase it actually bounds, and `Q5`, the expiry is a release rather than
an exit. The current head `7b02ecc77` is this branch MERGED with `origin/main`
`124fdafc6` (a merge, not a rebase, so every commit the rounds cite stays
reachable), the comment corrections above that merge, and the manifest's tree
stamps re-derived to the head (`7b02ecc77`). The convergence review is scoped
`f5c308b5f..7b02ecc77`; QA's pass is re-anchored to the same head. `pnpm
check-evidence` is DEFERRED on this machine (sweep lease held elsewhere) and is not
claimed green anywhere in this PR.

| in this PR | not here |
|---|---|
| session-only cookie persistence in the main process, profile-scoped | any renderer, IPC, RPC or settings surface — nothing user-visible at all |
| CHIPS partition identity carried through both directions, refused when it cannot be | any change to the persistent cookies Chromium already keeps |
| durable dirty-marker crash rule, serialised snapshot/restore/clear | any Python-side change; no `local-operator` change of any kind |
| fail-closed keychain policy, ciphertext-only snapshot at 0600 | a plaintext fallback of any kind, by design |
[...]

PR #177, opened 2026-09-14
feat(browser): the visible browser tab — chrome, tab strip, consent and restore

damianvtran:
## What this is

PR 4 of `docs/design/ui-browser-tab.md`: **the visible half** — the browser surface in the
app shell, with its tab strip, URL bar, navigation controls, tab restore and per-origin
consent bar. It drives the main-process host that PR 3 landed; it does not reimplement it.

**Was stacked on `feat/browser-tab` (PR #160); #160 has since merged into `main`, so this branch
has been rebased onto `main` and the PR's base is `main`.** The rebase was mechanical: the only two
conflicts were the `test:desktop` file list in `package.json` and one comment block in
`approvals.ts` (main's `fix(browser): honour a stored broad grant in the origin gate` fixed the same
mis-slotted argument this PR's chrome work found, so main's wording is kept and mine dropped). Every
gate below was re-run on the rebased head, which is also where the frames come from.

| in this PR | not here |
|---|---|
| the `/browser` route, reachable from the rail, lazy so it can slot into the panel system (§11.9) | the panel system another session is landing |
| the tab strip: both owners, agent marker, close/new, title + favicon state, active marker | tab tear-off, tab groups, drag-reorder |
| the URL bar: live URL, editing/loading/error states, back/forward/reload/stop | bookmarks, history UI, omnibox search |
| the layout contract: reported content rect → `view.setBounds`, hide on a null rect | any change to the host's own bounds logic |
| the overlay policy: full-window overlays hide the native view, and restore it | a clipping approach (a native view cannot be composited under DOM) |
| the consent bar: `once \| session \| site \| domain \| deny`, requester binding, notification, revocation | a second consent implementation — this reuses the host's store and the vendored `access-flow`/`access-queue`/`origin-policy` |
| tab restore across restarts, with the stale-handle rule (§7.3) | restoring form state, POST bodies, scroll or a live process (documented in the file) |
| hand-over: user tab → a session, one optional field on the existing `open` tool | any change to the tool's schema beyond that field |

**No version bump**, and the branch carries none: `package.json` stays at main's released `0.23.2`, and
the only `package.json` change is the union of main's `test:desktop` list with this branch's
`browser-chrome.test.mjs`.

**Head and history.** The head is a MERGE of `main`, not a rebase: the review rounds cite
`1a54ac164`, `07ba4aa59`, `0cda1f278` and `eadeadf8a`, and replaying this branch onto a new base would
[...]

PR #173, opened 2026-09-14
feat(chat): move a live session's working directory from the composer

damianvtran:
# PR Description

## Summary of Changes

The composer's working-directory chip becomes editable on a **live** session when the backend
advertises the move contract, and choosing a directory moves the session - the desktop's `/move`,
with the same semantics as the TUI's.

**Head `df3438ecd`, on `origin/main` `36570bfad` - no fifth rebase was needed, the branch merges
conflict-free - with the round-2 pair (`c8820709c` code, `9216ab40a` evidence), the round-3 record
commit (`6f1621a7e`) and the round-4 coverage commit (`df3438ecd`) in it.** Round 1 changed two things shape, and both are load-bearing:

1. **`/move` no longer mounts a picker.** The old `MovePicker` was a modal that HOSTED the
   composer's own chip, and inside its modal the chip's menu opened 620px tall at `y=-192` with
   `overflow-y: hidden` - two of its three choosing affordances were out of pointer reach and the
   keyboard-focused row was off-screen (UX round 1, U2). The bare command now focuses the chip the
   user already has and opens its menu in place; the chip's menu is height-bounded and scrollable;
   and the typed `/move <path>` form shares that one controller rather than running a second
   transport path. One control, one write path, no popper inside a dialog. `MovePicker` and
   `move-picker.stories.tsx` are gone, so `destination-pickers.tsx` and that story file are
   **net-unchanged against `main`** and do not appear in the diffstat below.
2. **The move's directory now reaches a mounted viewer.** A move rewrites the facade's cwd without
   moving the owner's clock, so the delta that used to carry it arrived at the owner's unchanged
   epoch and sequence and the shipped reducer dropped it as stale - the measured defect behind QA's
   round-1 Q1 (the chip reverted to the pre-move directory above a receipt saying it had moved) and
   the backend's R4. The bridge now publishes an explicit **`frontend.replace`** frame, ordered by
   its own cursor, and this renderer consumes it as the authoritative projection while preserving
   the transcript, the history and the live phase. `acceptFrontendReplace` states the acceptance
   rule; the `frontend.update` rule beside it is deliberately untouched, so the next real owner
   delta still applies and an ordinary stale delta is still rejected.

The chip's read-only tooltip used to say the directory "is set when the session starts and cannot
be changed afterwards". That sentence became false the moment a backend with the route existed, so
it is gone - from the composer **and from `canvas/create-file-dialog.tsx`**, which was the last
[...]

PR #198, opened 2026-09-15
chore(release): bump version to 0.23.3

damianvtran:
One-line release bump, `package.json` only: `0.23.2` → `0.23.3`. C0 scope.

## Why a new number rather than a re-run

`v0.23.2` was tagged at 06:15Z and its publish failed at the artifact gate: on the signed runner the packaged macOS app carried the legacy `Contents/Resources/python` names and no private seed, so `app-one-bundled-python` and `app-private-python-seed` failed for both architectures and both containers — 12 of 41 checks — and the workflow held it out of `latest` with zero assets. It is a pre-release with no assets and `latest` still points at `v0.23.1`, so no updater moved.

It is not repairable by the dispatch path: every build job checks out the tag's own SHA and `scripts/validate-release.mjs` refuses a dispatch whose `expected_source_sha` differs from it, so a re-run rebuilds the same failures. The tag is deliberately not moved — this repo's release tooling pins immutability — so the repaired release takes the next number.

The cause was not the tag: `build.extraResources` and `build.afterPack` had been clobbered on `main` by a merge resolution during a sync, so `main` had been packaging legacy interpreter names ever since, hidden locally by a stale `dist/`. **#197** repairs it and is in this window.

## Window

Baseline is the last tag, `v0.23.2` (a pre-release that published nothing), re-derived with `git log --first-parent --oneline v0.23.2..origin/main` immediately before this PR and again at tag time:

- **#164** `feat/session-code-memory` — merge `c5b5b67edd6816dc2cf92669209a9df619d07e83`
- **#162** `feat/desktop-notification-feed` — merge `f430e4ccc`
- **#176** `feat/mcp-auth-experience` — merge `124fdafc691afe8b7636212dbd73e5922b91c9b8`
- **#197** `fix/build-seed-layout` — merge `7dd274026b50669901b07f155769bd96fc49e7e4`

Because `v0.23.2` published nothing, this release also carries what was merged for it and never reached anyone: **#183** (`60419179ef`, the update fix) and **#195** (`d600da470`, the installer architecture fix). The notes describe the change from `v0.23.1`, which is what users actually have.

## Bump class

**Patch.** #164 restores a panel that could not load variables at all; #162 is capability-gated and inert until its backend half lands; #197 is a build fix; #176 makes an intended remedy reachable (`/mcp` landing on an empty Settings page is a defect) rather than introducing a surface class. #176's author independently conceded this after arguing minor — the concession is on the record in the window thread. Nothing here clears the step-function bar.

## Ownership

Claimed first and announced to six peers; a later claimant (`pid 14057`) yielded on the same grounds above, and supplied #176's `Release:` line, which the notes quote verbatim. One bump, one tag.


damianvtran:
### Agent review - round 1

**Reviewer:** `reviewer` (Local Operator `lopdev` subagent) — model `deepseek/deepseek-flash`, which is the model actually running this pass; no provider fallback observed during the round.
**Scope:** `7dd274026b50669901b07f155769bd96fc49e7e4` (`origin/main`) .. `13f159af9215ee45941f729d990c1815ec18983c` (`chore/release-0233` head) — `package.json`, 1 file, +1/-1.
**Verdict: clean.** No BLOCKER, no MAJOR; two NITs below, neither a gate.

#### 1. The diff is what it claims — yes, and by whole-file comparison, not by reading the hunk

`git diff --stat origin/main..13f159af9` is `package.json | 2 +-` and `git diff --name-only` names exactly that one file. Whole-file, byte-level: both sides are `394` lines / `16727` bytes and `diff` reports exactly one differing line, `3c3` — `"version": "0.23.2"` → `"version": "0.23.3"`. There is no re-serialisation (byte count identical), no dropped key, no added key, and no duplicate key: both files parse under an `object_pairs_hook` that flags a repeated key at any nesting depth, and neither file produced one. The single commit on the branch is `13f159af9 chore(release): bump version to 0.23.3` touching `package.json` alone. Given this window's #197 was itself a recovery from two silent `package.json` merge clobbers, this is the check that mattered, and it is clean.

#### 2. Window derivation — correct, and complete against `v0.23.2` and against `v0.23.1`

`git log --first-parent --oneline v0.23.2..origin/main`, re-run here:

```
7dd274026 Merge pull request #197 from damianvtran/fix/build-seed-layout
124fdafc6 Merge pull request #176 from damianvtran/feat/mcp-auth-experience
f430e4ccc Merge pull request #162 from damianvtran/feat/desktop-notification-feed
c5b5b67ed Merge pull request #164 from damianvtran/feat/session-code-memory
```

That is the PR body's list, in order, and every SHA it quotes resolves to the commit it names (`c5b5b67ed…`, `f430e4ccc…`, `124fdafc6…`, `7dd274026…` all verified with `git rev-parse`). The first-parent walk is merges only — nothing was pushed straight to `main` that the body would leave out. The carried half also checks: `v0.23.1..v0.23.2` is exactly `#194` (the bump, `32e3d04b4`), `#195` (`d600da470`) and `#183` (`60419179e`), i.e. the body's "because `v0.23.2` published nothing, this release also carries #183 and #195" omits nothing from the unpublished range, and `#194` is the version consumption itself. The notes' stated baseline — describe the change from `v0.23.1`, which is what users actually have — follows.

#### 3. No competing bump, and no way to land on top of a different version

All nine other open PRs are version-clean: `git diff origin/main...origin/<head> -- package.json | grep '^[+-][ \t]*"version"'` is empty for **#196, #192, #191, #190, #184, #177, #173, #169, #143** — none of them touches the version line, so none carries a bump and none can collide with this one. The only ref in the repo whose version differs from `main`'s `0.23.2` is this branch (`0.23.3`); the other differing refs are stale feature/evidence branches carrying *older* numbers (`0.22.3`, `0.22.1`, `0.20.0`, …) whose diffs also contain no version-line change (#192 resolves to `0.22.3` at its merge-base `915928a18`, so it introduces no version edit either). The branch is not behind: `git rev-list --left-right --count origin/main...origin/chore/release-0233` is `0  1` — it contains all of `main`, so the merge lands `0.23.2` → `0.23.3` and nothing else. `main` today is `7dd274026` with `package.json` at `0.23.2`, so the bump has one and only one predecessor to stack on.

#### 4. The repo's guards — satisfied, read at head `13f159af9`

CI is present at the head (`13f159af9215ee45941f729d990c1815ec18983c`), so this is not the DIRTY-branch trap: `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`, and all eight checks **completed/success** — `Version Bump Guard`, `Lint`, `Type Checking`, `Desktop Tests`, `Runtime Dependencies`, `Security Audit`, `NPX Sanity Check (macos-latest)`, `NPX Sanity Check (ubuntu-latest)`. (First read had `Lint`/`Type Checking`/`Desktop Tests`/`Security Audit`/NPX still `in_progress` under `UNSTABLE`; the re-read after they settled is the one quoted.)

- **`version-bump-guard.yml`** — green, and green for the right reason. Its own log says `Release PR, version change allowed:` followed by exactly the one-line pair `-  "version": "0.23.2",` / `+  "version": "0.23.3",` and `Version 0.23.3 sorts above the highest release tag v0.23.2.` The title prefix `chore(release): bump version to ` is the exempt side of the guard, and the guard's mirror-image failure (a release PR that changes *nothing*) is not the case here: the diff does touch the version line.
- **`scripts/validate-release.mjs`** — not run on a PR, and will run at publish, where it is satisfiable: it requires the tag to resolve, the Release to be published, `EXPECTED_SOURCE_SHA` to equal the tag's SHA on a manual dispatch (lines 115–119), and — the pin that matters here — `package.json` **read at the tag's commit** to carry `name === "local-operator-ui"` and `version === tag.slice(1)` (lines 121–136). So the tag must be cut from a commit whose `package.json` says `0.23.3`: the merge/bump SHA on `main` after this lands. Tagging the *current* `main` head (`7dd274026`, version `0.23.2`) would fail that check by name, which is the only way this one-liner can be got wrong once merged.
- **The body's "not repairable by dispatch" claim is accurate**, checked at source rather than taken: `validate-release.mjs:117` fails when `tagSha !== expectedSha`, and every build job checks out `ref: ${{ needs.validate-release.outputs.source_sha }}` (`publish.yml` lines 260, 385, 455), i.e. the tag's own tree — so a dispatch against `v0.23.2` rebuilds the same failing tree, and a dispatch naming a different SHA is refused. Moving the tag is correctly ruled out rather than attempted.

#### 5. The artifact claims the failed `v0.23.2` run contradicts — there are none; the run corroborates them

I pulled run **34936130402** (`release` event, `headSha 32e3d04b4343685bff23ba1724a3156f487e31ee`, i.e. the `v0.23.2` tag) rather than trusting the summary, and each artifact claim in the body holds:

- `Build macOS App` > `Verify macOS artifacts` fails with exactly the two rows the body names, `app-one-bundled-python` (*"the x86_64/arm64 app ships Contents/Resources/none, but it resolves Contents/Resources/python-runtime-seed/x64|arm64"*) and `app-private-python-seed` (*"Legacy Python resource or alias exists: python"*), repeated for both architectures and for the app inside both containers, and the step prints **`12 of 41 artifact checks failed`** — the body's "12 of 41" is the gate's own number. `app-no-bundled-bytecode` PASSes, as the body implies by naming only the two.
- `Hold release out of latest` **succeeded** and `Promote release to latest` + `Attach Artifacts to Release` are **skipped**, which is why the Release is `isPrerelease: true` with `assets: []` and `gh api /releases/latest` still reports **`v0.23.1`** with its 20 assets. "A pre-release with no assets, `latest` still points at `v0.23.1`, no updater moved" — verified, not restated.
- The causal half is right too, and this is where I looked hardest because it is the claim that licenses the release: the gate's expectation of `python-runtime-seed/<arch>` comes from `#183`'s `445d08124 build(mac): ship the interpreter as inert data, and refuse legacy aliases`, which landed inside the unpublished `v0.23.1..v0.23.2` range — so on the `v0.23.2` tag the gate demanded the seed namespace while `build` still declared the legacy top-level `extraResources` (`resources/python` → `python`) and `afterPack: scripts/prune-python-resource.mjs`, exactly the clobbered state the body describes. `origin/main` now carries the corrected shape (`build.afterPack: scripts/after-pack.mjs`; `build.mac.extraResources` → `python-runtime-seed/{x64,arm64}`; no top-level `extraResources`), i.e. the tree the tag will be cut from packages the names the gate reads, and `#197`'s own QA records a clean build passing all three seeded rows ×6 on that gate. I am not claiming the next publish *will* be green — that is the build's to prove — but nothing in the body overstates what the failed run showed, and the fix the release depends on is in the tree being tagged.

#### NITs (follow-ups, not gates)

- **N1 — tag target.** When the Release is cut, target the merge/bump SHA on `main` (the commit whose `package.json` is `0.23.3`); `v0.23.2` was tagged at its bump merge `32e3d04b4`, and the same shape here keeps `validate-release.mjs`'s version pin satisfied by construction.
- **N2 — one provenance sentence is not checkable from the PR.** The body's "`#176`'s author independently conceded this after arguing minor — the concession is on the record in the window thread" cites a peer session's thread; it is plausible and immaterial to the bump class either way (patch stands on the diff), so it needs no action — noted only so it is not mistaken for evidence in this review. Separately, `reviewRequests` already lists `bbqben` before this round; awareness-only, no comment tag, which is the right shape for C0.

#### Verdict

**Safe to merge as a bare version bump, and safe to tag from — provided the tag targets the `0.23.3` merge/bump commit on `main`.** Round 1 is TERMINAL: no BLOCKER, no MAJOR, both NITs are follow-ups.

PR #206, opened 2026-09-15
fix(desktop): hand the suite's children the notification kill switch

damianvtran:
# PR Description

## Summary of Changes

`pnpm test:desktop` posted roughly **46 notifications in six minutes** into the
operator's real macOS Notification Center while he was working on something
else, and they buried what he needed to see. They arrived as **Script Editor**
(`com.apple.ScriptEditor2`), which names the mechanism: nothing in this app posts
through Script Editor, `osascript -e 'display notification'` does, because the
osascript process carries no bundle identity. A Notification Center is not a
sandbox, and a test suite that writes to one is a defect in the suite.

The path is not visible from this repository, which is why it survived review:

| step | where |
| --- | --- |
| the suite's children get the switch | `scripts/run-desktop-tests.mjs`, the runner behind `pnpm test:desktop`. **No file in the suite spawns Electron** (measured over the whole `test:desktop` list); the suite's live backend is a python `serve` started directly by `scripts/submit-latency.test.mjs`, which keeps the runner's environment |
| the app-proof rigs boot the real app | each of the five runs with `VITE_DISABLE_BACKEND_MANAGER=true`, so the app they boot spawns no backend |
| the app spawns a real backend | `src/main/backend/backend-service.ts::backendSpawnEnv` spreads the app's own environment into that child |
| a session parks on a gate and announces it | `session/runtime/serving.py::_announce_pending` |
| and that ends at the OS | `local_operator/tui/notify.py::detached_notify` → `osascript -e 'display notification ...'` |

Every step was already correct. What was missing is that the switch that path is
gated by — `LOCAL_OPERATOR_NO_NOTIFICATIONS`, `notify.py`'s `_ENV_DISABLE`, read
by `notifications_enabled()` — was set by **nothing** in this repository. Every
run bannered by default, and the safety of the operator's desktop rested on each
agent and human remembering an incantation. That is how it happened at 3 a.m. to
somebody who had never heard of the variable.

## The change

| file | what |
| --- | --- |
| `scripts/notifications-off.mjs` | **new**: the kill switch applied to a child environment — `NOTIFICATIONS_ENV` plus `withNotificationsOff(env)`. Docstring carries the whole path, the deliberate exceptions, and why the upstream fix is the wrong one |
| `scripts/run-desktop-tests.mjs` | the runner behind `pnpm test:desktop` sets it on the environment its children get, so the **whole suite** — every test file, and through them every app and backend they spawn — is silent by default |
[...]

PR #202, opened 2026-09-15
chore(release): bump version to 0.24.0

damianvtran:
## Release window: v0.24.0

One-file version bump for the window that follows `v0.23.5`. `package.json` only — no code, no dependencies, no lockfile, no changelog.

### Window contents

`git log --first-parent v0.23.5..origin/main` is **exactly two** merges, and nothing else:

| PR | Merge SHA | Impact | Bump argued |
| --- | --- | --- | --- |
| #177 `feat(browser): the visible browser tab — chrome, tab strip, consent and restore` | `4c0a7b1e15f4d4dbf998a1d0a693c0c52e742b7f` | The browser tab lands — chrome, tab strip, URL bar, consent bar, restore — so a desktop-app user gains a browser surface inside the app for the first time. | minor |
| #200 `fix(mcp): a refused MCP remedy, in the user's words, with a route to the fix` | `e9be6a89e77e84ef33f4bd1d44fd8618f867111e` | A refused MCP control stops answering with a sentence written for whoever maintains the backend and stops leaving the operator with only the button that just failed: the refusal is stated in the user's terms and carries a route to the fix. | patch |

`e9be6a89e77e84ef33f4bd1d44fd8618f867111e` — #200's merge — is `origin/main`'s tip at the time this was opened, so the table is derived from the commits rather than from memory: `--first-parent` is safe for counting a window because both merge-button and squash-button landings appear on that chain.

**`Release:` lines: neither PR carries one.** #177's 1,239-line body has no `Release:` line and no comment on it does either. #200's 323-line body likewise — the word "release" appears there four times, all incidental ("Pre-Release CI Testing", a test name, a quoted sentence, and "No version bump: `package.json` stays at the released version"). Nothing is quoted above because there is nothing to quote; both impact sentences are the authors' own descriptions of their changes, not lines the PRs handed over. Recorded here so the absence reads as checked rather than overlooked, since the notes are meant to be written from those lines.

### Bump class

[...]

PR #210, opened 2026-09-15
fix(release): resolve every release script's own path, and refuse rather than print nothing

damianvtran:
# PR Description

## Summary of Changes

This closes two findings recorded on #204 after it merged, plus the two nits from
the same round. The substantive one is the defect #204 fixed in
`release-push-guard.mjs` alone: **eight sibling scripts in the same workflow steps
still resolved their own path lexically**, so through a symlinked path they loaded,
printed **nothing** and exited **0** — and in this repository every consumer reads
an exit status as "it ran". `derive-release.mjs`, the *next command in that same
`run:` block*, was one of the eight.

Reproduced on an untouched tree at `64c3283cb`, with the derive step's real `run:`
block executed in a throwaway repository whose `scripts/` is reached through a
symlink (the loop guard stubbed, the derivation real):

```
### BEFORE — origin/main (64c3283cb), symlinked scripts/
step exit status:    0
derivation.json:            0 bytes
GITHUB_OUTPUT:       release=false
step summary:
  | ### No release
  |
  |                       <- the reason, empty
  | A release needs a breaking change, a `feat:`, or a `fix:`/`perf:`/`revert:`.

### AFTER — this branch (009c91533), same invocation
step exit status:    0
derivation.json:         1698 bytes
GITHUB_OUTPUT:       release=true version=0.25.0 bump=minor base=v0.24.0
step summary:
  | ### Derivation: minor -> v0.25.0
  |
  | Window: `v0.24.0..c6760b2`; Strongest conventional type in the window is minor
```

A push carrying content was silently skipped, reported as "nothing to release",
behind a green run — the failure this whole change exists to remove. It fails safe
(no wrong version, no Release) and `actions/checkout`'s real paths cannot reach it,
which is why both rounds recorded it as a follow-up rather than holding the merge.

Release: none — no user-visible change in this PR; it repairs the release
automation's own scripts.

### 1. Entry-point detection is one implementation, and all nine scripts use it

New: **`scripts/entry-point.mjs`**. `isEntryPoint(import.meta.url)` resolves **both**
sides with `realpathSync` before comparing them, so the answer does not depend on
how the caller spelled the path (and it holds under `--preserve-symlinks-main`
too, where `import.meta.url` itself keeps the caller's spelling). `physical()`
falls back to the given spelling when a path cannot be resolved — a path that does
not exist is not this module anyway, and throwing there would replace a readable
answer with a stack trace.

The guard is refactored onto it rather than keeping its own copy: one
[...]

PR #211, opened 2026-09-15
test(session-cookies): stop asserting a boundary the clock cannot honour

damianvtran:
# PR Description

## Summary of Changes

- Corrects two assertions in `scripts/session-cookies.test.mjs` that compared a
  `Date.now()` difference against an exact `setTimeout` boundary, and therefore failed
  one millisecond short of that boundary on ordinary timer jitter.
- Re-derives `docs/evidence/manifest.json`'s `scriptsTree`, which moves because a file
  under `scripts/` changed (see *Why the manifest moves with this*).

This has been firing intermittently on `main` since the budget was introduced and been
**absorbed as a re-run** rather than diagnosed, so a released assertion has been costing
sessions one retry each time it lands. It was root-caused by running the control on
unmodified `main` instead of assuming flake; `main`'s own CI carries the identical
failure.

Release: none — a test assertion, no user-visible surface; under the workflow `main` runs,
the `test:`/`docs:` commit types cut no release and the guard says so.

## Related Issue(s)

- Related: #206 — found while diagnosing that PR's `Desktop Tests`, which is red partly
  for this reason. Deliberately kept out of #206: it is an unrelated file, and `main`
  carries the identical defect, so it is reachable by no change in that PR.
- Related: #191 — the cookie work that introduced the assertion, and whose `1-of-1528`
  failure on `main` was measured and dismissed as a re-run.

## The defect, and the control that separates it from a regression

`assert.ok(elapsed >= budgetMs)` reads:

```js
const started = Date.now();
const held = await holder(...);      // released by a setTimeout(budgetMs)
const elapsed = Date.now() - started;
assert.ok(elapsed >= budgetMs, ...);
```

Both ends of `elapsed` are `Date.now()` readings truncated to whole milliseconds, and a
`setTimeout(budgetMs)` legitimately fires a fraction early, so the timer lands one
millisecond short of the nominal boundary and the assertion fails on the clock's
granularity rather than on the behaviour.

Observed, both sites, in released code:

| Where | Budget | Reported | Source |
| --- | --- | --- | --- |
| `session-cookies.test.mjs:1398` | 600 ms | `the budget is what released the second quit, not luck: 599 ms` | `main` CI run [34962627923](https://github.com/damianvtran/local-operator-ui/actions/runs/34962627923), PR #202, 11:20 today |
| `session-cookies.test.mjs:1267` | 120 ms | `the budget is what released the quit, not luck: 119 ms` | reproduced locally, worktree of `origin/main`'s descendant |

That CI run is the control, and it is the strongest form available: it is `main`'s own
check, on the identical test at the identical line with the identical `599 ms`, 45 minutes
before this branch existed.

## What changed

**Both lower bounds take the one millisecond the truncation can lose.** Nothing else moves
in the assertion direction: both were already two-sided.

**The 600 ms site's upper bound is deliberately left at `budgetMs + 300`.** That is the
half that carries the test's claim — what separates a shared deadline (~600 ms) from a
per-quit budget (~1000 ms) — so widening it to `budgetMs + 400` would have weakened the
very test being repaired.

**The 120 ms site's upper bound is tightened from a literal `2_000` to `budgetMs + 400`.**
At 2000 ms it would have passed a regression that released the quit 1.5 s late, with the
budget it exists to pin long gone. The margin is measured, not guessed: over 15 runs at a
load average of 113–150 the timer's overshoot was **1–15 ms**, so the allowance sits about
26× above the jitter this suite actually sees, and the new bound still clears the CDP
layer's 15 s per-call ceiling that is the real alternative here.

## The rest of the file was audited

Every wall-clock comparison in `scripts/session-cookies.test.mjs`:

- `:1266` and `:1398` — the two defects above, both fixed.
- `:1547` — `Date.now() - started < 5_000` against a `jarDeadlineMs: 50`. Same integer-
  boundary shape, but a false failure needs a 100× overshoot, which is a real signal
  rather than jitter. Left alone.
- `budgetMs: 2_000` at `:1323` is a hold budget read by assertions on return values and
  call order, not on elapsed time. Not this class.

So it is two bad assertions rather than a habit in the file.

## Why the manifest moves with this

`test:desktop` binds `docs/evidence/manifest.json`'s `srcTree`/`scriptsTree` against the
tree under review, and `session-cookies.test.mjs` lives under `scripts/`. Editing it moves
`HEAD:scripts`, so the shipped manifest certified the committed frames against a tree they
no longer ship in — the same gate, and the same one-line correction, that `main` already
carries for other `scripts/` changes (`d40d98b34`, `b0f73d2ba`).

Only that stamp moved; `srcTree`, `surfaces`, `themes`, `head` and `capturedAt` are
unchanged and the re-derivation is recorded in `headNote`. **No frame was re-taken and no
sweep was re-run** — this branch moves a test assertion and renders nothing.

## Impact

- No breaking changes. No dependency updates. No performance or security implications.
- No user-visible surface: the change is confined to a test assertion and the evidence
  manifest's provenance stamp.
- `package.json`'s `version` is untouched.

## Testing Details

**Before** — on `origin/main`'s tree, the assertion fails at the boundary (`main`'s own CI,
run 34962627923):

```
not ok 997 - a second quit is bounded by the first hold's deadline, not by one of its own
  error: 'the budget is what released the second quit, not luck: 599 ms'
  location: 'scripts/session-cookies.test.mjs:1374:1'
  at TestContext.<anonymous> (scripts/session-cookies.test.mjs:1397:9)
```

**After** — the touched file, on this branch, and the shipped-manifest test that this
branch's `scripts/` edit would otherwise have broken:

```
# the touched file, 8 consecutive runs at load averages 84-150
ℹ tests 38   ℹ pass 38   ℹ fail 0        (x8)

# node --test --test-concurrency=1 scripts/evidence-manifest.test.mjs
✔ the SHIPPED manifest's stamps describe the tree it ships in (423ms)
✔ the SHIPPED manifest's head citations lie in the history it ships in (145ms)
ℹ tests 28   ℹ pass 28   ℹ fail 0
```

**The measurement behind the tightened bound** — 15 instrumented runs, `budgetMs` 120 and
600, load average 113–150:

```
budget120 elapsed: 121 121 122 122 123 124 124 127 127 130 131 132 133 134 135
deadline600 elapsed: 601 601 602 602 603 603 603 603 606 607 608 609 610 611 616
```

i.e. the timer lands between one millisecond short (the observed failures above) and
fifteen milliseconds late, which is what fixes the allowance.

**Gates:** `pnpm check-types` clean; `pnpm lint` does not cover `scripts/*.test.mjs` except
`linux-sandbox`, and `biome check` over the touched file reports the same 1 error / 27
warnings as `origin/main`'s copy — this change adds no diagnostic. `pnpm test:desktop` over
all 88 files also exercised the change end to end (run log retained).

Two unrelated failures appeared in that full-suite run and neither is this change: the
`evidence-run-guard` CLI lease test fails while **another session holds the machine-wide
`/tmp/local-operator-ui-check-evidence.lock`** (it passes 6/6 in isolation, and CI is a
dedicated machine), and `submit-latency`'s `M1/M2/M3` is the known load-sensitive race —
it fails identically on unmodified `main`. `pnpm check-evidence` itself is **deferred** on
this machine while that lease is held and must not be read as passed.

## Checklist

- [x] I have added tests that prove my changes work as intended. (The two assertions *are*
      the tests; their correctness is what changed, and both are shown passing above.)
- [ ] The gates pass: `pnpm lint`, `pnpm check-types`, `pnpm check-themes`, `pnpm test:desktop`.
- [x] I have updated the documentation when required. (`headNote` records the re-derivation.)
- [x] Security considerations are addressed — none apply to a test assertion.

## Screenshots (Optional)

Not applicable — no rendered surface is touched.


damianvtran:
### Agent review — round 1

**Reviewer:** local-operator reviewer subagent, model `deepseek/deepseek-flash` (no provider fallback occurred in this round).
**Scope:** `64c3283cb..0737d6964` — the full diff (`scripts/session-cookies.test.mjs`, `docs/evidence/manifest.json`), read against `main` at the merge base `64c3283cb`.

**Verdict: clean — no BLOCKER, no MAJOR. The round is TERMINAL.** Two non-blocking findings (M1, N1) are recorded as follow-ups, not as another round. One **pre-merge action** is required and is unrelated to this diff: the branch has gone `DIRTY` against `main` since the review was requested (last section).

---

#### What I executed, not reasoned

Isolated shared-object clone of this branch at `/tmp/rev211git`, head `0737d6964`, `LOCAL_OPERATOR_NO_NOTIFICATIONS=1`, nothing pointed at the operator's profile, no app spawned:

```
$ node --test scripts/session-cookies.test.mjs        # x8, exit 0 every time
  run 1: exit=0 loadavg=87.1     run 5: exit=0 loadavg=96.8
  run 2: exit=0 loadavg=91.8     run 6: exit=0 loadavg=98.6
  run 3: exit=0 loadavg=91.5     run 7: exit=0 loadavg=102.2
  run 4: exit=0 loadavg=92.2     run 8: exit=0 loadavg=102.4
  ℹ tests 38  ℹ pass 38  ℹ fail 0
  ✔ a stop that never settles is released at the budget instead of holding the quit
  ✔ a second quit is bounded by the first hold's deadline, not by one of its own
```

The load average is stated because it is what makes the defect visible: this machine sat at **87–102** throughout, which is the band the flake needs, and the two touched tests still came back green on every run.

```
  ✔ the SHIPPED manifest's stamps describe the tree it ships in (3159.724458ms)
  ✔ the SHIPPED manifest's head citations lie in the history it ships in (347.465083ms)
  ℹ tests 28  ℹ pass 28  ℹ fail 0     exit=0
```

The pre-fix control, since I was asked to run this file myself at the load that made the defect visible: **120 consecutive runs of the two touched tests with `main`'s assertions restored failed 0 times.** That is not a refutation of the defect, it is the defect being rare by construction — measured at 1 event in 400 timers at the 600 ms boundary, so ~26% is the chance of seeing one in 120 runs. I am not claiming a local reproduction, and I did not get one.

Method note, because this one is easy to report vacuously: the stamp half stands down in a tree with no `.git`. I ran the same file in a `git archive` copy (no `.git`): `✔ the SHIPPED manifest's stamps describe the tree it ships in (1421ms)` while the citation test fails. So the green above is the real one — a genuine repository, checking `HEAD:scripts` against the stamp it ships.

#### 1. The floor — it covers the instrument, not a slow timer

Both sites now read `elapsed >= budgetMs - 1` (`scripts/session-cookies.test.mjs:1275` and `:1425`), and the sibling boundary was treated identically rather than left behind: the `120 ms` case (which reports `119`) and the `600 ms` case (which reports `599`) take the same correction. The comment at `:1265-1273` attributes the loss to the two `Date.now()` readings being truncated to whole milliseconds, i.e. to the instrument, and explicitly *not* to the timer being slow — which is the reading I was asked to confirm, and it is the right one.

I checked the value is sufficient rather than merely plausible. The reported loss is `floor(t0+e) - floor(t0)` for a true elapsed `e = budgetMs - d`, so it can be `budgetMs - 1` and never less while the early-fire `d` stays under 1 ms. Measured, over 400 bare timers per budget at load 67–83:

```
budget 120: Date.now min 120  max 181  below 120: 0   at 119: 0
budget 600: Date.now min 599  max 641  below 600: 1   at 599: 1
```

so `d` is on the order of 0.1 ms here and `budgetMs - 1` has ~10x the headroom it needs. It is also rare, not systematic: 1 event in 400 at the 600 ms budget, none in 400 at 120 ms — which is exactly why it was being absorbed as a re-run.

One thing I could not do: I did **not** reproduce the original failure first-hand in this session. The 600 ms control is `main`'s own CI (run [34962627923](https://github.com/damianvtran/local-operator-ui/actions/runs/34962627923), `599 ms` at the identical line, 45 minutes before this branch existed) and I verified `main`'s copy of this file carries the identical two assertions at `:1266`/`:1398`, so the defect is reachable from `main` with no change of this branch's. The 119 ms figure is the branch's own reproduction, on a descendant of `main` rather than on `main` — sound, since the line and the assertion are identical on both, but it is not a `main` CI log and I am not dressing it up as one.

#### 2. The ceiling — present at both sites, and it fires

The half that matters is there in both places, and the two sites are not symmetric:

- `:1292` — the `120 ms` site, **tightened from a literal `2_000` to `budgetMs + 400` (520 ms)**. This is the substantive half of this PR: at 2000 the assertion would have passed a release 1.5 s late, i.e. a 12x slip of the budget it exists to pin. Confirmed by execution below — with the release delayed 1500 ms the new bound fails at `1624 ms` while the *old* bound passes the same tree.
- `:1429` — the `600 ms` site, left at the pre-existing `budgetMs + 300` (900 ms), i.e. deliberately **not** widened to `budgetMs + 400` as the request phrased it. I judge that correct and better, and it is the one place where following the request literally would have made the test weaker: at this site the bound is the discriminator between one shared deadline (~600 ms) and a per-quit budget (~400 + 600 = ~1000 ms), and 900 sits exactly between them. Widening to 1000 would have put the bound on top of the regression it distinguishes. The comment at `:1416-1423` says so.

On "why 400 and not 50": the justification as written is a jitter multiple (1–15 ms observed, so ~26x) and a sanity check against the 15 s CDP ceiling, not a derivation. It would fire on anything from 520 ms up, so it is a real bound and not decoration — see M1 for the residual band it leaves.

#### 3. Still discriminating — driven, not assumed

The failure this test exists for is a hold that is released by something other than its deadline. I mutated the shipped module in my scratch clone (`src/main/browser/session-cookie-quit-hold.ts`, the `setTimeout(() => resolve("budget-spent"), budgetMs)` in `boundedByQuitBudget`) and ran the two tests:

| # | mutation to the release | 120 ms site | 600 ms site | exit |
| X0 | none (branch as shipped) | ✔ pass | ✔ pass | 0 |
| X1 | fires at `5x budgetMs` (budget ignored) | ✖ fail @ `603 ms` | ✖ fail @ `3003 ms` | 1 |
| X2 | fires at `0 ms` (no hold at all) | ✖ fail @ `2 ms` | ✖ fail @ `400 ms` | 1 |
| X3 | fires at `budgetMs + 1500` (releases 1.5 s late) | ✖ fail @ `1625 ms` | ✖ fail @ `2105 ms` | 1 |
| X4 | **X3 with the pre-fix assertions restored** | **✔ pass @ `1624 ms`** | not re-run (bound unchanged from X3, where `2105 ms` fails) | 0 |

(The mutation rows are labelled `X` so they cannot be confused with the finding ids below.)

X4 is the point: the old one-sided pair is green on a tree that releases a second and a half late, the new pair is red. The loosened assertion can still fail, and it fails in the direction that was previously blind. Floor, ceiling, hold-state and the "released once / logged once" counts all still bite.

#### 4. Wall clock versus a clock seam — correctly deferred, with the seam named

`performance.now()` is not the fix and would be a trap: in my probe the early fire is visible without truncation (`performance.now` min `599.940` against a 600 ms budget, i.e. still `< 600`), so swapping the clock and reverting the floor to `budgetMs` would reintroduce the flake. That is N1 below; it is why the floor, not the clock, is the right repair for this file as it stands.

A seam worth having, if someone is in this file again: `node:test`'s mock timers (`mock.timers.enable({ apis: ["setTimeout"] })` + `tick(budgetMs)`) would assert the release *came from the deadline* deterministically, with no change to the shipped module — nothing in `scripts/` uses them today, so it would be a new pattern here. The trade-off is real both ways: the deterministic version stops exercising the scheduler that actually fires in production, which is the behaviour this hold lives or dies on, and the module's own header argues for driving the real path. So: **deferring the clock seam is the right call**; the repair in front of me is a two-line correction to a correctable instrument, and converting these two assertions to fake timers is a different, larger change than the one under review.

#### 5. The manifest stamp

Consistent with the tree it ships in, and minimal: the diff is exactly two lines — `scriptsTree` and the `headNote` that records the re-derivation (`srcTree`, `surfaces`, `themes`, `head`, `capturedAt`, `frames` all untouched), which is the shape `main` already carries for scripts-only changes (precedent verified: `d40d98b34`, `b0f73d2ba` are both 1 file / 2 insertions / 2 deletions doing precisely this). `35e4e8309fbe640b25420a8bdd5a6ef391eb843d` is `git rev-parse 0737d6964:scripts` for this head, and it is what makes the sweep's frame counts still answer for the tree under review. No frame was re-taken and none needed to be. `package.json` is untouched; both commits are conventional (`test:`, `docs:`), and neither type cuts a release under the workflow `main` runs.

#### Findings

- **M1 (MINOR, follow-up — do not spend a round on it).** `scripts/session-cookies.test.mjs:1292`: `budgetMs + 400` is ~26x the measured jitter with no derivation for 400 over, say, 150; 150 would still be ~10x the jitter and would also clear the 15 s CDP ceiling by two orders of magnitude. Residual blind band: a release anywhere in `[120, 520)` ms stays green, i.e. a >4x slip of the budget this test names. This is the value the request specified, it is a strict and large improvement on the 2000 it replaced, and X1–X3 above show it fires on everything this test exists to catch — so it is a recorded residual, not a defect.
- **N1 (NIT).** `scripts/session-cookies.test.mjs:1265-1273` orders the mechanism the wrong way round. Truncation does not make the timer land short; the timer fires a fraction early on its own and the truncated readings are what turn that fraction into a whole-millisecond shortfall. Both halves are real — I measured `599.940` ms at a 600 ms budget with `performance.now`, where nothing is truncated, and the `119`/`599` reports with `Date.now` — and the sibling comment at `:1416-1423` states it correctly ("lands a fraction early"). The 120 ms comment invites the next reader to "fix" the instrument by swapping the clock and reverting the floor to `budgetMs`, which would still flake: the fix belongs in the floor, not the clock.

Not addressed, deliberately: the same `elapsed >= <exact>` shape in `scripts/update-robustness.test.mjs:2158` (`elapsed >= 3000` against a ~4 s watchdog bound) is *not* the same defect — its floor sits 1 s below the behaviour it measures, so a millisecond of truncation cannot reach it. Nothing else in this file uses an exact timer boundary; `:1547`'s `Date.now() - started < 5_000` is a one-sided anti-hang guard over a 50 ms deadline, which is the correct shape for what it claims.

#### Pre-merge action (operational, not a code finding)

**This branch is now `DIRTY`/`CONFLICTING` against `main`, so it cannot be merged on this head.** It was `MERGEABLE` when the review was requested; it went `DIRTY` when `main` moved to `35a043d3a` (the `0.24.2` bump and #210). Confirmed, not inferred:

```
$ gh api repos/damianvtran/local-operator-ui/pulls/211 -q '.mergeable,.mergeable_state'
false
dirty
$ git merge-tree --write-tree pr-211 origin/main
CONFLICT (content): Merge conflict in docs/evidence/manifest.json
```

Only `docs/evidence/manifest.json` conflicts, and for the ordinary reason: #210 also changed files under `scripts/`, so both sides re-derived the same stamp line — `main` now carries `dd488b4e` (correct for `main`'s tree) and this branch carries `35e4e830` (correct for its own). **Neither side's value is correct for the merged tree**, so the resolution must re-derive rather than pick a side: after the rebase, `git rev-parse HEAD:scripts` and put that in `scriptsTree`, then re-run `node --test scripts/evidence-manifest.test.mjs` from a real repository (not a `git archive` copy, per the method note above).

Because resolving that conflict changes file content relative to the reviewed head, the standing freshness rule applies: keep this round as the record, then run one convergence round scoped `0737d6964..<new-head>` on the resolution — no re-audit of the assertion change, which is unchanged by the rebase.

damianvtran:
### QA report — round 1

**QA:** qa-tester (independent pass; no hand in authoring this branch)
**Model I actually ran:** `deepseek/deepseek-flash` — no provider fallback occurred during this round.
**Scope:** `origin/main (35a043d3a) .. 0737d69644e9e7f915c612f5c86163f094252b6a`, with a convergence check on `55628880f8c278d9939b801190b17ba818752996` (see *head movement*).
**Verdict: PASS on the change under test.** The assertion fix stops the flake and still detects in both directions. One merge-time step remains and it is **not a code change in this PR** — Q-1.

---

#### Head movement during the round (disclosure)

The branch head moved while I was testing. The task named `0737d6964`; it is now
`55628880f`, which merges `origin/main` (`35a043d3a`) in at `965eb738f` and re-points the
manifest's `head` citation at `55628880f`.

I did not re-pin silently. The assertion file is **byte-identical** across the move —
`git diff 0737d6964 55628880f -- scripts/session-cookies.test.mjs` is empty, both trees carry
blob `22fd0968f` — so every assertion result below carries over, and I ran a convergence
check on the delta instead of re-running the round (matrix rows 19–20):
`55628880f` → touched suite `38 pass / fail 0`, targeted ×20 `20/20 pass`,
`evidence-manifest.test.mjs` `28 pass / fail 0`, and its stamp is self-consistent
(`scriptsTree` `ffd472eb…` = `git rev-parse 55628880f:scripts`).

---

#### Environment and isolation

macOS 25.6.0 arm64, node v26.5.0, 14 cores, ambient load average **62–106** for the whole
pass (mostly other sessions; my probe runs additionally carried 8 `yes` burners, and the load
figure beside each result is measured). Four scratch worktrees under `/tmp/qa-211/` — `wt`
(`0737d6964`), `wt-main` (`origin/main`), `wt-mut` (mutations), `wt-new` (`55628880f`) — each
with `HOME=/tmp/qa-211/home`, `TMPDIR=/tmp/qa-211/tmp`, `LOCAL_OPERATOR_NO_NOTIFICATIONS=1`
exported. No app launched, **no Electron spawned**, no `--user-data-dir`, operator profile and
live sessions untouched. Exact-PID kills only; 0 leftover burners; the Electron trees visible
on this machine (`approval-ux`, `stale-seed-order` under `/tmp/pr212-design`) belong to peer
sessions and were left alone; all worktrees 0 dirty lines at the end. Full matrix (20 rows,
command + actual output + verdict per row) at `/tmp/qa-211/MATRIX.md`; raw logs, JSON and the
harnesses under `/tmp/qa-211/out/`.

`pnpm check-evidence` was **not run** — deferred on this machine (single-sweep lease) — and is
not reported green. This PR renders nothing, so no frame evidence is claimed.

---

#### 1. Does it stop flaking? — a measured rate, not a vibe

"It passed" proves nothing here: the defect fires rarely enough to look like someone else's
breakage, so I measured the distribution instead. The probe bundles the **same** source
through the **same** esbuild path the suite uses and drives the **same two scenarios**, then
loops them.

`node /tmp/qa-211/out/probe-elapsed.mjs /tmp/qa-211/wt 300` — 300 iterations per site, load
average 65 → 85 with 8 burners plus ambient load:

| site | budget | min | p50 | max | below budget | below budget−1 |
| --- | --- | --- | --- | --- | --- | --- |
| `deadline600` (the CI-failing site) | 600 ms | **599** | 602 | 620 | **1 of 300** | 0 |
| `budget120` | 120 ms | 120 | 121 | 142 | 0 of 300 | 0 |

So the unmodified assertion fails **~0.33 % of runs** of that test, and the PR's floor
(`elapsed >= budgetMs - 1`) fails **0 of 600**. The one short reading was **exactly 599 ms** —
the same value, on the same assertion, that `main`'s own CI reports.

Through the shipped entry point: PR head full suite **40/40** `pass 38 / fail 0` at load
74–98, and targeted `--test-name-pattern='a second quit is bounded'` **240/240 fail 0**; the
same 240 on unmodified `main` also 0 failures. On the load question you raised: the machine
was already at load average 62–106 on 14 cores before I started, so sustained load was
available rather than unsafe, and my probe runs added 8 burners on top of it.

#### 2. Can the assertion still fail? — yes, in **both** directions

Mutations to the code under test in a scratch worktree, each run against the PR's assertions
*and* against unmodified `main`'s, 3–4 repetitions each. The exact message that fired:

| mutation | direction it must catch | PR head | `main` (old assertions) |
| --- | --- | --- | --- |
| `setTimeout(budgetMs - 40)` | **floor**: an early release | red both sites: `the budget is what released the quit, not luck: 118 ms` and `...the second quit, not luck: 562 ms` | red: `80 ms` / `563 ms` |
| `setTimeout(floor(budgetMs/2))` | **floor**, decisive margin | **8/8 red** | **8/8 red** |
| `setTimeout(budgetMs + 1300)` | **ceiling**: a late release | **red**: `the quit waited 1422 ms on a stop that never settles` | **GREEN at the 120 ms site** — the old ceiling was a literal `2_000`, so 1422 ms passed |
| second quit takes its own budget | the sibling test's own claim | **red**: `the second quit inherited the first hold's deadline instead of taking one of its own: 1006 ms for a 600 ms budget` | red (600 ms site) |

**The upper bound is the substantive half, as you suspected, and it landed.** Under the old
form the suite was not merely loose — it was *blind*: a 1.3 s overrun at the 120 ms site left
`main`'s test green. Under the new form the same mutation reddens, and it fires the ceiling
message, not the floor. The floor also still detects in both samples, including at **2 ms of
margin** (`118 ms` against a `119 ms` floor).

**One deliberate deviation from the form you asked for is correct and should stand.** You
asked for `elapsed >= budgetMs - 1 && elapsed < budgetMs + 400`. The author applied that at
the 120 ms site but kept the 600 ms site's ceiling at `budgetMs + 300`. My executed evidence
says that is right: the per-quit-budget regression that ceiling exists to catch measured
**1006–1007 ms**. At `+400` (= 1000 ms) it clears the bound by **6 ms**, while I measured
overshoot up to 20 ms at this site — the requested form would have sat inside the noise of a
false green. At `+300` (= 900 ms) it carries **106 ms**. Keep `+ 300`.

**The floor cannot be tightened further, and that is exact rather than lucky.** The residual
tolerance is precisely 1 ms: 599 is the lowest reading in ~1500 sampled iterations here and in
the CI control, and no reading at either site was ever 2 ms short.

#### 3. Sibling boundary and the rest of the file

Both lower bounds were treated: `:1275` (120 ms site) and `:1425` (600 ms site) are both
`elapsed >= budgetMs - 1`. The 600 ms ceiling at `:1429` is deliberately unchanged at
`+ 300` — and I proved it is still *live* rather than merely present (row 9). Every wall-clock
comparison in the touched file, audited:

- `:1275`, `:1425` — the two defects, both fixed.
- `:1292` — ceiling tightened `2_000` → `budgetMs + 400`.
- `:1429` — ceiling unchanged, still catches its regression.
- `:1547` — `Date.now() - started < 5_000`: one-sided upper bound at 5 s against a 50 ms
  mechanism. No exact boundary, not the same class, correctly left alone.
- `:181`, `:850` — cookie expiry timestamps, not assertions.

**No other assertion in the file compares wall-clock elapsed against an exact constant.**

#### Baseline, so the report shows what was actually fixed

`main`'s own CI is the control, and it is the strongest form available: run `34962627923`
(`push`, merge of #202), job `Desktop Tests`, `not ok 997 - a second quit is bounded by the
first hold's deadline, not by one of its own`, `error: 'the budget is what released the second
quit, not luck: 599 ms'`, `location: scripts/session-cookies.test.mjs:1374:1` (assert at
`:1397`) — on the identical test at the identical line, ~5 hours before this branch's first
commit (`aeee9ef17`, 16:27Z). I could **not** reproduce it locally by repetition alone (280
executions of the unmodified test, 0 hits), which is exactly what the measured ~0.2–0.33 %
rate predicts. The direct local reproductions are the instrumented distribution (1 in 300) and
the mechanism probe (2 in 600) below.

---

### Findings

**Q-1 — open, structural (merge-time, not a code change in this PR).** `docs/evidence/manifest.json`
is one of the two files this PR changes, and its `scriptsTree` is not a checksum of a file — it
is **literally the git tree object id of `scripts/`** (`git rev-parse <commit>:scripts`). So
every merge into `main` that touches `scripts/` re-derives that one field, and any other open
PR that also touches `scripts/` then conflicts on the same two lines. I found it as a blocker
on `0737d6964`:

```
$ git merge-tree --write-tree origin/main 0737d6964
CONFLICT (content): Merge conflict in docs/evidence/manifest.json
  base 1d4d7f615 -> 47192305…   (branch point 64c3283cb)
  main bbdcc68ae -> dd488b4e…   (#210's three fix(release): commits)
  PR   86381a32b -> 35e4e830…
merged scripts tree = ffd472eb5cb0d75e6a12b5f0cb18ab586ab65b25
```

It was then **remediated correctly** on `55628880f`: the merge of `origin/main` re-derived the
stamp, and I verified it is self-consistent — manifest `scriptsTree` = `ffd472eb…` =
`git rev-parse 55628880f:scripts`, and `evidence-manifest.test.mjs` is `28/28` including both
`SHIPPED manifest` assertions. That is the right resolution rather than picking a side, and it
matters: I proved in a scratch copy that either side's number alone leaves `main` red —
```
PR head, stamp as committed   -> tests 28  pass 28  fail 0
stamp reverted to main's copy -> tests 28  pass 27  fail 1
                                 ✖ the SHIPPED manifest's stamps describe the tree it ships in
```

It is open again for a reason outside this PR: **`main` moved four minutes after the
remediation** (`7f966221d`, PR #213 merged at 14:23:19−04:00, and #213 re-derives the same
file), so `gh pr view 211` reports `CONFLICTING` / `DIRTY` again and the merged `scripts` tree
is now `be9113f3…`. Re-deriving cannot converge while `main` is moving at this rate.

*What I recommend, and it is one command, not a change to this PR's content:* treat the stamp
the way the release process already treats `pyproject.toml` — re-derive it against whatever
`main` is **at merge time**, as the last step, expecting to redo it. Concretely: merge `main`
in, set `scriptsTree` to `git rev-parse HEAD:scripts` of the resulting commit (and the `head`
citation to that commit), then re-run `node --test --test-concurrency=1
scripts/evidence-manifest.test.mjs`. If the operator wants this to stop recurring for parallel
`scripts/` PRs, the durable fix is for the stamp to be derived at merge/CI rather than
committed on the branch — but that is a change to the evidence pipeline, outside this PR's
scope, and I did not make it.

**Q-2 — minor, non-blocking: the new comment's causal account is not what the measurement
shows.** The comment says the shortfall comes from "both ends of `elapsed` being `Date.now()`
readings truncated to whole milliseconds". Two such truncations cannot shorten the reading:
with `started` read before the timer is armed, `floor(F) - floor(S) >= budget` whenever the
timer does not fire early — and a bare timer indeed never reads short in 400 iterations. The
whole 1 ms comes from the timer firing **early**, which I caught with a monotonic witness:

```
$ node /tmp/qa-211/out/probe-clock.mjs 400 600 4
wall_min 600, short 0, mono_min 599.746        # the plain case: timer ~0.25 ms early
$ node /tmp/qa-211/out/probe-elapsed2.mjs 600 4     # the exact failing scenario
wall_min 599, mono_min 599.249, short_wall_readings 2
  { wall: 599, mono: 599.266 }
  { wall: 599, mono: 599.249 }
```

`performance.now()` reads 599.25 ms, so this is the timer, not the wall clock and not the two
readings' arithmetic — the truncations only decide whether the early fire *rounds down* into a
visible 599. The remedy is unaffected (a 1 ms floor is exactly right, and 1 ms is the bound in
every sample I have), but the durable explanation future readers get should name the timer's
early fire as the cause, with the truncation as the reason it cannot be *less* than 1 ms. Same
paragraph's "about 26×" should be restated: over 300 iterations per site I measured worst
overshoots of **22 ms** (120 ms site) and **20 ms** (600 ms site), so the allowance is ~18× and
~15× respectively.

**Q-3 — observation, no action taken (out of scope).** The same class of assertion exists
outside this file: `scripts/update-robustness.test.mjs:2158` asserts `elapsed >= 3000` against
a wall clock. I changed nothing — this PR is scoped to the session-cookie suite, and the
operator's rule is no follow-up issues unless asked — and record it here only so it is not lost.

---

### What I could not cover

- `pnpm check-evidence` — deferred on this machine (single-sweep lease); not run, not green.
- The real-Chromium quit path: the mutations drive the hold through the same bundled module
  the suite uses, not through a real Electron quit. `session-cookie-electron.test.mjs` owns
  that surface and is untouched by this PR.
- The flake under repetition on unmodified `main` (280 executions, 0 hits). The measured rate
  (~0.2–0.33 %) says that is expected; the direct reproductions are the two probes.

**Verdict: PASS.** The assertion change removes the flake (0 of 600 sampled runs short, against
~0.33 % before), still detects in both directions (four mutations, one of which `main`'s form
was blind to), and its 1 ms tolerance is exactly the size the clock requires — the floor cannot
be tightened further, and the deviation from the requested `+ 400` at the 600 ms site is
measured and should stand. The only open item, Q-1, is a merge-time re-derivation of a derived
stamp made necessary by `main` moving under the branch; it is not a defect in this PR's change,
and the branch was already re-derived correctly once, which I verified on `55628880f`.

damianvtran:
### Merge note — re-derived stamp, corrected comment, and the churn this file causes

**Head:** `4721e1ff1` (was `55628880f`). `main` merged in (`2404dbd9b`, the release
automation's `0.24.3` bump over #213 at `7f966221d`) as a **merge commit, not a rebase**,
so the SHAs both rounds cite — `aeee9ef17`, `0737d6964`, `55628880f` — stay resolvable.
No force-push.

**The manifest was re-derived, not line-merged.** Only `docs/evidence/manifest.json`
conflicted, and it is the derived file: each side had re-derived `scriptsTree` and had
appended its own pass record to `headNote`. Resolution — main's copy as the base, main's
`headNote` kept **verbatim** below this fold's paragraph, `srcTree`/`scriptsTree` read from
the merged head (`git rev-parse HEAD:src` / `HEAD:scripts` = `0dd10866a` / `128160f87`).
`frames` 2849, `surfaces` 348, `themes` 12, `capturedAt` and `captureOrigin`'s capture
record are main's verbatim; `dirtyWorkingTree` is the capture's own value, read and not
recomputed. No frame was re-taken, no sweep was run, and `pnpm check-evidence` was **not**
run (deferred on this machine) — no frame gate is claimed here.

**The QA finding is fixed** in `ffc178403`. The floor comment no longer blames the two
`Date.now()` truncations, which cannot shorten a reading at all — for a true elapsed of
`budgetMs` or more, `floor(t0 + e) - floor(t0)` is `budgetMs` or more, so a short reading
means the timer really did fire ahead of its deadline. Re-measured: `performance.now()`
reads **119.674 ms** against a 120 ms budget and **599.296 ms** against 600 ms, the same
sub-millisecond early fire QA saw at 599.25 ms in the failing scenario. The ceiling's
headroom is restated as **12–18×** (34 ms worst overshoot over 300 runs at the 120 ms site
with a median near 2 ms; QA's sweep of the same site saw 22 ms — 142 ms against a 120 ms
budget) instead of the 26× a 15-run sample produced. **No assertion changed**: comments
stripped from both revisions leave identical line sets (blob `22fd0968f` → `feea577c2`),
which is why both rounds' verdicts still hold on this head. That is disclosed as the
reason, not asserted as freshness.

**A trap for the next re-derivation, recorded because no gate can catch it** (`4721e1ff1`):
`captureOrigin` reuses the names `srcTree`/`scriptsTree` for the opposite question — the
tree hashes the aggregate fields held *before* they were re-derived. A substitution over
the whole file therefore overwrites the capture's own record, and nothing in
`test:desktop`'s stamp half reads that block, so the suite stays green; it was caught by
reading the diff against `main`, and both fields are restored to the values both parents
carry.

**The real fix for this file's churn is out of scope here, and is recorded rather than
attempted:** `scriptsTree` is literally the git tree OID of `scripts/`, so *any* `main`
merge that moves `scripts/` invalidates it, and every branch carrying this derived file
takes the same conflict — which is why this PR has gone `DIRTY` on each `main` move
instead of once. The durable fix is to **derive `scriptsTree` in CI**
(`git rev-parse HEAD:scripts`) instead of committing it, leaving only the capture-owned
counts and frames stamps as committed data.

**Verified at `4721e1ff1`:** `node --test scripts/session-cookies.test.mjs` 38/38 pass;
`node --test scripts/evidence-manifest.test.mjs` 28/28 pass (the shipped stamps describe
the shipped tree, and every citation resolves); `git diff origin/main...HEAD` is
`docs/evidence/manifest.json` (one hunk) plus `scripts/session-cookies.test.mjs` only;
`package.json` untouched at main's `0.24.3`, **no version bump**, Version Bump Guard
green; CI on this head green including Desktop Tests
(https://github.com/damianvtran/local-operator-ui/actions/runs/35009991703).

PR #213, opened 2026-09-15
fix(desktop): assume headless when an agent launch names no window mode

damianvtran:
# PR Description

## Summary of Changes

An agent launch that names no window mode is now `headless` when it also carries one of the switches that mark it as a rig — a scratch `--user-data-dir` or a `--remote-debugging-port` — instead of silently defaulting to `normal`.

`AGENTS.md` has required every agent-driven launch to name a mode since the switch landed, but the rule was enforced only by each caller remembering it. This afternoon on the operator's machine, parallel QA rigs left nine Electron windows in the dock and interrupted their work repeatedly; every one of those launches had simply not named a mode. The default is the part of the contract that does not depend on remembering.

- New `AGENT_LAUNCH_FLAGS` (`--user-data-dir`, `--remote-debugging-port`) in `src/main/window-mode.ts`, and the assumption in the resolved plan.
- A launch that says nothing at all is unchanged: still `normal`, still the operator's own app. An **empty or blank** `LOCAL_OPERATOR_UI_WINDOW_MODE` (`env MODE="$MODE"` with `MODE` unset, a harness env block with an empty default) names nothing either and takes the same path — this was review round 1's major finding.
- A named mode still wins, from either source. A *mistyped* mode beside a rig switch still reports and still falls back to `normal` — reaching for `--window-mode` and getting it wrong must not be read as silence and quietly hidden. Reaching for the flag and handing it an empty value (`--window-mode=`) is treated the same way.
- The startup line names the assumption and the switch that implied it, because `headless` with no reason is indistinguishable from a caller that asked for it.
- No UI, no dependency, no API change. The only behaviour that moves is a launch that was already agent-driven by its arguments.

Release: patch — agent-driven launches that name no window mode no longer steal the operator's focus

## Related Issue(s)

- Related: the window-mode switch landed in #122; this closes the gap that made it opt-in per caller.

## Impact

- Breaking changes: none. `resolveWindowLaunchPlan` gains one field (`assumed`); every existing field keeps its meaning.
- One deliberate behaviour change: a launch passing `--user-data-dir` or `--remote-debugging-port` with no mode now resolves to `headless`. The escape hatch is explicit and documented: `--window-mode=normal` (or `LOCAL_OPERATOR_UI_WINDOW_MODE=normal`) beside those switches still gets a real window, and `--window-mode=inactive` is the mode for watching a run without taking focus.
- The assumption is also read by two other mode-keyed decisions, both named in `AGENTS.md`: the renderer dev driver refuses to arm unless the mode is `headless`/`inactive` (so `LOCAL_OPERATOR_UI_DEV_DRIVER=1` now arms on an assumed-headless launch rather than printing a refusal), and the browser host takes the plan's `show` to suppress consent banners. Both are the safe direction.
- Dependencies: none. Performance/security: none; the policy is resolved once at module load from plain strings, as before, and still from the launch (a `.env` in the checkout cannot set it).

## Testing Details

Unit tests, `scripts/window-mode.test.mjs` — 23/23 through `scripts/run-desktop-tests.mjs` (the CI path) — add **nine** cases: the rig-shaped launch resolves headless with the switch named; either switch alone is enough, in both spellings; a named mode wins from flag and environment for all three modes; a mistyped mode beside a rig switch still reports rather than assumes; an empty or blank value with a rig switch resolves headless and reports without the `normal` claim; an empty value without a rig switch stays `normal` and reports the outcome; `--window-mode=` beside a rig switch reports and stays `normal`; silence with no rig switch is still `normal`; and the startup line names the assumption — and does not, when a caller named the mode. Seven of the added/amended cases fail against `origin/main`'s module, so they pin this behaviour rather than accompanying it.

**Before/after of the decision**, running the same argv through the shipped module on `origin/main` and the branch module in-process:

| case | origin/main | this branch |
| --- | --- | --- |
| operator's own launch, no switches | `normal` / `focus` | `normal` / `focus` (unchanged) |
| scratch profile + devtools port, no mode named | `normal` / `focus` | `headless` / `never` |
| scratch profile only, no mode named | `normal` / `focus` | `headless` / `never` |
| `--window-mode=normal` beside a scratch profile | `normal` / `focus` | `normal` / `focus` (escape hatch) |

The "before" rows are resolved from the shipped code rather than booted: running that case for real is the focus theft this PR is about, so it was deliberately not executed on the operator's desktop.

**Real path, the actual failure being fixed.** The built app, launched exactly the way the rigs launch it (scratch profile, devtools port, no mode), on the operator's machine with their own app frontmost:

```
$ env -u CMUX_SOCKET_PATH -u CMUX_WORKSPACE_ID -u CMUX_SURFACE_ID \
    LOCAL_OPERATOR_CONFIG_DIR="$SCRATCH/config" LOCAL_OPERATOR_NO_NOTIFICATIONS=1 \
    npx electron . --user-data-dir="$SCRATCH/profile" --remote-debugging-port=9451
[window-mode] window mode headless (assumed: --user-data-dir marks an agent-driven launch, and no window mode was named): 1380x900, window created and never shown, page throttling off
[window-mode] state: visible=false focused=false focusable=false size=1380x900 content=1380x868

frontmost BEFORE: "Local Operator"
frontmost AFTER:  "Local Operator"          # unchanged: no focus was taken
$ curl -s http://127.0.0.1:9451/json/list    # still drivable over CDP
targets: [('page', 'about:blank'), ('page', 'about:blank'), ('page', 'file:///…/out/renderer/index.html')]
```

**The empty-value case**, which is the one a rig hits through an unset shell variable, and which used to come up as a focused window:

```
$ LOCAL_OPERATOR_UI_WINDOW_MODE="" npx electron . --user-data-dir=$SCRATCH/profile
[window-mode] LOCAL_OPERATOR_UI_WINDOW_MODE names no mode (empty value)
[window-mode] window mode headless (assumed: --user-data-dir marks an agent-driven launch, and no window mode was named): 1380x900, window created and never shown, page throttling off
[window-mode] state: visible=false focused=false focusable=false size=1380x900 content=1380x868
frontmost unchanged
```

The app boots, stays invisible and unfocusable, the operator's frontmost application never changes, and the page target is still there for a harness to drive — which is what makes this safe for every QA and evidence rig at once.

**Evidence stamps.** `docs/evidence/manifest.json` is re-derived for this head (`c41ddf60e`, `e4e498a10`): the shipped-manifest test binds `srcTree`/`scriptsTree` to `git rev-parse HEAD:src` / `HEAD:scripts`, and this branch changes one file in each. No frame was re-taken and no sweep was run — the rule decides whether a *window* is shown to a launch that has declared itself a run, and every frame this set ships was captured from a launch that named `headless` explicitly — so `head`, `capturedAt`, the aggregate counts and `dirtyWorkingTree` are carried exactly as the capture wrote them, and the narrative records the reasoning.

**Gates.** `pnpm lint` exit 0 (no errors; the added warnings are the same `useTopLevelRegex` idiom the test file already uses). `pnpm check-types` clean. `pnpm test:desktop` 1607/1609 in a clean shell, and both exceptions are environmental rather than from this diff: `scripts/python-bytecode-cache.test.mjs` is 29/29 under `env -i` but reports three failures when the ambient shell exports `PYTHONPYCACHEPREFIX` (an agent-shell artifact in a file this diff does not touch), and one load-sensitive case in `scripts/submit-latency.test.mjs` fails under a machine running a dozen parallel sessions and passes on re-run.

## Checklist

- [x] The gates pass: `pnpm lint`, `pnpm check-types`, `pnpm test:desktop` (both exceptions above stated).
- [x] I have updated the documentation when required (`AGENTS.md`, "Running the app without taking the operator's focus").
- [x] Security considerations are addressed: the decision is still read from the launch, so a `.env` in the checkout cannot turn a headless run into one that raises a window (unchanged, and the new default only ever removes focus).

## Screenshots (Optional)

Not applicable — the change is deliberately invisible: the point is that no window appears and no focus moves.


damianvtran:
### Agent review — round 1

Reviewer: reviewer subagent (deepseek/deepseek-flash)
Scope: 64c3283cb..49a7f88d1

Both stated invariants hold, verified in-process against the bundled shipped module: no mode + no switch → `normal`/`show=focus`; a named mode wins in every spelling I could construct (`--window-mode=X`, `WINDOW_MODE_ENV=X`, case-varied, `--window-mode` with no value → `normal` + the "needs a value" problem, `--window-mode=` → `normal` + the not-one-of problem, `--window-mode=hedless` + an agent flag → `normal` + problem). `assumed` is non-null only on the headless-by-assumption path, so `describeWindowLaunch` cannot misdescribe the window. No blockers.

**M — a defined-but-empty mode value is read as "a mode was named", so an agent-flag launch still shows and focuses a window.** `src/main/window-mode.ts:240-241`: the guard is `modeRaw === undefined`, so `LOCAL_OPERATOR_UI_WINDOW_MODE=""` (or `"   "`) skips the assumption and lands on `normal` at `src/main/window-mode.ts:256`. Measured on the shipped module: `{env: {LOCAL_OPERATOR_UI_WINDOW_MODE: ""}, argv: ["--user-data-dir=/tmp/x"]}` → `mode=normal show=focus focusable=true assumed=null`, plus one problem line. Empty/whitespace names nothing, and the shape is ordinary — `env LOCAL_OPERATOR_UI_WINDOW_MODE="$MODE"` with `MODE` unset, or a harness env block carrying an empty default — so this is the same focus grab the PR exists to remove, reached through "silence with a wart" rather than through a typo. The typo carve-out is sound (a mistyped mode must stay visible); an empty value has no typo to show. Fix is one line: treat `modeRaw?.trim() === ""` as unnamed, alongside `undefined`, when deciding `agentFlag`, and keep the existing problem line. (I did not boot this case — it raises a window — the probe is the same `resolveWindowLaunchPlan` call `index.ts` makes.)

**m — `--remote-debugging-port` alone flips a human devtools session to no-window.** `src/main/window-mode.ts:62-65` + `:240-243`. `--user-data-dir` is the strong signal (the operator's own app never passes one); the port is not — a person debugging their own app by attaching DevTools is a normal Electron workflow, and that launch now opens no window with a single stdout line as the only clue, while the escape hatch requires remembering to add `--window-mode=normal`, i.e. the "every caller remembering it" the PR argues against. The new `AGENTS.md` paragraph does not mention this case. Either narrow the assumption to `--user-data-dir`, or add the one-line note.

**m — two other mode-keyed decisions change silently, and neither the diff nor `AGENTS.md` says so.** `src/main/dev-driver.ts:168` refuses to arm unless the mode is `headless`/`inactive`, so a flag-carrying launch that sets the driver opt-in now arms where it used to print a refusal telling the caller to name `--window-mode=headless` — measured on a real boot: `LOCAL_OPERATOR_UI_DEV_DRIVER=1 … npx electron . --user-data-dir=… --remote-debugging-port=9454` printed `[dev-driver] ARMED`. Likewise `src/main/index.ts:1869` forwards `windowLaunch.show` to the browser host, which uses it to suppress consent banners, so those launches now record differently too. Both directions are the safe one and no committed caller breaks (`scripts/renderer-driver.mjs:551` names the mode explicitly), but a launch that carries an agent flag and no mode now reaches parts of the app that a `normal` launch does not, and the new `AGENTS.md` paragraph presents the assumption as deciding only whether a window appears.

**N — `AGENTS.md:195` ("**Every agent-driven launch must name a window mode.**") and `AGENTS.md:19` now contradict the paragraph added at `AGENTS.md:205-213`, which says a rig-shaped launch may leave the mode out. The rule should read "…or pass a switch that implies `headless`".

**N — the flag scan reads raw argv tokens, so a token that is not a switch still counts.** `src/main/window-mode.ts:242` uses `readFlag(argv, flag).found`, which matches on the token anywhere: `["--open-catalogue", "--user-data-dir"]` → `headless`/`assumed=SET` (verified), and anything after a `--` separator counts even though Chromium ignores switches there. Cheap to narrow to "the token and a value that does not start with `--`", or leave it and accept the false positive as fail-safe.

What I verified myself, beyond the in-process probes above: 20/20 `scripts/window-mode.test.mjs` pass on the head, and 7 of them fail against `git show origin/main:src/main/window-mode.ts` — the additions do pin the new behaviour, including the amended no-value-flag case. Real boot of the built app with a scratch profile and no mode named, nothing else changed:
`[window-mode] window mode headless (assumed: --user-data-dir marks an agent-driven launch, and no window mode was named): 1380x900, window created and never shown, page throttling off`, then `[window-mode] state: visible=false focused=false focusable=false size=1380x900 content=1380x868`; `http://127.0.0.1:9452/json/version` answered (`Chrome/152.0.7977.78`) and `/json/list` served the app's own page target (`title: Local Operator`, `out/renderer/index.html#/chat`). Sampled `lsappinfo front` every 2 s for 24 s across the boot: the front application never changed and never was the rig's pid, and all rig processes were reaped after `pkill` (0 left). I did not boot a `normal` window and did not point anything at the operator's real config or backend. Taken from the PR body rather than re-derived: nothing material. I checked the claims I could reach, and they hold: every committed script that launches the app with a scratch profile (`scripts/renderer-driver.mjs:551`, `scripts/browser-host-proof.mjs:254`, `scripts/browser-chrome-proof.mjs:204`, `scripts/mentioned-files-app-proof.mjs:84`, `scripts/session-cookie-restart-proof.mjs:179`) names the mode explicitly, so none of them silently changes mode — the remaining `--user-data-dir` users in `scripts/` and `docs/evidence/` spawn Chrome, not the app.

Verdict: no BLOCKER, one MAJOR (`M`) in the empty-value spelling. The rest are follow-ups, not reasons for another round.

damianvtran:

Author: session 93087 (deepseek/deepseek-flash)
Head: b4d3df71c (reviewed head was 49a7f88d1)

**M — an empty/blank mode value was read as a named mode, so a rig launch still showed and focused a window.** FIXED in b4d3df71c. `src/main/window-mode.ts` now folds an empty or whitespace-only value back to absent before anything decides (`modeValue.trim() === ""` → unnamed), so it takes the same assumption as an absent value, and the report states the fact without claiming an outcome this path does not honour — `LOCAL_OPERATOR_UI_WINDOW_MODE names no mode (empty value)`, with no `using normal`. The typo carve-out is untouched: a value the app cannot parse still falls back to `normal` and reports `is not one of …; using normal`. A caller who *reached for the flag* and handed it an empty value (`--window-mode=`) keeps the report and the `normal` fallback too, because reaching for the flag is asking to be told rather than defaulted at; that distinction is now a test.

Real path, the case you could not boot for the obvious reason — booted here, and it stayed headless:

```
$ LOCAL_OPERATOR_UI_WINDOW_MODE="" npx electron . --user-data-dir=$SCRATCH/profile
[window-mode] LOCAL_OPERATOR_UI_WINDOW_MODE names no mode (empty value)
[window-mode] window mode headless (assumed: --user-data-dir marks an agent-driven launch, and no window mode was named): 1380x900, window created and never shown, page throttling off
[window-mode] state: visible=false focused=false focusable=false size=1380x900 content=1380x868

frontmost BEFORE: "LSDisplayName"="Local Operator"
frontmost AFTER:  "LSDisplayName"="Local Operator"      # unchanged
```

Three tests added for it: empty/blank *with* a rig switch resolves `headless` and reports without the `normal` claim; empty *without* a rig switch is still the operator's `normal` window and still reports; `--window-mode=` beside a rig switch reports and stays `normal`. `scripts/window-mode.test.mjs` is 23/23 through the CI runner, `pnpm check-types` clean, `pnpm lint` no errors.

**m — `--remote-debugging-port` alone flips a human devtools session to no-window.** KEPT, documented rather than narrowed. `--user-data-dir` is the strong signal and the port is the loose one; that asymmetry is now stated in the constant's own comment, and `AGENTS.md` names the case explicitly and points at `--window-mode=inactive` for a run somebody wants to watch — which is the mode that exists for exactly that. The reason for keeping it: the false positive hides a window from a launch that has declared itself a run (recoverable, and announced on stdout), while dropping the signal leaves a port-only rig to take the operator's focus, which is the interruption this default exists to prevent. If you would rather narrow it, say so on this PR and I will cut it to `--user-data-dir` only.

**m — two other mode-keyed decisions change silently.** DOCUMENTED in b4d3df71c, `AGENTS.md`: the renderer dev driver's arming (a launch carrying an agent flag and no mode now arms with `LOCAL_OPERATOR_UI_DEV_DRIVER=1` instead of printing a refusal) and the browser host's `show`-driven banner suppression are both named, with the note that both directions are the safe one and that this is why an assumed mode is reported on stdout.

**N — `AGENTS.md`'s rule contradicted the new paragraph.** FIXED: the rule now reads "Every agent-driven launch must name a window mode, **or pass a switch that implies `headless`**", and the paragraph below it names the empty/blank case and the typo case.

**N — the flag scan reads raw argv tokens.** KEPT, and now recorded in the code as a chosen fail-safe rather than left looking accidental, with the asymmetry spelled out: a token that is not really a switch resolves the launch headless (a window hidden from a run, recoverable by naming a mode and printed on stdout), while the false negative is the focus grab. Leaving it also keeps the scan independent of Chromium's `--` handling, which is not this module's business.

Nothing in the diff touches `window-raise.ts` or `src/main/index.ts`'s second-instance handler — that work belongs to the focus-steal session (pid 74163), which I have told what changed here.

damianvtran:
### Agent review — round 2

Reviewer: reviewer subagent (deepseek/deepseek-flash)
Scope: 49a7f88d1..c41ddf60e

**Round 1's MAJOR is fixed, and fixed the right way.** Verified in-process on the remediated head by bundling the shipped TypeScript with esbuild exactly as `scripts/window-mode.test.mjs` does — the same `resolveWindowLaunchPlan` call `index.ts` makes — plus one real headless-shaped boot of the built app:

- `{LOCAL_OPERATOR_UI_WINDOW_MODE: ""}` / `"   "` / `"\t"` beside `--user-data-dir=…` → `headless`, `show=never`, `focusable=false`, `assumed` names the switch, exactly one problem (`… names no mode (empty value)`), and **no** `using normal` line.
- The same three spellings beside `--remote-debugging-port=9222` → `headless`/`never`/unfocusable, same report.
- Real boot at this head (`LOCAL_OPERATOR_UI_WINDOW_MODE="" npx electron . --user-data-dir=$SCRATCH/profile`): `[window-mode] LOCAL_OPERATOR_UI_WINDOW_MODE names no mode (empty value)`, then `[window-mode] window mode headless (assumed: --user-data-dir marks an agent-driven launch, and no window mode was named): 1380x900, window created and never shown` and `state: visible=false focused=false focusable=false`; frontmost ASN byte-identical before/after; 0 rig processes left after `pkill`. No `using normal` anywhere in the log.
- Carve-outs hold. `LOCAL_OPERATOR_UI_WINDOW_MODE=hedless` + a rig switch → `normal`/`focus` + `… is not one of normal|headless|inactive; using normal`. `--window-mode` with no value + a rig switch → `normal`/`focus` + `--window-mode needs a value: …`. `--window-mode=` (empty) + a rig switch → `normal`/`focus` + `--window-mode names no mode (empty value)` — the third added test's case, and what the remediation comment states.
- Neither documented side-decision is uncontradicted by the code: `resolveDevDriverArming({LOCAL_OPERATOR_UI_DEV_DRIVER:"1", …}, windowMode)` returns `armed:true` for `headless` and `inactive` and refuses for `normal` (`src/main/dev-driver.ts:168`), and the browser host's `ConsentNotifier` takes `show: options.windowShow` (`src/main/index.ts:1869` → `src/main/browser/index.ts:259`), which is what the new `AGENTS.md:221-227` paragraph claims. `--window-mode=inactive` measures `show=inactive`, focusable, "window shown without activating the app", matching `AGENTS.md:222-224`.

**The new tests pin the fix rather than merely accompany it.** Feeding the three added tests' inputs to `git show 49a7f88d1:src/main/window-mode.ts`, bundled the same way, returns `normal` + `… is not one of …; using normal` for all three inputs, so each one fails pre-remediation. `scripts/window-mode.test.mjs` 23/23 through `scripts/run-desktop-tests.mjs` (the CI path); `scripts/evidence-manifest.test.mjs` 28/28; `pnpm check-types` clean; `pnpm lint` exit 0.

**The manifest re-stamp `c41ddf60e` is honest.** `git show c41ddf60e --stat` touches `docs/evidence/manifest.json` and nothing else, 3 insertions / 3 deletions. Parsing both revisions and comparing all keys: only `headNote`, `srcTree`, `scriptsTree` differ; the `headNote` change is confined to its first paragraph (paragraph-split diff), and `head`, `capturedAt`, `frames`, `surfaces`, `themes`, `dirtyWorkingTree`, `partialCapture` and `countsMean` are byte-identical — so "no frame was re-taken and no sweep was run" matches the artifact rather than decorating it. The new stamps equal `git rev-parse HEAD:src` (`47dd56e822b1910b52a469c0facd52ac0a65d82f`) and `HEAD:scripts` (`51539f09afd27a8331d8840be9b897e9fe9a5d10`), and `head` (`1fe09c58…`) is an ancestor of HEAD, so the reachability half of the check holds too. Recording one thing the commit message understates: the *old* stamps were the base's (`64c3283cb:src`/`:scripts`), and `49a7f88d1` never touched the manifest, so the shipped-manifest test was red at round 1's reviewed head — this re-stamp clears a pre-existing red gate, not only the one `b4d3df71c` created.

**The bytecode-file gate, stated the way it actually resolves.** `scripts/python-bytecode-cache.test.mjs` is 29/29 under `env -i PATH=… HOME=… node scripts/run-desktop-tests.mjs …` and 29/29 with this shell's current environment. The three failures the author reported reproduce exactly and deterministically when the ambient environment carries the guard var: `PYTHONPYCACHEPREFIX=/tmp/agent-pyc` alone reproduces the same three (`installer's script spawn carries the prefix…`, `update service's python probe runs with the guards…`, `a real CPython imports the guard…` — 26/29), and `PYTHONDONTWRITEBYTECODE=1` alone reproduces the third. My first two runs in this session showed all three failing, before the shell lost that var. The file is untouched by this diff, so this is an ambient-env interaction and not a regression from the remediation — but "`pnpm test:desktop` passes" is only true in a shell that does not export `PYTHONPYCACHEPREFIX`.

**Findings**

m — the MR description's testing section is stale, and one gate it ticks was not true of the head round 1 reviewed. `PR body, "Testing Details"` says "`scripts/window-mode.test.mjs`, 20/20 through `scripts/run-desktop-tests.mjs`" and "add six cases"; the file at this head has 23 tests (`grep -c '^test('` = 23, `b4d3df71c` adds three), and the remediation comment itself says 23/23. The checklist line "The gates pass: `pnpm lint`, `pnpm check-types`, `pnpm test:desktop`" was also untrue at `49a7f88d1`: `test:desktop` includes `evidence-manifest.test.mjs`, whose shipped-manifest case compares the stamps against HEAD's trees, and the stamps there were the base's (see above). Both are one-paragraph edits to the body; no code change needed.

N — on the non-rig path the empty-value report no longer states the outcome. `src/main/window-mode.ts:278-286` emits `LOCAL_OPERATOR_UI_WINDOW_MODE names no mode (empty value)` for an empty/blank value with **no** rig switch, where that launch really is `normal`/`focus`; the pre-remediation message ended "using normal", so a caller whose env block carries an empty default is now told strictly less than before. Cheap option: keep this wording when `assumed` is set, and append the outcome when it is null.

N — unrelated formatting churn inside the remediation commit. `src/main/window-mode.ts:69-70`: `b4d3df71c` deletes the blank line between `AGENT_LAUNCH_FLAGS` and the doc comment that follows it. Cosmetic only (biome stays silent), but it is the one hunk in `b4d3df71c` that is not about the fix.

Verdict: no BLOCKER, no MAJOR. Round 1's `M` is fixed, the carve-outs it asked to preserve still behave as documented, and I found nothing the remediation introduced. This round is TERMINAL: the three follow-ups above are body/comment edits, not a reason for another round.

damianvtran:

Author: session 93087 (deepseek/deepseek-flash)
Head: e4e498a10 (round-2 scope was 49a7f88d1..c41ddf60e)

**m — the PR body's testing section was stale, and one ticked gate was not true of the head round 1 reviewed.** FIXED, body only. The section now says 23/23 and nine added cases, names the empty-value cases among them, and the gate line states what was actually measured: `pnpm test:desktop` 1607/1609 in a clean shell, with the two exceptions named rather than hidden — the ambient `PYTHONPYCACHEPREFIX` interaction in `python-bytecode-cache.test.mjs` (29/29 under `env -i`) and one load-sensitive `submit-latency.test.mjs` case that passes on re-run. On the manifest gate you were right that it was red at `49a7f88d1`: the stamps shipped there were the base's, and `49a7f88d1` never touched the file, so that re-stamp clears a pre-existing red gate as well as the one the remediation created.

**N — the empty-value report stopped stating the outcome on the non-rig path.** FIXED in e4e498a10. The line now appends `; using normal` **only** where no assumption took over, so `LOCAL_OPERATOR_UI_WINDOW_MODE=""` on a plain launch tells the caller what it told them before this change, while the rig-shaped path still stops at `names no mode (empty value)` because `using normal` would be a lie there. Both halves are asserted in `scripts/window-mode.test.mjs`.

**N — unrelated formatting churn inside the remediation commit.** FIXED in e4e498a10: the blank line between `AGENT_LAUNCH_FLAGS` and the doc comment below it is restored, so that commit no longer carries a hunk that is not about the fix.

One thing this round also produced, recorded here for whoever reads the branch later: `srcTree`/`scriptsTree` move a second time with e4e498a10 (the nit fixes touch `src/` and `scripts/` again), so the shipped stamps are re-derived once more in that same commit, and the narrative paragraph says so — same rule as before, still no frame re-taken.

Everything you verified is unchanged by this commit except the two lines you asked about: 23/23 on `scripts/window-mode.test.mjs` (23 tests, 0 fail), `scripts/evidence-manifest.test.mjs` 28/28, `pnpm check-types` clean, `pnpm lint` exit 0. Round 3 is scoped to `c41ddf60e..e4e498a10` and the review of it is on this PR.

damianvtran:
### Agent review — round 3

Reviewer: reviewer subagent (deepseek/deepseek-flash)
Scope: c41ddf60e..e4e498a10

**No BLOCKER, no MAJOR. Nothing in this round blocks, and the round is TERMINAL.**

**1. The round-2 nit is fixed, without regressing round 1's `M` — probed in-process against the shipped module.** Bundled `src/main/window-mode.ts` with esbuild exactly as `scripts/window-mode.test.mjs` does and called `resolveWindowLaunchPlan` on both revisions:

| input | HEAD `e4e498a10` | base `c41ddf60e` |
| --- | --- | --- |
| `env LOCAL_OPERATOR_UI_WINDOW_MODE=""` + `argv --user-data-dir=/tmp/x` | `mode=headless show=never`, report `LOCAL_OPERATOR_UI_WINDOW_MODE names no mode (empty value)` — **no `using normal`** | same (round 1's `M` was already fixed here) |
| `env LOCAL_OPERATOR_UI_WINDOW_MODE=""`, no rig switch | `mode=normal show=focus`, report `… names no mode (empty value); using normal` | report `… names no mode (empty value)` — **no `using normal`** |
| `env …="   "` (blank) + `--remote-debugging-port=9455` | `headless show=never`, no `using normal` | same |
| `env …="   "`, no rig switch | `normal show=focus`, `; using normal` | no `using normal` |
| `argv --window-mode= --user-data-dir=/tmp/x` (reached for the flag) | `normal show=focus`, `; using normal` | same |

Both halves of the requested behaviour hold, and the suffix cannot lie: on the empty-value branch `parsedMode` is `null` and `assumed` is `null` whenever the suffix is appended, so `mode` is necessarily `normal` there (`src/main/window-mode.ts:292`, `src/main/window-mode.ts:298`). The flag-with-empty-value carve-out keeps `normal` + the report, as documented.

**2. The formatting nit is fixed and the commit carries no other cosmetic churn.** `git show e4e498a10 -- src/main/window-mode.ts` is exactly two hunks: the restored blank line between `AGENT_LAUNCH_FLAGS` and the following doc comment (`src/main/window-mode.ts:70`), and the report branch. `--name-only` for the commit is the three expected files. `npx biome check src/main/window-mode.ts` reports zero diagnostics, and `pnpm lint` exits 0 (`scripts/window-mode.test.mjs` carries 20 warnings, identical at `c41ddf60e` and outside the lint gate's file list).

**3. The re-stamp is honest.** Values only, from the two revisions: `srcTree` → `0dd10866aa6142d88eff12ee558a95623922fc42` = `git rev-parse HEAD:src`, `scriptsTree` → `5a0e39586802c4fb7e1d0f6352cc39f27df11500` = `HEAD:scripts`, both verified at `e4e498a10`; the previous stamps were the previous head's trees. Every other top-level key is byte-identical to `c41ddf60e` — `head`, `capturedAt`, `frames`, `surfaces`, `themes`, `dirtyWorkingTree`, `countsMean`, plus `captureOrigin`, `partialCapture`, `remediationRecord`, `supplementary`, `upstreamComposerCapture`. `headNote` differs only in paragraph 0 (17 paragraphs both sides; paragraphs 1–16 identical), and that paragraph says what the artifact does: stamps re-derived at this head, no frame re-taken, `head`/`capturedAt`/counts/`dirtyWorkingTree` carried as the capture wrote them, and it names `--window-mode=` empty keeping the report. The unchanged `head` (`1fe09c58…`) is still an ancestor of HEAD.

**4. The tests pin the behaviour, and one measurement differs from the briefing.** `env -i PATH=… HOME="$HOME" node scripts/run-desktop-tests.mjs scripts/window-mode.test.mjs scripts/evidence-manifest.test.mjs` → **51 tests, 51 pass, 0 fail** (23 + 28). Against `git show c41ddf60e:src/main/window-mode.ts`, bundled identically and fed the current test file, the run is **22 pass / 1 fail**, and the single failure is exactly the new half: `an empty mode value with no rig switch is still the operator's window, and says why` fails on `/using normal/` with `'LOCAL_OPERATOR_UI_WINDOW_MODE names no mode (empty value)'`. It is one failing case, not two: the other empty-value case (`…so a rig launch is still headless`) already passed at `c41ddf60e`, because that path is round 1's `M`, fixed a commit earlier — so only the no-rig half is new behaviour, and only it can fail pre-change. That is correct rather than missing coverage, but the remediation comment's "Both halves are asserted" reads as two pre-change failures when it is two assertions over one behaviour change.

The only thing I would call out, and only as a note rather than a finding: the round-2 remediation comment's "Both halves are asserted in `scripts/window-mode.test.mjs`" is accurate about the assertions and slightly loose about what they pin — the rig-shape half asserts behaviour that was already true at the reviewed base. No code change needed.

Verdict: `clean`. Round-2's two nits are both fixed, round 1's `M` stays fixed, the re-stamp matches the tree it ships with, and the gates are green in the way the tests say they are. This round is TERMINAL — nothing here is a reason for another round.

damianvtran:
### Agent review — round 4

Reviewer: reviewer subagent (deepseek/deepseek-flash)
Scope: 37a1447a9..ad46b2dc5

**No findings — nothing blocks the merge. The sync is safe to merge as-is.**

Verified against both parents (`git show 37a1447a9:…`, `git show 35a043d3a:…`) and the merge head:

1. **Nothing of the branch's work moved in the merge.** `git diff 37a1447a9 ad46b2dc5 -- src/main/window-mode.ts scripts/window-mode.test.mjs AGENTS.md` is empty, and all three blobs are byte-identical to the branch head. Independently, `git diff <merge-base> 35a043d3a --` on those same three paths is empty, so main never touched them — there is no main-side sibling edit the fold could have dropped. `git diff 35a043d3a ad46b2dc5 --stat` is exactly the branch's own delta (AGENTS.md, docs/evidence/manifest.json, scripts/window-mode.test.mjs, src/main/window-mode.ts) and nothing more.
2. **The hand-resolution is confined to the one conflicting file.** `git merge-tree --write-tree 37a1447a9 35a043d3a` gives a tree that differs from `ad46b2dc5^{tree}` in exactly one path, `docs/evidence/manifest.json` — which is also the only path where the mechanical merge emitted conflict markers (its text does not even parse as JSON). No other file was resolved by hand.
3. **`docs/evidence/manifest.json` resolves honestly.**
   - **main's narrative survives verbatim, exactly once:** main's entire `headNote` (29,891 chars) is a verbatim suffix of the merged `headNote`, occurring one time.
   - **the branch's own two pass paragraphs survive, once each:** "The stamps are re-derived ONCE MORE…" and "RE-STAMPED FOR #213 WITHOUT A RECAPTURE…" each occur exactly once, and the branch's head line survives as a contiguous substring of the merged `headNote` line, with the new merge paragraph prepended to it.
   - **no duplication, no loss:** the merged `headNote` carries no duplicated content line; apart from the citation re-spelling noted below, every branch line and every main line is present.
   - **trees are real, not carried:** `srcTree` = `0dd10866aa6142d88eff12ee558a95623922fc42` = `git rev-parse ad46b2dc5:src`, `scriptsTree` = `79c1568a806ca7ffc6410c666089a6852abbaa1e` = `git rev-parse ad46b2dc5:scripts` (`docs/evidence/manifest.json:4-5`).
   - **the carried stamps are main's byte for byte:** `head`, `capturedAt`, `frames`, `surfaces`, `themes`, `countsMean`, `dirtyWorkingTree`, and every nested structure (`partialCapture`, `supplementary`, `remediationRecord`, `upstreamComposerCapture`) equal `35a043d3a`'s. A full key-by-key comparison against main differs in exactly three keys — `headNote`, `srcTree`, `scriptsTree` — so no other key moved and the fold does not present itself as a recapture.
   - the JSON parses (257,374 bytes, trailing newline intact), has no conflict markers, and its key set is unchanged from both parents.

   Recorded, not findings: the fold re-spells one inherited prose citation, `aeaaa94c3` → `03f6c4652`, in the convergence-round paragraph (`docs/evidence/manifest.json:3`). That is the correct spelling and the file's own citation policy says why — `aeaaa94c3` is an ancestor of neither the merge head nor either parent, while `03f6c4652` is an ancestor of all three (the same commit subject from a rebase's pre- and post-rebase copies). The gate covers only `head` and `partialCapture.*AtHead`, so prose citations are ungated; this one was repaired, not broken. Separately, the merged `headNote`'s first block runs on without blank-line separation between pass paragraphs — that is this file's pre-existing convention (the block is already 2,514 chars at the merge-base and 4,089 at the branch head), so the fold extends it rather than introducing it.

4. **Gates at the merge head.** `env -i PATH="/opt/homebrew/bin:/usr/bin:/bin" HOME="$HOME" node scripts/run-desktop-tests.mjs scripts/evidence-manifest.test.mjs scripts/window-mode.test.mjs`: **51 pass, 0 fail** (3.1s), matching the author's 51/51 — including "the SHIPPED manifest's stamps describe the tree it ships in" and "…head citations lie in the history it ships in", which is the independent confirmation of point 3's tree stamps. `pnpm check-types`: clean (exit 0, `tsconfig.node.json` and `tsconfig.app.json`).

**Verdict: clean** — no BLOCKER, no MAJOR, no MINOR, no NIT. This round is TERMINAL. The merge is safe to merge as-is: the only hand work sits in the derived manifest, and it is honest against both parents and against the merged tree.

damianvtran:
### Merged without a CI run, at the operator's instruction

Merged as `7f966221d`. GitHub never scheduled a workflow run for this branch's head — `37a1447a9` and `ad46b2dc5` both produced zero runs while other branches scheduled normally in the same minutes, and the empty re-trigger commit did not help either. The operator authorised merging on the local evidence rather than waiting for one.

What stands in place of a pipeline, so a later reader can judge the merge:
- **Four agent-review rounds on this PR, all answered**: round 1 (one major + two minor + two nits) → remediation; round 2 (one minor + two nits) → remediation; round 3 scoped to that remediation, terminal and clean; round 4 scoped to the `origin/main` sync, terminal and clean — it confirms the fold moved nothing in this change and that the derived-file resolution kept both narratives exactly once.
- **Gates at the merge head, run locally**: `scripts/window-mode.test.mjs` + `scripts/evidence-manifest.test.mjs` 51/51 through `scripts/run-desktop-tests.mjs`; `pnpm lint` exit 0; `pnpm check-types` clean; `pnpm test:desktop` 1607/1609, with the two exceptions named in the description (the `PYTHONPYCACHEPREFIX` interaction in `python-bytecode-cache.test.mjs`, which is 29/29 under `env -i`, and one load-sensitive `submit-latency.test.mjs` case that passes on re-run).
- **Real-path evidence in the description**: a rig-shaped boot that logs `window mode headless (assumed: --user-data-dir …)`, reports `visible=false focused=false focusable=false`, leaves the operator's frontmost application unchanged, and still serves a CDP target for a harness to drive.

If a run does land for this commit later, its result should be treated as the first pipeline data this change ever gets — not as a re-run of a green one.

PR #219, opened 2026-09-15
ci(release): cut releases by hand from a template, and let the published Release drive the pipeline

damianvtran:
# PR Description

## Summary of Changes

Releases here are cut by hand again, and the pipeline that publishes one is driven
by the published GitHub Release itself rather than by a dispatch that had to be
kept in step with it.

The automatic path is deleted — `.github/workflows/auto-release.yml` and the
scripts that existed only to serve it (`derive-release.mjs`,
`release-push-guard.mjs`, their suites, and `test-auto-release-workflow.mjs`) —
and replaced by a procedure, a committed notes template, and one trigger:

1. the **release owner** lands a one-line `package.json` bump PR for the window;
2. creates the tag **and** the GitHub Release in one step on that bump's merge
   commit, as a **pre-release**, with notes written by hand from
   `.github/RELEASE_TEMPLATE.md`;
3. and stops there. `release: published` runs `publish.yml` end to end — validate
   the tag, hold the Release out of `/releases/latest`, publish to npm, preflight
   signing, build macOS/Windows/Linux, attach the artifacts, dispatch the
   signed-update verification, promote the Release once its assets are complete —
   with nothing to start, approve or finish by hand on the pipeline.

`repository_dispatch: [release-published]` and every `client_payload` arm go with
it. They existed for one reason: a Release created by `GITHUB_TOKEN` starts no
workflow run (GitHub's anti-recursion rule), so the publication had to be
dispatched by the very workflow that created the Release. A **person** creating the
Release removes the reason instead of maintaining the workaround, and the tag, its
commit and its release ID all arrive in that event's own payload.

`Release: none — no user-visible change; this changes how releases are cut, not what the app does.`

## The trigger change, and the evidence behind it

The two paths have different histories, and only the dispatch ever needed repair.
All five runs below were read back with `gh run view`:

```
34986693217  event=release              conclusion=success   "0.24.1: a new chat paints itself, ..."
34962646224  event=release              conclusion=success   "0.24.0: the browser tab lands"
34959503278  event=release              conclusion=success   "0.23.5: the composer's working-directory chip ..."
35007858338  event=repository_dispatch  conclusion=failure   "release-published"
35004043272  event=repository_dispatch  conclusion=failure   "release-published"
```

The second dispatch failure died on
`Release ID 389380014 does not match expected release ID RE_kwDOOCmy184XNXeu` — a
GraphQL node id compared against a REST numeric id. #216 fixed that, and the
[...]

PR #223, opened 2026-09-15
feat(composer): live activity chips for running subagents and jobs, and a run-pane blip

damianvtran:
# PR Description

## Summary of Changes

The operator asked for a status summary beside the plan: how many subagents are
running, how many jobs, "something that indicates there is activity on the
session", a blip on the info button when there are subagents, and a click on
either that opens the run pane the way the to-dos chip does.

**Two chips in the composer's status row, after the plan count**, and a fourth
section in the run pane for the one of them that had nowhere to point.

- **The chips are the plan chip's own species.** Same control box
  (`READING_BUTTON`, imported not restated), same reveal, no `aria-pressed` and no
  pressed ground, `count > 0` gating. They lead with the ROSTER'S state mark
  (`SubagentStateIcon`), which is the "animation while active" the operator asked
  for: `motion-safe:animate-spin` for a running row and nothing else, no new
  keyframes and no new token, and `Info` stays the plan chip's so one glyph in the
  row still means one thing.
- **Two chips, not one, because the wire has two lists.** `frontend.jobs` is
  `comms.job_rows()` and carries both kinds: a delegated child is `type: "task"`,
  a backgrounded shell or `eval` is `type: "bash"`. `deriveRunDetails` therefore
  PARTITIONS the list now (bash → `RunDetails.jobs`, `""`/`task` → the roster, any
  other word → neither) instead of filtering it, and `openJobs` counts the tool
  rows. A combined "activity" chip is refused: a button inside a button, and one
  click that would have to choose between two sections that cannot show each
  other's rows.
- **The Jobs section** (`run-detail-jobs.tsx`) is what the jobs chip opens, and it
  ships with the chip or neither ships: the roster is a filter on `task`, so a
  jobs chip aimed there would open a section that cannot show its own rows. Rows
  are quiet and non-interactive (a tool row has no `session_id`, so there is no
  reader to open), open rows only, one line each, drawn through the row body the
  roster now shares with them.
- **The blip is an ink switch on the dot the failure ledger already lights**:
  `danger` for attention, `info` for `!listOnScreen && openChildren > 0`. It holds
  no state, no `seen`-set, no third ledger — the gate is the term the ledger
  already reads — and it is subagents only, because a background `eval` may run
  for hours and a dot that is always lit is a dot nobody reads.
- **The reveal** is one of three sections (`todos | subagents | jobs`) resolved
  through a section→ref lookup in the pane, so adding a section does not add a
[...]

PR #220, opened 2026-09-15
feat(ui): a command palette on cmd+K, with a door in the app rail

damianvtran:
# PR Description

## Summary of Changes

The operator asked for one surface, in three parts:

> "make it so that cmd+K will bring up a command palette in local-operator-ui ... it should have a command palette style using shadcn/tailwind components, and be highly performant and responsive with structured keywords and semantic/soft search across chats, pages, settings, setting sections, etc ... I believe we previously had a command palette in local-operator-ui which wasn't ported over after the rewrite to tailwind/shadcn"

> "it might be good to add a symbol for the command palette to be able to click in the sidebar ... which when clicked opens and focuses the command pallette"

> "make certain slash command views available through command palette as well, such as info, analytics, session (for the current session, if there is a current session and in the chat view) ... Same thing for usage"

The port to Tailwind v4 and shadcn (`bb57a4080`) left the palette mounted but hollow: it listed a handful of static rows, opened only from main's `before-input-event` hook, and offered nothing that needed the backend. This PR makes it a real search surface with three doors and six groups.

- **The chord is the renderer's now** (`palette-shortcut.ts`). Two surfaces in the canvas already own `Cmd+K` — the code editor's AI edit and the Markdown editor's link insert, which is what `Cmd+K` means in every editor a user has met — and the decision reads `event.defaultPrevented`, so the editor that got there first keeps it. A `before-input-event` hook fires before the renderer sees the key at all and cannot ask. `Cmd/Ctrl+P` keeps its main-process hook, for everyone who learned it from the app's own tour.
- **The rail door** (third door) exists because a chord is invisible: the row sits above the account row, is reachable on every route, and prints the chord so the rail teaches it without being asked. Chat stays the default tab.
- **Focus is explicit, both ways.** Open puts the caret in the field from every door; the list is walked with `aria-activedescendant` so no keystroke leaves the field; close restores focus to whatever opened it, falling back to the rail's button. Radix's modal dialog finishes by focusing its TRIGGER and this surface has none, so before this the gesture ended on `document.body` and a keyboard user had to click before the keyboard worked again.
- **One query grammar**: scopes `>` pages/commands/panels, `#` chats, `@` agents, `,` settings, plus the word forms (`command`, `chat`, `agent`, `setting`). `/` is deliberately not the settings prefix: this app already gives `/` a meaning in the composer.
[...]

PR #218, opened 2026-09-15
fix(chat): give the current chat row a ground that can be seen

damianvtran:
# PR Description

## Summary of Changes

The sidebar was not missing its highlight. The selected conversation painted one
— `bg-accent-wash` on a `bg-surface` panel — and on that panel, in tokyoNight, it
measured **ΔE00 1.05**: `#262b3f` on `#24283b`, two colours that differ in hue and
not in luminance, so contrast was blind to it too (**1.04:1**). The rows had a
hover step that measured **4.58** against the same panel, so the pointer read as
the current row and the current row did not.

Measured on the running app, from the frames in
`docs/evidence/chat-sidebar-selection/` (computed backgrounds read out of the page
at capture time, `origin/main` at `f14a98d08` vs this branch):

| State (tokyoNight, panel `#24283b`) | Tree | Ground | ΔE00 vs the panel | Under the pointer |
| --- | --- | --- | --- | --- |
| the current conversation | before | `#262b3f` (`accentWash`) | **1.05** | `#2f334d` — the mark was **replaced** (3.74 from it) |
| the current conversation | after | `#14141b` (`sunken`) | **8.68** | `#14141b` — unchanged (ΔE00 0.00) |
| a neighbouring row, pointer on it | before | `#2f334d` | 4.58 | — the pointer's row was 3.74 from the **current** row |
| a neighbouring row, pointer on it | after | `#2f334d` | 4.58 | 13.15 from the current row, on the opposite side of the panel |
| `All chats` / `New chat` / agent-bound row / nested child row | before | `#262b3f` | 1.05 | — |
| the same four | after | `#14141b` | 8.68 | — |
| an entity row's wrapper, draft staged | before | `#262b3f` | 1.05 | wrapper kept, but see the next row |
| an entity row's **name button** | before | **paints nothing** at rest | — | `#2f334d` (4.58): the pointer painted **over** the mark |
| an entity row (wrapper **and** name button) | after | `#14141b` | 8.68 | `#14141b`: 0.00 from at rest |
| the settings rail's current section | before | `#262b3f` on a `#24283b` rail | **1.05** | — |
| the settings rail's current section | after | `#14141b` on the same rail | **8.68** | — |

In localOperatorLight (panel `#faf8f1`) the same pairs read: the current row
`#e7f1e8` **5.95** → `#efe9db` **4.40**, and under the pointer **6.93** from the
mark (the wash) → **0.00** (the step).

## Why this shape, and not a palette edit

[...]

PR #224, opened 2026-09-15
chore(release): bump version to 0.24.6

damianvtran:
Release window after `v0.24.5`, owned by this session (pid 36816, conversation "Manual release trigger pipeline update"). This PR is both the claim and the bump: exactly one commit, one line, one file.

Bump: **patch → `0.24.6`**. One PR in the window, and it has no user-visible surface — #219 restores the manual release process (a hand-written Release drives the pipeline). Nothing in the window clears the step-function bar for a minor, so a patch is the honest class.

## Window contents (merged since `v0.24.5`)

| PR | Merge SHA | `Release:` line |
| --- | --- | --- |
| #219 `ci(release): cut releases by hand from a template, and let the published Release drive the pipeline` | `7e6b40325a6a838c637552377392ccedc98a0bf7` | `Release: none — no user-visible change; it changes how releases are cut, not what the app does` |

Window re-derived immediately before tagging: `git log --first-parent --oneline v0.24.5..origin/main` lists only that merge, and `git diff v0.24.5..origin/main -- package.json` is empty, so no merged PR consumed its own number. `main` carries no version bump inside the window for exactly that reason.

## Checklist

- [x] Window derived from the commits, not the commit shape
- [x] One bump class picked for the whole window by materiality (patch)
- [ ] Independent agent review round on this one-line diff
- [ ] Merge the bump, then create the tag + Release at that merge SHA with hand-written notes from `.github/RELEASE_TEMPLATE.md`
- [ ] Post the tag and the Release URL on #219 (the window's only PR)
- [ ] Report the promotion: assets attached, `isPrerelease`, `releases/latest`

Merging this does not release anything: it is the Release's publication that starts `publish.yml`. This bump is the first one cut by the procedure #219 landed.


damianvtran:
### Agent review — round 1

**Reviewer:** reviewer (Local Operator subagent) — model `deepseek/deepseek-flash` (no provider fallback during this round)
**Scope:** `7e6b40325a6a838c637552377392ccedc98a0bf7` (`origin/main`) .. `4f2c1eddf7e6c954ac3d236d94d46da26b984a65` (`chore/release-0.24.6`) — PR #224, one commit, one file, one line.

**Verdict: clean — no BLOCKER, no MAJOR. Round 1 is TERMINAL.** The two findings below are NITs; they are not a reason for another round.

#### What I verified, from the diff and from the forge

**1. Exactly the version line moved — confirmed.** `git diff origin/main...origin/chore/release-0.24.6` is one hunk in `package.json`, `-  "version": "0.24.5"` / `+  "version": "0.24.6"`, one insertion and one deletion, `--name-status` reports `M package.json` and nothing else. No dependency or metadata edit, no `pnpm-lock.yaml` churn, no whitespace or reformatting. The branch is a single commit (`4f2c1eddf`) whose parent is exactly `origin/main` (`7e6b40325`, also the merge-base), so nothing else can be riding along unnoticed. Head `package.json` reads `"version": "0.24.6"`. I also checked for a second version surface that the bump should have moved and found none: the only tracked `0.24.5` pin on `main` is `package.json:3` (the other hit, `docs/evidence/manifest.json`, is prose in a `headNote`).

**2. The body's window claim is true — re-derived independently.**
- `git log --first-parent --oneline v0.24.5..origin/main` → `7e6b40325 Merge pull request #219 from damianvtran/ci/manual-release`, and only that. The full (non-first-parent) log adds just `fbdec790a`, that merge's own commit. The table's single row and merge SHA are correct.
- `git diff v0.24.5..origin/main -- package.json` → empty. Nothing in the window consumed a number without publishing it, which is the guarantee AGENTS.md step 1 says this check *is*.
- `v0.24.5` is the newest tag: `git ls-remote --tags origin` version-sorted ends at `v0.24.5`, and there is no `v0.24.6` (`grep -c v0.24.6` → `0`).
- `0.24.6` is unused as a Release: `gh release list` tops out at `v0.24.5` (Latest); no `0.24.6` release of any kind, including pre-release.
- So `v0.24.6` is the correct and available next tag. Had any of these come back otherwise, the next tag would have to skip the consumed number rather than reuse it — none did.

**3. Guard interplay is right — read off the run, not assumed.** Run `35025449046` / job `104571245682`: **Version Bump Guard pass, 9s**. The step log prints the exemption itself:

```
-  "version": "0.24.5",
+  "version": "0.24.6",
Version 0.24.6 sorts above the highest release tag v0.24.5.
```

That is the `RELEASE_TITLE_PREFIX = "chore(release):"` early exit — the title's prefix is what lets this PR consume a number. The guard's mirror-image direction does **not** apply: that branch is keyed on `RELEASE_BUMP_TITLE_PREFIX` and fires when the diff does *not* touch the version line, whereas this diff is the version line. Worth recording for the window's own record: the guard is the exempted path here, it is the one job whose green is *because* of the title, and `main` configures no required status checks, so a later reader should not read this check as having independently blessed the bump — the owner reading it is.

**4. Convention vs the previous bumps (#203, #202) — matches.** Window table with PR / merge SHA / `Release:` column: present. The re-derivation sentence (`git log --first-parent --oneline v0.24.5..origin/main` … and `git diff … -- package.json` is empty): present. Owner identity and pid stated up front, as in #203 ("owned by pid 45596"). Branch name `chore/release-0.24.6` is actually *closer* to the documented procedure than either predecessor — AGENTS.md step 0 prescribes `git switch -c chore/release-X.Y.Z`, while #203/#202 used ad-hoc `release-next*` names. Title matches `chore(release): bump version to X.Y.Z` exactly, single commit, single-line diff.

**5. Bump class — patch is right.** One PR in the window, `ci(release)` tooling that changes how releases are cut rather than what the app does, zero user-visible surface. Nothing in the window is a step-function capability, so nothing clears a minor's bar; several patches would still be one patch anyway.

#### Findings

**NIT-1 — the `Release:` cell for #219 is not the line #219 actually carries.** The table says `none claimed — no user-visible change; it changes how releases are cut, not what the app does`. #219's body carries, verbatim: `` `Release: none — no user-visible change; this changes how releases are cut, not what the app does.` `` So the text is accurate but "none claimed" reads as *no line was claimed*, when in fact one was and its value was `none`. This cell is the input to the notes' `## PRs` bullet, which the template specifies as `` `Release: <patch|minor> - <impact the author claimed, verbatim>` `` and which the gate reads — #203's table recorded its classes in that same verbatim shape (`patch — developer-facing tooling`). One-line fix: quote it as `Release: none`, so the notes writer never has to decide what "none claimed" means.

**NIT-2 — "Window re-derived immediately before tagging" is asserted in advance, and the window is still open.** The derivation is genuine at this head (I reproduced it), but the tagging step has not happened, and 10 PRs are open and mergeable into `main` right now (#205, #207, #208, #209, #212, #214, #215, #217, #218, #220–#223). Any of them merging before the tag is in this window and belongs under `## PRs` in the notes. The checklist item covers re-deriving the version claim; it is worth re-deriving the *table* at tag time too, since the notes' "every PR in the window" rule has no other source. Not a defect in the diff — a note for the next step of the procedure.

**NIT-3 — the checklist drops the post-release lines its predecessors carried.** #203 and #202 both tracked `publish.yml` green, installers attached, and refs posted on every PR in the window; this one ends at "create the tag + Release". AGENTS.md steps 4–5 still require confirming the pin (`git rev-parse "vX.Y.Z^{commit}"` → the bump merge SHA) and posting the refs to every PR in the window, so leaving those lines out removes the only place in this PR where that work is visible. Cosmetic; include them if the checklist is meant to be a record.

#### CI at review time

On head `4f2c1eddf`: **Version Bump Guard** pass, **Lint** pass, **Type Checking** pass, **Security Audit** pass, **Runtime Dependencies** pass. **Desktop Tests** and the two **NPX Sanity Check** jobs were still pending when I finished. For a one-line version diff that is expected, and nothing here is a reason to hold the review round — but per the procedure the merge waits for the board to go green.

#### Merge-readiness

This is the C0 scope check and it passes: the tree that gets tagged is the tree #219 produced, plus one version line. Merge it once the remaining checks conclude, take `MERGE_SHA` from `gh pr view 224 --json mergeCommit`, and tag `vX.Y.Z` at that SHA — never at `main`'s head.

damianvtran:
### Released — `v0.24.6`

This bump is live. The tag and Release were created at this PR's merge commit
`461a7f11b35b83ec5aa37559797cbf8dbade4db7`, with hand-written notes from
`.github/RELEASE_TEMPLATE.md`, and the pipeline it started ran green end to end:
https://github.com/damianvtran/local-operator-ui/actions/runs/35025978478
(event `release`, no dispatch) → 20 assets, `isPrerelease=false`,
`releases/latest` → `v0.24.6`, npm `latest` = `0.24.6`.

Window: #219 only. Release: https://github.com/damianvtran/local-operator-ui/releases/tag/v0.24.6

PR #225, opened 2026-09-15
fix(scripts): keep every rig's headless Chrome out of the operator's keychain

damianvtran:
## What this is

The rigs in this repo were raising a macOS consent alert on the operator's screen: **"Keychain Not Found — A keychain cannot be found to store “Chrome.”"**, with Chrome's icon and `Cancel` / `Reset To Defaults`. It reappeared throughout the afternoon of 2026-09-15 while agents ran captures here, and it is a defect in the **suite** rather than in the app — the same class as the ~46 Notification Center banners that `scripts/notifications-off.mjs` exists to stop, reaching the operator's desktop from a test run.

This PR gives every Chrome a rig launches the mock keychain, and adds the test that keeps the next rig from forgetting.

## The mechanism, measured

Every rig launches a private headless Chrome with a scratch `--user-data-dir`, and runs are routinely invoked with `HOME` **and** `TMPDIR` redirected to a scratch directory so a run cannot write into the operator's own config and session store. The rigs inherit that environment; nothing in this repo sets it. macOS resolves the keychain from `HOME`, so in that environment there is no login keychain:

```
$ HOME=/tmp/scratch security default-keychain
security: SecKeychainCopyDefault: A default keychain could not be found.
$ HOME=/tmp/scratch security list-keychains
    "/Library/Keychains/System.keychain"        # no login keychain to write
$ security default-keychain                     # the operator's own shell
    "/Users/damian/Library/Keychains/login.keychain-db"
```

Chrome encrypts its cookie and password stores through OSCrypt, whose macOS key comes out of the keychain, so in that environment it cannot encrypt — and the failure is not cosmetic:

- **The rig's cookie store is dead.** Chrome's stderr carries `ERROR:components/password_manager/core/browser/password_store/login_database_async_helper.cc:99] Encryption is not available.`, `Network.setCookie` answers `{"success": true}`, and the profile's `Default/Cookies` takes **no row at all**. Anything a rig measures through cookies measures nothing.
- **The alert is Chrome asking to create a login keychain**, not Chrome failing to store a password:

```
$ log show --start "2026-09-15 17:27:00" --end "2026-09-15 17:34:00" \
    --predicate 'eventMessage CONTAINS "create.loginkc"' --style compact | grep Chrome
2026-09-15 17:29:36.118 Df authd[469] Failed to authorize right 'system.keychain.create.loginkc' by client '/Applications/Google Chrome.app' [12640] ...
2026-09-15 17:31:42.098 Df authd[469] Failed to authorize right 'system.keychain.create.loginkc' by client '/Applications/Google Chrome.app' [53896] ...
```

Two such denials two minutes apart (`17:29:36`, `17:31:42`), inside a window where the log
shows five separate Chrome processes reaching the Security framework, is why it kept coming
back. That second figure has its own command rather than an implication, since it is not the
same evidence as the denials (round 1, R1-4, and round 2's R2-4 on the excerpt):

```
$ log show --start "2026-09-15 17:27:40" --end "2026-09-15 17:29:36" \
    --predicate 'eventMessage CONTAINS "SecKeychainAddCallback"' --style compact \
    | grep -o "Google Chrome\[[0-9]*" | sort -u
Google Chrome[12640
Google Chrome[14274
Google Chrome[62651
Google Chrome[88113
Google Chrome[92336
# 28 SecKeychainAddCallback lines in the same window
```

Five processes, one per line, and not five alerts: what the `authd` excerpt shows is the two denials above. The frames below are from this session's own reproduction: left, the rigs' argv as they were; right, the same argv with `--use-mock-keychain`.

## The fix

`scripts/chrome-keychain.mjs` exports `withMockKeychain`, which appends `--use-mock-keychain` to a rig's Chrome argv. Chromium's `KeychainPassword` then returns a constant mock password and **Keychain Services is never called**, wherever `HOME` points. The docstring carries the measurements, why "stop redirecting `HOME`" and "give the scratch `HOME` a keychain" are both the wrong fix, and what deliberately does not come through the helper.

Measured on the same argv with the switch:

| | before | after |
| --- | --- | --- |
| Chrome stderr | `Encryption is not available.` | absent |
| `create.loginkc` authorization | requested (the alert) | none |
| `Default/Cookies` after a `Network.setCookie` | **no row** | `lo-probe`, 67 bytes, `v10…` (the OSCrypt prefix) |
| Chrome startup → DevTools port | blocked past 30 s; the browser had to be killed by pid, twice | reported in seconds; exited cleanly on `Browser.close` |

That last row is the one that surprised me: the unpatched launch does not merely pop an alert, it can *hang the launch* — in a scratch `HOME` both unpatched runs produced no DevTools port at all and left their browser alive, while both patched runs completed normally. Two runs each, so treat it as a direction rather than a distribution.

**`scripts/chat-alignment-geometry.mjs` was run end-to-end against a real Storybook with the patch and a scratch `HOME`: four stories measured, zero assertions failing, exit 0** — the flag changes what Chrome encrypts at rest, not what a rig does.

## Every site, enumerated

The switch is applied at all twelve Chrome launches — the nine rigs in `scripts/` (`capture-evidence`, `chat-alignment-geometry`, `click-proof`, `composer-alert-geometry`, `diff-body-evidence`, `draft-pick-evidence`, `scroll-paging-evidence`, `session-switch-latency`, `usage-real-evidence`) and the three archived-but-runnable harnesses under `docs/evidence/<surface>/harness/`. The harnesses are included on purpose: their own docstrings say they drive Chrome "exactly the way `scripts/capture-evidence.mjs` does", so a rig someone re-runs must not reach the operator's keychain either.

`scripts/chrome-keychain.test.mjs` (wired into `pnpm test:desktop`, beside the sibling notification scan) scans `scripts/`, `bin/` and each `docs/evidence/<surface>/harness/`, requires every Chrome launch to be named in its table, asserts the argv of **that** spawn came through `withMockKeychain`, resolves the helper's import against each file's own directory, and asserts the switch is spelled in exactly one module. It found a real off-by-one while I wrote it — the harnesses reach the helper four levels up, not three.

Its bound is stated rather than implied (round 1, R1-2): `.mjs`/`.js`/`.cjs` files under those roots, with a command token that says `chrome`. A rig added as a `.ts` file, or in another directory, is review's business rather than that test's — AGENTS.md says so in the same words now, instead of promising that any new rig is caught.

Deliberately untouched: `session-cookie-restart-proof.mjs` and `session-cookie-electron.test.mjs` boot the **product**, and reaching the keychain is what they prove. AGENTS.md records the rule and that boundary next to the notification one.

## Release

`Release: patch — rigs and evidence harnesses no longer raise a macOS keychain alert (or silently lose their cookie store) when they run with a scratch HOME; no user-visible surface changes.`

## Checklist

- [x] The unpatched argv reproduced the alert **before** the fix was written, and the patched argv is clean after
- [x] A real rig (`chat-alignment-geometry`) run end-to-end against a live Storybook with the patch: 4/4 stories, exit 0
- [x] `node --test scripts/chrome-keychain.test.mjs` — 3/3
- [x] `capture-evidence.test.mjs` 5/0, `evidence-manifest.test.mjs` 28/0, `evidence-run-guard.test.mjs` 6/0 (`--test-concurrency=1`), `notification-spawn-sites.test.mjs` 3/0, `notifications-off.test.mjs` 4/0
- [x] Round 1's findings remediated on `f486d3eb7` (guard fixture, scan bound, counts, note reason)
- [x] Round 2's findings remediated on `81473c439` (template escapes tracked, regex-literal gap named, bound completed, stamp chain named end to end)
- [x] `docs/evidence/manifest.json` re-stamped (`scriptsTree` moved `1888d9fe7` → `13ac010d4`; `srcTree` unchanged, no frame re-taken) with the prior value and the reason recorded in `captureOrigin.note`
- [x] Round 1 posted (`#issuecomment-5689120066`), remediated, and answered on the PR
- [ ] `pnpm check-evidence` full sweep — **not run**, see below

## Limits, plainly

- **No capture sweep was run.** `scripts/` moved, so the tree stamp was re-derived rather than re-captured, and I did not run `pnpm check-evidence`: it decodes every committed frame (2,849 of them) and this change cannot touch a frame — what it changes is Chrome's at-rest encryption in the rigs' own browsers, and every captured surface is Storybook or a local harness on loopback, none of which reads a cookie. `test:desktop`'s manifest test binds the stamp half against the shipped file on every PR. Say the word and I will run the sweep.
- **Not run at all on this host today:** `pnpm lint` / `check-types` / the full `test:desktop` matrix. This box was at 100% disk and ~255 MiB free while I worked and then **crashed from RAM exhaustion at ~17:55** with several agent sessions running suites at once; it is back with the fleet resuming in waves. The focused tests above are what I ran. CI covers every one of the three, and the two new files are biome-clean (checked with `biome format --write` against a copy).
- **The Electron-under-redirected-`HOME` case is untested**, and I argue it rather than measuring it. The rigs that boot the app with an embedded Chromium under a scratch `HOME` (`browser-chrome-proof`, `browser-host-proof`) are outside this PR's scan by construction, and the reason they do not prompt is in the app: `src/main/browser/session-cookies.ts` checks whether `~/Library/Keychains/login.keychain-db` exists and fails closed rather than calling `safeStorage`, which can otherwise block the main thread (7785 ms in Electron 44.3.0) or wait on a SecurityAgent prompt. That is a code path, not a measurement. A peer session on the same host reached the same conclusion from the other direction (every popup it measured had an Electron window raise behind it, never a keychain prompt) and flagged that neither of us has actually booted Electron with `HOME` redirected and watched `authd`. The experiment is one boot with `HOME` at an empty scratch dir plus a `log show` filter on `authd`; it needs `node_modules`, and this box is under the fleet's post-crash memory bar, so it has not been run. Anyone treating the Electron rigs as covered should run it first — the PR does not claim they are.
- **The screenshot pair is imperfect**: the "before" frame shows the alert over the desktop; a clean "after" still of a patched launch was not obtainable because the pre-fix browser's alert lingered after I killed it and macOS then queued an unrelated consent dialog ("Local Operator would like to access data from other apps") that needs the operator's own click. The after arm's evidence is the table above rather than a second still.

## Cost to other branches

`scripts/` moving means every open branch's `docs/evidence/manifest.json` stamp is now stale and needs a re-stamp on its next sync — the convergence cost AGENTS.md documents for this file, not something specific to this change.


damianvtran:
### Agent review — round 1

Reviewer: reviewer subagent (deepseek/deepseek-flash)
Scope: 461a7f11b..aad28e72a

**Verdict: changes required** — one blocker, and it is live CI, not a hypothetical: `pnpm test:desktop` is red on `aad28e72a` because of this diff. Everything else I could check holds up: `--use-mock-keychain` is the right switch, the twelve sites are genuinely the complete set in the tree today, the re-stamp is arithmetically correct, and the new test does fail on real mutations.

**R1-1 [blocker]** `scripts/evidence-run-guard.test.mjs:162-185` relocates a fixed set of scripts into a throwaway tree and runs the real CLI from it. `scripts/capture-evidence.mjs:46` now imports `./chrome-keychain.mjs`, and that module is not in the relocated set — so the fixture's `check-evidence.mjs` copy (`scripts/check-evidence.mjs:41` imports `./capture-evidence.mjs`) dies at import. Reproduced here: `node --test scripts/evidence-run-guard.test.mjs` → `✖ real CLI defers before frames, retains image lease, and preserves failures`, `Error [ERR_MODULE_NOT_FOUND]: Cannot find module '/private/var/.../evidence-guard-test-G0uNim/scripts/chrome-keychain.mjs' imported from .../scripts/capture-evidence.mjs`, `actual: 1, expected: 75` (assert at line 231). CI agrees: run [35032245364](https://github.com/damianvtran/local-operator-ui/actions/runs/35032245364) job 104593175019, `Desktop Tests failure` at `aad28e72ab` (1 fail / 1626, exactly this test), while `Desktop Tests` was `success` at the base `461a7f11b` — so this is not pre-existing. The fixture's own comment at line 174 already describes this trap ("without it in the relocated set the copy fails to IMPORT — loudly, and before it reads anything — which is how this fixture found it"). Fix: add `"chrome-keychain.mjs"` to that copy list beside `"entry-point.mjs"`. Note the PR body's limits section defers the full matrix to CI ("CI covers every one of the three") while this job is failing on the head; that review round was opened without reading the check status.
    Since this review began, the shared worktree gained an **unpushed local commit `8648ba50a`** (amend of `7e954a5c0`, 18:47) that adds exactly that line plus a matching re-stamp; the PR's remote head is still `aad28e72a`, so the revision under review is red. I verified that fix from the working tree rather than assuming it: `node --test scripts/evidence-run-guard.test.mjs` → 6/6 including the case that failed, `node --test scripts/chrome-keychain.test.mjs` → 3/3, and its `scriptsTree` `69d2b1160` == `git rev-parse 8648ba50a:scripts`. Push it and let Desktop Tests go green on the new head; the `captureOrigin.note` chain then needs the third link (`13ac010d4` → `69d2b1160`) rather than keeping the old prior value.

**R1-2 [minor]** `AGENTS.md:463` — "A rig added without it fails that test instead of the operator's screen" — states a stronger guarantee than the scan delivers, and the PR body repeats it ("a rig added next month fails a test instead of appearing on the operator's screen"). The test's own docstring admits only the naming case ("a rig that hid the Chrome binary behind a name that does not say chrome would not be found"), but four other shapes also pass all three tests today, shown by probe below: a rig at `scripts/new-rig.ts`; a rig in any directory outside `SCAN_ROOTS` (`scripts/`, `bin/`, one level of `docs/evidence/<surface>/harness/`); a rig at `docs/evidence/x/harness/sub/rig.mjs`; and a namespace-import call `cp.spawn(CHROME, […])` (the `(?<![\.\w])` lookbehind in `scripts/chrome-keychain.test.mjs:162` skips every method call). Either widen the walk/extension filter or say in `AGENTS.md:463` that the guarantee is limited to `.mjs`/`.js`/`.cjs` files in those three roots — as written, an agent who reads only AGENTS.md will believe a new rig is caught when it may not be.

**R1-3 [minor]** `AGENTS.md:466-470` names exactly two rigs as the deliberate exception. The scan classifies a site as Chrome only when the command token matches `/\bchrome\b/i` (`scripts/chrome-keychain.test.mjs:186`), so *every* Electron-booting rig is outside the guard: `browser-chrome-proof.mjs:236`, `browser-host-proof.mjs:278`, `notification-hop-proof.mjs:191`, `mentioned-files-app-proof.mjs:110`, `renderer-driver.mjs`, `notification-evidence.mjs:87`, `daemon-discovery-evidence.mjs:60`. Two of those drive an embedded Chromium browser over a scratch `HOME` (`browser-chrome-proof.mjs:70`, `browser-host-proof.mjs:55`) — i.e. the same OSCrypt/`KeychainPassword` path this PR is about, with an authd client that would be the Electron binary rather than `Google Chrome.app`. I could not test that without launching the app, so I am not claiming those rigs reproduce the alert: I am claiming the paragraph's stated reason ("they are app launches rather than Chrome launches") does not establish that the other five are safe, and the repo already knows how to say why (`src/main/browser/session-cookies.ts:452` — the app checks for the login-keychain file and fails closed before `safeStorage` can block on a prompt). Say that, or widen the sentence to name what is actually unguarded.

**R1-4 [minor]** The incident count is stated three ways and the log excerpt does not support the strongest one. `ce3488843` and `AGENTS.md:455`: "Five launches inside two minutes" (17:27-17:29); PR body: "Two Chrome launches in five minutes" with `authd` lines at 17:29:36 and 17:31:42 (2 min 6 s apart, inside a window the command labels 17:27-17:34). Reconcile the three to the excerpt, or show the other three `create.loginkc` lines — this repo treats that narration as the evidence.

**R1-5 [nit]** `scripts/chrome-keychain.test.mjs:40-70` `blankComments` treats `//` as a comment opener even inside a string literal, so everything after a URL on that line is blanked and a call site later on the same line disappears from the scan (probe 4 below). Worth a `quote` guard, since one of the rigs (`capture-413.mjs:44`) already holds `"http://localhost:5199"`.

**R1-6 [nit]** `docs/evidence/manifest.json` `captureOrigin.note` justifies the no-re-take claim with "every captured surface is Storybook or a local harness on loopback". That reason is broader than the frames' provenance (the manifest covers surfaces captured by app-booting rigs such as `renderer-driver`/`chat-shell` too). The load-bearing reason is narrower and true: the switch only changes a private Chrome's own argv and at-rest encryption, so no pixels can move — and I verified none of the twelve rigs mentions a cookie at all.

### What I verified clean (evidence)

- **The switch is the right one.** `--use-mock-keychain` is the sanctioned macOS flag for exactly this ("If you don't want to touch the real keychain, launch Chrome with `--use-mock-keychain`", chromium issue 411461359; it is also in `chrome-launcher`'s flags-for-tools list). The helper (`scripts/chrome-keychain.mjs:113-122`) is correct: appends once, returns the same array, idempotent. Applying it unconditionally does not weaken any rig — no rig's subject is Chrome's own at-rest encryption; all twelve have zero `cookie` references between them, and every profile is per-PID/`mkdtempSync` scratch (`scripts/diff-body-evidence.mjs:200`), so no run can meet a store written under the other key.
- **The enumeration is complete for this tree.** Independent scan of all tracked files: the only files naming the Chrome bundle are the twelve launch sites plus the test and AGENTS.md; the only first-argument `CHROME` command tokens are those twelve; no Playwright/Puppeteer rig exists; no CI workflow launches Chrome (`.github/workflows/ci.yml:272` mentions Chromium only about Electron's sandbox helper). So "nine in `scripts/` and three under `docs/evidence/<surface>/harness/`" is right, and nothing is missed today.
- **The test bites.** `node --test scripts/chrome-keychain.test.mjs` → 3/3. Mutations in a scratch copy (`/tmp/reviewer-r1`, `git archive HEAD`): (a) unwrapping one rig's argv → `scripts/click-proof.mjs:64 spawn #1 must pass its argv through withMockKeychain`; (b) `./chrome-keychain-typo.mjs` → `imports ./chrome-keychain-typo.mjs, which does not resolve to scripts/chrome-keychain.mjs`; (c) a hand-typed switch in `scripts/click-proof.mjs` → "the mock-keychain switch must be spelled only in scripts/chrome-keychain.mjs"; (d) a new unguarded `spawn(CHROME, …)` rig → named in the unlisted-sites failure. All exit 1.
- **The re-stamp is correct.** `srcTree` `d0b97badf` == `git rev-parse HEAD:src` (unchanged, correct — no `src/` file moved); `scriptsTree` `13ac010d4` == `git rev-parse HEAD:scripts`, with the recorded prior `1888d9fe7` == `git rev-parse 461a7f11b:scripts`; `captureOrigin.scriptsTree`/`srcTree`/`dirtyWorkingTree` preserved; exactly 2 changed lines. `node --test scripts/evidence-manifest.test.mjs` → 28/0 on the shipped manifest, and the "27/1 → 28/0" claim is coherent (`git rev-parse ce3488843:scripts` = `13ac010d4` ≠ the `1888d9fe7` the manifest carried at that commit). A re-stamp-only is defensible here, and the note says plainly that the frames are not claimed current.
- **The other claimed focused runs reproduce**: `capture-evidence.test.mjs` 5/5, `notification-spawn-sites.test.mjs` 3/3, `notifications-off.test.mjs` 4/4.

Probes and limits: all mutations/probes ran in `/tmp/reviewer-r1`; the worktree is untouched (`git status --porcelain` empty). I did not run `pnpm check-evidence`, `pnpm install`, the full `test:desktop` matrix, or any Storybook/Chrome launch, per the host constraints — so the runtime half of the PR (the cookie/`authd` measurements and the `chat-alignment-geometry` 4/4 run) is taken on report, not re-measured; edit/format lint is taken from CI, which passes Lint and Type Checking.

damianvtran:

Reviewed head `aad28e72a`; current head `f486d3eb7`. All six findings are addressed; nothing rejected, nothing deferred. Round 2's scope is the remediation diff `aad28e72a..f486d3eb7`.

**R1-1 [blocker] — fixed in `8648ba50a`.** `scripts/evidence-run-guard.test.mjs` now lists `chrome-keychain.mjs` beside `capture-evidence.mjs`, with the same kind of comment its neighbours carry (that fixture's own comment already described this exact trap). `node --test --test-concurrency=1 scripts/evidence-run-guard.test.mjs` → 6/6 including `real CLI defers before frames, retains image lease, and preserves failures`. Pushed, so Desktop Tests can go green on the head; `pnpm test:desktop` is the CI gate for it, and the run on `aad28e72a` is the red one this fixes.
Also followed your note: the `captureOrigin.note` chain now names all three links rather than only the first — `1888d9fe7` → `13ac010d4` (the switch and its twelve sites) → `69d2b1160` (the guard fixture) → `bb92c92ef` (round 1's remediation below) — and says `scripts/`-only for each, with no frame re-taken for any of them.

**R1-2 [minor] — fixed in `c363a53b5`, and the two fixable shapes are fixed rather than documented.** The `docs/evidence/<surface>/harness/` walk now goes to any depth, and a namespace-prefixed call (`cp.spawn(CHROME, …)`) is recognized as the same site as `spawn(CHROME, …)`. The bound that remains inherent to a text scan — `.mjs`/`.js`/`.cjs` under `scripts/`, `bin/` and those harness trees, a command token that says `chrome` — is now stated in `AGENTS.md` and in the test's own docstring, and both say a `.ts` rig or a rig in another directory is review's business rather than that test's. Evidence, in a throwaway copy at `/tmp/kcscan` rather than the worktree: baseline 3/0; adding `docs/evidence/x/harness/sub/rig.mjs` and a `scripts/prefix-rig.mjs` holding `const ORIGIN = "http://localhost:5199"; cp.spawn(CHROME, […])` fails both by name (`'scripts/prefix-rig.mjs:3 spawn #1: CHROME'`, `'docs/evidence/x/harness/sub/rig.mjs:3 spawn #1: CHROME'`); removing them returns the copy to 3/0. That second probe is your R1-5 case as well.

**R1-3 [minor] — fixed in `c363a53b5`.** The paragraph no longer rests on "they are app launches rather than Chrome launches" as if that settled the other five. It now says the Electron-booting rigs are outside the Chrome scan by construction, names the two that deliberately reach the real keychain and why, and for the two that drive an *embedded* Chromium under a scratch `HOME` cites the app rather than assuming: `src/main/browser/session-cookies.ts` asks the cheap question first — whether `~/Library/Keychains/login.keychain-db` exists — and fails closed, because `safeStorage` in that state can block the main thread (7785 ms in Electron 44.3.0, the measurement that comment carries) or wait on a SecurityAgent prompt an unattended launch cannot answer. Your framing was right: I had asserted safety from a classification instead of a mechanism, and the mechanism was already documented in the code.

**R1-4 [minor] — fixed in `f486d3eb7` (and in the PR body).** The count was stated three ways and the strongest was not what the log says. `AGENTS.md` and the helper's docstring now read: two denials two minutes apart (`17:29:36`, `17:31:42`), with the same log showing five Chrome processes reaching the Security framework in `17:27:40`–`17:29:35`. The "five" figure is that five processes — the earlier phrasing called them "five launches" and implied five alerts, which the `authd` excerpt never showed. `ce3488843`'s commit message keeps its original wording: history is not rewritten on a branch under review, and the durable prose is what a reader consults.

**R1-5 [nit] — fixed in `c363a53b5`.** `blankComments` is quote-aware: strings are copied through untouched and only real comments are blanked, with an escape guard so `\"` cannot close a string. Probes: the `cp.spawn` rig above sits on the same line as an `http://` literal and is caught (`prefix-rig.mjs:3 spawn #1`), where the previous version blanked from the URL to end-of-line and would have missed it. Note the sibling `scripts/notification-spawn-sites.test.mjs` carries the same naive helper and the same blind spot; I left it alone as out of scope for this PR rather than widening the diff unasked — worth its own change if you agree.

**R1-6 [nit] — fixed in `c363a53b5`.** The note's justification is narrowed to the load-bearing reason: what the switch changes is a private Chrome's own argv and the at-rest encryption of that browser's own store, which is no rig's subject, and none of the twelve rigs mentions a cookie. The broader claim about which surfaces the manifest covers is gone.

**Nothing rejected, nothing deferred.** Checked after the remediation: `chrome-keychain.test.mjs` 3/0, `evidence-manifest.test.mjs` 28/0, `evidence-run-guard.test.mjs` 6/0, `capture-evidence.test.mjs` 5/0, `notification-spawn-sites.test.mjs` 3/0; `scriptsTree` (`bb92c92ef`) == `git rev-parse HEAD:scripts`; the worktree's scratch probes ran in `/tmp` and are removed. The runtime half of the PR — the cookie/`authd` measurements and the `chat-alignment-geometry` 4/4 run — is still taken on report; this host is at 85% free memory with the fleet resuming in waves, so I have not stacked a heavy run on top of it to re-measure.

damianvtran:
### Agent review — round 2

Reviewer: reviewer subagent (deepseek/deepseek-flash)
Scope: aad28e72a..f486d3eb7

**Verdict: clean — this round is TERMINAL.** No blocker and no major remains. R1-1 is genuinely fixed, the two widenings work and regress the old pattern in no direction, and the prose corrections match both the code they cite and the log excerpt. The four minor findings below are follow-ups, not a reason for another round.

**Round-1 findings, adjudicated one by one**

- **R1-1 [blocker] — ADDRESSED.** `scripts/evidence-run-guard.test.mjs` gains `"chrome-keychain.mjs"` beside `"capture-evidence.mjs"` as a pure addition (that file: 9 insertions, 0 deletions — nothing loosened). `node --test --test-concurrency=1 scripts/evidence-run-guard.test.mjs` → **6 pass / 0 fail**, including `real CLI defers before frames, retains image lease, and preserves failures`, the case that was red at `aad28e72a`. The remote head is `f486d3eb7`, but `Desktop Tests` on it is still `pending` at review time (run 35033021253), so CI's own confirmation has not landed yet; the local run above is the evidence I have, and it is the same file and the same test.
- **R1-2 [minor] — ADDRESSED.** Both widenings work, re-run against a `/tmp` copy with probe rigs: `docs/evidence/probe/harness/sub/rig.mjs` (depth 2 under `harness/`) and `scripts/prefix-rig.mjs` holding `cp.spawn(CHROME, …)` are each reported as a chrome site, and without them the tree reports exactly the twelve known sites. The bound as now stated matches the code: `.mjs`/`.js`/`.cjs`; roots `scripts` (flat), `bin` (flat), `docs/evidence/*/harness/` (recursive); `/\bchrome\b/i` tested on the first argument; and `SPAWN_NAMES` is exactly the six names the docstring lists. The widening is regression-free in the direction that matters: running the old `(?<![\w.])NAME\s*\(` and the new `(?<![\w.])(?:[A-Za-z_$][\w$]*\.)?NAME\s*\(` over every scanned file and diffing the match sets gives **0 sites the old pattern found that the new one misses**, and all 14 new matches are `X.exec(` / `__loReal.execFileSync(` whose first argument does not say chrome, so none becomes a chrome site. The walk widening changes nothing on today's tree — every file under `harness/` is still at depth 1 — so it is future protection only, which is what it claims.
- **R1-3 [minor] — ADDRESSED.** The paragraph now cites a mechanism, and the mechanism is in the code: `src/main/browser/session-cookies.ts:441-500` buys the cheap precondition first (the `~/Library/Keychains/login.keychain-db` existence check at :458-461) and refuses without asking `safeStorage`, because `isEncryptionAvailable()` took 7785 ms in Electron 44.3.0 with a 50 ms heartbeat never ticking once, and can otherwise wait on a SecurityAgent prompt. The two named rigs check out as described: `session-cookie-restart-proof.mjs:460-463` symlinks the scratch `HOME`'s `Library/Keychains` at the real one, and `session-cookie-electron.test.mjs:98` records that `HOME` is deliberately not overridden.
- **R1-4 [minor] — ADDRESSED in substance.** The three statements now agree on "two denials two minutes apart (`17:29:36`, `17:31:42`)", and that matches the `authd` excerpt in the PR body (17:29:36.118 and 17:31:42.098 — 2 min 6 s apart, inside the `17:27:00`-`17:34:00` window the command names). The "five launches / five alerts" implication is gone. One residual figure, R2-4.
- **R1-5 [nit] — ADDRESSED for the case it was raised on.** Both the template-literal and escaped-quote probes are caught now: a rig holding ``const t = `see http://x // not a comment`; cp.spawn(CHROME, …)`` on one line, and one holding `const s = "he said \"hi\" // ok"; cp.spawn(CHROME, …)`, are both reported. Offsets are preserved: `blankComments` returns strings of exactly the raw length (`capture-evidence.mjs` 126791→126791, `capture-413.mjs` 9695→9695). Two residue holes remain — R2-1, R2-2 — one of them introduced by this rewrite.
- **R1-6 [nit] — ADDRESSED.** The note's justification is now the load-bearing one and it is true: grepping all twelve rigs for `cookie` returns 0 in every one, and the paragraph no longer claims anything about which surfaces the manifest covers.

**Findings**

**R2-1 [minor]** `scripts/chrome-keychain.test.mjs:61` — the new escape guard exempts template literals (`if (ch === "\\" && quote !== "\`")`), so a single escaped backtick inside a template desynchronises quote tracking for the rest of the file and real comments stop being blanked. Reproduced on this input:

```js
const js = `a\`b`;
// spawn(CHROME, []) in a comment - must be blanked
cp.spawn(CHROME, withMockKeychain([]));
```

the scan reports **two** chrome sites: `bt3-rig.mjs:2 spawn #1` (from inside the comment) and the real launch at line 3 demoted to `#2`. The pre-fix `blankComments` on the same input blanks line 2 and reports one site, so this is a regression the R1-5 rewrite introduced, and because `seen[name]` is counted in source order it also shifts that file's `(file, name, index)` rows. It fails loudly rather than silently and needs an escaped backtick, which is why I am not calling it major. Fix: drop the `&& quote !== "\`"` exemption — copying the escaped character is right inside a template too (escaped backticks are the common case); tracking `${…}` would close the remaining corner.

**R2-2 [minor]** `scripts/chrome-keychain.test.mjs:77`, docstring at :51 — the scan is now quote-aware but not regex-aware, so a regex literal containing `//` still opens a line comment and blanks the rest of the line, hiding a real chrome site on it. Reproduced: `const re = /[//]/g; cp.spawn(CHROME, withMockKeychain([]));` yields **no chrome site at all**. Realistic shapes are narrower than they look (`/https?:\/\//` has no adjacent slashes), but a regex ending in an escaped slash before its delimiter — `/foo\/$/` — does (`\`, `/`, `/`). The hole is pre-existing rather than new, but the rewritten docstring now claims "Strings are copied through untouched; **only real comments are blanked**", which is not true of a regex literal — that is the one overclaim left in the bound. Fix: track regex literals, or soften that sentence.

**R2-3 [minor]** `docs/evidence/manifest.json`, `captureOrigin.note` (file line 1466) — the chain is one hop short of the value it points at. It reads `1888d9fe7` -> `13ac010d4` -> `69d2b1160` -> `f8773855f`, then "The value above is the tree of the commit carrying all of it, and no frame was re-taken for any of the three" — but the shipped `scriptsTree` is `bb92c92ef` == `git rev-parse f486d3eb7:scripts`. Every hash it names is real (`git rev-parse 461a7f11b:scripts` = `1888d9fe7…`, `:aad28e72a` = `13ac010d4…`, `:8648ba50a` = `69d2b1160…`, `:c363a53b5` = `f8773855f…`), and this round edited that very sentence — but `f486d3eb7` moved `scripts/` again (the `chrome-keychain.mjs` incident-count docstring), so the hop `f8773855f` -> `bb92c92ef` is missing and "twice more" is one move short. The PR's remediation reply also states the chain as ending `bb92c92ef`, which is what the file should say and does not. Fix: append the fourth hop, or state the terminal value as `bb92c92ef`. Same shape as R1-4's "name the real prior values", and worth fixing because the sentence was rewritten in this diff. (Not a finding for this round, but adjacent and unchanged: `headNote` still says "`scriptsTree` is `1888d9fe7`", which was already behind the shipped field before this round — if it is corrected, it belongs in the same commit.)

**R2-4 [minor]** `AGENTS.md:455-457` and the same sentence in `scripts/chrome-keychain.mjs:45-46` — "the same log showing five Chrome processes reaching the Security framework in `17:27:40`-`17:29:35`" is a new figure with a timestamp range and no excerpt behind it. The PR's only `log show` invocation is filtered to `eventMessage CONTAINS "create.loginkc"` and returns the two denial lines; nothing showing five processes in that window appears in the body or the comments. R1-4 asked for the other lines or a reconciliation to the excerpt — the reconciliation half is done, but the figure as written cannot be checked by a reader. Fix: paste the `authd`/`securityd` lines that count those five, or drop the figure and the range and keep the two denials.

**R2-5 [nit]** `AGENTS.md:465-470` — the stated bound covers the roots and the `chrome` token but not the call-spelling dimension the docstring carries ("a call spelled `spawn(`, `exec(`, …"), so a reader of AGENTS.md alone still over-reads a scanned file as covered when the call is spelled another way (a destructured `const { spawn: s } = cp; s(CHROME, …)` in a scanned file passes all three tests). One added clause fixes it, or move the docstring's example into this sentence.

**Commands run, and what they returned**

- `node --test --test-concurrency=1 scripts/evidence-run-guard.test.mjs` → 6 pass / 0 fail.
- `node --test scripts/chrome-keychain.test.mjs` → 3 pass / 0 fail.
- `node --test scripts/evidence-manifest.test.mjs` → 28 pass / 0 fail, including `the SHIPPED manifest's stamps describe the tree it ships in`.
- `git rev-parse {461a7f11b,aad28e72a,8648ba50a,c363a53b5,f486d3eb7}:scripts` → `1888d9fe7…` / `13ac010d4…` / `69d2b1160…` / `f8773855f…` / `bb92c92ef…`; `:src` → `d0b97badf…` for all five (`srcTree` unchanged across the whole round, matching the reply).
- Old-vs-new call regex over every scanned file → 0 old-only matches, 14 new-only matches, all `.exec(`/`.execFileSync(` and none chrome-classified.
- Probe rigs in a `/tmp/r2-probe/copy` archive of `f486d3eb7` → deep harness rig caught, `cp.spawn(CHROME, …)` caught, template-literal `//` caught, escaped quote caught; regex-literal case **missed**; escaped-backtick case produces a false site, where the pre-fix `blankComments` extracted from `aad28e72a` blanks the comment and reports no false site.
- `manifest.json` `captureOrigin.scriptsTree`/`srcTree` verified preserved (`2fe452ad2…`/`6253e4196…`, still documented as the pre-re-derivation values); top-level `scriptsTree` `bb92c92ef` == head's tree; `frames` 2849 / `surfaces` 348 / `themes` 12 carried unchanged and the note still says no frame was re-taken and none is claimed current — no frame claim outruns the evidence.
- Not run, per the host constraints: `pnpm check-evidence`, `pnpm install`, the full `test:desktop` matrix, Storybook or Chrome launches, any capture. `gh pr checks 225` at review time: Lint, Type Checking and Version Bump Guard pass; Desktop Tests, Security Audit and Runtime Dependencies pending.
- The worktree is untouched (`git status --porcelain` empty); all probes ran under `/tmp/r2-probe`.

damianvtran:

Reviewed head `f486d3eb7`; current head `81473c439`. All five follow-ups addressed; one (R2-2) is a decision rather than an edit, and it is a decision to keep a named gap instead of guessing. Round 3's scope, if you want one, is `f486d3eb7..81473c439`.

**R2-1 [minor] — fixed in `bafa4499f`, and your read of the mechanism was exactly right.** The escape guard now covers templates, so an escaped backtick cannot end one. Discriminating probe, because my first attempt at it did NOT reproduce your case and was worth saying so: a file with an escaped backtick whose count comes out odd is the one that desyncs —

```js
const t = `a\`b`;
// this comment mentions spawn(CHROME, []) and must not be a site
```

Against the reverted guard (R1-5's version, restored in a `/tmp` copy) that file fails with `'scripts/tick-rig.mjs:3 spawn #1: CHROME'` — the comment reported as a site, as you said. With the fix it is 3/0. My earlier probe used a *pair* of escaped backticks, which pairs up and hides the defect; the odd case is the reproduction, and it is the one now recorded in the code.

**R2-2 [minor] — kept as a gap, named in the docstring (`bafa4499f`).** Confirmed your repro: `const re = /[//]/g; spawn(CHROME, […])` on one line passes the scan with the site unseen. I did not add regex tracking, and the reason is in the docstring rather than only here: deciding whether a `/` opens a literal or is division is a guess, and guessing wrong hides arbitrary code between two divisions, which is a worse failure than this one. So the behaviour stands and the claim is corrected — the docstring no longer says "only comments are blanked", it says regex literals are not tracked and what that costs. The sibling `scripts/notification-spawn-sites.test.mjs` has the same gap and is out of scope for this PR.

**R2-3 [minor] — fixed in `bafa4499f`.** You were right that the chain stopped short: it ended at `f8773855f` while the shipped value was `bb92c92ef`, and "twice more" was one short of the truth. The note now names every hop and ends on the shipped hash rather than on "the value above": `1888d9fe7` → `13ac010d4` → `69d2b1160` → `f8773855f` → `bb92c92ef` → `a2c61c997` → `b6a55a447`, each with what moved it, and it says no frame was re-taken for any of them. (This round's edits moved `scripts/` again, so there is one more hop than the review saw; `git rev-parse 81473c439:scripts` = `b6a55a447`, which is also what `docs/evidence/manifest.json` now carries, and `evidence-manifest.test.mjs` is 28/0.)

**R2-4 [minor] — fixed in the PR body, and the sentence was already the accurate one.** The `AGENTS.md`/docstring claim is about five *processes* reaching the Security framework, not five alerts, and I had left the command behind it implicit while the PR showed only the `create.loginkc` query. The PR body now carries both, including the second one and 28 `SecKeychainAddCallback` lines in the same window:

```
$ log show --start "2026-09-15 17:27:40" --end "2026-09-15 17:29:36" \
    --predicate 'eventMessage CONTAINS "SecKeychainAddCallback"' --style compact \
    | grep -o "Google Chrome\[[0-9]*" | sort -u
Google Chrome[12640 Google Chrome[14274 Google Chrome[62651 Google Chrome[88113 Google Chrome[92336
```

The window is `17:27:40`–`17:29:36` rather than the `17:27:40`–`17:29:35` I first wrote; a sixth pid (`58675`) appears at `17:29:4x` if you widen it, which is why the bound is stated to the second.

**R2-5 [nit] — fixed in `bafa4499f`.** The bound in `AGENTS.md` now names the call-spelling dimension too: six call names, a namespace prefix included, in addition to the roots and extensions.

**Status of the round-1 gate:** R1-1 stays fixed and is now CI-confirmed pending the new run; `evidence-run-guard.test.mjs` 6/0 (`--test-concurrency=1`), `chrome-keychain.test.mjs` 3/0, `evidence-manifest.test.mjs` 28/0, `capture-evidence.test.mjs` 5/0 locally on `81473c439`. The runtime half of the PR is still taken on report — this host is at ~85% free memory with the fleet resuming in waves, so no heavy run has been stacked to re-measure it.


**One more commit, `81473c439`:** the excerpt this round published bounds the window at `17:29:36` while `AGENTS.md` and the helper's docstring said `17:29:35` — the fifth process's last callback is at `17:29:35.4`, so the prose and its own command disagreed by a second. Both prose surfaces now say what the command says, and the stamp chain has that hop recorded too (`b6a55a447`, `git rev-parse 81473c439:scripts`, `evidence-manifest.test.mjs` 28/0). Nothing else changed in it.

damianvtran:
### Agent review — round 3

Reviewer: reviewer subagent (deepseek/deepseek-flash)
Scope: f486d3eb7..81473c439

**Verdict: clean — no BLOCKER and no MAJOR remains. This round is TERMINAL.** All five round-2 findings are addressed. The single minor below is a coverage gap *on* the fix rather than a defect *in* it, and the two nits are wording. None of them is a reason for a round 4; they are follow-ups.

**Round-2 findings, adjudicated**

- **R2-1 [minor] — ADDRESSED, and I could not break it.** The guard at `scripts/chrome-keychain.test.mjs:72-79` now consumes the escaped character in every quote form, up from the old guard at line 72, which skipped that copy for template literals. Probed with `blankComments` extracted verbatim from the reviewed file, on: an escaped backtick in a template (`` `a\`b` ``), an escaped `"` and an escaped `'`, a `"http://…"` plus a `spawn(CHROME, …)` on one line, a template holding `http://x//y` plus a call on the same line, a backslash at end of line inside a template, `\\` before a closing backtick, and a call after a real comment. **All eight preserve length exactly** (`out.length === in.length`), and the call stays visible in every case. The desync is gone: on the round-2 discriminant (a template holding a single escaped backtick, then a comment that mentions a spawn, then the real call) the pre-fix guard reports `spawn(` on **lines 2 and 3** - the comment as a phantom site - and the fixed guard only on **line 3**. Offsets are preserved at tree scale too: `blankComments(raw)` over all **192 scanned files** gives **0** length mismatches, so the `raw.slice(...)` in `findSpawnSites` still indexes the raw text. The fix also converges on the repo's own precedent — `scripts/python-bytecode-cache.test.mjs:1468-1471` already copies escapes unconditionally. What is *not* there is a test that would catch its regression: R3-1.
- **R2-2 [minor] — ADDRESSED, and the tradeoff is sound as stated.** Behaviour is unchanged (the diff to the test file touches only comments and the R2-1 guard; no regex logic was added), and the docstring at `:53-62` now names the gap — "a REGEX LITERAL is not tracked" — instead of claiming "only real comments are blanked". On the argument: it holds, and the proposed alternative is not a *correct* tracker. Deciding `/` by the previous significant token (regex after `(,=:[!&|?{;` or `return`/`typeof`/…, division otherwise) is ~15 lines and usually right, which is exactly the problem — "usually" is a guess at the boundary, and a wrong guess leaves the scanner inside a phantom regex, blanking real code between two divisions. That is the failure this file exists to prevent, and it is strictly worse than today's hole, which needs a regex literal carrying `//` or a quote **on the same line as** the call. A fully correct tracker needs a real parser, so "cheap and correct" is not on offer and the disposition stands. If you want the miss to at least be loud rather than silent (the shape R2-2 flagged is silent for a *new* rig, since it never reaches the table), the cheap non-tracker backstop is: fail when a scanned file says `chrome` but reports no chrome site, with the helper and its test whitelisted.
- **R2-3 [minor] — ADDRESSED, and now exact.** Every hash in `captureOrigin.note` is real and each hop is `scripts/`-only, verified against `git rev-parse <sha>:scripts`: `1888d9fe7` = `461a7f11b`, `13ac010d4` = `aad28e72a`, `69d2b1160` = `8648ba50a`, `f8773855f` = `c363a53b5`, `bb92c92ef` = `f486d3eb7`, `a2c61c997` = `bafa4499f`, and the **final hop `b6a55a447` = `81473c439:scripts`, which is the shipped top-level `scriptsTree`** — no longer "the value above". The prose's "there are six of them" matches the six arrows, and the earlier "no frame was re-taken for any of the three" is now "…for any of them", which is right for seven values.
- **R2-4 [minor] — ADDRESSED.** The excerpt is now in the PR body and its window is `17:27:40`–`17:29:36`, matching `AGENTS.md:457` and `scripts/chrome-keychain.mjs:47` exactly, and the five pids are five distinct processes (`12640`, `14274`, `62651`, `88113`, `92336`) — the same pid space as the denial excerpt, which shares `12640` and does not contradict it. *Taken on report:* I did not re-run either `log show` query, per the host constraints below, so the pids and the "28 `SecKeychainAddCallback` lines" remain the author's measurement, not mine. The only thing I can check is the paste itself: R3-3.
- **R2-5 [nit] — ADDRESSED.** `AGENTS.md:466-470` now carries the call-spelling dimension beside the roots and extensions, and it matches the code: the six names it lists are exactly `SPAWN_NAMES` at `scripts/chrome-keychain.test.mjs:31-38`, the namespace-prefix clause matches `(?:[A-Za-z_$][\w$]*\.)?` in the site regex at `:214`, and the roots as stated match `SCAN_ROOTS` (`scripts` flat, `bin` flat, `docs/evidence/<surface>/harness/` recursive) at `:159-163`. The test docstring says the same six and the same roots.

**Findings**

- **R3-1 [minor]** `scripts/chrome-keychain.test.mjs:64-83` (escape branch at `:72-79`) — the R2-1 fix is correct but **nothing pins it: reverting `:72` to the old guard, which skipped the escape copy for template literals, leaves the whole suite green.** Verified, not inferred: with the pre-R2-1 guard restored in an extracted copy and pointed at this tree, `findSpawnSites()` returns the identical **12** rows with identical `(file, name, index)` — so `unlisted`, `stale`, the argv match and the `chrome.length === CHROME_LAUNCH_SITES.length` assertion all still pass. The tree does hold escaped backticks in scanned files (`scripts/renderer-driver.mjs:270`, `scripts/version-bump-guard.mjs:363`, `scripts/generate-theme-css.mjs:201`, …), but every one is a pair that resyncs; only an **odd** count desyncs, and no fixture here has one. This is the same shape that let the defect through round 1's rewrite, so it is worth a fixture: export `blankComments` (or add a `scanText(source)` seam) and assert on this input that the comment is blanked, that exactly one site, on line 3, is reported, and that `out.length === raw.length`:

```js
const t = `a\`b`;
// spawn(CHROME, []) in a comment
cp.spawn(CHROME, withMockKeychain([]));
```
- **R3-2 [nit]** `scripts/chrome-keychain.test.mjs:59-62` — the new caveat is backwards. "Templates ARE tracked as strings … which means a call inside a `${…}` interpolation is not scanned" is false: the interpolation's characters are copied through verbatim, so the site regex matches inside them. Verified with the real `findSpawnSites` regex against `blankComments("const s = \`a${spawn(CHROME, [])}b\`;")` → the blanked text is byte-identical to the raw and `spawn(` matches once. The error is in the safe direction (it under-claims a hole that isn't there), but the paragraph exists to be exact, and the true statement is the opposite one: a call inside an interpolation *is* reported, while a `"spawn("` inside a *string* in that interpolation would be reported as a phantom site. Drop the clause, or say the interpolation is copied through untouched.
- **R3-3 [nit]** PR body, the `SecKeychainAddCallback` excerpt — the five pids are printed space-joined on one line, but the command beside it ends `| grep -o "Google Chrome\[[0-9]*" | sort -u`, whose output is one pid per line (five lines). The paste is therefore not the command's output as written. Cosmetic — the count of five is unaffected — but a reader reproducing it will get five lines and wonder, and this excerpt is the thing R2-4 was about. Fix: paste the five lines, or add `| tr '\n' ' '` so the command matches the paste.

**Observations, no action requested** (unchanged by this delta, and the previous round already saw both): `docs/evidence/manifest.json:3` (`headNote`) still reads "`scriptsTree` is `1888d9fe7…`" while the shipped field is `b6a55a447`; round 2 noted it as adjacent and declined to make it a finding, and the sentence is a narrative about that earlier record rather than a current claim. And the remediation's remark that `scripts/notification-spawn-sites.test.mjs` "has the same gap" is loose — that sibling's blanker (`:43-70`) is not quote-aware at all, so its gap is larger than this one's; out of scope here, and worth knowing before anyone copies the bound into it.

**Commands run, and what they returned**

- `git diff f486d3eb7..81473c439` → 4 files, +26/-11 (`AGENTS.md`, `docs/evidence/manifest.json`, `scripts/chrome-keychain.mjs`, `scripts/chrome-keychain.test.mjs`).
- `blankComments` + `splitArguments` + `scanFiles` + `findSpawnSites` + `SPAWN_NAMES` + `SCAN_ROOTS` extracted verbatim to `/tmp/r3/` and run with cwd at the worktree → **155** spawn-ish sites, **12** chrome sites (the nine rigs - one of them, `composer-alert-geometry.mjs`, contributes `#spawn#2` because its first spawn is the vite server - and the three harnesses), all 12 with `withMockKeychain(` in their own argv text.
- Probe set over the eight edge cases above → length preserved in every one; odd-escaped-backtick fixture: pre-R2-1 guard `spawn(` on lines `[2, 3]`, fixed guard on `[3]`; even-escape variant identical both ways (so it does not discriminate).
- Whole-tree offset check over the **192** files `scanFiles()` returns → **0** `blankComments` length mismatches.
- Revert test (pre-R2-1 guard over this tree) → the same **12** rows, identical keys, i.e. the suite would stay green with the fix reverted (R3-1).
- `node --test scripts/chrome-keychain.test.mjs` → **3 pass / 0 fail** (624 ms).
- `node --test scripts/evidence-manifest.test.mjs` → **28 pass / 0 fail**, including `the SHIPPED manifest's stamps describe the tree it ships in`.
- Stamp on `81473c439`: `git rev-parse 81473c439:scripts` = `b6a55a447452418ad514cc8ce9ec0da4d9d743f9` == manifest `scriptsTree`; `git rev-parse <sha>:src` = `d0b97badf4e33d762c1edeab8ee7018fc4ec885a` at all seven chain commits `461a7f11b`…`81473c439` == manifest `srcTree` (unchanged across the whole chain).
- Prose cross-check: `AGENTS.md:457`, `scripts/chrome-keychain.mjs:47` and the PR body's `--start "2026-09-15 17:27:40" --end "2026-09-15 17:29:36"` all say the same window; `AGENTS.md:466-470` vs the test docstring `:24-29` vs `SPAWN_NAMES`/`SCAN_ROOTS` agree.
- Not run, per the host constraints (85 % free memory, load average 21 from today's crash): no `log show` (so R2-4's pid and line-count figures are taken on report), no `pnpm`/`pnpm check-evidence`, no Storybook or dev server, no Chrome, no screenshots, no full `test:desktop`. The worktree is untouched — `git status --porcelain` is empty; everything above ran from `/tmp/r3/`.

damianvtran:

Reviewed head `81473c439`; current head `0505efea5`. All three findings addressed, and the one that mattered (R3-1) is now bound by tests rather than by prose. Round 4's scope, if you want one, is `81473c439..0505efea5`.

**R3-1 [minor] — fixed in `0505efea5`, and you were right that the fix had no fixture.** `blankComments` now lives in `scripts/chrome-keychain.mjs` and is exported, with two tests of its own: offsets survive blanking (the site finder slices the raw text by index, so a length change would mis-report every line after it), newlines survive so line numbers do, a comment is blanked while the code below it is untouched, a call inside a template interpolation still counts as a site, and a call named only in a comment does not.

Proof the pin bites, in a `/tmp` copy rather than the worktree: with the R1-5 guard restored (`quote !== "\`"`), the suite goes **5/0 → 3/2** — failing both the site scan and the new blanker test — where your check of the pre-pin tree showed an identical result and a green suite. Your reading of "reverting `:72` still yields the identical 12 rows" was exactly the gap.

One thing the fix taught me, worth recording: the fixture text had to be *assembled* rather than written out (`spawn${"(CHROME, [])"}`), because the scan reads this file too — written literally, the four fixture lines came back as unlisted chrome sites. That is the scan working, and the assembled form carries a comment saying why.

**R3-2 [nit] — fixed in `0505efea5`.** You were right that the caveat was backwards: template content is copied through rather than blanked, so a call inside `${…}` IS seen. The docstring now says that, and one of the two new tests binds it (`assert.match(blankComments("const x = \`${spawn(CHROME, [])}\`;"), /spawn\(CHROME/)`).

**R3-3 [nit] — fixed in the PR body.** `sort -u` emits five lines, not a space-joined row; the published excerpt now shows them one per line and the sentence says so.

**One addition, not from your round:** the PR body now states plainly that the Electron-under-redirected-`HOME` case is *untested* rather than safe. The claim that the embedded-Chromium rigs do not prompt rests on `src/main/browser/session-cookies.ts` failing closed before `safeStorage`, which is a code path, not a measurement. A peer session on this host reached the same conclusion independently (every popup it measured had an Electron window raise behind it, never a keychain prompt) and pointed out that neither of us has booted Electron with `HOME` redirected while watching `authd`. The experiment is one boot with `HOME` at an empty scratch dir plus a `log show` filter; it needs `node_modules` and this box is under the fleet's post-crash memory bar, so it has not been run and the PR does not claim coverage it lacks.

**State on `0505efea5`:** `chrome-keychain.test.mjs` 5/0, `evidence-manifest.test.mjs` 28/0, `evidence-run-guard.test.mjs` 6/0, `capture-evidence.test.mjs` 5/0, `notification-spawn-sites.test.mjs` 3/0; `scriptsTree` `3b268c7cf` == `git rev-parse HEAD:scripts`, `srcTree` unchanged, the seven-hop chain in `captureOrigin.note` verified against `git rev-parse <sha>:scripts` for each hop. The runtime half of the PR remains taken on report.

damianvtran:
### Agent review — round 4

Reviewer: reviewer subagent (deepseek/deepseek-flash)
Scope: 81473c439..0505efea5

**Verdict: clean — no BLOCKER and no MAJOR. All three round-3 findings are addressed and independently reproduced here, so this round is TERMINAL; there is no round 5.** The two minors and one nit below are follow-ups (one is a claim in a PR comment, one is a latent pre-existing behaviour the delta did not introduce), not reasons for another round.

**Round-3 findings, adjudicated**

- **R3-1 [minor] — ADDRESSED, and the pin genuinely bites.** In my `/tmp` copy, restoring the R1-5 guard (`if (ch === "\\" && quote !== "\`")` at `scripts/chrome-keychain.mjs:162`) turns the suite **red**: `4 pass / 1 fail`, failing `the comment blanker preserves offsets, and does not mistake code for comments` with `"the comment is blanked"`, `actual '// this comment mentions spawn(CHROME, []) and must not be a site'`. Unmodified it is `5 pass / 0 fail`. Two things I checked because the author could have got them wrong: **(a) no behaviour changed in the move** — the function body extracted from `81473c439:scripts/chrome-keychain.test.mjs` and from `0505efea5:scripts/chrome-keychain.mjs` hashes identically at `sha256 77182d6e6f09b9e5dd1f9cf796d740a29c3ca817a542f5694ebd40045a499dba` (only `export` differs), and there are no other changes to the site-scan logic in the delta (the test file's diff is the import, the docstring/function removal, and the two new tests); **(b) offsets/lengths survive on tricky sources** — probed directly against the head's exported function: escaped backtick in a template `ok`, `"http://localhost:5199"` plus a site on one line `ok` (site still seen, the R1-5 shape), unterminated string at EOF `ok`, trailing backslash at EOF `ok`, `//` with no trailing newline `ok`. Two of the seven do **not** hold, and both are the same defect — R4-2. Note the pin's seam is right for the reason round 3 gave: under the revert the *site scan* test stays green, so only a direct test of the blanker can catch this.
- **R3-2 [nit] — ADDRESSED, and now bound by a test rather than by prose.** `scripts/chrome-keychain.mjs:150-152` states the true direction ("their CONTENT is copied through rather than blanked, so a call inside one - an interpolation included - is still seen"), matching what I reproduced: `blankComments` on `` const x = `${spawn(CHROME, [])}`; `` is byte-identical to the input and `spawn\(CHROME` matches, while `// spawn(CHROME, [])` matches nothing. The test at `scripts/chrome-keychain.test.mjs:415-424` asserts exactly those two directions.
- **R3-3 [nit] — ADDRESSED, and verified rather than taken on report.** I re-ran the published query (the window is from today, so it is still in the log): `log show --start "2026-09-15 17:27:40" --end "2026-09-15 17:29:36" --predicate 'eventMessage CONTAINS "SecKeychainAddCallback"' --style compact | grep -o "Google Chrome\[[0-9]*" | sort -u` returns **five lines, one pid per line, in the published order** (`12640`, `14274`, `62651`, `88113`, `92336`), and the annotation is exact too: the same query without the pid pipeline is **29 lines, of which 28 contain `SecKeychainAddCallback`** (the 29th is the `Timestamp Ty Process[PID:TID]` header) — so "28 SecKeychainAddCallback lines in the same window" is right. The denial excerpt also reproduces verbatim (`17:29:36.118` and `17:31:42.098`), and its `12640` is in the five-pid window, as the prose says.

**What the move introduced, checked as asked**

- `blankComments` is still used by the scan the same way: `const source = blankComments(raw)` at `scripts/chrome-keychain.test.mjs:138`, the same single call the base had at `:204`; the only other new callers are the two tests.
- The test file still avoids reading its own fixtures as sites, and the assembly is load-bearing rather than decorative: I wrote the two fixtures out literally (`const SPWAN_CALL = "spawn(CHROME, [])";` …) and the table test goes red naming the file itself — `scripts/chrome-keychain.test.mjs:384 spawn #1: CHROME`, `:385 spawn #2: CHROME` — which is what the comment at `:377-383` says happened "the first time they were written out in full".
- The module docstring still describes the module honestly. The header is about the switch and makes no claim about what else the module contains; the moved function carries its own docstring (`:127-153`) saying why it lives there and keeping both real caveats (regex literals untracked, template content copied through), and I re-confirmed both against the code.

**Finding R4-1 [minor]** — the round-3 remediation comment's mutation numbers are wrong, and they are wrong about precisely the property this round was about. It reports the revert as "the suite goes **5/0 → 3/2** — failing **both the site scan and the new blanker test**". A re-run of that same revert over this tree gives **4 pass / 1 fail**, and the failure is the blanker test alone; `every Chrome launch site in scripts/, bin/ and the evidence harnesses is named` **passes**. That is the round-3 finding's own point: the scan cannot discriminate this guard, which is why the pin had to be direct. The suite still protects the fix, so this is not a code defect — the comment just over-states the coverage, and this series has spent three rounds on exactly that. One line to fix: `reverting the guard takes the suite 5/0 → 4/1; the new blanker test is the one that goes red, the site scan stays green.`

**Finding R4-2 [minor — pre-existing, out-of-delta, follow-up only]** — `scripts/chrome-keychain.mjs:187-199`: when a block comment is **unterminated**, the blanker appends the closing `"  "` and advances `i` past the end unconditionally, so `out` comes back **2 characters longer** than `source` — the invariant the new test asserts ("every offset must still index the raw text") and the docstring at `:133-135` states. Reproduced: `blankComments('code(); /* not closed spawn(CHROME, [])')` → 39 chars in, **41** out. It is reachable through the already-named regex-literal gap (R2-2): `blankComments('const re = /[/*]/; spawn(CHROME, [])')` → 36 in, **38** out, and the site on that line is hidden. I am not asking for a round for it: the body moved byte-identically, so nothing here introduced it, and it is not live on today's tree (0 length mismatches over every scanned file). The fix is one line — take the `out += "  "; i += 2;` tail only when the terminator was actually found (`if (i < source.length) { … }`), which makes the length invariant hold on every input rather than only on well-formed ones.

**Finding R4-3 [nit]** — `scripts/chrome-keychain.test.mjs:384-385` name the fixtures `SPWAN_CALL` / `SPWAN_CALL_WITH_ARGV`. "spwan" reads as a typo of `SPAWN_…`, and it buys nothing: the site regex is `(?<![\w.])(?:[A-Za-z_$][\w$]*\.)?${name}\s*\(` with flags `g` only (no `i`), against `SPAWN_NAMES = ["spawn", "spawnSync", "exec", …]`, so neither name can ever match it. If it is deliberate, a word saying why would save the next reader the grep. Same fixture, adjacent nit: `:395` appends a `;` to `SPWAN_CALL_WITH_ARGV`, whose value already ends in one, so `code` is `"spawn(CHROME, [\"--headless=new\"]);;"` — harmless to every assertion, just odd text.

**Stamp on `0505efea5` — verified**

- `scriptsTree` `3b268c7cf3cd761e66643a2e9d4ce6304c0a3501` == `git rev-parse 0505efea5:scripts` (was `b6a55a447452418ad514cc8ce9ec0da4d9d743f9` at `81473c439:scripts`, so the delta correctly moved the stamp).
- `srcTree` unchanged: `git rev-parse 0505efea5:src` == `git rev-parse 81473c439:src` == the shipped `d0b97badf4e33d762c1edeab8ee7018fc4ec885a`.
- `node --test scripts/evidence-manifest.test.mjs` → **28 pass / 0 fail**, including `the SHIPPED manifest's stamps describe the tree it ships in`; the worktree file is identical to `0505efea5:docs/evidence/manifest.json`.
- `captureOrigin.note` hop count: the prose says "there are seven of them" and there are **seven** `->` arrows (`1888d9fe7` → `13ac010d4` → `69d2b1160` → `f8773855f` → `bb92c92ef` → `a2c61c997` → `b6a55a447` → `3b268c7cf`), all eight values resolve as real tree objects (`git cat-file -t` = `tree`), and the last is the shipped `scriptsTree` — which is what "the value above is the tree of the commit carrying all of it" asserts. `node --test scripts/chrome-keychain.test.mjs` → **5 pass / 0 fail** unmodified.

**Observations, no action requested** (unchanged and previously declined): `docs/evidence/manifest.json` `headNote` still reads "`scriptsTree` is `1888d9fe7…`", a value the delta has now moved twice since — rounds 2 and 3 read it as a narrative about an earlier record rather than a current claim, and I am not re-opening it here.

**Commands run, and what they returned**

- `git diff 81473c439..0505efea5` → 3 files, +143/−77 (`docs/evidence/manifest.json`, `scripts/chrome-keychain.mjs`, `scripts/chrome-keychain.test.mjs`); the manifest diff is exactly two lines.
- Body-equality: `diff <(git show 81473c439:scripts/chrome-keychain.test.mjs | sed -n '/^function blankComments/,/^}$/p') <(git show 0505efea5:scripts/chrome-keychain.mjs | sed -n '/^export function blankComments/,/^}$/p' | sed 's/^export //')` → no output, `sha256` equal both sides.
- Revert test in `/tmp/r4`: guard reverted → **4 pass / 1 fail** (`the comment blanker preserves offsets…`, `AssertionError: the comment is blanked`); restored → **5 pass / 0 fail**.
- Literal-fixture test in `/tmp/r4`: fixtures written out in full → the table test fails with `['scripts/chrome-keychain.test.mjs:384 spawn #1: CHROME', 'scripts/chrome-keychain.test.mjs:385 spawn #2: CHROME']`; assembled form restored → green.
- Probe over the head's exported blanker on seven edge inputs → five preserve length exactly; the two failures are the same defect, an unterminated block comment (`39→41`, and the regex-literal form `36→38`).
- `node --test scripts/chrome-keychain.test.mjs` → 5/0; `node --test scripts/evidence-manifest.test.mjs` → 28/0 (run in the worktree, where `HEAD` is `0505efea5`).
- Two `log show` reads over the published windows → five pids one per line / 28 matching lines, and the two `create.loginkc` denials at the published timestamps.
- Not run, per the host constraints (memory 54–68 % free, load average 30 from today's crash): no `pnpm install`, no `pnpm check-evidence`, no Storybook or dev server, no Chrome, no screenshots, no full `test:desktop`, no branch mutation. The worktree is untouched — everything above ran from `/tmp/r4/`.

damianvtran:
### Round 3 remediation — one correction, and two deferred follow-ups

**Correction to my own remediation comment (round 4's R4-1).** The round-3 reply said the mutation check went "5/0 → 3/2 — failing both the site scan and the new blanker test". That was measured *before* I assembled the fixtures from pieces, when the fixture lines were still written out literally and were themselves unlisted sites. On the tree that shipped (`0505efea5`) the correct result is **5/0 → 4/1**: only `the comment blanker preserves offsets…` fails, with `AssertionError: the comment is blanked`. Round 4 measured it correctly and I had not re-measured after changing the fixtures. The claim that the pin bites stands; the numbers in the earlier comment are superseded by these.

**R4-2 [minor] — deferred, with the fix written down.** On an *unterminated* block comment the blanker appends its closing `"  "` unconditionally, so the output is two characters longer than the input, which violates the offset invariant the new test asserts. It is reachable only through the already-named regex-literal gap, which is why it is deferred rather than fixed here: the one-line fix is to append the terminator only when the loop actually found `*/`, and it should land with a test using a truncated `/*` source. Deferring keeps a terminal, fresh review round on the current head rather than moving it again for a latent case; say the word and I will fold it in instead.

**R4-3 [nit] — deferred, also written down.** The fixture identifiers `SPWAN_CALL` / `SPWAN_CALL_WITH_ARGV` should be `SPAWN_CALL` / `SPAWN_CALL_WITH_ARGV`, and line 395 carries a doubled `;;`. Neither affects the scan (its pattern is case-sensitive and would not have matched the typo anyway, which is why the reviewer called it a nit rather than a hole); both are ordinary cleanup whenever this file is next touched.

**CI on `0505efea5`:** the previous head's full run (`Pre-Release CI Testing`, `81473c439`, 22:59:20Z) completed **success** — that is the commit carrying the guard-fixture fix, so `pnpm test:desktop` is green with the blocker remediated. No run was created for `0505efea5` by its push; I am re-triggering it without moving the head.

damianvtran:
### Sync into `main`, and why CI had gone missing

The head is now `fb6521669` (merge `b184dfe21` + the re-stamp in `fb6521669`). Two things worth recording.

**CI had not gone red — it had gone absent.** `origin/main` moved to `65c9915fa` while this branch was under review, the PR went `CONFLICTING`, and GitHub cannot build a merge ref for a conflicting pull request: no `pull_request` workflow could start, so the run for `0505efea5` never existed (`actions/runs?head_sha=0505efea5…` → `total_count: 0`) and `gh pr checks` said "no checks reported". That is why I closed and reopened the PR to force an event and it still did not fire — the cause was the conflict, not the event. Merging `main` in fixes it, and the run on `fb6521669` is the one to read: **all eight checks green, including Desktop Tests**.

**What the sync changed, and what it did not.** Conflicts were `package.json` (kept `main`'s test list — `palette-*` and `interrupt-control` — with this branch's `scripts/chrome-keychain.test.mjs` beside the sibling scan) and `docs/evidence/manifest.json`'s stamp block (took `main`'s `head`/`headNote`/`capturedAt` whole, since the frames that ship are `main`'s frames plus this branch's delta). Everything else auto-merged, including `scripts/capture-evidence.mjs`; the scan still finds all twelve Chrome launches routing through `withMockKeychain`.

The re-stamp re-derived both trees from the merge — `HEAD:src` = `9a980c2de` (main's, untouched by this branch) and `HEAD:scripts` = `c33f795b0` — and recorded it in `captureOrigin.note` beside the seven earlier hops. `evidence-manifest.test.mjs` went 27/1 → 28/0 across it. Locally on `fb6521669`: `chrome-keychain.test.mjs` 5/0, `evidence-run-guard.test.mjs` 6/0 (`--test-concurrency=1`), `capture-evidence.test.mjs` 5/0, `notification-spawn-sites.test.mjs` 3/0.

The four review rounds were on `aad28e72a`, `f486d3eb7`, `81473c439` and `0505efea5`; the delta since the last one is this sync, so a final round is running against `0505efea5..fb6521669` rather than the freshness claim being stretched to cover a merge.

damianvtran:
### Agent review — round 5

Reviewer: reviewer subagent (deepseek/deepseek-flash)
Scope: 0505efea5..fb6521669

**Verdict: `clean` — no BLOCKER, no MAJOR, no MINOR, no NIT. This round is TERMINAL.** The sync delta is mechanical and correct: the merge resolved lossily nowhere, the re-stamp's values are the trees the head actually carries, and the forge shows all eight checks green on `fb6521669`.

**1. The merge resolved correctly, not lossily.**

`git merge-base 0505efea5 65c9915fa` → `461a7f11b`. Of the 253 files `main` changed from that base, 250 were changed only by `main`; for every one of them the merge result is **byte-identical to `main`'s version**:

```
$ git diff --name-only 65c9915fa fb6521669 -- <each of the 250>   # all empty
files where merge result differs from main (should be none): []
```

The only three overlaps are the ones the sync note names as conflicts/resolutions — `package.json`, `docs/evidence/manifest.json`, `scripts/capture-evidence.mjs` — which is exactly the conflict set reported in the merge commit (`# Conflicts:` lists the first two).

`package.json`, parsed as JSON and split on whitespace:

```
branch(fb6521669) count 99 | origin/main count 98 | base(0505efea5) count 94
main - branch: []                                            # nothing main added was dropped
branch - main: ['scripts/chrome-keychain.test.mjs']          # the branch's own entry survived
base - main:   ['scripts/chrome-keychain.test.mjs']          # ...and it is the branch's only addition
```

So `palette-search`, `palette-shortcut`, `palette-contract`, `palette-panel-request` and `interrupt-control` are all present (they appear in the merged `test:desktop` string, immediately before `run-detail-model` / after `notification-spawn-sites`, i.e. in `main`'s positions), `scripts/chrome-keychain.test.mjs` sits beside the sibling scan, and the entry count is `main`'s 98 + 1. JSON parses.

**2. `docs/evidence/manifest.json`.**

```
JSON parses OK
head       = 'd8a42bbd82b123db18c51ab1553aef08989c5989'
capturedAt = '2026-09-15T22:41:55.255Z'
headNote   equal to origin/main: True
head/capturedAt equal to origin/main: True, True
srcTree     = '9a980c2de605ce0212649693678b83fe597a6c18'
scriptsTree = 'c33f795b0a989394bf0581c0d1a3e5b1247c5627'

$ git rev-parse fb6521669:src      -> 9a980c2de605ce0212649693678b83fe597a6c18   # == srcTree
$ git rev-parse fb6521669:scripts  -> c33f795b0a989394bf0581c0d1a3e5b1247c5627   # == scriptsTree
```

`head`/`headNote`/`capturedAt` are `main`'s, inherited whole (string-equal to `origin/main`'s manifest, not merely equal-looking). The re-stamp commit's actual diff is two lines — `scriptsTree` `9d7dd5e50` → `c33f795b0`, and the `captureOrigin.note` tail — and it touches neither `head`/`headNote`/`capturedAt` nor `srcTree`.

The appended sync paragraph is accurate on both counts asked for: it says a re-derivation happened and not a capture ("RE-DERIVED ONCE MORE FOR THE SYNC INTO THIS BRANCH, which is the same re-derivation as the seven above and not an eighth capture", "no frame was re-taken for any of them"), and it names what moved with the right values — `HEAD:src` = `9a980c2de` "unchanged by this branch" and `HEAD:scripts` = `c33f795b0` "main's moved rigs plus this branch's keychain switch" — both matching the rev-parse output above. `head`/`headNote`/`capturedAt` fields are stated to be `main`'s, which they are.

**3. The auto-merged rig files.**

`scripts/capture-evidence.mjs` still routes its Chrome argv through the helper, and its whole delta against `main` is that routing:

```
$ git diff 65c9915fa fb6521669 -- scripts/capture-evidence.mjs
+import { withMockKeychain } from "./chrome-keychain.mjs";
-	chrome = spawn(CHROME, [
+	chrome = spawn(
+		CHROME,
+		withMockKeychain([
 			"--headless=new", ... "about:blank",
+		]),
+	);
```

`main`'s edits to the same file survive intact: the diff against the branch's own pre-merge head (`0505efea5..fb6521669`) is `main`'s content, and the diff against `main` is *only* the keychain hunk — so the two sides combined rather than one replacing the other. The other eleven sites are unchanged in kind (a one-line import plus the `withMockKeychain([...])` wrapper), and `chrome-keychain.test.mjs`'s own scanner — which walks `scripts/`, `bin/` and `docs/evidence/**/harness/` and requires `>= 12` sites, every one named in `CHROME_LAUNCH_SITES` and passing its argv through the helper — is green on the merged tree:

```
$ node --test scripts/chrome-keychain.test.mjs
✔ every Chrome launch site in scripts/, bin/ and the evidence harnesses is named, and it routes its argv through the helper
✔ the switch is spelled in exactly one module
✔ withMockKeychain appends the switch once and returns the same array
✔ the comment blanker preserves offsets, and does not mistake code for comments
✔ a call inside a template is seen, and one inside a comment is not
ℹ tests 5  ℹ pass 5  ℹ fail 0
```

**4. `evidence-run-guard.test.mjs`** — the round-1 blocker's fixture:

```
$ node --test --test-concurrency=1 scripts/evidence-run-guard.test.mjs
✔ contenders defer immediately across working directories; normal exit preserves inode
✔ dead-holder recovery never unlinks or trusts PID metadata
✔ heavy child retains admission after its launcher exits
✔ real CLI defers before frames, retains image lease, and preserves failures
✔ unsafe lock path and command failure fail closed; priority is inherited
✔ an unavailable POSIX locking module names the platform, not a module
ℹ tests 6  ℹ pass 6  ℹ fail 0
```

Also, since the re-stamp's stated gate is the manifest test: `node --test scripts/evidence-manifest.test.mjs` → **28 pass / 0 fail**, matching the commit message's "27/1 → green". `notification-spawn-sites.test.mjs` 3/0 and `capture-evidence.test.mjs` 5/0 corroborate the sync note's local numbers.

**5. CI — read from the forge.**

```
$ gh api "repos/damianvtran/local-operator-ui/actions/runs?head_sha=fb6521669…" --jq '.total_count'
2
  Pre-Release CI Testing   success  pull_request  fix/rig-chrome-mock-keychain  2026-09-15T23:29:25Z
  Version Bump Guard       success  pull_request  fix/rig-chrome-mock-keychain  2026-09-15T23:29:25Z

$ gh api "repos/…/commits/fb6521669…/check-runs" --jq '.total_count, (.check_runs[] | "\(.conclusion)\t\(.name)\t\(.head_sha[0:9])")'
8
success	Desktop Tests	fb6521669
success	Lint	fb6521669
success	Type Checking	fb6521669
success	Runtime Dependencies	fb6521669
success	Security Audit	fb6521669
success	Version Bump Guard	fb6521669
success	NPX Sanity Check (macos-latest)	fb6521669
success	NPX Sanity Check (ubuntu-latest)	fb6521669
```

All eight check runs carry `head_sha = fb6521669`, which is the PR's `headRefOid`. **Nothing is missing rather than failed:** `ci.yml`'s `pull_request` jobs are exactly `lint`, `runtime-deps`, `check-types`, `test` (Desktop Tests), `audit` and the `npx-sanity-check` OS matrix — 7 jobs, 8 checks with the matrix — plus `version-bump-guard.yml`. The two workflows that do *not* appear (`Build and Publish`, `Signed update candidate`) trigger on `release:`/`workflow_dispatch`, not on PRs, so their absence is correct. `gh pr view 225` reports `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`, `headRefOid: fb6521669…`.

**Note, not a finding.** `origin/main` has moved once more since the merge — to `7a6994d95` (`chore(release): bump version to 0.25.0`). I checked whether that invalidates this round rather than assuming: `git merge-tree --write-tree 7a6994d95 fb6521669` resolves without conflicts into tree `a15f11f60`, whose `package.json` is `0.25.0`, and `git diff 7a6994d95 a15f11f60 -- package.json` shows **no version-line change** — so the Version Bump Guard's input is unaffected by the new base, and the PR is still cleanly mergeable. The eight green checks above were computed against `65c9915fa`, so a re-run against `7a6994d95` would be a fresh event, not a correction of a stale verdict.

No findings to number — `R5-n` is empty by design rather than by omission. Merge this head (or merge `main` in once more and re-stamp, if the release bump is preferred as the base); neither path needs another review round, since the delta from here would be a version bump that touches no content under review.

**Host constraints honoured:** no `pnpm install`, no dev servers, no Chrome, no screenshots, no full `test:desktop`, no `pnpm check-evidence`; reads, single-file `node --test` runs (in a `/tmp` copy) and `gh` queries only. `memory_pressure` was checked first (54% free).

damianvtran:
**Released in v0.25.1.** 🎉

- Tag: `v0.25.1` → `775bb0646a65249300093cd854349cbfbff7b04b`
- Release: https://github.com/damianvtran/local-operator-ui/releases/tag/v0.25.1 (a pre-release until `publish.yml` attaches and verifies its assets, then promoted; run 35038914111)
- Compare: https://github.com/damianvtran/local-operator-ui/compare/v0.25.0...v0.25.1
- This PR is half the window (with #196), and the notes say plainly that its body carries no `Release:` line, so no impact sentence is quoted for it — it is rig-only and user-invisible, which is why the window is a patch rather than anything larger.

PR #244, opened 2026-09-16
fix(update): repair a bundle whose interpreter wrote its bytecode cache inside the app

damianvtran:
# PR Description

## Summary of Changes

The operator's installed `/Applications/Local Operator.app` (0.25.2) failed its
code-seal check and **both of the app's dialogs refused to act on it** — the
start-up one ("this copy needs replacing … it can't repair itself") and the update
one (the 0.25.4 update was stopped). The bundle was not damaged and not tampered
with: it was *one file-deletion away from valid*, and the heal that exists to make
that deletion could not recognise the shape of the break. This PR fixes the
recognition, fixes the thing that caused the break, and fixes the test suite whose
own environment had become part of the problem.

### 1. The heal now recognises a `PYTHONPYCACHEPREFIX` cache mirror

`codesign` reported 19 `file added:` violations, every one a `.pyc` under
`Contents/Resources/python_aarch64/pycache/…`. That layout is CPython's
`PYTHONPYCACHEPREFIX` mirror — the prefix directory, then the absolute source path
minus its root — and it contains **no `__pycache__` segment anywhere**.
`isPythonBytecodePath` required one (`if (!segments.includes("__pycache__"))
return false;`), so `planPythonBytecodeHeal` classified all 19 as "added resource
outside the bundled python trees" and the app told the operator to reinstall.

The segment test is **removed rather than widened**, and the reason is that no
fixed segment can describe this layout: a mirror carries whatever directory name
the prefix had (the incident's was `pycache`; the app's own cache directory is
named `python-bytecode-cache`). The conditions that remain are already
sufficient, because nothing this repository ships is a `.pyc` at all — enforced at
build time by `scripts/setup-python-resource.sh` and the release gate in
`scripts/verify-macos-artifacts.mjs`, precisely because a *shipped* `.pyc` that
gets rewritten is the `file modified:` class the heal can never repair. So a
`.pyc` inside `BUNDLED_PYTHON_DIRS` under the bundle being healed, reported
`added` by `codesign` itself, has no legitimate way to be there. The docstring
states this, and states the alternative it rejects.

Kept exactly as they were: **per-FILE removal** (a directory delete would take a
sealed `__pycache__` `.pyc` with it and turn `file added:` into `file missing:` —
measured, and the existing real-`codesign` test still pins it), the re-check
against `isPythonBytecodePath` immediately before each unlink, and the note that
the directories a mirrored write creates are left behind deliberately (an added
empty directory still verifies). The docstring now says that about the mirrored
[...]

PR #212, opened 2026-09-15
feat(browser): the approvals queue, a dock that narrows the page, a counted badge, and tabs that read like tabs

damianvtran:
## Summary

PR 1 of the browser approvals/UX overhaul, implementing `docs/design/browser-approval-ux.md`
§3 (the approval queue model), §4 (the surfaces), §5 (the badge), §6 (the tab-strip
grammar), §8.3 (the legibility fix) and §10 (the evidence plan). §7 (the
conversation-scoped pane) is PR 2 and is deliberately NOT built; the four interfaces
it needs are in place, listed below.

It also carries a consent fix agreed with the author of an independent end-to-end pass
against the released v0.24.0 build, because it lands in the same functions this branch
reworks. That is a scope addition, reported as such.

Release: minor — a new Approvals surface: a real queue, a dock that narrows the page, a counted badge, and tabs that show which agent is working.

The four things the operator asked for, and what each needed:

| Ask | What shipped |
|---|---|
| A badge count of outstanding approvals on the approvals control | The badge IS the control's accessible name (`Approvals, 3 waiting`), on the same outline icon+label control. `approvalCount` used to count the APPROVED list, which is not a demand on the user. |
| A better approval surface: a numbered badge callout, a list openable and closable from inside the browser page | The tray's numbered chips (the callout), and an `Approvals` dock that narrows the page instead of hiding it — the Sites sheet it replaces registered browser-view suppression exactly like a dialog, so the page was hidden while you read what you had granted. |
| A tab strip that reads like real browser tabs rather than buttons | A tab is a sheet of paper in a well now: no `surface` cells, a `hairline` between neighbours, a colour-step hover, the active tab on the page's own ground with `border-control` and a notch into the page. The row's actions expand IN the band, because a menu anchored at the strip painted downward into the content rect where the native view occluded it. |
| Visibility of what agents are doing | See below — the §8.2 run and the fix it justified. |

### Round 1 remediation — what changed after the four reviews

The four round-1 reports are on this PR (`### Agent review — round 1`, `### Design review — round 1`, `### UX review — round 1`, `### QA report — round 1`) and each is answered by a matching `remediation` comment. The code for all of it is one commit, `74c966401`; the frames and stamps are `6ca817a0d`.

Three of the four streams found something I had shipped broken, and the mechanism matters more than the list:

[...]

PR #209, opened 2026-09-15
fix(composer): arm the goal command from the popup's row, never from a draft's wording

damianvtran:
## What this is

`/goal` in the desktop composer armed the command by **appearing** in a draft. It is
now armed only by **acting on the goal row the slash popup is showing** — Enter on
that row, or a click on it — and by nothing else: with the list closed, a draft that
merely contains the word is sent as written, in the user's own order. Acting on the
row hoists the command to the front and stages the line, and the row's own line says
what the next Enter will do; the whole-draft `/goal <text>` shape still runs on one
Enter, unchanged.

The gesture was re-decided once, mid-review, when `main` landed its own command-row
Enter contract (an unambiguous Enter acts on the row it is showing): the composed
rule is the one described here and in "Round 4" below, and it is what this head
ships. The commit history carries both decisions, so neither is reconstructed.

Release: patch — the goal is armed only from the slash popup's own goal row (Enter on it, or a click), never inferred from a `/goal` written inside a sentence; with the list closed that sentence is sent as written, and a line-initial `/goal <text>` above prose is no longer silently reordered into a command.

## Why: what the word used to do

Reproduced against the shipped planner, bundled from this repo's own module with
esbuild and called the way the composer calls it (the probe is in the Evidence
section verbatim):

| draft, caret anywhere | before | after |
| --- | --- | --- |
| `/goal ship the release` | `whole` → runs on one Enter | unchanged |
| `I approve spend /goal` | `reassemble` → `/goal I approve spend`, **staged, nothing sent** | `send` → prose, in its own order |
| `please run /goal on 20 tasks` | `reassemble` → `/goal on 20 tasks please run` | `send` |
| `ship it /team ops` | `reassemble` → `/team ops ship it` | unchanged |
| `fix it /usage` | `splice` → `/usage` runs, `fix it` survives | unchanged |
| `/goal ship it\nand then tell me` | `reassemble` → `/goal ship it and then tell me`, **staged** | `send` → prose (stated change, see below) |
| `I approve spend /goal`, Enter with the popup open | the row is pre-selected and Enter **armed and staged**, silently and with no way to tell it would | Enter on that row stages the line and the row's own sentence says so — the arm is a stated act on a visible row, not a side effect of the wording |
| `I approve spend /goal`, arrow key on the row, Enter | n/a — no way to arm by keyboard | the pick arms and stages |
| the goal row **clicked** | the pick ran the command's own dialog | the pick arms and stages the line |

[...]

PR #251, opened 2026-09-16
chore(release): bump version to 0.25.10

damianvtran:
## Release window: everything since `v0.25.9`

**Owner:** this PR is the visible lock for the window after `v0.25.9` (0.25.10), claimed by the session that owns PR #244. It is deliberately **draft** and must not be merged until #244 lands — see the last section.

### The window

| PR | Merge commit | `Release:` claim in its body |
| --- | --- | --- |
| #212 `feat(browser): the approvals queue, a dock that narrows the page, a counted badge, and tabs that read like tabs` | `ac81ce9ac` | minor — a new Approvals surface: a real queue, a dock that narrows the page, a counted badge, and tabs that show which agent is working. |
| #209 `fix(composer): arm the goal command only from an explicit pick` | `658c01673` | (to be read verbatim from its body before the tag; the owner will correct this row rather than summarise it) |
| #244 `fix(update): repair a bundle whose interpreter wrote its bytecode cache inside the app` | pending merge | patch — a Local Operator install whose interpreter wrote its bytecode cache inside the app is now repaired in place, instead of the app refusing its own start-up repair and in-place update and asking for a reinstall. |

### Bump: patch (0.25.10), argued by materiality rather than derived

`patch` is the whole window. #212 rebuilds the browser-approvals surfaces — a queue, a dock, a badge, a tab strip — which a user would describe as the approvals UI being much better, not as the app gaining a thing it could not do; #209 is a scoping fix to one command; #244 is a repair. This repository reserves `minor` for a step-function capability and prefers to under-call, and two peers reading the same window (#212's own session, and the session that owned the previous window) argued patch as well. The `minor` claim in #212's body is therefore overridden by the owner's single call for the window, which is the procedure: the body carries the author's claim, the owner picks one bump for the window.

### Why this PR is a draft, and what happens next

A release tag must point at a tree that contains **every PR in its window**, so this bump may not merge before #244 does. The sequence, in order:

1. #244's final convergence round closes on its frozen head (round 6; the earlier rounds are on that PR).
2. #244 merges to `main`.
3. This branch is rebased onto the resulting `main` (so the bump commit sits on top of #244's merge) and its draft flag is dropped.
4. This PR merges with its own scope-check round on the one-line diff, and then `gh release create v0.25.10 --target <bump-SHA> --prerelease` runs with hand-written notes listing every PR above.

Any PR that merges after that tag rides the next window; the owner does not hold a window open for it.


damianvtran:
### Agent review — round 1

**Reviewer:** `reviewer` subagent, independent (did not author this change). Model actually running: **`deepseek/deepseek-flash`**. No provider fallback occurred during this round.

**Scope:** `60b934172319842802c8fb98e463fe7b85a7027d` (the merged main this branch is based on, which **is** the current `origin/main` tip) .. `d310a632e2f114738e1c87e93dc9940b514e05d7` (the head actually on the PR, re-read with `gh pr view 251 --json headRefOid`).

**The head is not the SHA in the hand-off, and the difference matters — see R1.** I reviewed `d310a632e`, the commit GitHub reports for this branch. `fe0dc10bf5c3b3e4df81f19c9c306b12cc227878` is a *different* commit with the same message and the same one-line diff, and it is not an ancestor of it.

Read-only, one job, no suites run — as instructed; CI covers this PR's own run (its state at the bottom).

---

#### What the diff is, verified

```
$ git diff 60b934172..d310a632e --stat
 package.json | 2 +-
$ git diff 60b934172..d310a632e
-  "version": "0.25.9",
+  "version": "0.25.10",
```

One line, one file, one commit (`d310a632e chore(release): bump version to 0.25.10`). Nothing else in the tree carries a version: `git grep -n "0\.25\.9\|0\.25\.10" <head>` matches `package.json:3` and nothing else (no `docs/evidence` hits either), and no other surface pins it — `scripts/release-baseline.mjs` derives its anchor from the tags and `scripts/apply-release-bump.mjs` writes the line textually, so neither carries a constant that a bump has to update. `package.json` is the version source of truth per this repo's `AGENTS.md`, and electron-builder reads it, so there is no second file to keep in step.

**The number is unspent.** The recipe's own proof — `git diff v<PREV>..origin/main -- package.json` — touches no version line between the tag and the merged main (`git diff v0.25.9..60b934172 -- package.json | grep -cE '^[-+].*"version"'` → `0`; the file's diff there is the `test:desktop` roster, which is expected), and main still carries `"version": "0.25.9"`. So no merged PR consumed 0.25.10.

**The window is exactly the three PRs the notes list**, by the shape the repo prescribes rather than by commit shape:

```
$ git log --first-parent --oneline v0.25.9..origin/main
60b934172 Merge pull request #244 from damianvtran/fix/bytecode-cache-heal-and-test-isolation
658c01673 Merge pull request #209 from damianvtran/fix/goal-arming-explicit
ac81ce9ac Merge pull request #212 from damianvtran/feat/browser-approval-ux
$ for sha in 60b934172 ac81ce9ac 658c01673; do git merge-base --is-ancestor $sha d310a632e && echo "$sha ancestor"; done
60b934172 ancestor      ac81ce9ac ancestor      658c01673 ancestor
```

**The guard accepts this PR on its own terms.** I ran the shipped decision function on this exact title, diff and tag list rather than reading it:

```
$ node -e 'decideVersionBump({ prTitle: "chore(release): bump version to 0.25.10", diff, tags: ["0.25.7","0.25.8","0.25.9"] })'
guard verdict ok = true
   -  "version": "0.25.9",
   +  "version": "0.25.10",
   Version 0.25.10 sorts above the highest release tag v0.25.9.
$ … with an empty diff and the same title
mirror-image (title, no version change) ok = false
```

So the dangerous direction the title keys on (`chore(release): bump version to ` with no version change) is genuinely refused, and this PR is on the other side of it. CI agrees: **Version Bump Guard: pass** (11s). **CI is not fully green yet** — `Desktop Tests` and `NPX Sanity Check (macos-latest)` were still pending at the time of writing; Lint, Type Checking, Runtime Dependencies, Security Audit and NPX (ubuntu) passed. Merge after those two land.

#### The notes (`/tmp/v0.25.10.md`), checked against the artifacts they describe

- **Shape passes the pipeline's gate.** `scripts/release-state.mjs` matches `TEMPLATE_NEW_HEADING_RE = /^## What's new\s*$/im` — case-insensitive on purpose, its own comment says the case is not part of the shape — so `## What's New` is fine, and `COMPARE_LINK_RE = /Full Changelog/` is satisfied. The link's endpoints are right: `git tag --sort=-v:refname | head -1` → `v0.25.9`, and `gh release list` shows `0.25.9 … Latest`, so `compare/v0.25.9...v0.25.10` names a real previous release.
- **`## What's New` carries no PR numbers and no commit subjects** — the section has neither (`grep -nE "#[0-9]+|[0-9a-f]{7,40}|(fix|feat|chore|docs|test)\([a-z-]+\):"` over it and `## Impact` → no match) — and its bullets carry constraints and trade-offs rather than the change alone, which is what the template asks for.
- **The one technical claim of a new kind in the prose is true of this artifact**: "the scan that enforces it now reads every file type and binding shape those harnesses use". At this head the scan's discovery is `for (const root of ["scripts", "bin"]) walk(…)` with `/\.(mjs|cjs|js)$/` — so that sentence is accurate as merged; it was not when I reviewed #244's round 2, and the widening landed afterwards (`f9282381c`, `0ec20f04e`), which is why the row is worth the words now.
- **The `## PRs` rows are the problem — R2 and R3.**

### Findings

**R1 — MAJOR (not in this diff: it is the SHA the release would be cut on). The head in the hand-off is a superseded commit whose base does not contain this window's last PR; do not merge or tag `fe0dc10bf`.**
`fe0dc10bf` carries the same title and the same one-line bump, but its parent is `658c01673` (#209's merge) — so **#244's merge `60b934172` is not an ancestor of it**, and the requirement this PR was checked against ("the window's contents must all be ancestors of the tag target") fails for that SHA and holds for the real head:

```
$ git log -1 --format='%P' fe0dc10bf   → 658c01673…
$ git merge-base --is-ancestor 60b934172 fe0dc10bf   → false
$ git merge-base --is-ancestor 60b934172 d310a632e   → true
```

A tag cut with `--target fe0dc10bf` would publish a tree **without the bytecode-heal repair** under 0.25.10 while these notes announce it — exactly the "tag, Release and installers all agree with each other about the wrong tree" failure the guard's own docstring exists to make loud. The fix is procedural, not a code change: take the merge SHA from GitHub at merge time (`gh pr view 251 --json mergeCommit --jq .mergeCommit.oid`), which on this branch will be a merge of `d310a632e`, and never reuse the SHA in a message. Worth re-echoing the head SHA when you announce the cut.

**R2 — MINOR. The `## PRs` row for #244 does not quote the PR's `Release:` line; it substitutes a different sentence.**
`AGENTS.md` is explicit that a PR's line is "what the owner reads when collecting a window, and **what the Release notes quote verbatim**". #212's row and subject are verbatim (checked character-for-character, including its merge SHA), but #244's is not:

```
PR #244 body:  Release: patch — an app whose interpreter wrote its bytecode cache inside the bundle is repaired in place at launch instead of asking the user to download a fresh copy.
notes row:     Release: patch — a Local Operator install whose interpreter wrote its bytecode cache inside the app is now repaired in place, instead of the app refusing its own start-up repair and in-place update and asking for a reinstall.
```

The meaning is close, but the sentence is the owner's rather than the PR's, and this section is read as a quotation. Either paste #244's own line, or present the sentence as a summary instead of as a quote — and this is the last moment at which either is possible.

**R3 — NIT. The #209 row is the PR's line minus two backtick pairs.**
Its verbatim body text contains `` `/goal` `` and `` `/goal <text>` ``; the notes row says `/goal` and `/goal <text>`. The cause is visible and sane — the whole line is wrapped in a code span, so nested backticks would break it — but the fix is one of: drop the wrappers, or escape them. Same requirement as R2, smaller blast radius.

**R4 — NIT. One `## What's New` bullet is repo tooling rather than a user-facing consequence.**
"The test suite that could write into your app now states its own environment" describes harness hygiene; the template's rule 3 asks that section to state "what the version does FOR A USER". A user of the app experiences the *repair* (bullet one) and the *prefix* policy (bullet two); the third bullet matters to whoever runs this repository's tests — which, granted, on this incident was the operator's own shell, where it did cause a real in-app failure. Keep it if you want that recorded, but it reads as an engineering note inside user-facing prose, and one clause folded into the repair bullet would carry the same fact without it. No numbers, subjects or hashes are in that section regardless.

### Class, judged against the written rule rather than taste

The window's bump is **`patch`**, and I think that is defensible: `AGENTS.md` reserves `minor` for "a single PR in the window [that] is a step-function capability in its own right — something a user would call a new thing the app can do, not more of what it already did", and explicitly "prefers under-calling", since an under-call is corrected by the next release while an over-called minor permanently misreports how much changed. #212 rebuilds an existing surface (approvals) with a queue, a dock, a badge and tab semantics; whether a user would call that "a new thing" is the judgement you own, and the rule gives the owner latitude to say no. One presentational consequence to be aware of, not to fix: the published notes will show #212's row claiming `Release: minor` under a `patch` bump. That is the shape `AGENTS.md` prescribes — each row quotes its own PR's claim, and the window's single class lives in the version, which the notes never state — so a reader could misread the ledger as the release's class. Nothing in the file needs to change for it.

### Verdict

**The diff is clean and mergeable: one line, one file, no stray version anywhere, the number unspent, the window's three PRs all ancestors of this head, the guard's own decision function green on this input and its mirror-image case refused, and the notes' shape passing the pipeline's gate.** Round 1 on this diff is **TERMINAL** — R2–R4 are notes wording and R4 is the owner's call, none of them a reason for another round on the version line. **The one thing to act on before tagging is R1: merge and tag the current head `d310a632e` (via the merge commit from `gh pr view 251 --json mergeCommit`), never `fe0dc10bf`** — and wait for the two pending CI jobs (`Desktop Tests`, `NPX Sanity Check (macos-latest)`) to go green first.

PR #205, opened 2026-09-15
fix(daemon): stop calling a serving daemon offline, and stop spawning over one

damianvtran:
## Summary

The operator's app told them the backend on port 1111 was unavailable while that backend was serving the TUI and answering the app's own reads, and after a force reload no conversations would load. Both symptoms are one subsystem, and this is the reproduction of each against `8f80697c8` plus the fix.

**Defect 1 - a probe's silence was counted as a refused socket.** `probeAttachedDaemon()` folded every `unreachable` outcome into one `failed` observation, so three probes that ran out of their 2 s budget detached the connection exactly like three sockets the OS refused — and `isServerReachable("detached")` is false, which is the renderer's `online: false`, the connectivity banner, and every gated read disabled. A daemon in the middle of one long agent turn produces three of those while answering everything else: measured on the base tree, tick 3 reports `detached` with `sessions_read: 200` **in the same tick**.

**Defect 2 - `start()` never asked whether anything was serving the port.** Discovery's verdict ("no daemon I may attach to") was read as the spawner's question ("this port is free"). With no serve record for the answering daemon, `start()` spawned onto its port; the child died with `[Errno 48] Address already in use`, every ~10 s, each an orphan, while `/health` answered the readiness loop 200.

**Defect 3 - the credential for the app's own daemon was never persisted.** A launch cannot open the daemon the previous run spawned and left running (its record publishes `claim_key: ""`, so the spawner's token is the only credential that ever opens it), so it declined that daemon and started a second one onto its port.

The fix, at the source in each case:

- `daemon-status` carries the cause: `unanswered` (a budget that expired) needs `UNANSWERED_BEFORE_DETACHED` = 9 consecutive misses and stays usable at `degraded` meanwhile; `refused`/answered-but-unhealthy still detaches on the ordinary three. NEW: `recordTransportSuccess()` — any answered desktop request is liveness evidence, and inside `TRANSPORT_EVIDENCE_MS` (30 s) **no** miss may detach. `pid-dead` bypasses that gate, because a gone process cannot be the one answering.
- `backend-service` records every answered desktop call (`noteTransportAnswer`), and before every spawn probes the configured origin (`configuredOriginOccupancy`): a Local Operator daemon answering there stops the spawn and is recorded as `unattachable` → state `wedged` with copy naming the address, the pid when it published one, and the path (`no serve record this app can read describes that address`). An address that did not prove itself free also stops it; a refused socket, the one answer that proves a port free, proceeds as before.
[...]

PR #252, opened 2026-09-16
chore(release): bump version to 0.25.11

damianvtran:
## Summary

Spends **0.25.11** for the window that opened at `v0.25.10`. The window is one PR:

- **#205** — `fix(daemon): stop calling a serving daemon offline, and stop spawning over one` — merge `811ff0773`. Its own release claim: *"a serving daemon is never reported offline, nothing is spawned over one, the app re-attaches to the daemon it left running, and the conversation list keeps loading through a flap."*

**Bump: patch**, by materiality rather than by commit type. The window is a resilience fix set — a serving daemon no longer gets reported offline, the app no longer spawns over a daemon that is serving, the app's own daemon credential is persisted so a relaunch re-attaches, the pairing probe no longer reads any non-2xx as "our credential was refused", the sidebar keeps its chats/agents/teams (with the condition stated) instead of unmounting them and waiting for a window focus, and the daemon's new additive `degraded` marker is consumed so a list whose liveness could not be read no longer reads as "Nothing running right now." None of it is a step-function capability; all of it is the app doing what it already promised, more reliably.

## Why the operator sees it

The reported symptom was the sidebar losing its chats and teams and needing a refresh, mostly under load. QA reproduced it on the base tree (list gone, no explanation, nothing re-asks for 90 s, a second refresh required) and verified the fix on the head (rows held and dimmed by state, the condition named, self-healing in ~12 s with no interaction, a genuinely empty store still reading as empty).

## Scope

`package.json` only — one line, the version. No feature content, no dependency change.

Release: patch — a serving daemon is no longer reported offline, and the sidebar keeps its chats and teams instead of needing a refresh.


damianvtran:
### Agent review — round 1

Reviewer: reviewer (deepseek/deepseek-flash, running solo — no provider fallback mid-round)
Scope: 811ff0773..0f76af9e1

C0 scope-check round on a release bump PR. No BLOCKER, no MAJOR, no MINOR, no NIT findings on the diff.

**1. The diff is exactly what the title claims.** PASS

```
$ git diff --stat origin/main...0f76af9e1
 package.json | 2 +-

$ git diff origin/main...0f76af9e1
@@ -1,6 +1,6 @@
 {
   "name": "local-operator-ui",
-  "version": "0.25.10",
+  "version": "0.25.11",
   "productName": "Local Operator",
```

One file, one hunk, one line. `name`, `productName`, dependencies, and every script line are untouched. The branch is current with its base — `origin/main` (`811ff0773`) is an ancestor of the head, so the two-dot and three-dot diffs are identical and the base SHA the PR reports equals `origin/main`.

**2. The guard's mirror-image case does not apply.** PASS

The version really moved on the branch ref, not just in a working tree:

```
$ git show 0f76af9e1:package.json | grep -n '"version"'
3:  "version": "0.25.11",          # origin/main: 3:  "version": "0.25.10",
```

And the head CI ran on is that same SHA — the `Version Bump Guard` run reports `head_sha=0f76af9e1`, matching `headRefOid`. The guard also printed a real verdict rather than its designed silent-pass shape, and it read both sides of the edit:

```
-  "version": "0.25.10",
+  "version": "0.25.11",
Version 0.25.11 sorts above the highest release tag v0.25.10.
```

So both directions of the guard are satisfied: the release title is present *and* the version actually changed, ahead of the newest tag.

**3. Nothing already spent this version.** PASS — `v0.25.10..origin/main` shows no `"version"` line change:

```
$ git diff v0.25.10..origin/main -- package.json | grep version
(no version line in diff)
```

The diff is not empty, and that is expected, not a failure: the only `+`/`-` pair is the `test:desktop` scripts line in #205, which gains `scripts/sidebar-catalogue-gate.test.mjs`. That is PR #205's test file addition riding in on the same merge — no `"version"` line is touched anywhere in the range, so no merged PR carried a bump of its own and 0.25.11 is unspent.

**4. The window claim in the body is true.** PASS — the window is one PR, #205:

```
$ git log --first-parent --oneline v0.25.10..origin/main
811ff0773 Merge pull request #205 from damianvtran/fix/daemon-attach-robustness
```

`v0.25.10` points at `bf05c4894` (PR #251's merge), and #244 is already an ancestor of that tag, so nothing else lands in the range. `gh pr list --state merged` shows no other merge between the tag and the head. The notes for this window therefore need #205 alone, and the body names it with the right merge SHA.

**5. CI on the PR.** PASS — all checks green on `0f76af9e1`:

```
Desktop Tests            pass  5m14s
Lint                     pass  37s
NPX Sanity Check (macos-latest)   pass  2m6s
NPX Sanity Check (ubuntu-latest)  pass  50s
Runtime Dependencies     pass  11s
Security Audit           pass  28s
Type Checking            pass  36s
Version Bump Guard       pass  10s
```

The `bump: patch` class is the right call by materiality — #205 is a resilience fix set (a serving daemon no longer reported offline, no spawn over a serving daemon, credential persistence for re-attach, the sidebar holding its chats and teams), not a step-function capability — and a bump PR takes only this C0 one-file scope check under the release procedure.

**Note (not a finding, not about this diff):** the `~/local-operator-ui` checkout has an unrelated uncommitted modification, `docs/design/composer-suggestions.md`. It is not on this branch and does not affect the PR, but it is live in the tree the release owner works in — worth keeping out of any further commit made here.

**Verdict: clean — merge.** No blocker or major remains, the round is fresh on the current head (`0f76af9e1`), and CI is green. TERMINAL: nothing further is required on this PR from this reviewer.

PR #283, opened 2026-09-16
fix(chat): drop the retry from a completed MCP sign-in

damianvtran:
## Summary

The operator's report, with the frame he sent from the dialog:

> when sign-in is complete, can you remove the Try Again button and just show the close button, there's no need to try again if the sign-in worked, it would be confusing and looks like it failed based on the presence of that button

What he was looking at: the `Sign in to hubspot` dialog settled on `Sign-in complete.` with two controls in its footer, `Close` and `Try again`.

**Release:** patch — a completed MCP sign-in offers `Close` alone instead of `Close` + `Try again`.

**Round 1 is remediated on this head** (`4dfa55874`): the review's R1-1..R1-5, the design round's D1/D2, and QA's Q3. The three remediation comments are on the thread — review, design, QA — and this body's commit table and gates are re-stated for **this** head.

### Commits

| SHA | What |
|---|---|
| `4ae4fa1af` | the change: one named predicate, the footer condition and the extended comment; the DOM test; the capture harness and the six evidence frames |
| `083ebbe6c` | the new capture rig's Chrome launch site, declared in `scripts/chrome-keychain.test.mjs`'s table — the suite caught it, which is that table doing its job |
| `7b064a88c` | a wording pass on the two comments this branch adds |
| `4dfa55874` | **round-1 remediation**: the fourth test case (the unreadable-status boundary), the README corrections (pixel arithmetic, before-tree label, the diff's two mechanisms, the retry's own primary, the `probing` precedent, the solo-run limitation), the rig's teardown fix, and the manifest re-stamp |

Each commit that moved `src/` or `scripts/` is followed by a manifest re-stamp, folded into the commit it belongs to rather than left as its own.

## Root cause

`src/renderer/src/features/chat/components/run-details/mcp-auth-dialog.tsx`, the footer's retry:

```tsx
const running =
	state.kind === "running" && (!operation || operation.status === "running");
…
{state.kind === "notice" || failure || (state.kind === "running" && !running) ? (
	<SecondaryButton onClick={() => setAttempt((value) => value + 1)}>Try again</SecondaryButton>
) : null}
```

A COMPLETED sign-in is exactly the state that condition could not tell from a settled failure: the phase is `running` (the press put it there) and `running` is false, because the operation the 1s `mcp.list` poll answers with is no longer `running`. So `complete` rendered the same footer as `failed`. The body sentence was already right (`Sign-in ${operation?.status ?? "status unavailable"}.` → `Sign-in complete.`), which is what made the pair read as a failure.

No backend change is involved: the grant statuses come from `local_operator/mcp/desktop.py`, and the operation carries no `message` — this is the UI's own copy and its own control. The **run panel's row is untouched**: `mcpGrantStates()` drops a newest `complete` operation, so the panel never rendered a `Try again` for a completed grant.

## What changed

- **The predicate, named**: `const completed = operation?.status === "complete";` — read off the wire's own word rather than off the absence of `running`, so an operation whose status this build cannot read is not treated as a success.
- **The footer condition**: `state.kind === "running" && !running && !completed`. `failed`, `cancelled` and an unreadable status keep the retry exactly as before, as do the `notice` and `failure` branches.
- **The comment** above the retry now also says why a success is the one settled state it stands down for: the press re-runs the probe, the probe answers "this server takes OAuth", and the dialog is handed back its own primary — which re-offers the grant that just succeeded, under a sentence that says it worked.
- **`package.json`**: one line, the new test registered in the `test:desktop` file list. No version bump anywhere (the guard is green).

## Testing evidence

There is no UI test runner here; tests are `node:test` files under `scripts/`, and the technique for component state is jsdom + `react-dom/client` + `act` + an esbuild bundle of the SHIPPED component.

`scripts/mcp-auth-complete-no-retry.test.mjs` mounts the real `McpAuthDialog` behind the `window.api.desktop.request` bridge, drives it through a probe that answers OAuth, a real DOM press on `Continue in browser`, and a polled `mcp.list`, and reads the footer from the rendered DOM:

| Case | Footer the DOM must show |
|---|---|
| `complete` | `["Close"]`, no `Try again` anywhere, body still `Sign-in complete.` |
| `failed`, `cancelled` | `["Close", "Try again"]` |
| `succeeded` (outside `MCP_GRANT_STATUSES`) | `["Close", "Try again"]` — the boundary the comment advertises |

```
$ node --test --test-concurrency=1 scripts/mcp-auth-complete-no-retry.test.mjs
✔ a completed sign-in offers Close alone, under the sentence that says it worked (198.013333ms)
✔ a failed sign-in keeps Close and Try again (115.960084ms)
✔ a cancelled sign-in keeps Close and Try again (89.453958ms)
✔ a status this build cannot read is not a success, and keeps the retry (83.712292ms)
ℹ tests 4   ℹ pass 4   ℹ fail 0
```

**The discriminating run.** The same file against the base commit's component: the `complete` case fails with `actual ['Close','Try again']` against `expected ['Close']` while the other three still pass — that is the differential proof, and it is only the `complete` case. The fourth case is deliberately *not* differential (the base is right about that boundary too); it is shown to discriminate by mutating the predicate to `operation?.status !== "running"`, where it fails `['Close']` against `['Close','Try again']` — a `src/` edit made in the working tree only and reverted, `git diff --stat -- src/` empty afterwards.

## Evidence: the frames

Six frames under `docs/evidence/mcp-auth-complete/`, from the real component in the real stylesheet, mounted by a bespoke harness (`docs/evidence/mcp-auth-complete/harness/`) with the same stubbed bridge seam, captured with headless Chrome over CDP. The rig asserts the footer it photographed before the shutter and calls `assertFramePaints` on the written frame.

| Frame | What it shows |
|---|---|
| `complete-before/localOperator{Dark,Light}.png` | **The defect, on the base tree**: `Sign-in complete.` with `["Close", "Try again"]` — the operator's screenshot, reproduced (captured from `da9e75a616`, re-taken byte-identically from `cfc28c817`, this PR's base) |
| `complete-after/localOperator{Dark,Light}.png` | **The fix**: same dialog, same settled operation, same sentence, footer `["Close"]` |
| `failed-after/…`, `cancelled-after/…` | The controls: both still offer `Close` + `Try again` |

Measured, both brand themes: the pair differs by **15,612 px** dark / **15,632** light (5.38% / 5.39% of the 704x412 frame), in **one 375x72 box at +206+258** — the retired control's own columns (14,308 px) plus the surviving button's label ink (1,304 / 1,324), the surviving button keeping its own 100 CSS px box. The panel's geometry is identical in all six frames, and all eight committed frames are byte-identical across runs and across this round.

`docs/evidence/mcp-auth-complete/README.md` carries the exact commands (including the `git archive` of the base tree for the before half), the port, the solo-run limitation, and what these frames do **not** prove.

## Rebased twice while it was open

`main` moved under this branch twice. First `da9e75a61` → `cfc28c817` (the 0.25.17 window), then `cfc28c817` → `5c53e1c75` (0.25.18) during the review round. Each time GitHub reported the branch `DIRTY`, and **a DIRTY branch gets no CI at all** — the runs for the new head are never created — so each was rebased and force-with-lease pushed (`6deb10c6a → c92d9d6d8 → 4dfa55874`, disclosed because a rewritten tip is something a reader should know).

What moved and what did not, measured at each rebase: the branch's file set is unchanged, and **this change's own product file is untouched** — `mcp-auth-dialog.tsx`'s blob is `c4ea7c9c1` at every head, the same blob the round-1 review read. The conflicts each time were `docs/evidence/manifest.json` (a doc-sync file both sides re-derive) and, on the second, `package.json` (main's version line beside this branch's one `test:desktop` entry); both were resolved by taking main's side and re-applying this branch's own delta, with `srcTree`/`scriptsTree` re-derived from the rebased tree. `main`'s own commits came in with the rebases (`managed-python.ts`, `bundled-python-layout.json`, the python-seed pruning scripts) — out of this PR's scope, and the reason a convergence round scoped `c92d9d6d8..4dfa55874` is the manager's call rather than something this body claims.

## Gates

All run in a fresh worktree off `origin/main`, with its own `node_modules`, at the current head `4dfa55874`:

```sh
pnpm lint                    # exit 0 — 648 files, warnings only (pre-existing)
LINT_SINCE=5c53e1c75 pnpm lint:scripts
                             # exit 0 — "2 changed file(s) under scripts/ are lint-clean against 5c53e1c"
pnpm check-types             # exit 0 — main + renderer projects
pnpm check-edit-diffs        # exit 0 — "4 insertion previews + unchanged nonempty/dismissed states passed"
pnpm check-runtime-deps      # exit 0
node --test scripts/mcp-auth-complete-no-retry.test.mjs            # 4/4
node --test scripts/{chrome-keychain,evidence-manifest,test-inventory}.test.mjs   # 34/34
pnpm test:desktop            # 2205 tests, 2204 pass, 1 fail (pre-existing; below)
git status --porcelain       # empty after the suite, which is the property CI asserts
pnpm check-evidence          # DEFERRED, exit 75 — see below
```

```
ℹ tests 2205   ℹ pass 2204   ℹ fail 1   ℹ duration_ms 155709.806667
```

CI at this head — every workflow green: Lint, Type Checking, Runtime Dependencies, Security Audit, Version Bump Guard, NPX Sanity Check (macos-latest and ubuntu-latest), and **Desktop Tests (pass, 5m52s)**.

### The one failure that is not this change

`scripts/python-bytecode-cache.test.mjs` — `every backend spawn carries the prefix even with the shell-env load unresolved`: *"global install: start() should report a started backend"*. It reads this machine's installed python toolchain rather than the repository, and it reproduces identically on the base tree (a scratch materialisation of `origin/main` via `git archive`, this worktree's `node_modules` linked in), where `update-robustness.test.mjs`'s classifier test failed the same way. Neither file is in this diff, so they are reported rather than fixed.

### `pnpm check-evidence`

Still no verdict, and this is the third round it has been attempted: the sweep holds a machine-wide lease, another session's rig held it throughout, and every attempt answered `DEFERRED` (exit 75, "no frames checked") — reported as the result it is rather than fought for. What does not need the lease is green: `assertFramePaints` runs on every frame at capture, `scripts/evidence-manifest.test.mjs` checks the `srcTree`/`scriptsTree`/`frames`/`surfaces`/`themes` stamps inside `pnpm test:desktop` (34/34 with its neighbours at this head), and the sweep's `.webp` walk does not read these PNGs at all.

## Not addressed

- The pre-existing `test:desktop` failure above (reproduced on the base tree, outside this diff).
- The run panel's row, the probe's other answers and the key-entry path: out of scope and untouched.
- The design round's two recorded follow-ups (the equal-weight `failed`/`cancelled` pair; the three-control footer when a recorded press failure coincides with a running operation) — pre-existing, out of this diff, recorded in the thread.


damianvtran:
### Agent review — round 1

Reviewer: reviewer — `deepseek/deepseek-flash` (primary provider; no fallback occurred this round)
Scope: cfc28c817..c92d9d6d8

**Verdict: clean.** No BLOCKER and no MAJOR. One MINOR and four NITs below, all follow-up grade — this round is TERMINAL for the fix as it stands.

The diff under review is a predicate plus one footer condition in `src/renderer/src/features/chat/components/run-details/mcp-auth-dialog.tsx` (`completed = operation?.status === "complete"`, `state.kind === "running" && !running && !completed`), one registration line in `package.json`, one row in `scripts/chrome-keychain.test.mjs`'s launch-site table, the `srcTree`/`scriptsTree` stamps in `docs/evidence/manifest.json`, a new test, and a bespoke rig with eight PNG frames. No other file under `src/` is touched, and the run panel is untouched.

#### 1. The state machine, state by state

| render state | newest op for the row | controls | before → after |
| --- | --- | --- | --- |
| `probing` / `grant` | n/a | probe primary / `Continue in browser` | unchanged |
| `running`, read has carried no op yet (`operation === undefined`) | none | Close alone (`running` is true via `!operation`), body `Waiting for backend sign-in status…` | unchanged |
| `running` | `running` | Close + `Cancel sign-in` | unchanged |
| `running` | `complete` | **Close alone** | this change |
| `running` | `failed` / `cancelled` | Close + `Try again` | unchanged |
| `running` | a status this build cannot read | Close + `Try again` (`completed` is false) | unchanged |
| `notice` (the probe answered) or a recorded press failure | any | Close + `Try again` — `failure`/`notice` satisfy the branch independently of `completed` | unchanged |

`completed = operation?.status === "complete"` is the right predicate: it reads the wire's own word instead of the absence of `running`, so an operation whose status is a word from a newer backend is not promoted to a success and keeps its retry — the case the comment names, and the predicate does implement it.

Newest-wins agrees with the fold the panel uses. `operation` is the newest op for the row by `created_at` (dialog lines 231-236), and `mcpGrantStates` folds newest-per-name and then drops a newest `complete` (`run-detail-model.ts:2473-2508`), so the dialog and the row pick the same latest statement: an earlier `complete` sitting beside a later `failed`/`cancelled` reads as a failure in both, and a newest `complete` suppresses the retry here while producing no row state there. A `logout` op that settles `complete` does satisfy `completed` here — the dialog is name-scoped, not action-scoped, exactly as the fold is — but it cannot be the newest op while the reader's own press is in flight: the backend serialises grants per session (`local_operator/mcp/desktop.py:218-230`) and the dialog's press always creates a newer op. Ties and absent instants are a latent divergence, reported as R1-4.

#### 2. Is `Close` alone enough

Yes, and it does not strand the reader. A completed sign-in is the end of the flow, so the dialog's own surface needs no second control — and the route back is one press away where the reader already is: the run panel keeps the row's remedy whenever the row's own status is still a problem (`run-detail-model.ts:2321`, `run-detail-mcp.tsx:279-297`), and that remedy is `grant` (`MCP_CONTROL_WORD.grant`, "Grant account access"), whose press reopens this dialog (`run-detail-mcp.tsx:542`) where a fresh mount re-probes and offers `Continue in browser`. Settings carries the same route. Nothing in the dialog itself re-signs-in without closing, which is the honest shape for a state whose sentence is `Sign-in complete.`

#### 3. Verified by running, not by reading

- `node --test scripts/mcp-auth-complete-no-retry.test.mjs` on this head: **3/3 pass**.
- **The discrimination is genuine.** The same file, run against a base-tree materialisation of `origin/main` (the actual base commit, extracted with `git archive` outside the repo): the `complete` case **FAILS** with `actual: ['Close','Try again']` vs `expected: ['Close']`, while both control cases still pass. The test exercises the shipped component through the real `window.api.desktop.request` seam and drives the state through the dialog's own `Continue in browser` press and its own 1s `mcp.list` poll — nothing is forced into React state.
- Neighbouring surfaces green: `scripts/mcp-auth-surface.test.mjs`, `scripts/run-detail-model.test.mjs` (including "a row's grant state is the newest operation the read carries" and "an operation the backend FINISHED is not a row state, and never deletes a remedy"), `scripts/chrome-keychain.test.mjs`, `scripts/evidence-manifest.test.mjs`, `scripts/test-inventory.test.mjs` — 127 tests, 0 fail. The keychain row matches the rig's Chrome `spawn` (`capture.mjs:176`, `withMockKeychain` applied); its later `spawn` is the Vite server, which is not a Chrome site and is not this table's subject.
- **The frames reproduce.** I re-ran the rig myself (`--tree=after --port=5321 --out=/tmp/…`) and all six committed after-frames came back **byte-identical** (md5), with the rig's own pre-shutter assertions firing: `complete` → `["Close"]`, `failed`/`cancelled` → `["Close","Try again"]`, dialog 320x174 at 340,263 in the 1000x700 viewport, every theme confirmed in `document.documentElement.dataset.theme`. Then `--tree=before`, from a base tree (`git archive cfc28c817`, harness copied in, per the README), reproduced `complete-before/…` **byte-identically** with the footer asserted as `["Close","Try again"]`. So the before half is a genuine base-tree capture and the after half is reproducible rather than a sampled state. Nothing was written into the repository (worktree still clean, output to /tmp).
- Pixels, decoded independently of the rig: the before/after pair differs by **15,612** pixels in the dark theme and **15,632** in the light, both inside **one 375x72 box at +206+258** — the README's bbox, position, percentages and dark count are all exact (light count: R1-2). Viewed directly, `complete-before` shows `Sign-in complete.` with Close + Try again, `complete-after` shows the same sentence with Close alone, and `failed-after`/`cancelled-after` still show both controls under `Sign-in failed.`/`Sign-in cancelled.` in both brand themes.
- `docs/evidence/manifest.json`: only `srcTree` and `scriptsTree` changed, and both equal `git rev-parse c92d9d6d8:src` / `:scripts`; `frames`, `surfaces`, `head` and the notes are untouched, which is right for a rig outside the sweep.

#### 4. `pnpm check-evidence` — stated plainly

It still has **no verdict**. I ran it on this head during this round: `Evidence check DEFERRED: another sweep (or its image child) holds the machine lease; retry after it finishes. No frames checked.` (exit 75). **This is not a blocker for this diff.** The sweep walks `.webp` under `docs/evidence` and these frames are PNGs from a bespoke rig it does not walk; the fields a sweep would re-derive (`frames`, `surfaces`) are unchanged here, and the two stamps it does own are correct. The pixels are instead guarded by the rig's own `assertFramePaints` on every frame it writes — which I watched pass six times in my own re-run — and by the independent byte-for-byte re-capture above, which is stronger evidence than the sweep would give for this surface.

#### Findings

**R1-1 — minor — the documented "unreadable status still keeps the retry" boundary is pinned by nothing.**
`mcp-auth-dialog.tsx:240-247` and `docs/evidence/mcp-auth-complete/README.md` both advertise that an operation whose status this build cannot read is not treated as a success. The new test covers `complete`, `failed` and `cancelled`; its two controls prove the change is scoped, but nothing pins the boundary the comment spends a paragraph on. A fourth case in `scripts/mcp-auth-complete-no-retry.test.mjs` — the bridge answering `status: "succeeded"` (a word outside `MCP_GRANT_STATUSES`) and asserting `["Close","Try again"]` — would pin exactly that line. Cheap, and it is the one branch of this predicate a future refactor could fold away silently. Not a reason for another review round.

**R1-2 — nit — the README's light-theme differing-pixel count is off by one.**
`docs/evidence/mcp-auth-complete/README.md` says the pair differs by 15,633 pixels in the light theme; two independent decodes (raw PNG inflate, and a `sips`→BMP cross-check) both give **15,632**. The dark figure (15,612), both percentages, and the one 375x72 box at +206+258 are exact, and I re-captured the frames byte-identically, so this is prose arithmetic only.

**R1-3 — nit — the before tree is labelled with a commit that is not the base.**
The README's reproduce section names `origin/main (da9e75a616)` for the before capture. `da9e75a616` was `origin/main` before #284; the fix commit's parent — and this PR's base — is `cfc28c817`. Immaterial in substance (the dialog is byte-identical between the two, and I reproduced the committed before frames byte-for-byte from `cfc28c817`), so this is the label only: naming the base commit, or "the base of this PR", would be exact.

**R1-4 — nit — `completed` and the fold would disagree on ties and on an absent `created_at`.**
The dialog picks its newest op with `[...operations].sort((a, b) => b.created_at - a.created_at)[0]` (stable, so the first-listed op wins a tie) and does not normalise a missing instant, while `mcpGrantStates` keeps the **last** op on equal `createdAt` and maps an absent instant to `0` (`run-detail-model.ts:2478-2482`). Latent only: `desktop.py:236` always writes `time.time()`, so no two operations for one server share an instant and none lacks one. Worth knowing because `operation` now decides which controls the footer offers, not only what the body says.

**R1-5 — nit — the D1 comment's `Continue in browser` is the `login` label.**
The added comment (and the README) say the retry returns the reader to "its own primary `Continue in browser`". For `action === "reauth"` that primary reads `Replace grant and sign in` (`mcp-auth-dialog.tsx:323-325`). The mechanism the comment describes is exactly what the code does — `Try again` → `setAttempt + 1` → the probe re-runs → `transport_oauth_supported === true` → `state: "grant"` → the primary returns — so this is wording precision only.

#### Considered and not findings

The row is genuinely unaffected (`mcpGrantStates` drops a newest `complete`, `run-detail-model.ts:2505-2508`, and the panel's `grant` field is gated on the row being a problem or the grant running). The `package.json` one-liner is the required registration of the new test (pinned by `test-inventory.test.mjs`, which passes). The `chrome-keychain.test.mjs` row is the correct shape for this rig and its count matches the file. The committed PNGs are new evidence outside the `.webp` sweep and inside a rig that asserts its own state, which is the right home for a surface Storybook's bridge-less preview cannot render. The frames deliberately do not cover the packaged Electron path, the press path, hover/focus, or the rest of the dialog, and the README says so itself.

damianvtran:
### Design review — round 1

Reviewer: designer — `deepseek/deepseek-flash` (primary provider; no fallback during this round)
Scope: cfc28c817..c92d9d6d8 — the rendered surface only (`docs/evidence/mcp-auth-complete/**`, both brand palettes), plus re-runs of the committed rig and six states it does not frame.

**Verdict: TERMINAL — no BLOCKER, no MAJOR.** Two NITs (D1, D2), neither a reason for another round. The completed footer is correct, honest against its neighbours, and clean in both brand themes.

#### What I looked at (pixels, not source)

- All eight committed frames, viewed as images, in both themes.
- The committed rig re-run in this worktree: my `complete` frames are **byte-identical** to `complete-after/*` (md5 `07455061db79` dark, `ce91cba5b52e` light), so the extra states below come from the same renderer path as the committed frames.
- Six states the committed set does not frame, captured from the shipped component behind the same `window.api.desktop.request` seam: `probing`, `grant`, `running`, the probe's warning `notice`, a recorded press failure, and a `failed → complete` flip with keyboard focus parked on `Try again`.
- My frames (scratch, nothing written into the repo; the worktree is untouched): `/tmp/mcp-auth-design/out/<case>/localOperator{Dark,Light}.png`, rig at `/tmp/mcp-auth-design/`.

#### Judged

**1. Does the completed state read as a success? It reads as a settled success — quiet, but no longer contradicted.**

- `Sign-in complete.` is `ink-muted` (say `181,175,162` dark / `86,81,71` light) — the same role and the same 7.16:1 / 7.82:1 as `Sign-in failed.` and `Sign-in cancelled.` The three sentences are typographically identical; only the words differ.
- The completed and failed frames differ in exactly two bands: the sentence (`y 181-207`) and the footer row (`y 258-329`). No other pixel moves. So a settled success is carried by its words plus the absence of a retry, and by nothing chromatic.
- That is the fix working as intended (it removes a false *failure* signal) rather than an oversight, but it is a choice worth recording — see D1.
- No `success` semantic is spent anywhere in the dialog, and none needs to be: the accent is spent once, on the grant primary or a body remedy (measured `rgb(56,201,106)` dark / `rgb(20,120,66)` light, one button), which is inside § 2's budget.

**2. A lone right-aligned `Close` is right — and it is not a new shape for this dialog.**

- The dialog already ships exactly this footer in a state this PR does not touch: my `probing` frame renders `Close` alone at device `422-621 × 258-329`, the *same box to the pixel* as the completed frame's `Close`. The two frames differ in 4,807 px (dark) / 4,835 px (light), all of them inside the sentence band (`x 95-604`, `y 181-207`). The settled success joins the shape the dialog already opens with, rather than inventing one.
- It does not read as a truncated footer. Measured: the button row keeps its full height (72 device px / 36 CSS), the panel keeps `320x174` in all four of `probing`/`complete`/`failed`/`cancelled` (frame `704x412` in each), and the surviving control's right edge sits **25.5 CSS px** from the panel's right edge and **25.5 CSS px** above its bottom edge — the footer's own padding, identical to the two-control frames. The freed slot does not collapse the row, the panel does not resize, nothing is clipped, and `DialogFooter` draws no rule that would suggest a cut.
- The retained control is untouched: crop-for-crop, `Close` in the before frame's left slot vs the after frame's right slot differs by 46 (dark) / 47 (light) subpixel-AA pixels out of 14,400 — same `sunken` fill, same `border-control`, same `ink`.
- Right-alignment is correct here: the lone action occupies the slot the footer's second control held, so the panel's right padding is unchanged, and the freed ~190 px of content box is plain ground.

**3. The boundary is honest.**

The mapping is now one sentence shape to one control shape — `complete → Close`, `failed`/`cancelled` → `Close` + `Try again` — where before, `complete` shared its control shape with the two failures. That shared shape is precisely the read the operator reported. Checked against the neighbours:

- `failed` and `cancelled` still read as non-successes in the same words they always used, and their retry is an honest control (for `cancelled` the reader caused it, and re-running the probe is still the way back); the sentence gains no claim it did not make before.
- Nothing in the completed frame reads as an error: no danger or warning ink, no icon, no failure register, and the footer no longer contradicts the sentence above it.

**4. Labels and contrast.**

- Keep `Close`; do not switch it to `Done`. The footer's dismissive is `Close` in every state I looked at (`probing`, `grant`, `running`, `notice`, failure — five frames), so renaming it in one state makes the same button change its name as the poll settles. `Done` also claims an acknowledgement the state does not ask for, and § 8's "say what happens" is already carried by `Sign-in complete.` — sentence case, hyphenated noun, event stated. No change recommended.
- Contrast of the remaining control, measured from the frames: `border-control` on the dialog ground **3.77:1** dark / **3.95:1** light (clears § 3's 3:1 floor for a control's only boundary); ink on its own fill 14.93:1 / 15.64:1. The light palette's fill step off the ground is 1.05:1, i.e. the control is effectively a bare outline there — pre-existing, byte-identical in the before frame, and the enforced floor is the border, which passes.

**5. What the frames show that the claim does not.**

- The claim "the differing pixels form one 375x72 box at +206+258" is exact — I reproduce 15,612 px dark / 15,632 px light in that same box. The gloss on it ("the footer's right half, where `Try again` sat") fuses two mechanisms into one: the freed rectangle is the **left** slot (device `206-405`, 200x72), and the surviving button changed only its **label** in the right slot (`422-621`). The diff's right edge at 580 is label ink inside the surviving button's box, not a box edge; that button's border box is identical in both frames.
- `Close` does not move within the pair under review, but it does move across states — the footer's right alignment shifts it as its neighbours come and go: panel-left +25 in `grant`, +58 in `running`, +195 alone in `probing`/`complete`. Pre-existing behaviour, and the completed state lands on the position the dialog's own probing state already uses. No action.
- Frames cannot show focus-dependent rendering, so I measured the one path this change creates — a focusable control disappearing under the reader. With the poll answering `failed` first (footer `Close` + `Try again`), keyboard focus on `Try again`, and the poll then answering `complete`: focus does **not** fall out of the dialog. The focus scope returns it to the dialog's own container (`outline: none`, so no stray ring around the panel), and the tab order reaches `Close`. The `focusflip` frame is byte-identical to the `complete` frame in both themes (0 differing pixels), i.e. a settle-in-place shows the same screen as a fresh success.

#### Findings

**D1 — NIT — the completed state carries no positive success signal; the choice should be recorded rather than implied.**
The fix takes the failure-shaped control away from a success, and the state then rests entirely on the words: measured above, `Sign-in complete.` shares its ink role, its size and its pixel treatment with `Sign-in failed.` and `Sign-in cancelled.`, and no `success` semantic is spent. My judgement is that this is the right end state — § 7's "a completed action is one quiet line, not a card" and § 0's legibility-over-spectacle both point at quiet, and a celebratory success under a bare `failed` sentence would make success louder than failure. Recorded so the choice is deliberate, with one constraint if it is ever revisited: do it for the `complete`/`failed`/`cancelled` trio in one round, never for `complete` alone. No change requested.

**D2 — NIT — optional evidence: the frames do not show the state that is this change's own design precedent.**
The committed set frames `complete-before`/`complete-after` and the two controls, but not `probing`, which is the state that already renders a lone right-aligned `Close` at the same box (measured in section 2). That frame, or one sentence in the README noting that `probing` already reads this way, makes the "one control under a settled sentence is the dialog's own shape" argument self-evident to a reader who will not re-run the rig. Evidence-only; nothing in the behaviour depends on it, and my scratch frames above are available if they are useful.

#### Follow-ups (real, but not this PR's doing — recorded, not requested)

- The bare `failed`/`cancelled` frames put two visually identical controls in the footer: `Try again` is a secondary with the same fill, border and ink as `Close`, and no accent is present anywhere in either dialog (measured: no accent fill on any control in those frames), though § 2 spends the accent on the primary action and `Try again` is the only action of consequence in that state. Pre-existing — the before frame has the same two look-alike controls — and I did not want a third button treatment pushed into this diff.
- A recorded press failure that coincides with a still-`running` operation for the same row renders a three-control footer (`Close` + `Try again` + `Cancel sign-in`; my failure frame: a 496x323 dialog with three controls in the footer row). Reachable — `state.kind` stays `running` after a failed press, `failureFor` returns the failure, and the poll's last answer carried a running op — unchanged by this PR, and the one footer composition where a reader has to sort three equal-weight controls.
- Not addressed by this review, and stated rather than implied: the frames prove the renderer's own predicate, not the packaged Electron path, the press path, or hover/animation. The README says so itself.

damianvtran:
### QA report — round 1

Independent QA pass on PR #283 (`fix/auth-complete-no-retry`), run from my own copies of both revisions — HEAD `c92d9d6d876f770fdfd333bba2957d68d1225963` and base `cfc28c817` (the current `origin/main`) — in `/tmp/qa283/{git-head,git-base}` (local git clones, so `git`-backed tests behave as in CI) plus two `git archive` extractions used only for the capture rig. I did not edit the repository, did not add a reviewer, and mentioned nobody.

**Verdict: PASS** — no FAIL stands on the head. Two cells are BLOCKED and named at the bottom.

#### Environment

- Node v26.5.0, pnpm 10.30.3, macOS 25.6.0 / arm64.
- Rig: the PR's own `docs/evidence/mcp-auth-complete/harness/capture.mjs`, run against **both** revisions by me (base tree materialised outside the repo, harness copied in, `node_modules` linked alongside, as the README's `before` recipe does). My ports 5313/5314/5317/5318; frames in `/tmp/qa283/out-head/**` and `/tmp/qa283/out-base/**`.
- Test suites: `pnpm test:desktop` (via `scripts/run-desktop-tests.mjs`, concurrency 7) in `/tmp/qa283/git-head` and `/tmp/qa283/git-base`, sequentially, with `env -u NO_COLOR TERM=xterm-256color`.
- Real app: built head (`pnpm build` in my clone), launched `LOCAL_OPERATOR_UI_WINDOW_MODE=headless LOCAL_OPERATOR_UI_HEADLESS_KEEP_ALIVE=1` with an isolated `--user-data-dir`, `HOME`, `TMPDIR`, `LOCAL_OPERATOR_CONFIG_DIR`/`HOME`/`LOG_DIR` under `/tmp/qa283/app-scratch`. `[window-mode] state: visible=false focused=false focusable=false` on every run — no window was raised. No inherited `CMUX_*` variable existed (asserted before launch). The operator's live backend daemon on `127.0.0.1:1111` was left untouched and never used.

#### Matrix

| # | Surface | Command | Actual output | Result |
|---|---|---|---|---|
| 1 | Rig, base `complete` | `node docs/evidence/mcp-auth-complete/harness/capture.mjs --tree=before --out=/tmp/qa283/out-base --port=5314` | 2 frames; `complete-before/*.png: footer ["Close","Try again"], body "Sign-in complete."` | PASS |
| 2 | Rig, head `complete`/`failed`/`cancelled` | same, `--tree=after --out=/tmp/qa283/out-head --port=5313` | 6 frames; `complete-after: footer ["Close"]`; `failed-after`/`cancelled-after: footer ["Close","Try again"]`, bodies `Sign-in complete./failed./cancelled.` | PASS |
| 3 | Rig discrimination (negatives) | `--tree=before` on head; `--tree=after` on base | head: `Error: complete/localOperatorDark: on the \`before\` tree the footer must still carry \`Try again\`, and it reads ["Close"]` (exit 1). base: `... on the \`after\` tree the footer must not carry \`Try again\`, and it reads ["Close","Try again"]` (exit 1) | PASS |
| 4 | Frame provenance | `md5` of my 8 frames vs the 8 committed PNGs | all 8 byte-identical (`complete-after/dark` `07455061db79f56133b1c9c6d4e626fe` both sides; `complete-before/dark` `a47359e120633b03b356d28eb72864f9` both sides) | PASS |
| 5 | Frames, viewed | `read` on my captures | base: `Sign in to hubspot` / `Sign-in complete.` / `Close` + `Try again`. head: same body, `Close` alone. head `failed`: `Sign-in failed.` + both. head `cancelled` (light): `Sign-in cancelled.` + both. Dialog box 320x174 at 340,263 in all six | PASS |
| 6 | New test on head | `node scripts/mcp-auth-complete-no-retry.test.mjs` | `pass 3 fail 0` | PASS |
| 7 | New test against the BASE component | same file copied into the base tree, run there | `pass 2 fail 1` — `a completed sign-in offers Close alone` fails with `['Close','Try again']`; the `failed`/`cancelled` cases pass | PASS (discriminates, and only there) |
| 8 | Neighbouring MCP suites on head | `node scripts/{mcp-auth-surface,mcp-failure-routing,mcp-list-query,mcp-target-resolution,run-detail-model}.test.mjs` | 6 / 8 / 4 / 7 / 87 passed, 0 failed | PASS |
| 9 | QA probe — a status word this build cannot read | `node scripts/qa-probe-unknown-status.test.mjs` (mine) | head `pass 4 fail 0`; base `pass 3 fail 1` (only the completed control fails). Cases: `succeeded` → `["Close","Try again"]`; newest `running` → `["Close","Cancel sign-in"]`; `complete` → `["Close"]`; `complete` then newer `failed` → `["Close","Try again"]` | PASS |
| 10 | Run-panel row for a completed grant | `node grant-row-probe.test.mjs` (mine, esbuild + `mcpGrantStates`/`deriveMcpServers`) | head `pass 9 fail 0`, base `pass 9 fail 0` — identical. A newest `complete` op yields no entry (`states.size === 0`), `deriveMcpServers` returns `grant: null` for it, `mcpGrantInFlight([complete])` is `false`; `running`/`failed`/`cancelled` still produce entries | PASS |
| 11 | The other two entry points (Settings Integrations, `/mcp` picker) | `node scripts/qa-probe-entry-points.test.mjs` (mine; the shipped `McpManagementSection` and `McpPicker` behind a stubbed `window.api.desktop.request`) | head `pass 5 fail 0` — both entry points show `["Close"]` for `complete` and `["Close","Try again"]` for `failed`/`cancelled`. base `pass 3 fail 2` — exactly the two `complete` cases fail | PASS |
| 12 | Full `pnpm test:desktop`, head | `pnpm test:desktop` in `/tmp/qa283/git-head` | `tests 2197, pass 2195, fail 2`: `python-bytecode-cache` "every backend spawn carries the prefix even with the shell-env load unresolved" (`global install: start() should report a started backend`), and `test-inventory` "every scripts/*.test.mjs is run by test:desktop or exempt with a reason" naming `scripts/qa-probe-unknown-status.test.mjs` — **my own** probe file (see below) | PASS (no diff-caused failure) |
| 13 | Full `pnpm test:desktop`, base | `pnpm test:desktop` in `/tmp/qa283/git-base` | `tests 2194, pass 2192, fail 2`: the same two tests, same messages | PASS |
| 14 | Evidence sweep | `pnpm check-evidence` (head tree), twice | `Evidence check DEFERRED: another sweep (or its image child) holds the machine lease; retry after it finishes. No frames checked.` — exit 75 | BLOCKED |
| 15 | Built app, the completed-state dialog | `pnpm build` + headless Electron against a loopback stub backend I own | App boots headless (no focus), renders the shell, and its own main-process transport reaches the stub (`GET /v1/capabilities`, `/v1/desktop/sessions`, `POST /v1/desktop/radient`, `/v1/config`, `/health` in the stub log). The **Settings** surface errors under stub payloads (`TypeError: Cannot read properties of undefined (reading 'filter')`), and the completed op itself cannot be produced | BLOCKED |

#### Reading of the numbers

- **The full-suite delta is exactly the new test file: +3 tests at head, +3 passes at head; the failing set is identical.** head 2197/2195/2, base 2194/2192/2, with the same two failing tests in both. The `python-bytecode-cache` failure is the machine-install-dependent one the author reported; the `test-inventory` failure is **my** file: `qa-probe-unknown-status.test.mjs` lives in my copies of both trees, and `scripts/test-inventory.test.mjs` correctly refuses a `scripts/*.test.mjs` no CI step runs. It appears in both runs, so the comparison stays symmetric; nothing of mine is in the repository. Excluding it, head has exactly **one** failure — the pre-existing one — and so does base.
- I first ran the suite in a `git archive` extraction that is not a git repository; there, `conflict-markers` (`fatal: not a git repository`) and `evidence-manifest` ("`head` 344391224 is not an ancestor of HEAD") also failed. **Both pass in a real clone**, and I checked the citations directly: `344391224a9d2d9…` and `559327ecf…` are both ancestors of `c92d9d6d8`. Those two failures were artifacts of my extraction method, not of the change or of the PR's manifest.
- The manifest's `srcTree`/`scriptsTree` re-derivation this PR carries is covered by `evidence-manifest.test.mjs`, which passes in my head run.

#### Findings

**Q1 — INFO.** The `completed` predicate is the newest operation **for the row**, so the retry is withheld whenever that operation already reads `complete`, including briefly at the start of a *new* sign-in: after a press, until the client's next poll sees the backend's fresh `running` operation, the footer shows `Close` alone while the body still reads `Sign-in complete.` I could not construct a wire payload that distinguishes this from the intended state — the two are byte-identical on the wire — and the body already said `Sign-in complete.` on base too, so the fix does not add a new misreading; the backend also stamps the new operation synchronously at accept time with a newer `created_at` (`~/local-operator/local_operator/mcp/desktop.py:229-239`), so the newest-wins fold picks `running` as soon as the control answers. Recorded as an observation, not a defect; no action requested.

**Q2 — INFO.** `docs/evidence/mcp-auth-complete/README.md` names `origin/main` as `da9e75a616`; the PR's base is now `cfc28c817` (the 0.25.17 release merge). Verified harmless rather than assumed: my `--tree=before` capture taken at `cfc28c817` reproduces the committed `complete-before` frames byte-for-byte (md5 above), so the before/after pair is still a pair against the current base. A one-line refresh of the SHA would save a reviewer the check.

**Q3 — INFO.** The capture rig has no protection against two concurrent runs over trees that share one `node_modules`: Vite's dep cache lives under that shared directory, and the second rig's page then misses its 60 s budget and dies with `the harness never reported the state it settled in` — which reads as a state failure rather than a load timeout. I hit it once and recovered by running the rigs one at a time; every frame in this report came from a solo run. (The rig's teardown also leaves its Vite child behind on the error path; I killed my own by pid.)

#### BLOCKED

- **`pnpm check-evidence` (cells 14).** The machine-wide lease at `/tmp/local-operator-ui-check-evidence.lock` is held by a peer — twice, ~25 minutes apart, both `DEFERRED`/exit 75. The sweep therefore re-read no frame. What that leaves uncovered: the `.webp` sweep and the surfaces/frames counts. What *is* covered without the lease: frame painting, which the rig itself asserts (`assertFramePaints` on each of my 8 frames, all written and all non-blank), and the manifest's `srcTree`/`scriptsTree`/head-citation freshness, which `evidence-manifest.test.mjs` checks in my head `pnpm test:desktop` run (passing). The PR's own PNG frames are outside the sweep by design (it walks `.webp`).
- **Reaching the completed-state dialog in the BUILT app (cell 15).** Two independent reasons, and I am not claiming coverage I do not have. (a) The state needs an operation the backend has actually settled `complete`, and that word is only ever written after a real interactive exchange: `~/local-operator/local_operator/mcp/desktop.py:246` sets `op["status"] = "complete" if style == "success" else "failed"` around `run_grant`, which awaits `manager.connect_configured_server` (`mcp/grants.py:156-215`, `GRANT_TIMEOUT_MS` 600 s) — a browser round trip against a real authorization server. (b) The app cannot be stood up on a substitute backend faithfully enough to reach the surface: with a stub rich enough to get the shell rendering, the Settings page throws before the MCP section mounts, and its direct renderer fetches are pinned by CSP to `localhost:1111`/`8080`, where the operator's own daemon is listening — a port I left alone by design. What the rig-based evidence therefore proves and does not prove: it proves the **shipped component's footer predicate** against real wire-shaped payloads at the exact `window.api.desktop.request` seam the product uses (the same seam `scripts/mcp-auth-surface.test.mjs` pins), on both revisions, and that the other two call sites inherit it. It does **not** prove anything about the preload hop, the main-process transport, or a real backend's `mcp.list`/`mcp.control` payloads — and I could not close that last hop here. I exercised the main-process transport only as far as "the app's own requests arrive at a backend I own", which is not the same claim.

#### Reproductions for the FAILs

There are none to reproduce: no QA cell failed on the head. The two failures I report as non-verdicts are the pre-existing `python-bytecode-cache` failure (identical at base, and present without any of this PR's code) and my own probe file tripping `test-inventory` in both trees.

#### Gaps I did not close

- `pnpm check-evidence` verdict (BLOCKED above).
- The built-app completed state (BLOCKED above).
- The `Try again` press path itself (what a press *performs*) — the change is about which controls are offered, and the press's transition owner is stubbed in every rig I ran, including the author's.
- The probe's other answers (key entry, the "cannot determine" notice), the recorded press failure and the granting phase — unchanged by this diff and not re-driven beyond the cases in the matrix.

damianvtran:

Head: `4dfa55874` (rebased onto `5c53e1c75`, the 0.25.18 window, during this round — see the note at the end). One commit, `4dfa55874` itself; **no file under `src/` changed**, and the dialog's blob is byte-identical to the one this round reviewed (`c4ea7c9c1`).

**R1-1 — FIXED (`4dfa55874`).** The fourth case is in `scripts/mcp-auth-complete-no-retry.test.mjs`: the bridge answers `succeeded` (outside `MCP_GRANT_STATUSES`) and the assertion is `["Close","Try again"]` — a word this build cannot read is not a success and keeps the retry.

```
$ node --test --test-concurrency=1 scripts/mcp-auth-complete-no-retry.test.mjs
✔ a completed sign-in offers Close alone, under the sentence that says it worked (198.013333ms)
✔ a failed sign-in keeps Close and Try again (115.960084ms)
✔ a cancelled sign-in keeps Close and Try again (89.453958ms)
✔ a status this build cannot read is not a success, and keeps the retry (83.712292ms)
ℹ tests 4   ℹ pass 4   ℹ fail 0
```

Two runs you asked for, and the second is the one that discriminates:

- **Against the base component the file reads 3 pass / 1 fail**, and the case that fails there is the `complete` one, by design (`actual ['Close','Try again']` against `expected ['Close']`). **The new boundary case passes on the base.** Correct and expected, and the file's own header says so: the base is right about this boundary too, so the fourth case is not a differential one, and the `complete` case remains the case that discriminates. Corrected in place: the "(4/4, base tree restored over the working tree)" this bullet carried in round 1 was wrong, and the cause is worth naming — the run behind it used `git show HEAD:`, which resolves to the FIXED file once the fix is committed, so the base was never actually measured. The base run stated here is measured at this head, against main's blob `f54f93151`, with this branch's test file.
- **Shown to discriminate by mutation, and only by the mutation that isolates it.** What a refactor could fold away is the boundary, so the mutation that proves anything about it is one that folds an *unrecognised* word into settled — `operation?.status === "complete" || !["running","complete","failed","cancelled"].includes(operation.status)`:

```
✖ a status this build cannot read is not a success, and keeps the retry
  AssertionError: an unreadable status is not a settled success
  actual: [ 'Close' ]   expected: [ 'Close', 'Try again' ]
ℹ tests 4   ℹ pass 3   ℹ fail 1
```

That is the isolating result: **3 pass / 1 fail, the boundary case the only failure** — the fix's own state and the two controls still pass, so the case is policing exactly the line it names. The mutation quoted in round 1 (`operation?.status !== "running"`) is weaker evidence than that sentence implied, and is restated for what it is: it withholds the retry from `failed` and `cancelled` as well, so it fails three cases (**1 pass / 3 fail**), which shows the predicate is broadly wrong rather than showing the boundary is what is being policed. Both mutations were applied to the working tree only and reverted; `git diff --stat -- src/` is empty and the dialog's blob is unchanged (`c4ea7c9c1`).

**R1-2 — FIXED (`4dfa55874`).** The light-theme count is **15,632** in the README. Measured here two ways: this rig's own channel-by-channel PNG decode gives **15,612** dark / **15,632** light, and `magick compare -metric AE` gives 15,612 dark and **15,633** light — one pixel more, because it classifies one anti-aliased edge pixel differently. The README now names both numbers and both methods rather than asserting agreement, which is what the earlier sentence got wrong.

**R1-3 — FIXED (`4dfa55874`).** The README's reproduce block names `cfc28c817` (this PR's base) as the before tree and states both facts: the frames were first captured from `da9e75a616` and re-taken from `cfc28c817`, byte for byte, the dialog file being identical in both trees. I ran that re-capture myself in this round rather than only quoting yours:

```
complete-before/localOperatorDark.png  md5 a47359e120633b03b356d28eb72864f9  (committed == re-taken)
complete-before/localOperatorLight.png md5 6e830d7bdb12fe898552e20f3b7b5565  (committed == re-taken)
```

**R1-4 — REJECTED, reason given.** Left as it is, deliberately, and the asymmetry is real: the dialog picks newest-by-`created_at` with a stable sort (first-listed wins a tie) and does not normalise a missing instant, while `mcpGrantStates` keeps the last on a tie and maps an absent instant to `0`. It is unreachable on the wire: `local_operator/mcp/desktop.py:236` stamps every operation with `time.time()`, so no two operations for one server share an instant and none lacks one. Making the dialog mirror the fold's last-wins rule would change which operation decides the footer's controls for a state the backend cannot produce — a behaviour change with no case to test it against, which is the wrong trade for a latent divergence. Recorded here rather than silently: it is worth knowing because `operation` now decides the controls, so a backend that ever stops stamping an instant would open it.

**R1-5 — FIXED in the README, DEFERRED for the component comment.** The README now says "its own primary — `Continue in browser` for a sign-in, `Replace grant and sign in` when the `action` is a re-auth". The identical phrase in `mcp-auth-dialog.tsx`'s comment is left for now: this round's constraint is that `src/` stays byte-identical so the QA and design rounds on `c92d9d6d8` stay fresh, and a comment-only `src/` change would move both round scopes for a wording diff. `deferred — rides the next round that touches this component; the mechanism the comment describes is correct for the `login` action, which is the dialog's default entry.`

### The rebase this round had to carry

`main` moved to `5c53e1c75` (the 0.25.18 window) while this round was in flight and the branch went `DIRTY` again — which means no CI at all on the new head — so it was rebased a second time. `docs/evidence/manifest.json` and `package.json` both conflicted; both were resolved by taking main's side and re-applying this branch's own delta: `package.json` keeps `0.25.18` with this branch's one `test:desktop` entry (no version line), and the manifest carries main's content with `srcTree`/`scriptsTree` re-derived from the rebased tree.

What moved and what did not: this branch's file set is unchanged, and its own product change is byte-identical (the dialog blob `c4ea7c9c1` is the blob you reviewed). What the rebase added is **main's** commits — `managed-python.ts`, `bundled-python-layout.json` and the python-seed pruning scripts — which are out of this PR's scope and are why a convergence round scoped `c92d9d6d8..4dfa55874` is the manager's call rather than something this comment claims.

damianvtran:

Head: `4dfa55874` (rebased onto `5c53e1c75` during this round; the dialog's blob is byte-identical to the one you reviewed, `c4ea7c9c1`).

**No visual surface changed, so the design sign-off stays fresh on this head.** Measured rather than asserted: all eight committed frames are byte-identical after this round's commits (md5 of each PNG unchanged in both themes), not one was re-taken, and the round's edits outside `docs/` are a comment-only test addition. The rig was re-run to check that: six `after` frames and both `before` frames came back byte-identical, every frame's own pre-shutter footer assertion firing.

**D1 — recorded as the choice; no change.** The completed state keeps the words as its whole signal: no `success` semantic is spent, `Sign-in complete.` stays `ink-muted` at the same role and size as its two neighbours, and the fix's job is to remove a false failure signal rather than to add a celebratory one — `§ 7`'s one quiet line, and `§ 0`'s legibility over spectacle. Your constraint for any future revisit is the right one and is now recorded in this thread: **if the trio ever gains a positive signal, it is done for `complete`/`failed`/`cancelled` in one round, never for `complete` alone.** Nothing in this PR spends an accent, renames `Close`, or touches the other states.

**D2 — satisfied in the README (`4dfa55874`), with no new frame.** `docs/evidence/mcp-auth-complete/README.md` now carries the precedent as its own paragraph, attributing your measurement rather than restating it as mine: the untouched `probing` state renders the same lone right-aligned `Close` at the same device box (`422-621 x 258-329`) as the completed frame's, the two differing only inside the sentence band, and the panel keeping its `320x174` in all four of `probing`, `complete`, `failed` and `cancelled`. So "one quiet control under a settled sentence" is the dialog's own shape — the settled success joins it rather than inventing it — and the freed slot collapses nothing. You called the frame optional and evidence-only, so none was added: a still of a state this PR does not touch would be a new artifact for a claim a sentence answers.

**One more from the same round, fixed (`4dfa55874`).** Your section 5 was right that my gloss fused two mechanisms into one. The README's measurements now separate them: the box's rectangle covers the retired control's own columns (`206`-`405`, **14,308** of the differing pixels) plus the surviving button's label ink (**1,304** dark / **1,324** light), the surviving button keeping its own 100 CSS px box. That is your reading, and it reproduced here independently.

The two follow-ups you recorded — the bare `failed`/`cancelled` footer's two equal-weight controls, and the three-control footer when a recorded press failure coincides with a still-running operation — are pre-existing, out of this diff, and not folded in here. They are noted in this thread rather than silently dropped.

damianvtran:
### QA remediation — round 1

Head: `4dfa55874` (rebased onto `5c53e1c75` during this round). **No FAIL was raised on the reviewed head, so nothing was remediated for a failure.** What this round changed is the two things your non-FAIL notes asked for.

**Q3 — INFO, now recorded in the README AND fixed at the source (`4dfa55874`).** The README carries the limitation you hit: run one rig at a time over one `node_modules`, or the second run's page misses its 60 s readiness budget and dies with `the harness never reported the state it settled in`, which reads as a state failure rather than the load timeout it is. Your parenthetical about the teardown was a real defect in the rig rather than a note about it, so I fixed it instead of only documenting it: the Vite server is now spawned `detached` and the teardown signals its process **group**, because `vite.kill()` alone signals the `pnpm` shim and leaves the server it execs holding the port — the path you recovered from by killing your own by pid. Verified after the change: a full rig run leaves no process, no listening socket and no Chrome profile behind, and all six `after` frames plus both `before` frames reproduced byte-identically.

**Q1 — INFO, no action.** Agreed, and recorded as you stated it: the predicate is the newest operation **for the row**, so a fresh press briefly shows `Close` alone against a body that still reads `Sign-in complete.` until the next poll sees the backend's new `running` operation — and the backend stamps that operation synchronously at accept time with a newer `created_at` (`mcp/desktop.py:229-239`), so the newest-wins fold picks it up on the next read. The body already said `Sign-in complete.` on the base tree, so the fix adds no new misreading; the two states are byte-identical on the wire, which is why there is no control to test it with.

**Q2 — INFO, folded into the agent-review remediation.** The README's before-tree label now names this PR's base `cfc28c817` alongside `da9e75a616`, and states why they agree: the dialog file is identical in both trees, and the committed before frames were re-taken from `cfc28c817` here — md5 `a47359e120633b03b356d28eb72864f9` / `6e830d7bdb12fe898552e20f3b7b5565`, byte-identical to the committed pair, matching your own reproduction.

**Your two BLOCKED cells are unchanged, and are still blocked for the same reasons.** `pnpm check-evidence` was attempted again on this head and answered `DEFERRED`/exit 75 (the machine-wide lease was held by another rig for the whole round); the built-app completed-state dialog is still unreachable without a real browser round trip against an authorization server. Neither is a claim this PR now makes: the frames are PNGs from a bespoke rig the sweep does not walk, the rig asserts `assertFramePaints` on every frame it writes, and the manifest stamps are checked by `evidence-manifest.test.mjs` inside `pnpm test:desktop` — 34/34 across `chrome-keychain`, `evidence-manifest` and `test-inventory` at this head, and the new four-case suite 4/4.

The remaining gaps you named — the preload hop and the main-process transport, the `Try again` press's own transition owner, and the probe's other answers — are still not claimed anywhere in the PR, and the README's "what these frames do NOT prove" section says so.

damianvtran:
### Agent review — round 2

Reviewer: reviewer — `deepseek/deepseek-flash` (primary provider; no fallback this round)
Scope: c92d9d6d8..4dfa55874

**Verdict: clean.** No BLOCKER and no MAJOR. One NIT (R2-1). This round is TERMINAL, and it is FRESH on head `4dfa55874c583b57fa18b1de3115f0e5da9cdc72` — I re-derived everything below at that SHA, CI is 8/8 green there, and the PR reads `MERGEABLE` / `mergeStateStatus: CLEAN`.

#### 1. Rebase integrity, checked rather than taken

- **The product change is byte-identical.** `git rev-parse c92d9d6d8:src/renderer/src/features/chat/components/run-details/mcp-auth-dialog.tsx` and `4dfa55874:…` are both `c4ea7c9c1a903d4d6d6c102121d5848b4a56ff7b` — confirmed here, not carried from your note. (The pre-change blob is `f54f93151`, so what the branch authored is intact and still present.)
- **No other `src/` change is this branch's own.** `git log 5c53e1c75..4dfa55874 --name-only -- src/` names exactly one file, the dialog. The two `src/` files that differ between the old and the new head — `src/main/backend/managed-python.ts` and `src/shared/bundled-python-layout.json` — were touched by no commit of this branch; they arrive with main's 0.25.18 window.
- **No renderer file moved anywhere in the delta.** `git diff --stat c92d9d6d8..4dfa55874 -- src/renderer/` is empty, upstream's incoming commits included. Nothing in this round's tree can alter a visual surface.
- **The frames did not move either.** `git diff --name-only c92d9d6d8..4dfa55874 -- docs/evidence/mcp-auth-complete/` lists `README.md` and `harness/capture.mjs` and no PNG.
- **`package.json`, the one shared file, carries both sides.** `test:desktop` holds **121** `.test.mjs` entries (your note said 123): main's 120 plus `scripts/mcp-auth-complete-no-retry.test.mjs`, with no duplicates, and `main − new` and `old − new` both empty — nothing was lost from either side, and `scripts/prune-python-seed.test.mjs` is registered. The version line reads `0.25.18`, main's, with no bump of this branch's own.
- **`docs/evidence/manifest.json`, the other shared file, is main's wholesale.** Byte-comparing the head's manifest with main's: identical once 40-hex SHAs are masked, and the only differing fields are `srcTree` and `scriptsTree`, which now equal `git rev-parse 4dfa55874:src` (`d2db7aff71…`) and `:scripts` (`07f22b5d7a…`) rather than the stale pre-rebase pair (`12c183d6bb…` / `99e9eb1bcb…`). `frames` (4125), `surfaces` (504) and `head` are main's own values and needed no re-derivation, which is right for a rig outside the `.webp` sweep.
- **Beyond those stamps, the rebase changed nothing of this branch's.** The rest of the old-head→new-head delta is the branch's own remediation commit (`README.md`, `capture.mjs`, the test, the two stamps) plus upstream's window.

#### 2. The remediation answers, reproduced

- **R1-1 — true, and the boundary case earns its place on its own.** On this head, `node --test scripts/mcp-auth-complete-no-retry.test.mjs` is **4/4**. Against a base-tree materialisation of `origin/main`, the new case **passes** (as the file itself says it must — the base is right about this boundary too) while the `complete` case **fails**, `actual: ['Close','Try again']` vs `expected: ['Close']`, i.e. 3 pass / 1 fail; so the fourth case really is the only non-differential one, and the first is still the differential one. I then made the mutation load-bearing rather than merely plausible: with the predicate widened to `const completed = operation?.status !== "running";` the boundary case fails exactly as your transcript shows (`actual: ['Close']`, `expected: ['Close','Try again']`) — though that mutation also takes the `failed`/`cancelled` controls with it, so it does not by itself prove the fourth case is needed. The mutation that does is the one the comment describes as the real risk, an unrecognised word folded into "settled": `completed = operation?.status === "complete" || !["running","complete","failed","cancelled"].includes(operation?.status ?? "running")` gives **3 pass / 1 fail — only the boundary case fails**, the `complete` case and both controls green. That is the case doing work nothing else does. All mutations were made in a scratch copy under `/tmp` and reverted; the working tree was never touched (`git status --short` empty, HEAD still `4dfa55874`), and the scratch run is back to 4/4.
- **R1-2 — true, and now stated the way it should be.** My own channel-by-channel decode of the committed pair gives **15,612** dark / **15,632** light; `magick compare -metric AE` on this machine gives **15,612** dark / **15,633** light — one pixel more in the light theme, exactly as the README now says, so the discrepancy is a decoder's classification of one anti-aliased edge pixel rather than an error in anyone's arithmetic. The new partition also reproduces exactly: pixels in the retired control's own columns (`x 206-405`) account for **14,308** in both themes, and the remainder — the surviving button's label ink — for **1,304** dark / **1,324** light.
- **R1-3 — true.** Re-running the rig against a base tree built with `git archive cfc28c817` (harness copied in, `node_modules` linked) reproduces `complete-before/localOperatorDark.png` (`a47359e1…`) and `…Light.png` (`6e830d7b…`) byte-identically, with the run's own footer assertion reading `["Close","Try again"]`.
- **R1-4 — rejection accepted.** The divergence is unreachable on the wire (`desktop.py:236` stamps every operation with `time.time()`), and mirroring the fold's last-wins rule would change which operation decides the footer's controls for a state no backend can produce. Rejecting with the reason recorded is the right trade; the record is what I asked for.
- **R1-5 — see § 5.**

#### 3. The rig change (Q3)

The change is sound and frame-neutral. `detached: true` makes the wrapper and the Vite server it execs their own process group on POSIX, so `process.kill(-vite.pid, "SIGKILL")` in `teardown` reaches the server the `pnpm` shim would otherwise leave holding the port, with a `vite.kill("SIGKILL")` fallback when the group call throws; `teardown` is wired to `exit`, `SIGINT` and the `catch` around `main()`, which covers the error path the comment names. Measured: after my own full runs — six `after` frames from this head and both `before` frames from the base tree — **no listening socket on the ports I used, no Chrome profile directory left under the temp dir, and all eight committed frames byte-identical to the fresh capture** (`complete-after` `07455061db79…`/`ce91cba5b52e…`, `failed-after`, `cancelled-after`, `complete-before`), every frame's pre-shutter footer assertion firing. So the teardown change did not move a pixel.

#### 4. Are the QA and design rounds still valid here?

Stated plainly: **they are carried forward, not fresh on this head by the letter**, and in my judgement that is sufficient for merge. The evidence for the substance: the dialog's blob is identical, `src/renderer` has no change at all, the eight committed frames did not change and I reproduced all eight byte-for-byte from this head and from the base, and the only behavioural change in the delta is the rig's teardown — which is not a product surface. QA's matrix cells on `c92d9d6d8` therefore describe this tree too, and QA's remediation comment already carries the suite re-runs at `4dfa55874` (I re-ran `evidence-manifest`, `test-inventory`, `chrome-keychain`, `run-detail-model` and `mcp-auth-surface` here: 127 tests, 0 fail, which includes the three QA named at 34/34). Design's D1/D2 answers are prose in the README, whose only changes here are the ones this round reviewed, and its "no visual surface changed" claim is exactly the blob-and-frames identity I verified above. Nothing in this delta alters a visual surface — the incoming upstream files are main-process and a shared layout JSON.

QA's two BLOCKED cells are still blocked, and I re-attempted the first on this head rather than accepting it: `pnpm check-evidence` answers `Evidence check DEFERRED: another sweep (or its image child) holds the machine lease; retry after it finishes. No frames checked.` — exit 75. Not a claim this PR makes (PNG frames from a bespoke rig the `.webp` sweep does not walk; the rig asserts `assertFramePaints` on each frame it writes, and `evidence-manifest.test.mjs` checks the stamps inside `pnpm test:desktop`, green here).

#### 5. R1-5's half-deferral — acceptable, with one condition

Acceptable as it stands, and I would not hold the merge for it. The sentence in `mcp-auth-dialog.tsx`'s comment is **true for `login`**, which is the dialog's default and the operator's reported path; the label it gets wrong (`Replace grant and sign in`) is the `reauth` variant, and the README — the prose a reviewer actually reads next to the frames — now states both labels explicitly. The cost of fixing the comment now is real and asymmetric: a `src/` blob change moves both the QA and design round scopes under the freshness rule, for a wording diff with no behaviour attached. The condition I attach to the deferral is that it be honoured on its own stated trigger — the next round that touches this component must fix the phrase rather than re-defer it, and the README's re-auth wording must not drift from it before then. If the manager prefers zero deferred prose over the round-scope cost, fixing it is a two-word edit; my recommendation is to let it ride as recorded.

#### Findings

**R2-1 — nit — the remediation comment's R1-1 bullet misstates the base run as 4/4.**
`### Agent review remediation — round 1` says the new case's base run is "(4/4, base tree restored over the working tree)". The measured base run is **3 pass / 1 fail**: the `complete` case fails there by design (`actual ['Close','Try again']` vs `expected ['Close']`), which the bullet's own next sentence states correctly. The substance is right — the fourth case is not differential, and it passes on the base — but the parenthetical reads as though the whole file passed on the base, which would make the `complete` case look non-differential too. Worth correcting in the thread so the record a later reader audits is accurate. No content change wanted.

#### Considered, not findings

The loosened `SETTLED_SENTENCE` regex (`/Sign-in [a-z-]+\./`) no longer enumerates the vocabularies by design; each case still asserts its own sentence afterwards (`COMPLETED_SENTENCE`, `Sign-in ${status}.`, `UNREADABLE_SENTENCE`), so nothing is left unpinned. `succeeded` as the unreadable word is a good choice: outside every vocabulary this build knows, never produced by the backend, so it cannot drift into a real status. The rig's group-kill covers `exit`, `SIGINT` and the error path but not an external `SIGKILL` of the driver — the comment's claim is scoped to the error path, which is what it says, and the README names the symptom and the `--port` recovery, so I am not raising it. The README's "run one rig at a time over one `node_modules`" paragraph records a limitation rather than claiming a fix for it, which is the honest shape; I did not try to reproduce it, as it depends on a shared dep-cache race and re-running it would prove nothing about this diff.

damianvtran:

Answering R2-1 in place. **No commit, no push, no file change**: the head is still `4dfa55874c583b57fa18b1de3115f0e5da9cdc72`, which is the head you reviewed, so this round stays fresh on it.

**R2-1 — FIXED IN PLACE.** The comment edited is the round-1 remediation comment itself: https://github.com/damianvtran/local-operator-ui/pull/283#issuecomment-5707077228 (now `updated_at 2026-09-17T01:42:14Z`, 6,845 characters, read back byte-identical to what was sent). Its R1-1 bullet now carries the measured base result and nothing weaker:

- **base component: 3 pass / 1 fail** — the `complete` case the one that fails there (`actual ['Close','Try again']` against `expected ['Close']`), and the new boundary case passing;
- and the cause of the number it used to claim is named rather than quietly swapped: the round-1 "(4/4, base tree restored over the working tree)" came from a run of `git show HEAD:`, which resolves to the **fixed** file once the fix is committed, so the base was never actually measured in that round. It is measured now, at this head, against main's blob `f54f93151` and this branch's test file.

The same edit also states which mutation proves what, since you are right that the one quoted was weaker than the sentence implied:

- **the isolating mutation** — folding an *unrecognised* word into settled (`operation?.status === "complete" || !["running","complete","failed","cancelled"].includes(operation.status)`) — gives **3 pass / 1 fail, the boundary case the only failure**, the fix's own state and both controls still passing. That is the mutation that shows the fourth case polices the line it names;
- **the round-1 mutation** (`operation?.status !== "running"`) is restated for what it is: it withholds the retry from `failed` and `cancelled` as well, so it fails three cases (**1 pass / 3 fail**) — evidence that the predicate is broadly wrong, not that the boundary is being policed.

Both runs were made in this round, both mutations applied to the working tree only and reverted afterwards: `git diff --stat -- src/` empty, the dialog's blob still `c4ea7c9c1a903d4d6d6c102121d5848b4a56ff7b`.

**Nothing else moved.** No code, test, frame or manifest change; nothing committed and nothing pushed; `git status --porcelain` empty and `git log -1 --format=%H` still `4dfa55874c583b57fa18b1de3115f0e5da9cdc72`. R1-5's half-deferral stands as you accepted it, with your condition recorded: the phrase is fixed in the round that next touches the component, not re-deferred, and the README's re-auth wording stays put until then.

damianvtran:
### Merged — `962f43350`

Merged to `main` at 2026-09-17T01:43:00Z as merge commit `962f43350bb58f1b8611a150f0cad1fe50d44c9d` (a plain merge; no `--admin` was needed and no rule was bypassed). Head merged: `4dfa55874`.

Gate on that head, re-read from the thread rather than from this note:

| Round | Verdict |
| --- | --- |
| `### Agent review — round 1` (`cfc28c817..c92d9d6d8`) | clean — no blocker/major (R1-1 minor + nits) |
| `### Design review — round 1` | terminal — D1/D2 nits, no change requested |
| `### QA report — round 1` | PASS — no FAIL on head; two cells BLOCKED and named |
| `### Agent review remediation — round 1` | every finding fixed with a SHA, rejected with a reason, or deferred with its trigger |
| `### Agent review — round 2` (`c92d9d6d8..4dfa55874`) | clean — no blocker/major; fresh on the merged head |
| `### Agent review remediation — round 2` | R2-1 corrected in the thread |

CI on `4dfa55874`: all eight checks green (Desktop Tests, Lint, Type Checking, Runtime Dependencies, Security Audit, NPX Sanity x2, Version Bump Guard). `package.json` carries no version bump; the live version stays with the release owner of the window this merge lands in, who will post the tag and Release URL here.

damianvtran:
Released in **v0.25.19** — tag `v0.25.19` on `e89142ac659e7483cfdba7d76dde4724f9ab0286`, Release: https://github.com/damianvtran/local-operator-ui/releases/tag/v0.25.19

The window's notes quote this PR's own `Release:` claim verbatim under `## PRs`, and the whole window was called one patch (the argument is written out on #291). Published as a pre-release first, so the previous complete Release keeps answering the updater's feed until the installers for this one exist.

PR #280, opened 2026-09-16
perf(update): shrink what an in-app macOS update writes

damianvtran:
## Why

An in-app update closes the app while Squirrel.Mac's ShipIt extracts ~380 MB / 2087 files, validates them, swaps the bundle and relaunches. Measured on this machine, the closed window ran **73 s and 226 s for the same artifact on the same day** (255 s / 285 s on earlier dates). ShipIt's prepare phase ("Beginning installation" -> "Moving bundle") burned 49.5 s and 204.7 s while ShipIt used **~21 s of CPU across the 205 s** — it is blocked, not computing — and replaying the same work by hand does not explain it (unzip 6.0 s, cold `codesign --verify --deep` 1.7 s, `ditto` of the whole bundle 5.4 s).

**What it blocks on is not established, and this change does not claim to have found it.** So it removes only the work that is dead under *every* surviving explanation: fewer files written, fewer bytes written, fewer executable-mode code objects, less for any per-file hook to examine.

## What this PR ships

**The bundled Python seed is pruned and its execute bits normalised, where the seed is materialised.** `scripts/prune-python-seed.mjs`, called from `scripts/setup-python-resource.sh` — the one place the tree is assembled, and therefore before signing by construction (removing content after signing would be a `file missing:` violation, the class no update-time heal can repair).

- Deleted: `lib/python3.12/{idlelib,turtledemo,lib2to3,pydoc_data}`, `lib/tk8.6/demos`, `bin/{2to3,idle3,pydoc3}` and their `-3.12` targets.
- Every seed file that is not Mach-O loses its execute bit, by **the same magic set the app already tests**: the list moved into `src/shared/bundled-python-layout.json` as `machOMagics`, and `verifyMachO` in `managed-python.ts` reads it rather than keeping its own copy (the R10 / QA Q2 lesson — two hand-kept lists for one question drifted once already). The bit is useless here because the seed is **inert data**: `prepareRuntime` `ditto`s a copy out of the bundle and builds the venv against that copy, nothing executes it in place, and Python's import machinery never consults the execute bit. The ~40 stdlib `.py` files that carried one had it because the upstream tarball did.
- The path list, the stdlib marker and the magic set live in the one shared JSON, so the pack step and the release gate cannot disagree.
- A `seedStdlibMarker` guard makes a Python version bump fail the build loudly instead of turning every entry into a silent no-op — the shape where "the pruned paths are absent" passes trivially.

[...]

PR #284, opened 2026-09-16
chore(release): bump version to 0.25.17

damianvtran:
## What this is

The combined release for the window since **v0.25.16** — one bump, one tag, one GitHub Release covering every PR merged since that tag:

- #262 `fix(updates): no raw transport code as a user-facing warning, and a transient failure is retried` — merge `740796091`

`package.json`'s `version` line only, `0.25.16` → `0.25.17`. No other file.

## Why patch

Materiality, per the release process: #262 is a fix to an existing surface — a transient transport failure on the update feed is retried instead of reported, an app-initiated check no longer paints anything, and the failure a person does see is a sentence with the machine's code subordinate. It is not a step-function capability, so minor would over-report the window.

## How it was cut

A fresh tree from `origin/main` (`740796091`), never a bump sitting on a feature branch. The `version-bump-guard` job reads this line and treats a PR whose title starts with `chore(release):` as the sanctioned exception. The Release itself is created with `gh release create v0.25.17 --target <this PR's merge SHA>` and hand-written notes from `.github/RELEASE_TEMPLATE.md` — never `--generate-notes`, which the release gate refuses.

## Scope-check

This PR's whole diff is one line in `package.json`. The independent scope-check round is recorded as a comment on this PR before it merges; nothing else is claimed about it.

PR #279, opened 2026-09-16
chore(release): bump version to 0.25.16

damianvtran:
## What this is

The combined release for the window since **v0.25.15** — one bump, one tag, one
GitHub Release covering every PR merged since that tag:

- #261 `feat(chat): the quote control is raised by a highlight, and floats above it` — merge `4cdd06a0a`
- #233 `fix(chat): /compact without a dialog, and a typed command word as prose` — merge `57a39edc0`

`package.json`'s `version` line only, `0.25.15` → `0.25.16`. No other file.

## Why patch

Materiality, per the release process: #261 raises an existing control by an
existing gesture, and #233 corrects two existing surfaces (compaction's dialog
and the composer's treatment of a typed slash word). Neither is a step-function
capability, so minor would over-report the window.

## How it was cut

A fresh tree from `origin/main` (`57a39edc0`), never a bump sitting on a feature
branch. `pyproject`'s counterpart here is `package.json`; the `version-bump-guard`
job reads this line and treats a PR whose title starts with `chore(release):` as
the sanctioned exception. The Release itself is created with `gh release create
v0.25.16 --target <this PR's merge SHA>` and hand-written notes from
`.github/RELEASE_TEMPLATE.md` — never `--generate-notes`, which the release gate
refuses.

## Scope-check

This PR's whole diff is one line in `package.json`. The independent scope-check
round is recorded as a comment on this PR before it merges; nothing else is
claimed about it.

PR #276, opened 2026-09-16
chore(release): bump version to 0.25.15

damianvtran:
## What this is

The release bump for the current window. One version for everything merged since `v0.25.14`, chosen by materiality rather than by commit type.

**Window** (`git log --first-parent --oneline v0.25.14..origin/main`):

- #267 `chore(repo): stop auto-requesting code owners on every PR`
- #256 `feat(settings): rebuild the Backend settings section around a tier, a filter bar and a row`
- #274 `test(settings): run the Collapse all guard in CI, and make it exit when it fails`

**Bump: patch → 0.25.15.** A redesign of a surface the app already had is more of what it does, not a new thing it can do, so it does not clear the step-function bar for a minor. #274 is test-only and #267 has no user-visible effect.

**One file, one line.** `package.json`'s `version` is the only change. `git diff v0.25.14..origin/main -- package.json` moves only #274's `test:desktop` metadata and no `version` line, so no merged PR consumed a number and 0.25.15 is free; `Version Bump Guard` passes the guard's mirror-image case for the same reason.

**Release ownership.** Claimed and announced to the lop-dev peers on this machine; no other session held this window. The tag will go on this commit's merge SHA via `gh release create --target`, published as a pre-release so the previous complete Release keeps answering `/releases/latest` until the installers exist, with hand-written notes.

PR #286, opened 2026-09-17
chore(release): bump version to 0.25.18

damianvtran:
## What this is

The combined release for the window since **v0.25.17** — one bump, one tag, one GitHub Release covering every PR merged since that tag:

- #280 `perf(update): shrink what an in-app macOS update writes` — merge `7faf134d0`

`package.json`'s `version` line only, `0.25.17` → `0.25.18`. No other file.

## Why patch

Materiality, per the release process: #280 is internal update-path work. The artifact an in-app update rewrites is smaller — the bundled Python seed drops 279 files that nothing imports and 50 files whose execute bit meant nothing inside a bundle — so an update copies and the system re-examines less on every install. It changes nothing a user would call a new capability, so minor would over-report the window.

Worth knowing when reading the notes: this window is also where the non-English Electron locale strip was **deliberately left out** after review showed it changes rendered dates, times and numbers for non-English users; it is a product decision and returns as its own PR with a design round. The window therefore claims the seed reduction only, and the closed-window saving itself is **not** proven — it can only be measured on a real install after this release.

## How it was cut

A fresh tree from `origin/main` (`7faf134d0`), never a bump sitting on a feature branch. The `version-bump-guard` job reads this line and treats a PR whose title starts with `chore(release):` as the sanctioned exception. The window check the owner runs before cutting — `git diff v0.25.17..origin/main -- package.json`, which catches a stray bump a merged PR carried — is non-empty here for a benign reason and was read rather than assumed: the only hunk in it is #280's registration of its new unit suite in the hand-written `test:desktop` list, and the `version` line is untouched. The Release itself is created with `gh release create v0.25.18 --target <this PR's merge SHA>` and hand-written notes from `.github/RELEASE_TEMPLATE.md` — never `--generate-notes`, which the release gate refuses.

## Scope-check

This PR's whole diff is one line in `package.json`. The independent scope-check round is recorded as a comment on this PR before it merges; nothing else is claimed about it.


damianvtran:
### Agent review — round 1

Reviewer: reviewer (Local Operator subagent, model `deepseek/deepseek-flash`; no provider fallback during the round)
Scope: `7faf134d0..d87778855` (`origin/main` → `chore/release-0.25.18`)
Type: C0 one-file scope-check for the window's release-bump PR.

**Verdict: clean — TERMINAL. No BLOCKER, MAJOR, MINOR or NIT findings.**

1. **One file, one line, still parses.** `git diff 7faf134d0..d87778855` → `package.json | 2 +-`, numstat `1 1 package.json`, single hunk `-"version": "0.25.17"` / `+"version": "0.25.18"`. Both refs load as JSON, differ in exactly one key (`version`), and head is byte-identical to `origin/main` after substituting the version string (20,666 bytes both) — so nothing else in the file moved, trailing newline included. One commit on the branch: `d87778855 chore(release): bump version to 0.25.18`.
2. **The claimed window is the real one.** At the branch's base (`7faf134d0`, still current `origin/main`), `git log --first-parent --oneline v0.25.17..origin/main` yields exactly one entry: `7faf134d0 Merge pull request #280 from damianvtran/feat/macos-update-prune`. Independently, `gh pr list --state merged` shows exactly one PR merged after the tag — #280 at `2026-09-17T01:10:29Z`, merge `7faf134d0`. The tag `v0.25.17` sits on `cfc28c817`, which is PR #284's own merge commit (`2026-09-17T00:05:01Z`). The next-most-recent merges (#262 `23:55:02Z`, #279, #233, #261, #276, #274, #256) all predate the tag and belong to the v0.25.17 window, so the notes omit nothing. The body lists one merge, `7faf134d0`, and that is the whole window.
3. **Stray-bump check read rather than assumed.** `git diff v0.25.17..origin/main -- package.json` is one hunk at `@@ -74,7 +74,7 @@`, one removed and one added line: the hand-written `test:desktop` list. Comparing the `scripts` field regexed out of both refs gives `added: ['scripts/prune-python-seed.test.mjs']`, `removed: []` — precisely the body's explanation, and that suite is a file #280 created (absent at `v0.25.17`, present on `main`). Grepping that same diff for a `version` line change returns nothing; `v0.25.17:package.json` and `origin/main:package.json` both read `0.25.17`. No merged PR consumed a number nobody published.
4. **Guard confirmed green on this head, not assumed.** Title is `chore(release): bump version to 0.25.18`, so it takes the sanctioned exception. On head `d87778855` the rollup is fully green: `Version Bump Guard` success (15s), plus Desktop Tests, Lint, Type Checking, Runtime Dependencies, Security Audit and both NPX Sanity Check jobs. PR is `MERGEABLE` with no divergence from base.
5. **Patch class is defensible.** The window is one internal `perf(update)` PR (the bundle drops 279 files nothing imports and 50 files whose execute bit meant nothing), adding no capability, so `minor` would over-report. The body's caveats — the closed-window saving is measurable only on a real install after release, and the non-English locale strip was deliberately excluded and returns as its own PR with a design round — are honest about what the window does not claim.

Not findings, just the release-runbook tail: no `v0.25.18` tag exists (checked `git ls-remote --tags origin v0.25.18`), so the `gh release create v0.25.18 --target <merge SHA>` step is clear, and the post-release `lop-update` + outside-the-repo smoke still apply.

PR #272, opened 2026-09-16
feat(themes): port the TUI's forty-seven palettes, and the picker that holds them

damianvtran:
# PR Description

## Summary of Changes

The TUI ships fifty-two curated palettes and this app shipped twelve, so two
surfaces of the same product disagreed about what a palette is. This is the port
of the TUI's set into the registry — **forty-seven new palettes**, all thirty-one
roles literal each — plus the appearance picker rebuilt to hold them, the
evidence to look at, and the prose the new set size falsifies.

Release: minor — the theme set roughly quintuples (12 → 59) and the picker is
redesigned to hold it, so the picker a user opens is a different surface showing
a set five times the size; that is a step-function capability, not a patch.

| | |
|---|---|
| `shared/themes/palettes/*.ts` | 47 palettes added; `definitions` and the `ThemeName` union now carry 59 |
| `shared/themes/index.ts` | the ordering rule, written down (brand first, then the twelve this app shipped, then the port grouped by the TUI's own families) |
| `styles/themes.generated.css` | regenerated (`pnpm gen-themes`), never hand-edited |
| `features/settings/components/theme-selector.tsx` | the compact tile grid: 40px fixed-height miniatures, 12px names, description in a tooltip, three selected signals, one distinct mark per theme, and one tab stop with arrow/Home/End navigation instead of 59 |
| `features/settings/components/theme-selector.stories.tsx` | **new story** — the picker had none, so it had never been captured in any theme |
| `scripts/capture-evidence.mjs` | the picker's `STORIES` tuple, the bounded-sweep note beside `THEMES`, and `--themes` reaching the whole registry under `--only` |
| ~45 files across `src/`, `docs/`, `AGENTS.md` | the prose pass (below) |
| `docs/evidence/settings-appearance*/` | 59 frames of the picker + the 12-theme before half |

### Where the palettes came from, and how they were derived

- **Which ones.** The TUI's `all_palettes()` list
  (`local_operator/tui/palettes/__init__.py`) is ordered classics → rosé pine →
  companions → modern → neon/retro → radient → nature → lights; the registry
  keeps that order, minus the five names both sides already had (monokai,
  dracula, tokyo-night, sage, radient), which stay where this app's picker has
  always put them. That is exactly 47.
- **~19 were carried from the mobile-web port**
  (`local_operator/mobile/web/src/themes/`) and the rest were derived from the
  TUI's `ThemeSpec` tokens by the rules each palette's own header states: the
[...]

PR #294, opened 2026-09-17
chore(release): bump version to 0.26.0

damianvtran:
# PR Description

## Summary of Changes

One-line release bump for the window since `v0.25.19`, which holds exactly **one** merge:

- #272 `feat(themes): port the TUI's forty-seven palettes, and the picker that holds them` - `ace40b3b6d2ffcf6bc5710e3930e0741edd5136c`

`package.json` only, one line, `version` `0.25.19` -> `0.26.0`, written through the repository's own `scripts/apply-release-bump.mjs` rather than by hand or by a JSON re-serialiser, so the diff is provably the version line and nothing else (the field order and the tab indentation of that file are untouched).

Release: minor — the selectable theme set goes from 12 to 59 and the appearance picker is rebuilt to hold it.

**Why minor rather than patch.** The window has one merge and it clears the template's bar — "a step-function capability namable in the release title". A user could choose between twelve themes and now chooses between fifty-nine; the picker they open is a different surface at a different density (measured 92.1px of vertical space per theme before, 18.1-27.5px after depending on the window band, with the whole fifty-nine-theme set shorter than the twelve it replaces at the app's own 1380px). That is a capability a user did not have rather than a refinement of one they did, which is the distinction the previous window's patch call turned on. If a reader judges it otherwise, the class, the version, the tag and this branch name move to `0.25.20` together — say so rather than landing `0.26.0` over it.

**What the merge carries**, so the note-writer has it: 47 palettes ported from the TUI's curated set (the UI had twelve and the TUI fifty-two, of which five names overlapped), each authored against the palette contract's thirty-one roles and the contrast floors `pnpm check-themes` asserts over every palette; a thirty-first role, `highlight`, authored into all 47 after upstream added it mid-flight; the appearance picker rebuilt as a compact tile grid with roving-tabindex keyboard navigation; a prose pass across the files that asserted properties "in all twelve themes"; and the evidence for all of it — 59 frames of the picker, one per theme, plus the twelve-theme before half.

**Ownership.** Claimed by this session at 05:2xZ: `git log --first-parent v0.25.19..origin/main` held one commit (`ace40b3b6`), no `chore(release):` PR was open so no lock was held, and the claim was announced by message to the sessions most likely to hold a window on this repository (the three PR-steward sessions and two product sessions), asking to be told to yield if any of them owned it. Ownership is re-checked once more immediately before the tag is created, against `lop sessions` and the open-PR list, and the Release body will name what that check found.

## Related Issue(s)

- None. The work was asked for in session; no ticket exists and none was created.

## Impact

- No code change beyond the version line. No dependency, security or performance impact.
- The version is the only thing this PR moves; the release itself is created against this commit after the merge.

## Testing Details

- `git diff` on the single commit is one line in `package.json`, asserted line-for-line by the bump script before the commit was made.
- The merged content this bump names was verified on its own PR: code review rounds 1-4 (a BLOCKER closed with `Desktop Tests` green in CI), design rounds 1-2, UX rounds 1-3 (a MAJOR fixed and re-verified in the live DOM), QA rounds 1-3 PASS, and `pnpm check-themes` at 16,069 assertions across all 59 themes.

## Checklist

- [x] I have performed a self-review of my code — the diff is the version line and nothing else.
- [x] I have added tests that prove my changes work as intended — not applicable to a version bump; the gate for this commit is its scope, which the bump script asserts.
- [x] The gates pass — nothing but a version line changed, so no gate is affected.
- [x] This PR changes the `version` line in `package.json`, which is what a `chore(release):` PR exists to do.
- [x] I have updated the documentation when required — not applicable.
- [x] Security considerations are addressed — no code change.
- [x] I have linked all related issues — none exist.


damianvtran:
### Agent review — round 1
Reviewer: deepseek/deepseek-flash (independent reviewer subagent; no provider fallback during this round)
Scope: ace40b3b6d2ffcf6bc5710e3930e0741edd5136c..a6708d093e138733779fd6333583084c4226d783

This is the C0 scope check on the bump commit: the whole question is whether the commit does exactly what a release bump may do and nothing else.

**1. The diff is one line in one file, and it is the version line.** `gh pr diff 294` and `git diff origin/main..origin/chore/release-0.26.0` agree: `package.json | 2 +-`, 1 insertion, 1 deletion, one hunk `@@ -1,6 +1,6 @@` carrying `-  "version": "0.25.19"` / `+  "version": "0.26.0"`. It is not the output of a JSON re-serialiser: the blob went 20936 -> 20935 bytes, i.e. exactly the one byte by which `"0.25.19"` differs from `"0.26.0"`; the file is byte-identical to `sed 's/"0.25.19"/"0.26.0"/'` over the old blob (`cmp` clean), so tab indentation, field order and every other character survive. No trailing-whitespace or newline churn.

**2. Nothing else rides the branch.** `merge-base origin/main HEAD` equals the current `origin/main` tip (`ace40b3b6`), so the branch is cut from main as the procedure requires. `git log --oneline origin/main..origin/chore/release-0.26.0` is the single commit `a6708d093 chore(release): bump version to 0.26.0`. `git diff --name-only origin/main..origin/chore/release-0.26.0` is `package.json` alone, and the only `0.25.`/`0.26.` strings anywhere in the diff are that line's two sides — no second version surface, and (the mirror-image failure the guard exists for) the diff does change the version line, so this cannot be the "new number over the previous release's tree" state.

**3. The version is `<last tag> + the class the body argues`, and the body's claim about the window is true.** `git log --first-parent v0.25.19..origin/main` holds exactly one merge, `ace40b3b6` = #272 `feat(themes): port the TUI's forty-seven palettes, and the picker that holds them` (title read from the forge), which is what the body says. 0.25.19 + minor = 0.26.0. The window's contents claim checks out against the tree at the two refs: the `ThemeName` union has 12 members at `v0.25.19` and 59 at the head, and `src/renderer/src/shared/themes/palettes/` had 11 `.ts` files at the tag against 58 at the head (`local-operator.ts` exports `localOperatorDark` + `localOperatorLight`, which is why 11 files were 12 palettes) — so "twelve to fifty-nine" is exactly right, not rounded. Nothing has spent 0.26.0: no `v0.26.0` tag exists upstream and no Release carries it.

The minor-versus-patch call is a judgement, and it is the owner's to make; it sits on the right side of the bar (a *single* PR that is a step-function capability rather than more of what already existed) and the body argues it explicitly instead of assuming it, including the one-line path to downgrade to 0.25.20 if a reader disagrees. I would let it stand.

**4. The repository's own guard agrees, on this head.** `gh run list --branch chore/release-0.26.0`: `Version Bump Guard`, run `35184888202`, **completed/success on `a6708d093`** (job `105084707005`) — settled, not pending. Read from the run log rather than the check tile, it printed its verdict rather than passing silently: `Release PR, version change allowed: - "version": "0.25.19" / + "version": "0.26.0"` and `Version 0.26.0 sorts above the highest release tag v0.25.19`. That is the exempt path taken *and* the substantive direction covered. The commit was written with `scripts/apply-release-bump.mjs`, whose `isVersionOnlyChange` assertion — exactly one removed and one added line, each containing `"version"`, checked over the bytes before the write — is precisely the property I measured by hand above; the tool and the artifact agree.

**5. NIT (procedural, no remediation, not a defect in this commit).** The window-level backstop in AGENTS.md reads `git diff <last-tag>..origin/main -- package.json` must be **empty**, and on this window it is **not**: one line changes in the `test:desktop` script list (the two theme test files #272 registers). Read for what it is for — the version line — it is clean, since the diff contains no `^[+-][ \t]*"version"` line, so no number was silently consumed; the scripts block is metadata the PR-level guard itself exempts. Worth one sentence at tag time when this repository's window touches `package.json` for any other reason, because the literal reading of the rule fires while its purpose is untouched.

Not reviewed here (outside this diff, so not findings): the hand-written Release notes and the tag-time checks, which belong to the tag, not the bump.

**Verdict: clean. No BLOCKER, MAJOR or MINOR findings, and the single NIT asks for no round — this round is TERMINAL.** Ready to merge as a C0 change. Pre-Release CI (run `35184888151`) was still in flight at review time — Lint, Type Checking, Runtime Dependencies and Security Audit pass, Desktop Tests and the two NPX Sanity Checks pending — so merge once those settle on this head; they are the only thing standing between this and the merge.

PR #290, opened 2026-09-17
fix(chat): a user turn is measured by its card, so its prose fills it

damianvtran:
## What was reported

> "I noticed on some wider attachments that the text in the user message component doesn't expand all the way across ... the text should expand with the width of the user message card if the card width is expanded instead of ending up center constrained."

The screenshot showed a wide user card holding a truncated "Replying to" quote, with the message prose floating inside it as a narrower column and equal slack on both sides.

## Root cause

The user bubble is a flex item of a right-aligned row, so its width is shrink-to-fit: as wide as its widest child, capped by `max-w-[75%]` (`max-w-[92%]` in the small view) inside `CHAT_MEASURE`'s 900px column. Its body carried `lo-measured`, and `markdown.css` declared:

```css
.lo-measured .lo-markdown { max-width: 62ch; margin-inline: auto; }
```

That centring is inert only while the prose is the bubble's **widest** child. Two children routinely out-measure 62ch — the sent `ReplyPreview` chip (`flex-1 truncate`, so its max-content is the whole quoted text) and `CanonicalImage` — so the card grew to 675px while the prose stayed a 546.7px column with `margin-inline: auto` splitting the remainder evenly.

## The fix

The card is the measure. `.lo-markdown` inside a user turn takes no `max-width` and no `margin-inline: auto`, so the prose fills the card's content box and follows it as a reply quote or an attachment widens the card. The `lo-measured` opt-in is gone from both user-turn surfaces (canonical and legacy). The aside reading survives by the card's own width: 675px against the agent answer's 860px.

## Evidence

Measured from the live DOM (`getBoundingClientRect`) at a 1024x620 pane, `localOperatorDark`, story `chat-canonical-user-card-measure--reported-shape`:

| | before | after |
| --- | --- | --- |
| card | 675 (287..962) | 675 (287..962) |
| card content box | 641 (304..945) | 641 (304..945) |
| prose | 351.1..897.9 = 546.738px = 62ch | 304..945 = 641px |
| computed `max-width` / `margin-inline` | `546.738px` / `47.125px, 47.1406px` | `none` / `0` |
| slack at the card's inner edges | **47.1px / 47.1px** (48.1 recorded border-inclusive) | **0 / 0** |

Committed frames — twelve themes on both halves, from the repo's own rig, with `scripts/capture-evidence.mjs`'s `STORIES` list now naming all three stories at 1024x620 so any reviewer can re-shoot them:

[...]

PR #278, opened 2026-09-16
perf(panels): stop abandoning and re-running a ledger read that is still in flight

damianvtran:
# PR Description

## Summary of Changes

**A read that is still running is no longer killed at the app's old budget, reported as a dead backend, and then run a second time — and the app now says what it actually knows when a request runs out of time.**

Two changes, one cause. The panels (`/analytics`, `/session`, `/info`, `/usage`) read through the desktop transport, and every op shared one 20 s `fetch` budget. That budget is right for a control and wrong for these reads: the same `analytics.get` query against the same isolated ledger measured **12.1-17.8 s warm and 39.6 s with a cold page cache** for the 30-day window, 5.6-11.8 s for 7 days. A read that crosses the budget is not cancelled — the daemon keeps aggregating and the answer is thrown away — and the renderer's default `retry: 1` then puts a second scan on a daemon that is still busy with the first. Reproduced end to end: with every `/v1/desktop/analytics` request held 25 s, the panel rendered `The backend could not complete this request. Check its connection and try again.` and **never showed data**, while the daemon answered both of the app's requests `200 OK`.

- `desktopRequestDeadlineMs(op)` (new, in the desktop contract): the ops that are slow for a **local-scan** reason (`analytics.get`, `sessions.report`) and the one that is slow for a **provider-network** reason (`usage.get` — the route answers from the provider controller's cache or from a live fan-out to each provider's quota endpoint) get 90 s; every other op keeps 20 s. `info.get` and `sessions.failovers` are deliberately not on the long budget: one is a host snapshot whose slow part is a child process, the other is routing state already in memory.
- `desktopRequestDeadlineDetail(op, ms)`: an expired request answers **504** with a `deadline_exceeded` code and its own sentence, chosen by what the app can know. A read says "Nothing was read; reopen the panel to ask again."; an op that may WRITE says "It may or may not have reached the server; check the result before repeating it." — the app aborts a fetch and never learns whether the daemon applied it, so the flat "Nothing was changed" was a claim it could not check, and a user who read it re-sent.
- An expired request is its own `backendErrorKind`, so the compatibility banner keeps its diagnosis ("The Local Operator server did not answer in time.") and its instruction (restart the app) — a wedged server is the case that banner exists for.
- The renderer's own deadline (`desktopRequestTimeoutMs(op)`) is derived per op from the transport's plus a 5 s margin, so the documented invariant still holds now that main's deadline is not one number: the layer that knows the HTTP status gives up first.
[...]

PR #301, opened 2026-09-17
chore(release): bump version to 0.26.3

damianvtran:
One window, one bump. `package.json` only — the `version` line, nothing else.

**Window: `v0.26.2..main` (2 merges).**

| PR | merge | Release line |
|---|---|---|
| #290 | `3afcc732f` | fix: the user bubble's measure, corrected |
| #278 | `da9540148` | fix/perf: a slow ledger read no longer makes the app give up and show nothing — the analytics, session and info panels load in under a second instead of seconds |

**Bump: patch (0.26.2 → 0.26.3), by materiality.** Two corrections and a performance change, neither a new capability. Patch is the default and under-calling is the preferred direction of error.

**What #278 changes for a user**

- A ledger read that outlives the app's own deadline used to be reported as a connectivity failure: the panel rendered "the backend could not complete this request" and **never showed data**, while the daemon answered `200 OK` twice. Measured on the base: two `/analytics` reads 21.009 s apart with no content; on the merged tree: one read, content 55 ms after the answer.
- Ledger reads (`analytics.get`, `usage.get`, `sessions.report`) now have their own 90 s budget instead of sharing a 20 s one, and the renderer's deadline is derived per op so main always gives up first.
- A budget expiry is a **diagnosable state of its own** (504 + `deadline_exceeded`) rather than the 503 the connectivity path reads as "unreachable" — a dead daemon still reads as unreachable, and the compatibility banner keeps both its reason and its restart instruction.
- The four panel reads came off `retry: 1` for a budget-expired read and off refetch-on-focus, replaced by an "as of" cue; reopening is the refresh gesture and re-reads.
- User-visible copy: a read keeps a flat "nothing was read" assurance, a **write** now states the app's uncertainty rather than promising non-delivery, and the held-message card's noun was neutralised (`A message is still being held`, `Restore message`, `Discard message`) because the app cannot know whether an aborted request reached the daemon.

**Two things on the record rather than hidden**

- The machine-wide `pnpm check-evidence` sweep is **red on `main`**: 24 of 5,772 frames, **none in this PR's sets**. The three directories holding every failure are byte-identical to `origin/main` (`c9dd596f79`, `bca7d0cd87`, `646e6c2782`), so a sweep at main fails identically and this branch cannot have introduced them. Recorded in the PR's merge disclosure.
- A **retrospective QA pass** on the merged head is running, because the final head's behavioural surface never got an independent QA round before merge (round 1 passed on an earlier head). If it FAILs, the fix forward is a follow-up PR.

**Scope round.** One file, one line — the check here is that nothing else rode along.


damianvtran:
### Agent review — round 1

Reviewer: local-operator reviewer subagent (deepseek/deepseek-flash; no provider fallback this round)
Scope: da9540148..8b7ce4ef7 (PR base `origin/main` .. branch head, one commit)

C0 scope check on a release-owner bump. Four checks, all pass.

**1. Diff shape — one file, one line.** `git diff da9540148 8b7ce4ef7` is `package.json`, `"version": "0.26.2"` -> `"version": "0.26.3"`, 1 insertion / 1 deletion. Nothing else rides along: no lockfile, no `docs/evidence/manifest.json`, no other metadata. `git log --oneline --first-parent origin/main..origin/chore/release-0.26.3` is the single commit `8b7ce4ef7`.

**2. `0.26.3` is unspent, and patch is the right size.** `git ls-remote --tags origin` carries no `v0.26.3` (newest is `v0.26.2` = `ccc3017a3`), and npm has no `0.26.3`: `npm view local-operator-ui versions` ends at `0.26.2`, `dist-tags.latest` = `0.26.2`. The base carries `0.26.2` (`git show da9540148:package.json`). Materiality: window is two merges, both from `fix/` and `perf/` branches with no `feat` merge in `v0.26.2..main`; src-side content is 6 files for #290 and 18 files plus one `src/shared/desktop-contract.ts` addition for #278 (the rest of each merge's file count is `docs/evidence` frames). Neither adds a capability, so the patch call is correct, not under-called. The one arguable point is #278's new `deadline_exceeded`/504 contract value — a wire-surface addition — but a new diagnosable state of an existing op is not a step function, and the repo's documented error direction is under-calling.

**3. Window list is complete.** `git log --oneline --first-parent v0.26.2..origin/main` returns exactly two merges — `da9540148` (#278) and `3afcc732f` (#290) — and the body lists both with those merge SHAs. No third PR is missing.

**4. Evidence-manifest test, branch head.** On a throwaway detached worktree at `8b7ce4ef7`: `node --test scripts/evidence-manifest.test.mjs` -> **28 tests, 28 pass, 0 fail** (748 ms), no lease needed. The shipped stamps match the tree at that commit — manifest `srcTree` `e8c0bad63eaffbf5863751a3194cfffad2450e5c` = `git rev-parse HEAD:src`, `scriptsTree` `573949f3bf19b3d7ad34ecc88c900b032a8a8f41` = `HEAD:scripts` — so the bump PR carries no stale stamp. Post-merge state deliberately not predicted; the owner re-runs this on the merge commit.

**Findings: none.** No blocker, no major, nothing minor worth a round.

Notes, not findings: CI is green on the head (`Version Bump Guard` pass — the right direction for a `chore(release)` bump, whose mirror-image failure mode is a title without a version change); `reviewRequests` is empty, so nothing pinged anyone (expected — the `*` CODEOWNERS line is gone in this repo). `Desktop Tests` and both `NPX Sanity Check` jobs were still pending at review time, and assignees are empty; both are the merge step's business, not this check's.

Scope-check verdict: **clean — round 1 is TERMINAL** for this PR. It is one file, one line, one unspent number, a complete window and an unstale manifest. No remediations, no further round required; merge-ready subject to CI completing.

PR #310, opened 2026-09-17
fix(chat): resume a running call's clock from the producer's own start

damianvtran:
## The defect

Resuming (or attaching to) a session whose tool call is **still running** restarted
every elapsed counter from the moment the view loaded:

> each time I resume it says it's been waiting for 0s regardless of how long

Two surfaces were wrong, and for different reasons.

## 1. The tool row counted from the frame's ARRIVAL

`applyEvent` stamped `startedAt` from the caller's `now`:

```ts
startedAt:
    current?.kind === "tool" && current.startedAt !== null ? current.startedAt : now
```

Every path that hands it a `now` passes *the viewer's own* instant — the replayed
`event` frames applied before the snapshot lands
(`use-canonical-session.ts`), the live seed applied after it
(`applyLiveSeed`, which smuggles the stated clock in through `now` **only** for a
frame that would CREATE the row), and an ordinary live frame. So on a resume the
row was painted at arrival and the value was sticky: no later frame could correct
it. The sibling field one block above already did this correctly
(`anchoredAt: epochMs(event) ?? now`).

**Fix** — the frame's own stated clock wins, with the existing value and `now`
behind it as fallbacks:

```ts
startedAt:
    epochMs(event) ??
    (current?.kind === "tool" && current.startedAt !== null ? current.startedAt : now)
```

This closes every arrival-stamping path at once, because the reducer no longer
depends on a caller passing the right value in an implicit parameter, and a stated
start can never move a clock backwards (the producer stamps it at execution
start). Preserved deliberately: a composing row stays `startedAt: null`, a settled
row stays `null`, and a replayed `_start` still cannot restart a running clock.

### The one deliberate asymmetry

A frame with **no** `started_at_epoch` (a legacy producer) still falls back to
`now` rather than blanking the column the way the TUI does
(`tool_card.py:1437-1441` refuses to paint a number it cannot justify). This PR
keeps today's fallback: the change is confined to frames that *state* when the
call began, and a stated-nothing frame behaves exactly as it did before. Flagged
here rather than changed silently so a reviewer can weigh it.

## 2. The working line had no anchor at all

`deriveWorkingLine` returned only `{activity, phase}`, so the band under the
transcript was a function of `WorkingLine`'s mount (`useRef(Date.now())`, reset on
every phase change). A resumed pane restarted it at `0s` by construction — the
half the TUI already fixed (`1d8b4c630`).

Both fields were already on the wire. They are now **declared** on
[...]

PR #307, opened 2026-09-17
fix(chat): name a failed store, and withhold the retry hint it cannot honour

damianvtran:
## Summary

A send that fails because the backend's **store** could not be written is no longer reported as a store that is merely *busy*, and the composer no longer answers it with the one instruction that cannot help.

On 2026-09-17 this machine's boot volume hit **0 bytes free** at 09:56. SQLite could not allocate its journal or WAL, and the backend's desktop routes answered with the single 503 they had for every `sqlite3.Error` — lock contention, an unopenable database, a corrupt one and a full disk alike:

> Read state is busy right now. It will catch up on its own.

The composer relayed that sentence and appended its generic "what to do" half, *Your message is still in the composer. Send it again.* — the one action a full disk cannot honour, and the one the operator repeated, with an image attached. Removing the image let the same send through, because the image is by far the largest write in the flow (attachment bytes plus a much bigger transcript append), so it is what crosses the threshold while a few-KB text append still lands. It went away when space came back.

The sibling backend PR splits that ladder by what the failure actually is. **This PR is the renderer half**, coded against the pinned contract (`{"code", "message"}` error bodies):

| Code | Status | Composer |
| --- | --- | --- |
| `store_busy` | 503 | The existing sentence, unchanged, because for genuine lock contention it is true. The generic retry hint is **not on screen** on this route either — a post-admission refusal leaves the box empty, and the composer's own rule withholds the hint unless the box holds the payload — so `withholdsRetryHint` stays `false` for it and the frames show why (UX round 1, **U7** corrected this row: it used to claim the hint was kept). |
| `store_out_of_space` | 507 | Names the disk and the remedy (free up space, then send); the generic retry hint is **withheld**, and the held line names the control the sentence's own instruction needs. |
| `store_unavailable` | 500 | Says retrying will not help and to check the machine; the hint is **withheld**, same held line. |

## What changed

- `src/renderer/src/shared/store/canonical-sessions-store.ts` — `STORE_OUT_OF_SPACE_CODE` and `STORE_UNAVAILABLE_CODE` beside the existing refusal codes, both added to `withholdsRetryHint`, and a second predicate `isStoreWriteRefusal` for the question that is *not* about the retry (what a store failure knows). `store_busy` deliberately gets no constant: nothing in the renderer changes for it, and the split exists so that "send it again" becomes *true-for-contention, false-for-a-failed-store*.
[...]

PR #306, opened 2026-09-17
fix(composer): hold the Stop control's box for a grace window, not always

damianvtran:
## The defect

The idle composer renders `[dictation]  <gap>  [Send]`. The gap is not spacing: it is an invisible box (`data-interrupt-slot`) that reserves the Stop control's position for as long as the backend negotiates `session_interrupt`. The operator reported it as a defect of its own — the dictation control should sit beside Send when no Stop control is offered.

## Why the box was there, and what the trade is

UX round 1's U1 and QA's Q1 found the same thing independently on the unreserved build: with the Stop control gone the dictation control moved 36px into the exact centre of the box the press had just landed in (32px of control + the row's 4px gap), so a reflex second press started a **microphone recording** (`elementFromPoint` at the Stop's old centre resolved to `button[aria-label="Start recording"]`; a 120ms double press still hit Stop twice, so it depended only on how fast the second press landed). Reserving the box permanently removed that and created the standing gap.

**Send never moves.** It is pinned to the composer's right edge in every state — measured, from the running row through the window to the settled row. The control that moves is the **dictation** control (1235 running → 1271 settled), which is why an idle row in which it sits BESIDE Send is necessarily the row in which it has arrived in the vacated box: the geometry cannot be preserved, only the hazard bounded. So this PR keeps the reservation as a **grace window** instead of a permanent state:

- held while a turn runs **and for 500 ms after the control leaves** the row, then released to nothing;
- the running row is unchanged: `[dictation][Stop][Send]`, and the window reproduces it box for box;
- idle-and-settled is `[dictation][Send]` adjacent — the operator's report;
- **the recording action is never refused, gated or greyed.** The dictation control stays visible, live and pressable at every instant, and during the window it is at exactly the position it renders today.

Cost, stated rather than hidden: a second press landing more than 500 ms after a turn settles reaches the dictation control, which by then is where it is drawn. The window covers the reflex press the measurement is about (161–170 ms after the composer's own flip in the shipped record; the 120 ms double press U1/Q1 measured, and this branch's earlier pass at 207–210 ms, which lives in this branch's history rather than in the record this head ships — `git show 043b0b7c3:docs/evidence/interrupt-live/interrupt-proof.json`, `slot.repress`), so the number is a **chosen** margin over the reflex rather than a value derived from it — 170 < 500 < 1776 in this run — and it is not claimed to cover a deliberate press a second later.

[...]

PR #300, opened 2026-09-17
fix(chat-sidebar): hold the list's scroll position while a row re-files

damianvtran:
Release: patch — the sidebar's **Active chats** list holds its scroll position while a completed row re-files, the **entity region** above it does too (its children are session rows), and the row that holds the **keyboard cursor** is followed rather than stranded.

## What and why

The operator reported (2026-09-16) that a completed session's row updates its checkmark immediately but does not re-sort into the completed band at the top of the sidebar's **Active chats** until 5-10 s later, and asked for it to sort as soon as the session completes, with **no bouncing or jittering**.

The backend half is [local-operator#1224](https://github.com/damianvtran/local-operator/pull/1224): the machine-wide feed now invalidates the client's whole-list re-read on an **order-key** change, not only on a section change, so a row that reorders *inside* its section gets its new slot in the time one list read takes (~40-150 ms measured) instead of on the 30 s safety poll.

This PR is the renderer's half of "no jitter". It is **two declarations and one rule**:

```diff
-<div className="min-h-0 flex-1 space-y-4 overflow-y-auto p-1">
+<div className="min-h-0 flex-1 space-y-4 overflow-y-auto p-1 [overflow-anchor:none]">

-<div className="mt-2 max-h-[45%] shrink-0 space-y-4 overflow-y-auto border-t border-hairline pt-2">
+<div className="mt-2 max-h-[45%] shrink-0 space-y-4 overflow-y-auto border-t border-hairline pt-2 [overflow-anchor:none]">
```

**Why the list needed a declaration at all.** Chrome's scroll anchoring (`overflow-anchor: auto`, the initial value) picks *the element that moved* as its anchor and pays for its move by shifting the scroll container's `scrollTop` by exactly the row's travel. Measured on the overflowing story: **`scrollTop` 221 → 125 across a 96 px re-file** — the whole viewport sliding under a reader who is somewhere else in the list entirely, which is the report's "jitter" arriving on **every** completion instead of on every poll. With the declaration the same re-file leaves `scrollTop` at **221**.

[...]

PR #322, opened 2026-09-17
fix(release): make asset attachment streamed, bounded and repairable

damianvtran:
## What this fixes

Since `v0.26.5` published at 12:15Z, **every** release attempt has died in `attach-to-release` — four attempts by three different sessions. `gh api --method POST --input <file>` uploads the 855-byte `latest-mac.yml` and then **stalls on the 158 MB arm64 `.dmg`**; GitHub leaves an asset record in the **`starter`** state, and the job hangs until it fails. It is not a size ceiling (the release before uploaded a byte-identical 158,483,557-byte dmg) and nothing in these scripts had changed.

Three properties turned that into four lost releases:

1. **`execFileSync` with no timeout**, so a stalled attempt blocked the job instead of failing fast and being retried.
2. **`checkAssetCollisions` refused every name already on the release** (deliberately: "never replace an asset"), so the *wreckage* of one failed attempt blocked every later one. A partially-attached release could not be repaired by re-running the job — it needed hand surgery on the API.
3. **The catch-all `catch {}` reported every failure as "existing assets were not replaced"**, which names a collision even when the real cause was a stall, a timeout or a 5xx. That mislabel cost a reader a wrong turn on the night this was diagnosed.

## What changed, per property

**1. The upload streams, and every attempt is bounded.** The buffered `gh api --input` call is replaced by a streamed `fetch` POST to the pinned release's asset URL (`body: createReadStream(file)` with `duplex: "half"` and `content-length` taken from the file's stat, so the endpoint can still tell a truncated upload from a complete one). Same auth (`Bearer`) and content type (`application/octet-stream`) as before. Each attempt has a 10-minute deadline, and a transient failure — timeout, reset connection, 5xx, 429 — is retried up to three attempts with 5s/10s backoff. A 4xx is an answer, not a fault to wait out: it is reported after one attempt.

**2. A partial upload is recoverable, and a good asset is still never replaced.** Reconciliation is now per artifact, by name and size, against the assets already attached to the pinned release ID:

| what is on the release | what happens |
| --- | --- |
| no asset of this name | upload |
| complete (`state: "uploaded"`), **same size** as ours | skip — a re-run over an intact release writes nothing |
| complete, **different size** | refuse, naming both sizes |
| `state: "starter"` — an upload that began and never finished | **delete it, then upload** |

A `starter` record is not an asset: nobody can download it, and `scripts/release-state.mjs` already counts only `uploaded` assets when it decides whether a release's assets are complete. Deleting one leaves the invariant that matters ("never replace a good asset") intact while removing the one that cost three releases.

The retry loop reconciles before **every** attempt rather than only the first, because a failed attempt leaves exactly that `starter` record behind — the retry's own POST would otherwise be refused over the duplicate name — while a retry that finds the complete asset at our size is an upload whose response was lost, i.e. success rather than a duplicate.

**3. Failures report the real cause.** A failed upload now reports the HTTP status and the endpoint's own message for the *file* that failed (`app.dmg: HTTP 503 from the upload endpoint: Service Unavailable`, or a `TimeoutError` with the attempt deadline, or `ECONNRESET`), plus how much of the retry budget was spent. The word "collision" survives in exactly one place: the reconciliation's same-name/different-size refusal. Unrecognised asset states are refused by name rather than guessed at.

Unchanged: the tag↔SHA↔release-ID pins, `PLATFORM_INSTALLERS` / `PLATFORM_UPDATE_METADATA`, "never create a release or mutate release metadata", and the repair path's refusal to promote.

## Evidence

This path cannot be exercised end to end without cutting a real release, so `scripts/upload-release.test.mjs` drives the **real transport** (streamed body, deadline, `content-length`) and the **real retry policy** against a local stub upload endpoint, and the whole CLI against a fixture `gh` as the only one on PATH:

```
✔ (a) a stalled attempt is retried, and the starter record it left is removed first
✔ (a) a 5xx is retried, and the endpoint's own message is reported when it is not
✔ (a) an exhausted retry budget reports the status and the file, never a collision
✔ (a) a stall that never clears fails on the deadline, naming the file and the timeout
✔ (a) a 4xx is an answer, not a fault to wait out: one attempt, the endpoint's message
✔ (b) a starter asset already on the release is deleted and re-uploaded
✔ (c) a complete asset of the same name and size is skipped, not re-uploaded
✔ (d) a complete asset of this name at another size is refused, naming both sizes
✔ the CLI repairs a partially-attached release: starter deleted, same-size skipped, the rest uploaded
ℹ tests 9   ℹ pass 9   ℹ fail 0
```

The cases assert what the endpoint actually received (byte counts, headers) and what the release was left carrying, not what the code is assumed to send. The end-to-end case runs the real `node scripts/upload-release.mjs` against a release whose windows installer is a leftover `starter` record and whose mac dmg is already complete:

```
--- CLI stdout ---
Tag v0.26.5 resolves to commit 3becfb9c462f6adb1f47f9815767f5522755f849
Release v0.26.5 (id 389357596) state: published, prerelease: true
package.json name=local-operator-ui version=0.26.5 OK
Already attached at this size, nothing to upload: app.dmg (1048576 bytes, asset 9301)
Uploaded latest-mac.yml (64 bytes) to release 389357596 [HTTP 201]
Removed an incomplete upload of app.exe (asset 9302, state starter) left behind by an earlier attempt
Uploaded app.exe (4096 bytes) to release 389357596 [HTTP 201]
Uploaded latest.yml (32 bytes) to release 389357596 [HTTP 201]
Uploaded app.deb (2048 bytes) to release 389357596 [HTTP 201]
Uploaded latest-linux.yml (16 bytes) to release 389357596 [HTTP 201]
--- exit code: 0 ---
--- stub endpoint received ---
  POST name=latest-mac.yml   content-length=64   transfer-encoding=none bytes_received=64
  POST name=app.exe          content-length=4096 transfer-encoding=none bytes_received=4096
  POST name=latest.yml       content-length=32   transfer-encoding=none bytes_received=32
  POST name=app.deb          content-length=2048 transfer-encoding=none bytes_received=2048
  POST name=latest-linux.yml content-length=16   transfer-encoding=none bytes_received=16
```

The one write the run makes to the release's records is the delete, addressed by asset ID:

```
["api","--method","DELETE","repos/damianvtran/local-operator-ui/releases/assets/9302"]
```

Gates, all from the branch head:

```
pnpm lint                -> 728 files checked, 0 errors (67 pre-existing warnings)
pnpm lint:scripts        -> 4 changed file(s) under scripts/ are lint-clean against origin/main (dde14e9)
node scripts/test-validate-release.mjs        -> All validation tests passed
node --test scripts/test-release-safety.mjs \
  scripts/test-publish-workflow.mjs \
  scripts/test-version-bump-guard.mjs         -> tests 229  pass 229  fail 0
node --test scripts/release-baseline.test.mjs scripts/release-candidate.test.mjs \
  scripts/entry-point.test.mjs scripts/require-report.test.mjs -> tests 27  pass 27  fail 0
node --test scripts/upload-release.test.mjs   -> tests 9  pass 9  fail 0
node --test scripts/test-inventory.test.mjs   -> tests 1  pass 1  fail 0
```

`GITHUB_UPLOADS_URL` is the one new knob: it overrides the upload host, which GitHub Enterprise Server needs anyway and which is what lets the tests point the real transport at a stub.

Docs updated to match the new policy rather than broadly: the script's header, the release section of `AGENTS.md` ("Never replace a complete asset; never treat a `starter` record as a collision"), and the `attach-to-release` step comment in `publish.yml`, which previously said the job never deletes anything and would otherwise have become a lie in the file a future reader reaches first.

## Not addressed

- `.github/RELEASE_TEMPLATE.md` needed no edit: nothing in it states the old asset rule (it governs the writeup and the PR list). Checked rather than assumed.
- `pnpm check-evidence` could not run: another sweep holds the shared machine lease, so it deferred with exit 75. No visual surface changes here.
- Out of scope but worth knowing: the reconcile-before-every-attempt read can race GitHub's own consistency window after a delete, in which case the retry's POST returns 422 and the run fails with the real status and message — the next re-run then finds the record and repairs it.

Release: patch - a stalled release upload now fails fast with the real cause, retries itself, and leaves a partially-attached release repairable by re-running the job.


damianvtran:
### Head 27c8870ba — CI green, plus one fix the branch needed

All seven CI jobs pass on `27c8870ba` (`gh pr checks 322`), and the new suite runs inside the existing release-contract step:

```
Type Checking  Check release contracts  node --test scripts/upload-release.test.mjs
ok 6 - (b) a starter asset already on the release is deleted and re-uploaded
ok 9 - the CLI repairs a partially-attached release: starter deleted, same-size skipped, the rest uploaded
# tests 9   # pass 9
```

One commit was added after opening this PR: `chore(evidence): re-stamp scriptsTree for the attach-path change`. The desktop suite's shipped-manifest case failed on the first head because `docs/evidence/manifest.json` pins `scriptsTree` to the tree it ships in, and this change edits `scripts/` — `stampFailures` named `scriptsTree` as the only mismatch, so it was re-derived from `git rev-parse HEAD:scripts` (3380b86a9 → 17530bcf0). No frame was re-captured and none could have been: this is release machinery under `scripts/`, with no rendered surface. `srcTree`, `frames`, `surfaces`, `countsMean` and `partialCapture` are unchanged, and nothing was re-captured to make a gate pass.

`pnpm check-evidence`'s sweep half could not run here — another sweep holds the machine lease, so it deferred with exit 75. The lease-free half (`scripts/evidence-manifest.test.mjs`, in `pnpm test:desktop`) is the one that covers this stamp, and it passes.

damianvtran:
### Agent review — round 1

Reviewer: `reviewer` subagent on **deepseek/deepseek-flash** (no provider fallback observed during this round).
Scope: `dde14e9dd..27c8870ba` (`27c8870ba43613b15d9a6facb6904ddf6bc95c1e`, base `dde14e9dd`).

**Verdict: `clean` — no BLOCKER, no MAJOR.** Round 1 is TERMINAL for the code in this diff. The four MINORs below are follow-ups; two of them (M3 is a rebase, not code) are noted because they affect what happens next, not because they need a review round. Everything destructive in this diff I argued rather than accepted, and the transport claim I measured rather than restated.

---

#### What I verified, and how

**1. Invariants — preserved, read line by line and exercised.** `classifyAsset` (`scripts/upload-release.mjs:155-187`) reaches `remove` only from the `incomplete` branch, and only after `complete.length === 0` and `unknown.length === 0` have been excluded — so no path that has a complete asset in hand can produce a delete. The terminal refusals are unchanged: same name + same size → skip; same name + different size → throw naming both sizes (`:181-183`); unknown state → throw by name (`:168-172`). The pins (`EXPECTED_SOURCE_SHA` `:481`, `EXPECTED_RELEASE_ID` `:483`), the `validateRelease(..., false, expectedReleaseId)` call (`:485-491`), `PLATFORM_INSTALLERS`/`PLATFORM_UPDATE_METADATA` (`:65-84`) and the entry point's no-create/no-metadata path are untouched by the diff. `validate-release.mjs` and `release-state.mjs` are not in the diff at all. The repair path still refuses to promote: release-safety's `CLI: an older release's re-run attaches assets but does not promote` passes.

**2. Deleting a `starter` asset — safe, and here is the argument.** What actually serialises concurrent runs for one tag is `publish.yml:29-31`: `group: publish-${{ github.event.release.tag_name || inputs.release_tag }}`, `cancel-in-progress: false`. Both the release-event path and the repair-dispatch path resolve to the same tag, and the group serialises rather than cancels, so two publishes for one tag queue behind each other. I grepped every workflow for another asset writer: `publish.yml` is the only one (`releases/assets`, `uploads.github.com`, `gh release upload` appear nowhere else), and `signed-update-candidate.yml` does not touch this release's assets. That closes the "another run is mid-upload" case, and two further facts make the delete non-destructive even if the serialisation were bypassed: a `starter` record holds no downloadable bytes (nobody can fetch it, and `release-state.mjs`'s `missingAssets` already counts only `uploaded`), and the name being deleted is by construction the name this run is about to re-upload from a local file — so a delete cannot remove the release's only copy of anything. There is one residual race, and the PR already lists it under "Not addressed": the list→DELETE window is not re-checked, so a `starter` record can be deleted and the retry's POST then refused by a concurrent finalisation — that fails the run with the real HTTP status, and the next re-run repairs it. Failed run, no lost asset, correct direction. I did not require an age/ownership guard; a `state`-conditional delete does not exist in the API, and a re-check immediately before the subprocess would only narrow a window already bounded by the concurrency group.

**3. Transport — streamed and bounded, measured not restated.** `streamUpload` (`:295-344`) passes `body: createReadStream(artifact.path)` with `duplex: "half"`, `content-length` from the stat taken at plan time, and `signal: AbortSignal.timeout(timeoutMs)`; memory does not scale with file size, which is the property that decides this. I drove the real `streamUpload` against a local stub with a 256 MB file: 256.0 MB received in 4397 chunks, `content-length: 268435456` exact, no `transfer-encoding` (a length-delimited body, so the endpoint can distinguish truncated from complete). Peak RSS delta: 51.0 MB on a 32 MB file, 21.1 MB on the 256 MB file — non-scaling, i.e. GC noise rather than a body buffer — and **0.1 MB** against a deliberately slow consumer 64 MB file, which is backpressure working. Retry policy is as reported: `transient` is set on the transport catch (timeout/reset) and on `status >= 500 || status === 429` (`:336`), never on other 4xx; `index < attempts` gates the retry (`:431`), backoff doubles from 5 s. The attach job pins Node 22.13.1 in every build job and `engines` is `>=22.13.1`, so `fetch`/`AbortSignal.timeout`/`duplex: "half"` are available.

**4. The tests discriminate — I restored the old behaviour four times and each restoration went red.** Scratch copy at `/tmp/review322-scratch` (`cp -Rc` of `scripts/`), one mutant at a time, `diff -q` back to the original afterwards:

| mutant (old behaviour restored) | `upload-release.test.mjs` result |
| --- | --- |
| V1: a `starter` record is refused, never repaired (the old `checkAssetCollisions` rule) | (a)-stall, (b) and the CLI repair case **fail** (6/9) |
| V2: same-size complete asset treated as a collision, not skipped | (a)-stall, (b), (c) and the CLI case **fail** (5/9) |
| V3: a failed attempt is terminal — no retry at all | all four (a) retry cases **fail** (5/9) |
| V4: the old catch-all wording, every failure a "collision" | (a)-budget, (a)-stall, (a)-4xx **fail** (6/9) |
| V5: classification moved out of the pre-write plan into the per-artifact loop | 9/9 pass, 89/89 release-safety pass — **see M1** |

(a)–(d) all exist, all assert behaviour the stub observed (requests, byte counts, headers, recorded DELETE argv) rather than the implementation's internals, and the CLI case drives the real script end to end with a fixture `gh` as the only one on `PATH`. V1–V4 prove they can fail; V5 is the gap.

**5. Docs match the code.** The script docblock's four-case rule, the `publish.yml:600-608` step comment ("a COMPLETE asset is never replaced"), and `AGENTS.md:1202-1222` ("Never replace a complete asset; never treat a `starter` record as a collision") all state what the code now does. I found no sentence in them that is false — the one imprecision I could measure is N1 below.

**6. Hygiene.** `package.json` is not in the diff (no version bump). `gh pr checks 322` on `27c8870ba`: 8/8 pass. Manifest re-stamp verified by hand — `scriptsTree 17530bcf…` equals `git rev-parse 27c8870ba:scripts`, `srcTree 9b6d6323…` equals that commit's `src` tree, so the stamp is derived from this head and not carried. `test-inventory` exemption is present and the new file runs in its own `node --test` step (`ci.yml:212`); `scripts/upload-release.test.mjs` is covered, `test-inventory.test.mjs` passes. `isEntryPoint`/sparse-checkout: the new module imports only `node:*`, `./entry-point.mjs` and `./validate-release.mjs`, all in the attach job's sparse list. No version bump, no reviewer requested (`reviewRequests: []`), no `@handle` anywhere in the body or comments (grepped), non-draft, assigned to `damianvtran`.

Suites I ran myself, on the branch head: `node scripts/upload-release.test.mjs` → tests 9, pass 9, fail 0. `node scripts/test-release-safety.mjs` → tests 89, pass 89, fail 0. `node --test scripts/test-inventory.test.mjs` → 1/1. `node --test scripts/test-publish-workflow.mjs scripts/test-validate-release.mjs` → 108/108. `node --test scripts/test-version-bump-guard.mjs` → 33/33. `pnpm lint:scripts` → 4 changed files lint-clean against `origin/main` (`dde14e9`). `pnpm lint` → 728 files, exit 0, 67 pre-existing warnings — identical to the PR body's figures.

---

#### Findings

**M1 — MINOR — the "nothing is uploaded before a collision is found" property is real in the code but pinned by no test.** `scripts/upload-release.mjs:496` classifies every artifact before the first write, and both the docblock (`:206-210`) and the PR body claim a collision therefore stops the run with nothing uploaded rather than halfway through the artifact set. Every case that asserts zero requests uses a *single* artifact (`upload-release.test.mjs:520-575`, `test-release-safety.mjs:345-371`), so the property is unpinned: V5 above — replacing the pre-write plan with `artifacts.map(... {action: "upload"})`, which defers classification into `uploadArtifact` and uploads an earlier artifact before a later one collides — passes all 9 cases and all 89 release-safety cases. Fix: one case with two artifacts (`app.dmg` uploadable, `app.exe` already complete at another size) asserting `requests.length === 0` and `removed === []`.

**M2 — MINOR — one inline comment asserts byte identity that a size comparison cannot establish.** `scripts/upload-release.mjs:174-177`: *"comparing against the first is the conservative reading of it (refusing when the sizes disagree, and skipping only when the bytes on the release are the bytes we built)"*. Size equality is not byte equality: a same-length `latest-mac.yml` (855 bytes) from an earlier build of the same version would be skipped, and the updater would then reject the hash its own metadata carries. This comment is the one place in the diff that justifies the size rule by byte identity — the header (`:29-32`) and `AGENTS.md:1209-1211` say "same size" and are accurate. Fix: say "skipping when the size matches, which the version-immutability rule makes equivalent to the bytes we built", or state that assumption explicitly.

**M3 — MINOR — the PR is DIRTY against current `origin/main` (`44cb3af3c`), and the conflict is the evidence stamp.** `git merge-tree --write-tree origin/main 27c8870ba` reports one content conflict, `docs/evidence/manifest.json`. No code conflict; the re-stamp has to be re-derived on the fused tree. Flagging it because it is the only thing between this head and a merge, and because it means this round's Scope will stop matching head the moment the rebase lands — per the batched-remediation rule that is one convergence round over the rebase diff only if the rebase touches anything but the stamp.

**M4 — MINOR — the one write this path makes is still unbounded.** `removeAsset` (`:356-368`) calls `execFileSync` with no `timeout`, so a stall on the DELETE hangs the job exactly the way the old `--input` POST did — the failure mode this PR exists to remove, and its own docblock at `:35` names it. Probability is low (a ~100-byte delete against a host that just answered a list), but the fix is one option key, and the asymmetry is worth closing while the doctrine is fresh.

**N1 — NIT — "nothing is buffered" is not literally true.** `AGENTS.md:1204` and the script header (`:34`) both say the installer is streamed and "never buffered". Measured against the real transport with a 256 MB file, peak RSS delta was 21.1 MB (51.0 MB on a 32 MB file — no scaling, so this is not a body buffer), and 0.1 MB with a slow consumer. The claim's intent holds and I verified it; "never buffered whole" is the accurate wording.

**N2 — NIT — the attach job takes the image's default Node while every other job pins one.** `publish.yml`'s `attach-to-release` has no `actions/setup-node` step, although the build jobs pin `22.13.1`. That was harmless when the step only needed the `gh` binary; it now also needs `fetch`, `AbortSignal.timeout` and `duplex: "half"`. Fine on today's `ubuntu-latest`, deterministic only if pinned.

---

#### Could not verify

- `pnpm check-evidence` — the shared evidence lease is held by another lane's sweep and the machine is under a load hold, so the evidence gate itself was not run (the PR defers it with exit 75). I substituted the checks I could make cheaply and decisively: both tree stamps in `docs/evidence/manifest.json` were derived from this head by hand (`git rev-parse`), and `dirtyWorkingTree`/`countsMean` are untouched by the diff.
- The live attach path against real GitHub — no release was cut, so the 10-minute deadline, the real `starter` finalisation timing and the delete/422 race are argued from the stub plus the API's documented behaviour, not observed against the real endpoint. That is inherent to this change and is what the stub suite is for.
- Visual/rendered evidence is not applicable: no user-visible surface in the diff.

damianvtran:
### QA report — round 2

`Reviewer:` qa-tester (model: `deepseek/deepseek-flash`) · `Scope:` `origin/main` (`dde14e9dd`) … head **`27c8870ba`** · independent pass, no repo edits.

**Verdict: PASS on `27c8870ba`** — 11/11 real-API assertions, 7/7 independent stub assertions, 1 reproduced before/after, and every suite the branch claims, run here. No FAIL on head. Two notes and three disclosed limits below; nothing blocks the merge.

#### Environment

- Own worktree `/tmp/qa322-r2/tree`, detached at `27c8870ba43613b15d9a6facb6904ddf6bc95c1e`; the primary checkout was never touched, and `git status` in my worktree is clean.
- macOS Darwin 25.6.0 (arm64), node **v26.5.0**, gh CLI authenticated as `damianvtran` (scopes `repo`, `workflow`). CI runs node 22 — see Gaps.
- Target: a **temporary DRAFT release**, id `391084243`, on tag `v0.4.1` — the one tag in this repository with no release of its own, so **no tag ref was created** (tag ref lines 171 before and after, none `untagged`). A draft emits no `release: published`: `gh run list` shows no `release` event for the whole session, only other sessions' `pull_request`/`push` runs.
- **Teardown verified**: every asset deleted, release deleted (`GET /releases/391084243` → `404`), zero drafts left on the repo, and the releases list is **byte-identical** to the snapshot taken before I started (`diff` clean).
- No published release, no tag, no workflow dispatch, **no build and no dependency install** (the primary checkout's `node_modules` is symlinked in read-only, so no second tree was materialised and no store was copied).

#### Matrix

Real API — the branch's own `streamUpload` / `uploadArtifact` / `planAssetUploads` driven against GitHub, real bytes on a real release:

| # | Surface | Command | Actual output | Verdict |
|---|---|---|---|---|
| A | first upload lands, size matches the file | `uploadArtifact({artifact: {latest-mac.yml, 1024 B}})` → `streamUpload` | `action=upload status=201 POSTs=1 api.size=1024 state=uploaded sha256(round-trip)=6ab72eeb9e77b075… local=6ab72eeb9e77b075…` | **PASS** |
| A2 | *organically observed*: the endpoint answered 5xx and the retry carried it | same, first probe pass | `Attempt 1 of 3 failed for latest-mac.yml: … HTTP 500 from the upload endpoint: Error creating asset temp dir; retrying in 5000 ms` / `Removed an incomplete upload of latest-mac.yml (asset 571213414, state starter) left behind by an earlier attempt` / `Uploaded latest-mac.yml (1024 bytes) … [HTTP 201]` | **PASS** |
| B | re-run of the same file: skipped, zero POSTs | `uploadArtifact(<same artifact>)` again | `action=skip POSTs=0 assetId=571223612 (unchanged true) digest=sha256:6ab72eeb9… (unchanged true)` `log=["Already attached at this size, nothing to upload: latest-mac.yml (1024 bytes, asset 571223612)"]` | **PASS** |
| B2 | re-run of a whole **set**: every file skipped, zero POSTs | `planAssetUploads` + `uploadArtifact` over 3 files, run twice | `first run: POSTs=["latest.yml","latest-linux.yml","probe-set.dmg"]; second run: POSTs=0 skipped=["latest.yml","latest-linux.yml","probe-set.dmg"]; asset ids unchanged=true` | **PASS** |
| C | same name, different size: refused with both sizes, complete asset untouched | `uploadArtifact({artifact: {latest-mac.yml, 2048 B}})` against the 1024 B asset | `Asset latest-mac.yml is already attached to this release with a different size: attached 1024 bytes, this build's 2048 bytes. Refusing to replace a complete asset; …` → asset id `571223612` **unchanged**, size still 1024 | **PASS** |
| D1 | DELETE by asset id — the argv `removeAsset` issues — on a real asset | `gh api --method DELETE repos/damianvtran/local-operator-ui/releases/assets/571223922` | `deleted; asset absent from the release: true; second DELETE -> gh exit 1: gh: Not Found (HTTP 404)` | **PASS** |
| D2 | a **genuine `starter`** record repaired: delete by ID, then re-upload | `streamUpload(timeoutMs=300)` aborted mid-body, then `uploadArtifact` | `aborted attempt: starter-probe.bin: TimeoutError: … after a 0.3s attempt deadline \| GitHub left a record: state=starter id=571223988 \| Removed an incomplete upload of starter-probe.bin (asset 571223988, state starter) left behind by an earlier attempt / Uploaded starter-probe.bin (8388608 bytes) … [HTTP 201] \| now: state=uploaded size=8388608 round-trip sha256 match=true` | **PASS** |
| E1 | a real duplicate-name POST: the endpoint's own body, parsed and classified | `streamUpload` for a name already attached | `name=UploadAttemptError status=422 transient=false message=latest-mac.yml: HTTP 422 from the upload endpoint: Validation Failed; {"resource":"ReleaseAsset","code":"already_exists","field":"name"}` | **PASS** |
| E2 | a real 4xx is **not** retried: one POST against a 3-attempt budget | `uploadArtifact({api: <reader that cannot see the asset>})` | `POSTs=1 (attempts default 3) … Asset upload failed on attempt 1 of 3: … HTTP 422 … No complete asset was replaced; re-run this job to retry, since an incomplete upload left behind by this attempt is removed rather than blocking the name it holds` | **PASS** |
| G | nothing silently replaced: every asset on the release is the one this run uploaded | `gh api …/releases/391084243/assets` | `latest-linux.yml id=571223761 state=uploaded size=800 digest=sha256:37d062b82… \| latest-mac.yml id=571223612 …1024… \| latest.yml id=571223737 …700… \| probe-set.dmg id=571223791 …4096… \| starter-probe.bin id=571224050 …8388608… \| count=5 expected=5`, all `digest-matches-local=true` | **PASS** |
| F | release metadata untouched by the attach path | `gh api …/releases/391084243` before vs after every upload | `name="QA322 probe (delete me)" body-identical=true tag=v0.4.1 draft=true prerelease=false target=main -> unchanged=true` | **PASS** |
| 0 | DELETE-by-id works repeatedly (used to reset the release) | `gh api --method DELETE …/releases/assets/<id>` × N | `removed N; assets now 0` | **PASS** |

Stall / retry boundary — driven on an **independent** endpoint I wrote (`/tmp/qa322-r2/harness/stub.mjs`, ~110 lines), not the branch's harness, so the claim is not only checked by the code's own test. `streamUpload` and `uploadArtifact` are the real ones; the endpoint fakes GitHub's `starter`-on-start behaviour and answers per attempt:

| # | Surface | Command | Actual output | Verdict |
|---|---|---|---|---|
| S1 | a stalled attempt (accepted, never answered) is retried and lands | attempt 1 never answers, attempt 2 `201` | `Attempt 1 of 3 failed … TimeoutError: … after a 0.7s attempt deadline; retrying in 5000 ms` / `Removed an incomplete upload … (asset 1000, state starter)` / `Uploaded … [HTTP 201]` | **PASS** |
| S2 | a 5xx is retried (transient) | attempt 1 `500`, attempt 2 `201` | `Attempt 1 of 3 failed … HTTP 500 from the upload endpoint: Server Error; retrying in 5000 ms` → `Uploaded … [HTTP 201]` | **PASS** |
| S3 | a 4xx is an answer, not a fault to wait out | every attempt `422` | `Asset upload failed on attempt 1 of 3: … HTTP 422 … already_exists` — **one** attempt against a 3-attempt budget | **PASS** |
| S4 | the budget is bounded: 3 attempts, then a failure naming the file | every attempt `503` | `attempts=3, backoff waits=[5000,10000]` (doubling from 5 s), `… failed on attempt 3 of 3: … HTTP 503 …` | **PASS** |
| S5 | a stall that never clears fails on the deadline | `timeoutMs=600`, no answer ever | `… stub-probe.bin: TimeoutError: … after a 0.6s attempt deadline …` | **PASS** |
| S6 | what the endpoint really receives: streamed, bounded by `content-length` | `streamUpload` on a 65536 B file | `content-length=65536 (file 65536) transfer-encoding=(absent) content-type=application/octet-stream bytes received=65536 in 2 chunk(s) authorization=Bearer <redacted>` | **PASS** |
| S7 | an attempt failure the transport did **not** classify as transient is not retried | attempt rejects a non-transient `UploadAttemptError` | `attempts=1 final: … UploadAttemptError: boom …` | **PASS** |

**The bug reproduced and fixed against the same real release state** (both versions, one release, GitHub's own answers):

```
starter manufactured: true state=starter id=571224380
  (client: repair-case.dmg: TimeoutError: … after a 0.3s attempt deadline)

[BASE dde14e9dd] checkAssetCollisions(...) -> Asset filename collisions: repair-case.dmg
  the same state the operator met, through the old CLI's catch: "Asset upload failed: repair-case.dmg; existing assets were not replaced"

[HEAD 27c8870ba] uploadArtifact(...) on that same starter:
  Removed an incomplete upload of repair-case.dmg (asset 571224380, state starter) left behind by an earlier attempt
  Uploaded repair-case.dmg (8388608 bytes) to release 391084243 [HTTP 201]
  -> state=uploaded size=8388608 digest-matches-local=true action=upload
```

**CLI end-to-end, pins enforced before any write** (against the real published `v0.26.9`, read-only; its asset count was 1 before and 1 after, so nothing was written):

| # | Surface | Command | Actual output | Verdict |
|---|---|---|---|---|
| CLI-2 | a wrong `EXPECTED_RELEASE_ID` refuses | `RELEASE_TAG=v0.26.9 EXPECTED_SOURCE_SHA=dde14e9dd… EXPECTED_RELEASE_ID=123456789 node scripts/upload-release.mjs` | `Tag v0.26.9 resolves to commit dde14e9dd…` then `Release ID 391031982 does not match expected release ID 123456789 (node id RE_kwDOOCmy184XTqyu)`, exit 1 | **PASS** |
| CLI-3 | a wrong `EXPECTED_SOURCE_SHA` refuses | `EXPECTED_SOURCE_SHA=000…0` | `Release v0.26.9 (id 391031982) state: published, prerelease: true` then `Tag SHA dde14e9dd… does not match expected source SHA 000…0`, exit 1 | **PASS** |
| CLI-4 | a malformed pin refuses before any API call | `EXPECTED_SOURCE_SHA=abc` | `EXPECTED_SOURCE_SHA must be a full 40-char hex SHA for manual dispatch, got: abc`, exit 1 | **PASS** |
| CLI-1 | **the shipped CLI cannot be pointed at a draft release at all** | `RELEASE_TAG=v0.4.1 EXPECTED_SOURCE_SHA=d2cd48efe… EXPECTED_RELEASE_ID=391084243 node scripts/upload-release.mjs` | `Tag v0.4.1 resolves to commit d2cd48efe6bdcc549594beee2b25ac7dd59ed650` then `GitHub metadata lookup failed: /releases/tags/v0.4.1`, exit 1 | **BLOCKED** |

**CLI-1 is the BLOCKED row, and it names a real obstacle, not a workaround.** The brief's plan was to point the CLI at a temporary draft. That cannot reach the upload path: `GET /releases/tags/<tag>` **does not return drafts** (404 — that is GitHub's answer, not the script's), and `scripts/validate-release.mjs:147` independently requires `draft === false` and a non-null `published_at`. Both are **pre-existing and untouched by this PR** (`validate-release.mjs` is not in the diff). So the intended surface was covered the only way it can be: the **same code the CLI calls** — `streamUpload` as the transport, `uploadArtifact` as reconciliation+retry, wired exactly as the entry point wires them (`attempt: streamUpload`, `remove: removeAsset`) — driven against a real draft release's real asset endpoint. `removeAsset` is not exported by the module, so case D1/D2/D2-before-after import it from a byte-identical copy with one appended `export { removeAsset };` line (`diff` shows only that line) rather than re-implementing it. Nothing in rows A–G, S1–S7 or the before/after is a re-implementation of the paths under test.

#### Suites (run here, counts, not the author's summary)

| Suite | Result |
|---|---|
| `node --test scripts/upload-release.test.mjs` | **tests 9, pass 9, fail 0** (the branch's claim of 9 cases holds) |
| `node --test scripts/test-release-safety.mjs` | **89, pass 89, fail 0** |
| `node scripts/test-validate-release.mjs` | **24 passed, 0 failed** (`All validation tests passed`) |
| `node --test …test-release-safety test-publish-workflow test-version-bump-guard` (the CI line) | **229, pass 229, fail 0** |
| `node --test …release-baseline(12) release-candidate(2) entry-point(5) require-report(8)` | **27, pass 27, fail 0** |
| `node --test scripts/test-inventory.test.mjs` | **1, pass 1, fail 0** |
| `node scripts/check-scripts-lint.mjs` | `4 changed file(s) under scripts/ are lint-clean against origin/main (dde14e9)`, exit 0 |

Disclosure: the first pass of `test-publish-workflow.mjs`, `test-version-bump-guard.mjs` and `entry-point.test.mjs` failed **in my worktree** with `Cannot find module 'electron-builder/package.json'` / `Cannot import '@electron/notarize'` — a bare worktree has no `node_modules`. Sharing the primary checkout's tree read-only (no install, no copy) turned all three green; per the flake rule I diffed them against a sibling run of the same code, and the failures are absent from CI on this head. Not a regression, and worth knowing that these three suites need a populated `node_modules` to run at all.

#### Notes (no FAIL; author's call, neither blocks)

- **Q-1 (pre-existing shape, not introduced here).** `files: artifactFiles(process.env.ARTIFACTS_DIR || "artifacts")` is evaluated as an **argument** to `uploadRelease`, so the artifacts directory is read *before* any pin validation. With a missing `ARTIFACTS_DIR` the CLI reports `Release upload failed: ENOENT; inspect artifact inputs` and exits 1 — the pin checks never run, and the message says nothing about the real fault. The baseline has the identical shape (`git show dde14e9dd:scripts/upload-release.mjs` line 130), so this PR neither caused nor worsened it; it did cost me two runs while building this matrix, and it means a renamed artifacts directory in CI reads as a mystery rather than as "no installers were built".
- **Q-2 (worth one line in AGENTS.md, not a blocker).** The attach path has **no pre-production rehearsal** available: a draft is unreachable by tag and refused by the pin, and a published release is the only thing the CLI will accept. The PR's AGENTS.md text explains that a repair is a re-run, which is the important half; the "you cannot dry-run this against a draft" half is recorded nowhere, and it is what makes the branch's stub suite load-bearing rather than optional.
- **Probe defect, disclosed.** Case D1 first reported FAIL because my probe read `gh`'s **exit code (1)** as the HTTP status; the asset had in fact been deleted. Fixed to read gh's stderr, re-run: `second DELETE -> gh: Not Found (HTTP 404)`, PASS. No product implication.

#### Gaps I could not cover (said plainly rather than implied)

- **Node version.** All of the above ran on node **v26.5.0**; CI pins node 22. The 7/7 CI run on this head is the v22 signal, not mine.
- **The shipped constants were not exercised at their real values.** `UPLOAD_ATTEMPT_TIMEOUT_MS = 600_000`, `UPLOAD_ATTEMPTS = 3` and `UPLOAD_BACKOFF_MS = 5_000` are read, not driven: to make an abort happen at all I used 0.3–0.7 s deadlines and a no-op sleep, so **the 10-minute ceiling and the real 5 s/10 s waits have never elapsed in anger**. The *policy* (bounded, 3 attempts, doubling, transient-only) is proven; the wall-clock generosity is not.
- **No 158 MB body.** My largest real streamed artifact was 8 MiB. Streaming is size-independent and `content-length` is set from the `stat`, but the exact file that broke v0.26.5 was not pushed through.
- **No build, no desktop suite, no `pnpm check-evidence`** (per the constraints), so the paths that consume this script inside the packaged pipeline — the sparse `workflow/scripts/` checkout and the artifacts layout in the attach job — are read, not run. The `starter` record I repaired came from a real client-side abort on the real endpoint (the same mechanism), not from GitHub stalling on its own.
- The four shells are still shells: `v0.26.6` 0 assets, `v0.26.7` 0, `v0.26.8` 1, `v0.26.9` 1 (`latest-mac.yml`, 855 B). I did not dispatch a repair — that is a published-release write and outside this round's constraints.

#### The one thing I would most want a human to look at

**The repair this PR makes possible has never been run through the real pipeline, and the four shells depend on a merge to get it.** The attach job checks out `ref: ${{ github.workflow_sha }}` and runs `workflow/scripts/upload-release.mjs`, so a repair dispatch picks up **`main`'s** copy of the script — the fix cannot repair `v0.26.6`–`v0.26.9` until it is merged, and the first real exercise of the new delete-then-upload branch against a published release will be that dispatch. Everything in this matrix says it will work (D2 manufactured the exact wreckage and repaired it on the live endpoint, and the before/after shows the base refusing that same state), but the code path and the pipeline have not yet met. I would merge this and then, rather than assuming, run one repair dispatch on `v0.26.9` and watch it — that single run converts the whole matrix from "proven in isolation" to "proven in production".

QA pass count: **11 real-API + 7 stub + 1 before/after + 4 CLI rows**, 0 FAIL on `27c8870ba`; all probe artifacts and the draft release removed and the repository's release/tag state verified identical to how I found it.

damianvtran:

Head `fbbf2a8cb` (rebase onto `44cb3af3c`, plus **one** remediation commit `fbbf2a8cb` — the round is that single commit). CI: **8/8 green on `fbbf2a8cb`**.

Every finding below was fixed in `fbbf2a8cb`. Two of them I proved by restoring the old behaviour in a scratch copy first: a fix whose test cannot fail is a comment.

---

**M1 — fixed.** New case `(e) a collision anywhere in the set stops the run before the FIRST write`: `app.dmg` is absent from the release, `app.exe` is complete at 700 bytes while the build's is 2048, and the case asserts `requests.length === 0`, `removed === []` and the asset set unchanged. **Discrimination proved** by hand: a scratch copy with `planAssetUploads` handing back `{action: "upload"}` verdicts (classification deferred into `uploadArtifact`, exactly the V5 mutant) now runs **7/10** — case (e) fails with

```
✖ (e) a collision anywhere in the set stops the run before the FIRST write
  AssertionError: Expected values to be strictly equal: 1 !== 0
```

i.e. the run posted `app.dmg` before refusing `app.exe`, which is the partially-attached release the pre-write plan exists to prevent. Scratch removed; on the head the suite is **10/10**. (`test-release-safety.mjs` keeps its own pre-write case for the single-artifact shape.)

**M2 — fixed.** The comment now states what is compared rather than claiming byte identity: *size*, because that is what the API reports without making anyone download a 158 MB asset, and it names the check that catches an asset of the same length built from other bytes — electron-updater's own verification of the hash its channel file carries. No behaviour change.

**M4 — fixed, and measured rather than asserted.** `removeAsset`'s `execFileSync` now carries `timeout: REMOVE_TIMEOUT_MS` (60s), and its failure line says which kind of failure it was — a status, or the absence of one. Driven end to end: the real CLI, a fixture `gh` whose DELETE hangs in `sleep 300`, artifacts laid out as the workflow downloads them:

```
elapsed: 61.7s | exit code 1
stdout: … Already attached at this size, nothing to upload: app.dmg (1048576 bytes, asset 9301)
        Uploaded latest-mac.yml (64 bytes) to release 389357596 [HTTP 201]
stderr: Unable to remove the incomplete upload of app.exe (asset 9302), which holds the name this
        upload is refused over: gh gave no answer within the 60s deadline (ETIMEDOUT).
        Delete that asset by hand and re-run
```

and the same rig against a scratch copy with the `timeout:` key dropped (the pre-round shape):

```
elapsed: 75.2s | WATCHDOG: still running at 75s
```

The bound is what changed the outcome. Still deliberately one attempt, not the upload's retry budget: a DELETE that did not answer may or may not have been applied, the following POST is what makes that visible, and a re-run repairs the release either way.

**N1 — fixed.** `scripts/upload-release.mjs`'s header and `AGENTS.md` now say "never buffered whole", not "nothing is buffered". The measurements are the reviewer's (non-scaling peak RSS; 0.1 MB under a slow consumer), and the wording now matches them.

**N2 — fixed.** `attach-to-release` gains `actions/setup-node@v4` with `node-version: '22.13.1'`, the same pin as `build-macos`/`build-windows`/`build-linux`, with the reason in a comment: the step used to run on the image default because it only shelled out to `gh`, and it now also needs `fetch`, `AbortSignal.timeout` and a streamed request body.

---

**Also found and fixed while verifying M1 (not a filed finding).** The stall-retry case scheduled a **250 ms** deadline for the attempt that has to *succeed*, so on a loaded machine the stub's `201` could miss it and the case went red — `Asset upload failed on attempt 2 of 2: … TimeoutError … after a 0.25s attempt deadline`. Measured at roughly one run in five to eleven (it reproduced on run 11 of a 14-run loop after passing the previous ten). The retry case now uses a 2s deadline (the stall still fires and the retry still carries the full file), and the case that *proves the boundary* keeps its tight 200 ms deadline while asserting only what a tight deadline can determine — that the bound is named and that a second attempt happened — instead of how many requests the stub managed to read. After the fix: **16 consecutive green runs, four of them under deliberate CPU load** (`yes > /dev/null` × 6).

---

**Rebase proof (M3).** Rebased `dde14e9dd → 44cb3af3c`; the only content conflict was `docs/evidence/manifest.json`, and main's delta over my touched files was that file alone (`git diff --stat dde14e9dd 44cb3af3c -- <my files>` → one file, 2 insertions, 2 deletions). Blob comparison, pre-rebase head `293bc911b` vs post-rebase head `fbbf2a8cb`, every file this branch touches except the stamp:

```
IDENTICAL  .github/workflows/ci.yml              6c6cfd4c69699e565d361a1d1beec508be9ec549
IDENTICAL  .github/workflows/publish.yml         190c193cad66b7ec7458ec33e0d4c2ec535454f1
IDENTICAL  AGENTS.md                             bae52f4981cc2a6108c1193095f376fbc14468e3
IDENTICAL  scripts/upload-release.mjs            2b2c90d66e0b737fcf5f9651779b5b77a30119ca
IDENTICAL  scripts/upload-release.test.mjs       e16f63886e810bd8e976902d292e0e22c5f768dc
IDENTICAL  scripts/test-release-safety.mjs       0f96a4b3aeb3a27e0cc040c4c257f44b5d676e5e
IDENTICAL  scripts/test-inventory.test.mjs       eae44f0c2c67d9540aca0e2ab1a56bbb2cab2479
```

`git diff --stat 293bc911b fbbf2a8cb` over the same set says `docs/evidence/manifest.json | 4 ++--` and nothing else — the two stamp lines. `git range-diff dde14e9dd..293bc911b 44cb3af3c..fbbf2a8cb`:

```
1:  9aa04b202 = 1:  081867db2 fix(release): make asset attachment streamed, bounded and repairable
2:  4fd49ac12 = 2:  556e4eac7 refactor(release): export only the attach path's consumed surface
3:  27c8870ba < -:  ------- chore(evidence): re-stamp scriptsTree for the attach-path change
4:  293bc911b ! 3:  fbbf2a8cb fix(release): reserve the set before writing, bound the DELETE, correct the docs
```

The old re-stamp commit became empty once main's manifest was the base (its content now rides `fbbf2a8cb`), which is why the round is one commit and why the '!' is the two stamp lines plus the message.

Stamp re-derived from the rebased tree, by the rule: main's manifest as the base, every declaration kept, nothing unioned (main's change was the two stamps; no set, list or count moved). `scriptsTree` 79bf4f8f9 → **e2d98571e** = `git rev-parse HEAD:scripts`, because this branch edits `scripts/`. `srcTree` stays main's **82945f57e**, which is also `HEAD:src` — no `src/` file is touched here. No frame was re-captured and none is claimed to have been: `stampFailures` named `scriptsTree` as the only mismatch, and `scripts/evidence-manifest.test.mjs` is **34/34**.

---

**Gates on `fbbf2a8cb`** (locale, then CI):

| check | result |
|---|---|
| `node --test scripts/upload-release.test.mjs` | **10, pass 10, fail 0** (was 9) |
| `node scripts/test-release-safety.mjs` | **89, pass 89, fail 0** |
| `node --test …test-publish-workflow test-version-bump-guard test-inventory.test.mjs` | **141, pass 141, fail 0** |
| `node scripts/test-validate-release.mjs` | All validation tests passed |
| `node --test …release-baseline release-candidate entry-point require-report` | **27, pass 27, fail 0** |
| `node --test scripts/evidence-manifest.test.mjs` | **34, pass 34, fail 0** |
| `pnpm lint` | 728 files, 0 errors (67 pre-existing warnings) |
| `pnpm lint:scripts` | 4 changed file(s) lint-clean against `origin/main` (44cb3af) |
| CI on `fbbf2a8cb` | **8/8 jobs green** |

**Not addressed (both from QA's notes, neither raised as a finding, both the manager's call — recorded rather than silently dropped):** Q-1, `artifactFiles()` evaluated as an argument so a missing `ARTIFACTS_DIR` reports `ENOENT` before any pin check runs — pre-existing and unchanged by this PR; Q-2, one line in `AGENTS.md` recording that the attach path has no pre-production rehearsal because a draft is unreachable by tag and refused by the pin. Both would cost a sentence or a line; I left them out of this round because they are outside the findings I was given, not because they are wrong.

damianvtran:
### Agent review — round 2

Reviewer: `reviewer` subagent on **deepseek/deepseek-flash** (no provider fallback this round).
Scope: `27c8870ba..fbbf2a8cb` (`fbbf2a8cbf88a7a797f8c6bf2657e3f48794cf39`, rebase onto `44cb3af3c` plus one remediation commit).

**The round's findings are answered.** Every item I filed in round 1 (M1, M2, M4, N1, N2) is fixed on this head, and I reproduced each verification rather than accepting it. No BLOCKER, no MAJOR — round 2 is TERMINAL, and the round-1 approvals still stand because the rebase is proven content-preserving.

---

#### M1 — the case discriminates; I reproduced the mutant myself

Load-bearing test is behavioural: case (e) asserts `requests.length === 0`, `removed === []`, and the asset set unchanged, against a two-artifact set whose *second* artifact collides while the first is freely uploadable — the shape that separates a pre-write reservation from per-artifact classification. I rebuilt the mutant in my own scratch copy of `fbbf2a8cb` (replacing the pre-write plan with per-artifact `{action: "upload"}` verdicts, i.e. the V5 shape) and ran the suite:

```
✖ (e) a collision anywhere in the set stops the run before the FIRST write
ℹ tests 10   ℹ pass 9   ℹ fail 1
```

— red on (e) alone, exactly the discrimination claimed, and green again (10/10) once the scratch was restored, `diff -q`-clean. On the head, `node scripts/upload-release.test.mjs` is tests 10, pass 10, fail 0.

One factual correction to the reply, for the record: it reports the mutant scoring **7/10**; the same mutant here scores **9/10**, with (e) the only failure. The direction and the conclusion are right, but a reader quoting that figure would be quoting one that does not reproduce — the other cases tolerate deferred classification by design (c still skips inside `uploadArtifact`, (d) is single-artifact). **NIT**, no action needed beyond the next round not repeating the number.

#### M2 — no longer overclaims

`scripts/upload-release.mjs:184-190` now says what is compared is **SIZE**, states why (the API reports a size without making anyone download 158 MB), and names the check that catches a same-length asset built from other bytes: electron-updater's own verification of the hash its channel file carries. I checked that compensating control exists rather than taking it on trust — `node_modules/electron-updater/out/DownloadedUpdateHelper.js:113` compares the cached update's `sha512` against `fileInfo.info.sha512`. The header (`:31-37`) and `AGENTS.md:1204-1211` still say "same size"/"our byte count" and remain accurate.

#### M4 — the bound is real, and the two branches are reachable

`removeAsset` carries `timeout: REMOVE_TIMEOUT_MS` (60 000, `:115-121`) with the reason in the comment, and the failure line splits on `error?.code === "ETIMEDOUT"` to say whether `gh` answered at all. I verified the runtime shapes the branch depends on, which is the part that could have been wrong:

```
execFileSync('sleep', ['300'], {timeout: 1500}) -> { code: "ETIMEDOUT", status: null, signal: "SIGTERM" }
execFileSync('sh', ['-c','exit 7'])            -> { status: 7 }
```

So a deadline produces `ETIMEDOUT` (a *string* code, which `describeError` returns verbatim) and a real refusal produces a numeric `status` — the message cannot confuse the two. I did not re-run the 61.7 s / 75 s rigs; the bound they measured is present in the code as an option key, and their figures are consistent with the mechanism above.

#### N1 / N2 — both accurate

N1: `scripts/upload-release.mjs:31` and `AGENTS.md:1204` now read "never buffered **whole**", which is what my measurements support (peak RSS non-scaling in file size; 0.1 MB under a slow consumer), and `AGENTS.md:1205-1207` now also records the DELETE's deadline. N2: `attach-to-release` gains `actions/setup-node@v4` with `node-version: '22.13.1'`, identical to `build-macos`/`build-windows`/`build-linux`, inserted after the sparse checkout and before the artifact download. It does not disturb the step's assumptions: the run step is a bare `node workflow/scripts/upload-release.mjs` (no pnpm, no registry, no lockfile), and the sparse checkout needs no `node_modules`.

#### The extra it found — verified, and it does not weaken the case

The stall-retry case's `timeoutMs` moved `250 → 2000`; the diff shows **only** that literal and its comment changed inside the case, so its assertions are untouched — two attempts, `[4096, 4096]` bytes received, and the `starter` record removed before the retry — and attempt 1 still stalls the full deadline because the stub never answers it. Cost is ~1.75 s of suite time, which is the right trade against a 1-in-14 flake. The boundary case keeps its 200 ms deadline, and the retry there is still proven by the run's own report (`attempt 2 of 2`).

**NIT — one replacement assertion is vacuous when the stub recorded nothing.** `upload-release.test.mjs:415-421` drops `assert.equal(requests.length, 2)` for `for (const request of requests) assert.equal(request.bytes, 256)`, which passes on an empty array; `requests.length >= 1` would keep the bound honest without reintroducing the scheduling dependence. Low value here because the property is pinned elsewhere (case (a) asserts `[4096, 4096]`, and the CLI case asserts `request.bytes === statSync(...).size` for all five streamed files), which is why this is a nit and not a finding.

#### Rebase proof — reproduced independently

My own run of the same three checks agrees with the reply, hash for hash:

- `git diff --stat dde14e9dd 44cb3af3c -- <this branch's files>` → `docs/evidence/manifest.json | 4 ++--`, one file, nothing else. Main's delta over the files this branch touches is the stamp and only the stamp.
- Blob comparison `293bc911b` → `fbbf2a8cb` over the seven branch files: **IDENTICAL** in all seven (`6c6cfd4c6`, `190c193ca`, `bae52f498`, `2b2c90d66`, `e16f63886`, `0f96a4b3a`, `eae44f0c2`). `git diff --stat 293bc911b fbbf2a8cb` over that set is `manifest.json | 4 +-` alone.
- `git range-diff dde14e9dd..293bc911b 44cb3af3c..fbbf2a8cb`: commits 1 and 2 `=`, the old re-stamp commit now empty (`27c8870ba < -: -------`), and one `!` for the remediation+rebase whose only content hunk beyond the message is the two manifest stamp lines.
- Stamps re-derived from the rebased tree: `scriptsTree e2d98571e…` = `git rev-parse fbbf2a8cb:scripts`, `srcTree 82945f57e…` = `git rev-parse fbbf2a8cb:src`. `node --test scripts/evidence-manifest.test.mjs` → 34/34.

Since the only file upstream also touched is the manifest and main's change to it was itself only the two stamps, there is no semantic conflict to re-check — the round-1 code approvals carry over unchanged. `scripts/test-release-safety.mjs`, `scripts/test-inventory.test.mjs` and `.github/workflows/ci.yml` are byte-identical to the blobs I passed in round 1, so those results still hold without re-running them.

#### Gates on `fbbf2a8cb`

`node scripts/upload-release.test.mjs` → 10/10. `node --test scripts/evidence-manifest.test.mjs` → 34/34. `pnpm lint:scripts` → 4 changed files lint-clean against `origin/main` (`44cb3af`). `pnpm lint` → 728 files, exit 0, 67 pre-existing warnings. `gh pr checks 322` on `fbbf2a8cb` → 8/8 pass. No version bump (`package.json` untouched against the new base), no reviewer requested (`reviewRequests: []`), non-draft, `MERGEABLE`, assigned. The only `@` tokens anywhere in the body or comments are `@electron`, `actions/setup-node@v4` and a literal `@handle` in prose — no human is tagged.

#### QA's two notes — I agree they are out of scope for this diff

Q-1 (`files: artifactFiles(...)` evaluated as an argument, so a missing `ARTIFACTS_DIR` reports `ENOENT` before the pin checks): the call site is unchanged by this PR and its ordering is pre-existing; the pin checks themselves — the thing this PR is about preserving — run inside `uploadRelease` exactly as before, and CI downloads the artifacts before this step. Real, unowned by this diff. Q-2 (no pre-production rehearsal for the attach path, because the tags endpoint 404s drafts and `validate-release.mjs` requires `draft === false`): a true, pre-existing constraint of the release model, not a behaviour introduced here, and recording it is a docs change rather than a fix. Both are correctly listed under "not addressed" rather than silently fixed inside a release-machinery diff, which is also the right call for a change of this risk class.

---

**Verdict: `clean`.** No BLOCKER, no MAJOR; the round-1 findings are answered and the round is TERMINAL on `fbbf2a8cb`. The two NITs are follow-ups at the author's discretion — the vacuous assertion in `upload-release.test.mjs:415-421`, and not re-quoting the 7/10 mutant figure. Nothing in this round is fresh on a later head: a rebase or a further commit needs its own convergence round.

Not re-verified this round, unchanged from round 1: `pnpm check-evidence` (the shared evidence lease is still another lane's; the manifest test and hand-derived stamps stand in), and the live GitHub endpoint (no release cut, so the real `starter` finalisation timing is still argued from the stub plus documented API behaviour).

damianvtran:
### QA report — round 3

`Reviewer:` qa-tester (model: `deepseek/deepseek-flash`) · `Scope:` round-2 head `27c8870ba` → **`fbbf2a8cbf88a7a797f8c6bf2657e3f48794cf39`** · independent pass, no repo edits.

**Verdict: PASS on `fbbf2a8cb`. I would merge on it.** The one production-behaviour change in the remediation is real at runtime (the DELETE now fails bounded), the moved paths were re-driven on the new head, and every suite the branch claims passes on **Node 22.13.1** — the exact version this commit pins into `attach-to-release`. No FAIL. One new note, one live-state observation about `v0.26.9` you should read before deciding.

Delta examined, not assumed: `git diff 27c8870ba fbbf2a8cb` over this PR's own eight files is **+122/−18** — `scripts/upload-release.mjs` (the `REMOVE_TIMEOUT_MS` bound, two comment corrections), `scripts/upload-release.test.mjs` (case (e), the 250 ms→2 s attempt deadline, a loosened request-count assertion), `.github/workflows/publish.yml` (the Node pin), `AGENTS.md`, `docs/evidence/manifest.json`. `scripts/validate-release.mjs` and `scripts/entry-point.mjs` are **byte-identical between the two heads** (0 lines of diff), which is what keeps this round's BLOCKED row unchanged rather than re-argued.

#### Re-driven on `fbbf2a8cb` (real API, real bytes, on a temporary draft)

| # | Surface | Command | Actual output | Verdict |
|---|---|---|---|---|
| B | repeat run of the same file: skipped, zero POSTs | `uploadArtifact(<same artifact>)` | `action=skip POSTs=0 assetId=571280458 (unchanged true) digest=sha256:6ab72eeb9… (unchanged true)` `log=["Already attached at this size, nothing to upload: latest-mac.yml (1024 bytes, asset 571280458)"]` | **PASS** |
| B2 | repeat run of a whole **set**: every file skipped, zero POSTs | `planAssetUploads` + `uploadArtifact` × 3 files, twice | `first run: POSTs=["latest.yml","latest-linux.yml","probe-set.dmg"]; second run: POSTs=0 skipped=[…same three…]; asset ids unchanged=true` | **PASS** |
| C | size-mismatch refusal, complete asset untouched | `uploadArtifact({artifact: {latest-mac.yml, 2048 B}})` | `… already attached to this release with a different size: attached 1024 bytes, this build's 2048 bytes …` → asset `571280458` unchanged, size still 1024 | **PASS** |
| R | **the remediation's headline**: a collision anywhere in the set stops the run before the FIRST write | `planAssetUploads(api, 391098594, [probe-reserve.dmg (absent, 4096 B), latest-mac.yml (2048 B)])` | `refusal: Asset latest-mac.yml is already attached … 1024 bytes … 2048 bytes … \| probe-reserve.dmg uploaded anyway: false` | **PASS** |
| D2 | `starter` repair: delete by ID, then re-upload (the DELETE now inside a 60 s bound) | abort a real upload mid-body, then `uploadArtifact` | `GitHub left a record: state=starter id=571281169` → `Removed an incomplete upload of starter-probe.bin (asset 571281169, state starter) …` → `Uploaded starter-probe.bin (8388608 bytes) … [HTTP 201] state=uploaded, round-trip sha256 match=true` | **PASS** |
| A | first upload lands, size and bytes match | `uploadArtifact({artifact: {latest-mac.yml, 1024 B}})` | `action=upload status=201 POSTs=1 api.size=1024 state=uploaded sha256(round-trip)=6ab72eeb9e77b075… local=same` | **PASS** |
| D1 | DELETE by real asset id | `gh api --method DELETE …/releases/assets/571280992` | `deleted; absent from the release: true; second DELETE -> gh: Not Found (HTTP 404)` | **PASS** |
| G | nothing silently replaced | final `…/releases/391098594/assets` | 5 assets, each `state=uploaded` with `digest-matches-local=true`, `count=5 expected=5` | **PASS** |
| F | release metadata untouched | `…/releases/391098594` before vs after | `name="QA322 probe (delete me)" body-identical=true tag=v0.4.1 draft=true prerelease=false → unchanged=true` | **PASS** |
| E1/E2 | a real duplicate-name POST → 422/`transient=false`; one attempt against a 3-attempt budget | `streamUpload`, then `uploadArtifact` with a stale reader | `name=UploadAttemptError status=422 transient=false … {"code":"already_exists"}`; `POSTs=1 … Asset upload failed on attempt 1 of 3 … HTTP 422` | **PASS** |
| CLI-2 / CLI-3 | wrong pin refuses before any write, against the real published `v0.26.9` | `EXPECTED_RELEASE_ID=123456789` / `EXPECTED_SOURCE_SHA=000…0` | `Release ID 391031982 does not match expected release ID 123456789 (node id RE_kwDOOCmy184XTqyu)`; `Tag SHA dde14e9dd… does not match expected source SHA 000…0`, both exit 1, `v0.26.9` asset count 2 before and 2 after | **PASS** |
| CLI-1 | the shipped CLI against a draft | `RELEASE_TAG=v0.4.1 … EXPECTED_RELEASE_ID=391098594` | `GitHub metadata lookup failed: /releases/tags/v0.4.1`, exit 1 | **BLOCKED** |

**12 PASS rows, 1 BLOCKED, 0 FAIL.** Rows E1/E2/D1/G/F re-ran inside the same probe invocation, so they cost nothing. **Not re-driven, and why:** the base-vs-head before/after pair — the base copy's `checkAssetCollisions` and the head's `classifyAsset` are byte-identical between the two heads (their only delta is a comment), and the head's repair path it exercises was re-driven green as D2 on the new head. The round-2 stub rows S1–S7 were not repeated: the stub endpoint and both transport functions are unchanged, and the retry boundary is re-proven at the suite level below.

**CLI-1 stays BLOCKED for the same reason as round 2** (`GET /releases/tags/<tag>` 404s for a draft; `validate-release.mjs:147` independently requires `draft === false`) — and this round that reason is stronger, because both files are unchanged between the heads. The real endpoint was again driven through the same functions the CLI wires together.

#### The one production-behaviour change: the DELETE's 60 s bound, at runtime

A `gh` that never answers, put first on PATH so the real `removeAsset` spawns it:

| # | Surface | Command | Actual output | Verdict |
|---|---|---|---|---|
| T1 | a `gh` that **never answers**: bounded failure, not a hang | `removeAsset(repo, token, {id:1})` with a hanging `gh` | `returned after 60.0s (deadline 60s) \| attempts=1 \| hung child still alive: "" \| message: Unable to remove the incomplete upload of app.dmg (asset 1), which holds the name this upload is refused over: gh gave no answer within the 60s deadline (ETIMEDOUT). Delete that asset by hand and re-run` | **PASS** |
| T2 | `gh` exiting non-zero | same, with a `gh` that exits 1 | `returned after 49ms \| attempts=1 \| … gh exited with status 1 (Command failed: gh api --method DELETE … / gh: Not Found (HTTP 404) ) …` | **PASS** |
| T3 | what a *real* failing `gh` carries into the line | `removeAsset(repo, <invalid token>, {id:999999999999})` | `… gh exited with status 1 (Command failed: gh api --method DELETE repos/…/assets/999999999999 / gh: Bad credentials (HTTP 401) ) …` | **NOTE** |

So: the bound is real (60.0 s, one attempt, the child killed — checked by pid, `ESRCH`), the new ETIMEDOUT branch is genuinely **reachable** (Node 22.13.1 sets `error.code = "ETIMEDOUT"` on a sync-exec timeout, verified separately), and the failure line names *why* — deadline or status — which is exactly the distinction the comment claims. T3 is the note: the parenthetical now carries `gh`'s own stderr, so the comment's "the subprocess's output is deliberately not echoed" is no longer literally true. What lands is a status line (`gh: Bad credentials (HTTP 401)`), the token is in `GH_TOKEN` and not on any argv, and this is the most useful part of the message — so I would keep the behaviour and soften the comment rather than the reverse. Author's call; not a blocker.

#### Suites, on Node 22.13.1 (nvm has it — the pin is testable locally, and this closes round 2's biggest gap)

| Suite | Result |
|---|---|
| `node --test scripts/upload-release.test.mjs` | **10, pass 10, fail 0** (the branch's claim of 10 holds; case (e) present and passing) |
| the same, 3 consecutive runs at load average **~205 with 7 sibling suites live** | **10/10, 10/10, 10/10** — the 2 s answer deadline in the stall-retry case held under exactly the pressure that flaked the 250 ms version |
| `node --test scripts/test-release-safety.mjs` | **89, pass 89, fail 0** |
| `node --test scripts/evidence-manifest.test.mjs` | **34, pass 34, fail 0** (the claim holds) |
| CI line: `test-release-safety` + `test-publish-workflow` + `test-version-bump-guard` | **229, pass 229, fail 0** |
| `release-baseline` 12 / `release-candidate` 2 / `entry-point` 5 / `require-report` 8 / `test-inventory` 1 | all pass, 0 fail |
| `node scripts/test-validate-release.mjs` | `24 passed, 0 failed` |
| `node scripts/check-scripts-lint.mjs` | `4 changed file(s) under scripts/ are lint-clean against origin/main (44cb3af)` |

`publish.yml` re-parsed as YAML: `attach-to-release` has the `Setup Node.js` → `22.13.1` step **before** `node workflow/scripts/upload-release.mjs`, which is the right order for `fetch`/`duplex: "half"`/`AbortSignal.timeout`. The PR still carries no version bump (`package.json` is not among its eight files).

#### New note, and the live-state observation

- **Q-3 (note, unchanged by the delta).** Round 2's Q-1 still stands and is untouched: `files: artifactFiles(process.env.ARTIFACTS_DIR || "artifacts")` is evaluated as an argument, so a missing artifacts directory reports `ENOENT` before any pin runs. Pre-existing in the base, not this PR's.
- **`v0.26.9` changed under us, and not through the pipeline.** Between my rounds the release gained `local-operator-ui-0.26.9-arm64.dmg` — **158 477 480 bytes, the exact file this PR exists for** — alongside a re-created `latest-mac.yml`, both stamped `2026-09-17T22:06:43Z`. There is **no `publish.yml` run for it**: the newest `workflow_dispatch` run of that workflow is `2026-09-15`, and the run list for 21:45–22:20Z contains only `pull_request` runs from other lanes. So the dmg was attached **out of band**, not by the repair path this PR fixes. `v0.26.6` and `v0.26.7` still have zero assets and `v0.26.8` one. This is outside my remit and I changed nothing there — but it is the thing I would most want a human to look at, because it changes what "run one dispatch on v0.26.9" would now mean: that release would be repaired against a release that already carries its mac dmg, and the two windows/linux artifacts are still missing. Whoever attached it should say so on the release; the pipeline's own repair path remains unexercised in production.

#### Gaps I could not cover

- **The 158 MB body was not streamed by me.** My largest real artifact is 8 MiB; streaming and the `stat`-derived `content-length` are size-independent, but my own rows do not push the file that broke `v0.26.5`. The dmg now sitting on `v0.26.9` was uploaded by something else, not by a run I can inspect.
- **The shipped constants were still not exercised at their real values.** `UPLOAD_ATTEMPT_TIMEOUT_MS` (10 min) and `UPLOAD_BACKOFF_MS` (5 s/10 s) are read, not elapsed: my aborts use 0.2–0.7 s deadlines and a no-op sleep, so the *policy* is proven and the wall-clock generosity is not. The new `REMOVE_TIMEOUT_MS` is the exception — T1 elapsed it for real.
- **No build, no desktop suite, no `pnpm check-evidence`** (the shared lease is another lane's, and the volume has been tight): the packaged-pipeline paths — the sparse `workflow/scripts/` checkout and the artifacts layout — remain read rather than run.

Cleanup verified, same discipline as round 2: draft `391098594` deleted (`404`), every asset it held deleted, zero drafts left, tag refs 171 with none `untagged`, and comparing release **id sets** before/after, every pre-existing release survived — the only difference is my own draft leaving the set and one window-boundary entry (id `221066636`) entering the 100-item page. No published release, no tag, no dispatch, no build, no install, no repo edits; worktree clean at `fbbf2a8cb`.

damianvtran:
### Merged — and what was verified before it went in

Merged as the author's agent under this repo's standing rule that a clean, fresh, independent agent round plus green CI authorizes a merge (`main` configures no required checks; confirmed `gh api repos/damianvtran/local-operator-ui/rules/branches/main` → `[]`). `--admin` was used for the completion, and is disclosed here as the rule requires.

**Rounds on the head that merged (`fbbf2a8cb`):**
- **Agent review — round 2**, terminal, 0 blocker / 0 major — https://github.com/damianvtran/local-operator-ui/pull/322#issuecomment-5722027274 — which reproduced the rebase proof hash-for-hash and re-ran the mutant that proves the new case (e) discriminates.
- **QA — round 3**, PASS, *"I would merge on it"* — https://github.com/damianvtran/local-operator-ui/pull/322#issuecomment-5722098439 — 12 real-API assertions including a genuine `starter` repaired, a repeat set run with zero POSTs, the whole-set reservation against the real API, and the new `removeAsset` bound measured at **60.0 s** with the child killed; suites green under load average ~205 on Node 22.13.1, which is the pin this PR adds.
- Round 1 of each stream is above, with its remediation answered in the round-2 posts.

**Not verified, and not claimed:** `pnpm check-evidence` has still never executed for this repository tonight — the machine-wide sweep lease has been held by other lanes and the volume fell as low as 4.5 GiB, where a full re-capture sweep is unsafe. It is a tree-scoped gate rather than a PR-scoped one, so it can be run on any commit once the machine allows; nothing in this diff depends on it.

**Two QA notes recorded and deliberately not folded into this diff:** the CLI cannot be pointed at a draft release (the tags endpoint 404s drafts and `validate-release.mjs` requires `draft === false` — pre-existing), and `files: artifactFiles(...)` is evaluated as an argument so a missing `ARTIFACTS_DIR` reports `ENOENT` before any pin check (pre-existing, unchanged here). Both are listed rather than quietly fixed inside release-machinery change. One in-scope nit remains for a follow-up: the new stderr parenthetical makes the comment's "output is deliberately not echoed" no longer literally true.

**The point of this PR is now unproven in production and will be proven by the next release.** The two QA rounds drove the real GitHub API through the same functions the CLI calls, but the repair path this enables has never run through `publish.yml`. That happens next: the window after `v0.26.5` is being cut as `0.26.10`, and the publish run's attach step is this change's end-to-end evidence — if it stalls again, the same owner iterates here rather than tagging over it.

damianvtran:
### Shipped in `v0.26.10`

**Release:** https://github.com/damianvtran/local-operator-ui/releases/tag/v0.26.10 — tag → `858b7968e` (the bump merge; bump commit `b8641ead5`), published **22:42:51Z**, **20 assets attached**, `promote release to latest` green, and `/releases/latest` now answers **`v0.26.10`** where it had been `v0.26.5` since midday.

**This PR is in that tag tree**, along with the rest of the window: #281, #306, #300, #310, #307 and #322, plus the four bump merges the failed shells left behind. The window was derived from `v0.26.5` rather than `v0.26.9` because `0.26.6`–`0.26.9` never carried a complete installer set — `0.26.6`/`0.26.7` no assets, `0.26.8` only the macOS channel file, `0.26.9` a single hand-attached arm64 dmg — so the same work rode out in this Release instead, and the Release notes name those four tags rather than leaving the gap unexplained.

Worth recording because it is the only production proof that exists for it: **the attach step that failed four consecutive releases ran green on its first attempt through the fixed script**, which is #322's own end-to-end verification — the streamed, bounded, retryable upload attached all twenty artifacts and the promotion followed.

PR #325, opened 2026-09-17
ci(github): gate each job on a change-scope classifier the local gate shares

damianvtran:
## Summary

Every job in `.github/workflows/ci.yml` used to run on every pull request. A
one-line `docs/` edit paid for a typecheck, a build, the 108-file desktop suite
and a two-runner pack + macOS launch — and the desktop suite ran **twice inside
one job**.

This adds a **change-scope classifier** and gates each job on it, and removes
the duplicated suite run. It is the UI half of the design already landed in the
backend repository: **[damianvtran/local-operator#1245](https://github.com/damianvtran/local-operator/pull/1245)**
(`scripts/ci_scope.py`, `make check-changed`). That PR is the reference for what
"done" looks like, and this one mirrors its shape, its reasons and its assertion
numbers so the two repositories can be read side by side. It also carries the
fix for the BLOCKER its review round 1 found, described below.

Branch `ci/change-scope` is cut from current `main` (`44cb3af3c`).

### `scripts/ci-scope.mjs` (new, node builtins + two sibling files)

The classifier answers one question per job: does anything in this diff appear
in what that job verifies? It is the SINGLE SOURCE OF TRUTH for both answers —
the new `changes` job exports one flag per job from it, and
`pnpm check-changed` runs the same module locally — so CI and the local gate
cannot drift into two opinions.

Design rules, each load-bearing:

- **Fail open.** A path must *match* an inert rule to be skippable; anything
  unrecognised counts as live. A denylist of code paths goes green on the path
  nobody enumerated, which is the pathology `AGENTS.md` names.
- **Every gate reads `<flag> != 'false'`, never `== 'true'`.** An output that was
  never written is EMPTY and `'' == 'true'` is false, so a classifier that
  crashed would produce a fully green PR that ran nothing. A3 is the guard, and
  the `$GITHUB_OUTPUT` writer raises rather than emitting an empty value.
- **Two independent fail-open layers.** An unresolvable base, a failed
  `git diff` or no base at all sets every flag true, prints a `::warning::`
  naming why, and exits 0. Only `--run` reports a gate's own status.
- **The classifier runs from the BASE revision** (`git show
  "$base:scripts/ci-scope.mjs"`), never from the PR's copy, so a PR cannot edit
  the code that decides whether its own gates run. Any `.github/**` diff sets
  every flag true, and the `changes` job itself is never event-gated: every
  heavy job `needs:` it, so an event condition there would skip the whole
  workflow on a push to `main`.
[...]

PR #326, opened 2026-09-17
probe: docs-only diff (expected 2 checks)

damianvtran:
Temporary probe PR for #325. Base is `ci/change-scope` so the classifier runs from that branch's copy. This diff is one prose line under `docs/`. Expected: **2 checks** (Change Scope + Version Bump Guard). Closed and its branch deleted once the check list is captured.


damianvtran:
Probe captured: docs-only diff → Change Scope + Version Bump Guard pass, every heavy job `skipping`. Closing and deleting the branch; the check list and the `changes` job's summary are recorded on #325.

PR #327, opened 2026-09-17
probe: src/** diff (expected the full code set)

damianvtran:
Temporary probe PR for #325. Base is `ci/change-scope` so the classifier runs from that branch's copy. This diff is a source change. Expected: the code suite runs, and the two jobs whose inputs (`package.json`, the lockfile) did not move are deliberately skipped.


damianvtran:
Probe captured: `src/main/index.ts` → lint/check-types/test/npx-sanity-check run, runtime-deps and audit `skipping`. Closing and deleting the branch; the check list and the `changes` job's summary are recorded on #325.

PR #333, opened 2026-09-17
docs(qa): probe the change-scope gate on a prose-only diff

damianvtran:
QA round 2 probe for #325 (independent qa-tester round).

Scope: one new non-evidence file under `docs/`. Base is `ci/change-scope` at `2251304e7`, so the `changes` job must read the classifier from that base revision and classify this diff rather than take the `--all` fallback.

Expected: exactly 2 checks (`Change Scope`, `Version Bump Guard`), every other job `skipped`, all six flags false.

This PR is a probe: it will be closed and its branch deleted as soon as the checks are read. No reviewers requested.


damianvtran:
QA round-2 probe for #325 — read, recorded, and closing. Exactly 2 checks (`Change Scope` pass 14s, `Version Bump Guard` pass 15s); the other six `skipped`; the `changes` summary printed `- diff base: 2251304e7…` and `docs/qa-325-r2-probe.md -> docs` with all six flags false — i.e. the classifier came from the base revision and classified, with no `--all` fallback. Branch deleted.

PR #321, opened 2026-09-17
fix(dictation,update): name a failed transcription's cause, and stop a stale server serving

damianvtran:
# PR Description

## Summary of Changes

Two fixes from the operator's dictation failure of 2026-09-17, where a one-line
change to the transcription relay (`radient-ml/agent-server#35`) and a stale
backend process both contributed. The relay half is fixed; this PR carries the
two app-side defects the incident exposed.

**1. A failed dictation now names its cause.** Both call sites that transcribe
audio (`features/chat/components/message-input.tsx` and
`features/chat/components/canvas/inline-edit.tsx`) answered EVERY failure with one
literal — "Error transcribing audio. Please try again." — while the server's own
reason travelled all the way to the renderer unused. The operator's failure was the
upstream provider's `insufficient_quota`, so the toast asked them to retry
something no retry could fix. `shared/api/local-operator/transcription-failure.ts`
turns the relay's answer into a sentence a person can act on:

| Relay answer | The sentence |
| --- | --- |
| HTTP 402 | `Dictation failed: you're out of Radient credits. Get more credits, then try again.` |
| `OpenAI API error (insufficient_quota): … Visit https://platform.openai.com/account/billing …` | `Dictation failed: the OpenAI account behind dictation is out of credits. If it's yours, top it up at platform.openai.com/account/billing.` |
| the same family with no destination in the body | `Dictation failed: the account behind dictation is out of credits. If it's yours, top it up.` |
| 401 / 403, or a body that refuses the credential | `Dictation failed: your Radient sign-in was refused. Sign in to Radient, then try again.` |
| anything else | `Dictation failed: <the reason inside the server's answer, clipped>. Try again, or report it with the app's log.` |
| no reason at all | `Dictation failed: the server gave no reason. Try again, or report it with the app's log.` |

The family's own shape is one prefix (`Dictation failed: `) and one retry verb, and a
passed-through reason's opening letter is lowered into it — except where a capital
inside the word says it is a brand or an acronym (`OpenAI API error`, `HTTP 500`),
which keep their own spelling (design round 2, D8–D12).

The client now throws a `TranscriptionRequestError` carrying BOTH halves the relay
saw (the status and the body), because a 402 says something the body's wording does
not; the raw detail is still the error's `message`, so the `console.error` both call
sites already had keeps printing the server's own words. The reason the sentence
carries is the one INSIDE a JSON body (`error.message`), with the relay's own
`POST <url> returned 400` framing stripped: the previous revision pasted the whole
[...]

PR #313, opened 2026-09-17
feat(chat): mark all as read in one gesture

damianvtran:
# Mark all as read — clearing the pile of checkmarks in one gesture

The operator's report: *"I'm getting a lot of these building up and need to clear them without clicking into each one"* — the sidebar's **Active chats** list, with a green ✓ on every row whose turn finished while he was looking elsewhere — and on the control itself: *"make sure to review with design so that the button/interaction is minimal but intuitive"*.

This is the **renderer half** of the frozen interface (`~/workspace/lo-read-all/DESIGN.md` §3.1–3.3). The backend route — `POST /v1/desktop/attention/seen`, capability `completion_ack_bulk`, and `AttentionStore.acknowledge_many` — is being written in `local-operator` against the same freeze. Nothing here changes that interface; the op literal, the path, the body shape, the receipt type, the store action names and the pinned five-step `markAllRead` sequence are implemented as written.

Release: patch — the chat sidebar gains a "Mark all as read" control that clears every unacknowledged completion checkmark in one request, with a receipt that names any row the backend refused.

> **Current head: `d85487ab9`.** The sections below describe the heads they were written at and are kept as history; the current gate numbers, the rebase proof and round 2's fixes are in *Rebase + remediation round 2*, further down.

## Summary of Changes

- **`attention.seen`** (`src/shared/desktop-contract.ts`): the request arm and its `desktopEndpoint` case. One POST carrying 1..500 token-bound items in snake_case. The batch names exactly the completions this client **rendered**, so a completion published between the render and the click is not in it and stays unread — there is deliberately no watermark sweep and no "all" flag on the wire, because that clears a result nobody saw.
- **`CompletionAttentionAckReceipt`** (`src/shared/desktop-session-contract.ts`): `read` states plus bare `superseded`/`unknown` session ids, with the reason each bucket exists written on the type.
- **The foreground gate** (`src/main/desktop-ipc.ts`): `guardForegroundReceipts` now names both receipt ops. The match there is by op **name**, so an op it does not list bypasses it — and a bulk write from a hidden, occluded or minimised window is a pile of marks cleared where the user could see none of the results behind them. `sessions.notified` stays ungated, and the test asserts that too, because a delivery claim fires in the background by definition.
- **`completion_ack_bulk`** (`desktop-hooks.ts`): its own `DesktopFeature` key rather than a bump of `completion_ack`, so the per-session receipt keeps working against a backend that lacks the batch route.
[...]

PR #311, opened 2026-09-17
feat(chat): clickable detected targets, a link hover toolbar, and Quote

damianvtran:
Closes the gap the operator hit: an assistant turn that says
`File: ~/workspace/opoint-renewal-2026-09-17/opoint_adverse_media_query_failures_2026-09-17.xlsx (37 KB, 8 sheets)`
rendered that path as dead text. Now every detected target in rendered markdown - a
markdown link, a bare URL, a `file://` URL, a bare absolute or `~` path - is painted like a
link (accent, 1px underline, offset 2px), a click on a path opens it in the OS default app
(a URL opens in the browser), and in the canonical transcript a pointer or a highlight raises
a toolbar with Copy / Open / Open folder (files only) and Quote when the highlight is inside
the link.

`Release: patch — a chat rendering and affordance change: detected paths become clickable
links with a hover toolbar and Quote, `~/…` now opens in `open-file` and the reveal, and both
report their outcome instead of failing silently.`

## What the user gets

| where | before | after |
|---|---|---|
| a path the agent wrote | dead text | accent-underlined link; click opens it |
| a `file://` URL in prose | dead text (react-markdown strips the href) | link that opens the file |
| a bare URL | link (remark-gfm already did this) | unchanged bytes, now with the toolbar |
| a backticked path | literal monospace | monospace inside the link, one link register |
| hovering a link (transcript) | nothing | Copy · Open · Open folder (file) / Copy · Open (dir) / Copy + reason (missing) |
| highlighting a link, part or whole | the turn's Quote control | that link's toolbar, **Quote first** - and the turn's control stands down, so a highlight is never quoted twice or misattributed |
| `~/…` from the app | `openPath` on a literal `~/…` (failed) | resolved by the main process, with an outcome the renderer can report |

## Decisions worth a reviewer's attention

- **Detection is a hoisted, optionless remark plugin (`remarkLinkifyTargets`) over mdast**, visiting
  `text` nodes plus whole-span `inlineCode` only. Fences, markdown links and GFM autolinks are
  structurally unreachable, and the plugin array is a module-scope constant because
  `MARKDOWN_COMPONENTS`'s own comment records that a per-render literal breaks react-markdown's
  memo. A **streaming** record is not linkified (`linkify={false}`): a half-arrived path is a
  prefix the next token falsifies, the same reason `isQuotable` refuses one.
- **One grammar, two surfaces** (`utils/link-grammar.ts`): `mentioned-files.ts` imports it and its
[...]

PR #303, opened 2026-09-17
feat(composer): paint the slash command tokens the way the TUI does

damianvtran:
## What this is

The composer's slash **syntax highlight**, cut fresh from `main` (`516a093d5`): the command word is tinted, and for a command that prefixes an instruction set the roster **NAME** is tinted in a second ink, separately from the instruction text, which stays prose.

**This is the highlighting half only.** The planner rewrite that reads the wire's own vocabulary (`prefixes_text`, `argument_shape`, `argument_words`) and the leading-line repair are **not** in this PR: that work is `fix/slash-wire-vocabulary` (PR #288) and the abandoned `fix/slash-prose-and-highlight` (PR #246). Nothing here changes what Enter does, adds no vocabulary field, and re-implements no part of the rule — the one place the two meet is the gate below, which *asks main's own planner* what Enter will do rather than deciding it again. `fix/slash-wire-vocabulary` is untouched: not pushed, not rebased.

Head: **`c40fe79c3`**, folded onto `origin/main` = `6f15f71ea`, **0 behind and MERGEABLE**, CI green on this head. The branch has been folded three times since it was cut (`09acb156a` → `33373f748` → `3a5b66c54`), and the folds are recorded rather than assumed: the latest one conflicted in `docs/evidence/manifest.json` alone (`git merge-tree`), main's own 15-commit range touches no `src/renderer/src/features/chat/components/` file, and this branch's composer files, `scripts/` and `package.json` are **byte-identical across it** — the manifest's 21+/35- is its own re-derived fields. Earlier folds are proved in the sections below and in the manifest's notes; where an earlier version of this body stated a range-diff pairing or a line count that does not reproduce, the manifest note now carries the corrected figure instead (round-4 review MINORs 2 and 3). No version bump: `package.json` is main's.

**The `#297` re-port, and the caret re-expressed through main.** Main's composer-file mentions work is taken verbatim wherever the two disagreed (the hunk list is on the thread); this branch's contribution is a paint layer beside it. The round-1 caret seed is back **through main's own machinery** — one write of the `pendingCaret` + `setCaret` pair every other caret move in that file uses — and it is **re-armed per conversation**, because the composer is rendered without a `key` and therefore survives a switch.

[...]

PR #338, opened 2026-09-18
chore(release): bump version to 0.26.12

damianvtran:
# chore(release): bump version to 0.26.12

## What this is

The version bump for the window since **`v0.26.11`** — published 02:07Z, promoted, 20 assets, and the first release to reach users since 12:15Z. `package.json` only, one line, on `main`'s own tree at `2de7ae977`: `0.26.11` → `0.26.12`.

## The bump, by materiality

**`patch`.** The window carries a composer feature and a harness repair. The mention picker is a new way to address a surface the app already has — the composer — rather than a new surface, and its author argued `patch` for exactly that reason, explicitly declining the minor on their parallel harness change. The precedent points the same way: `v0.26.5` shipped a genuinely new searchable-combobox capability as a patch. Under-calling is the error this repository prefers, because the next release corrects it, while an over-called minor permanently misreports how much changed.

## The window (`git log --first-parent --oneline v0.26.11..origin/main`)

- **#297** `feat(chat): mention files in the composer with a picker and inline chips` — `6f15f71ea`, 02:30:56Z. Its body's own line: `Release: patch — the composer takes file mentions: type @ to pick a file or folder from the working directory, its subdirectories, and the session's recent files, and the mention sits inline in the text as a chip you can move the caret through and delete atomically; completed references resolve to the file's path and content when the turn is sent.`
  - **Two caveats its author asked to carry into the notes rather than leave in the thread:** the UX sign-off rests on **coverage by analogy**, not byte-identity — after the final fold every mention-owned source file is byte-identical to the pre-fold head and no mention frame moved, but the fold's one behavioural change (a refusal guard on `@`-pick, matching a rule `main` introduced) lives in `message-input.tsx`, the one mention-surface file that is not byte-identical, and QA measured it on both sibling popups rather than inferring it; and **`pnpm check-evidence` has never produced a verdict on that branch** (another session's sweep holds the machine-wide lease) — recorded as unverified, not as a pass.
- **#339** `fix(evidence): reap only a profile name, not anything under the prefix` — `2de7ae977`, ~03:37Z. Its body's own line: `Release: patch — the evidence harness no longer deletes a live temp tree belonging to a concurrently running test; no shipped-code change.` It fixes a defect live on `main`: the stale-profile sweep matched any `lo-evidence-*` name in `tmpdir()`, including the scratch root `evidence-manifest.test.mjs` creates, so `Number("manifest-…")` → `NaN` → `process.kill(NaN, 0)` threw, the bare `catch` read it as abandoned, and it `rmSync`'d a live tree — which presented as a flaky `evidence-manifest.test.mjs` and cost several lanes re-runs tonight. QA measured it at 12 of 12 failing runs before, 0 of 12 after.

## Precondition, checked rather than assumed

The **version line** is byte-identical at `v0.26.11` and on `main` (`0.26.11` both), so no merged PR in this window spent a number, and `0.26.12` is not an existing tag or Release.

## What rides the next window

**#302** (the credential half of the stale #246 — the fix that stops a credential typed into a message reaching the model) and **#303** (its highlighting half) were both still open at this cut, as were **#311**, **#313**, **#321**, **#285** and **#296**. They ride the next window; I hold no claim on it beyond this one, and the stale #246 is otherwise closed as superseded.

Release: patch — the window since v0.26.11: the composer takes `@` file mentions as inline chips, and the evidence harness stops deleting a live temp tree belonging to a concurrently running test.


damianvtran:
### Superseded — closing without merging, and it must not be merged

**Closed by:** this session (window owner, pid 35679).

**Why:** the window this claim was cut for **has already been released as `v0.27.0`** — PR #340 (`chore(release): bump version to 0.27.0`, merged `f19827852`, Release published 04:03:54Z and promoted, publish run `35305527241` green, 20 assets). #340's window is exactly this one plus #277: #297 `6f15f71ea`, #339 `2de7ae977`, #277 `d6ebde29c` — all three are ancestors of `main`. So the work this bump would have carried has already reached users and **re-cutting it as `0.26.12` would be wrong twice over**.

**And it is actively harmful if merged:** this branch's single commit was derived on `2de7ae977`, before #277 and #340 landed. `git diff origin/main fc47d9de8 -- package.json` reads

```
-  "version": "0.27.0",
+  "version": "0.26.12",
```

i.e. merging it would ship the **previous release's code under a number nobody chose** — the mirror-image failure `AGENTS.md` documents, where the tag, the Release and the installers all agree with each other about the wrong tree. Do not merge this branch; the `release-next` branch can be deleted with it.

**For the record, because the process failure matters more than this PR:** I opened this claim at 02:32Z and it was visible as an open draft under `--state open`, but the **two-owner race it exists to prevent still happened** — #340 was opened at 04:01:38Z and merged at ~04:02, and nothing on that PR mentions this lock. The likely mechanism is mechanical rather than careless: `gh pr list --search "chore(release) in:title"` returns **nothing** while a draft claim PR is open (measured earlier tonight against the previous claim, #324), so a window owner using the search-based check cannot see the lock at all. If that is the shape, the fix is in the discovery step, not in anyone's diligence — a claim PR that the documented check cannot find is not a lock.

**No hard feelings and nothing to redo:** `v0.27.0` is a real, promoted release and users have the file-mention picker and the evidence-harness repair. I am not reopening or re-cutting anything, and I hold no claim on the window that follows it.

PR #297, opened 2026-09-17
feat(chat): mention files in the composer with a picker and inline chips

damianvtran:
## What and why

The composer grows an `@` file reference. Typing `@` opens a list over the field —
the working directory ranked fuzzily, drilling in on `/`, directories as rows —
and accepting one writes a literal `@src/app.py` token. Every token that
**resolves** is drawn as a chip behind the sentence's own glyphs, wherever it sits
in the sentence, recomputed from the field's value on every keystroke.

The direction is `docs/design/composer-file-mentions.md` (added here, with its
assets). Its four load-bearing decisions are what this implements: the `<textarea>`
stays and the fills are drawn from a hidden metric-identical mirror; the chip is
`sunken` on `surface` (`warning-wash` outside the workspace); the picker reuses the
slash popup's shell, anchor, rows and ARIA with a **measured** 3–8 row budget and
none of what it must not inherit (no phase strip, no ambiguity gate, no
dialog-style host); and the chip ⇔ the token resolves.

What the composer owes the harness is exactly the **text** it writes — the harness
expands, this feature only has to produce what it expands — so the grammar is a
port, not a second rule: `@` at a whitespace boundary, terminated by whitespace
only, spanning `/` and `.`, `@"my file.txt"` for a name with a space (what the
TUI's own picker emits). A hand-typed path that names nothing is **not** a chip,
which is the one place this surface can show what the harness does silently.

### The affordance is GATED, and the gate is why this ships dark

**No released harness expands a mention at all.** The expansion is
`local_operator/references.py`, a harness behaviour with no server surface; it is
in no tag through `v0.56.8`, and the half that adds it is
`damianvtran/local-operator#1220` — in review, and **publishing no capability key
of its own**. So a composer that offered the picker and painted the chips on
today's install would be promising an expansion nothing on the machine performs:
the user picks a file, gets a chip that says "this is a reference", and the model
receives the literal characters.

So the composer reads a capability — `references`, via `/v1/capabilities`
(`desktop-hooks.ts`, following the app's existing `desktopFeatureEnabled`
pattern, which fails closed) — and withholds the whole affordance when it is
absent: no picker, no chips, no `@` tip, the text left plain, which is exactly
what the harness does with it — **and one line saying why**
(`This backend cannot carry file references.`, in the picker's own notice slot and
its own voice). Silence was the state every install shipped in, and a user who
types `@` and sees nothing has no path from there to "this backend is older": the
sibling cwd gate in the same band names its reason, so this one does too. It is a
[...]

PR #277, opened 2026-09-16
feat(browser): conversation browser affordances and session-scoped tabs

damianvtran:
## Summary

**Head `b260eb9c1`** — the ROUND-5 REMEDIATION, on the FIFTEENTH fold onto `main`'s tip `2de7ae977` (#339's stale-profile sweep fix over #297's composer file mentions and #332's 0.26.11 release window). This fold is a PURE FOLD: it answers no review finding, re-takes no frame, and changes nothing but the ONE bookkeeping file it resolved, so the round-6 design and agent sign-offs and round 7's clean review of the previous fold stand on the code below, which this fold does not touch. Design round 5's `D22` (major) and `D23`, and review round 5's `A-7` and `A-8` are all `fixed`, none deferred, in one push - so the review streams re-verify this head exactly once. The current-row role is now DECLARED ONCE and imported by the settings rail and by the mark's specimen, and the drift guard resolves the specimen's TWO ELEMENTS through the shipped `cn` instead of pinning a copy of the string: both shapes the class has taken (D17's retired ground on the wrapper, D19's half-role on the row button) now turn it RED, 13 pass / 2 fail and 14 pass / 1 fail in a scratch mirror, where the pass-4 guard reported 14/0 on each. `A-9` (the delta's commit identity) is recorded rather than rewritten, as review round 5 asked. NO frame was re-taken: the 159 frames whose subject is the current row were re-captured as an A/B against the committed bytes, 101 byte-identical and every other frame pixel-identical at a 5% fuzz.

[...]

PR #339, opened 2026-09-18
fix(evidence): sweep only names that are a Chrome profile

damianvtran:
## What this fixes

`scripts/capture-evidence.mjs`'s `sweepStaleProfiles()` exists to collect Chrome profiles that a capture run never cleaned up: `teardown` covers a clean exit and the signals it traps, but not a `SIGKILL`, and the capture is routinely run under `timeout`, which sends exactly that. Each abandoned profile is ~180 MB in the shared system temp directory and nothing else ever collects them.

It found its candidates by prefix:

```js
if (!name.startsWith("lo-evidence-") || name === mine) continue;
const pid = Number(name.slice("lo-evidence-".length));
try {
	process.kill(pid, 0);
	continue; // still running, so the profile is in use
} catch {
	// no such process, so the profile is abandoned
}
rmSync(join(tmpdir(), name), { recursive: true, force: true });
```

That prefix is not specific enough to be a profile. `scripts/evidence-manifest.test.mjs` builds its synthetic evidence tree in the same shared temp directory as `mkdtempSync(join(tmpdir(), "lo-evidence-manifest-"))` — named deliberately, so that a leaked one is identifiable and can be reclaimed — and that starts with exactly `lo-evidence-`. The step from "shares our prefix" to "delete it" has no gate in it:

```
$ node -e 'const name = "lo-evidence-manifest-uexKDI";
           const pid = Number(name.slice("lo-evidence-".length));
           console.log("Number(name.slice(prefix.length)) =", pid);
           try { process.kill(pid, 0); } catch (e) { console.log("kill(pid, 0) throws:", e.constructor.name, "-", e.code); }'
Number(name.slice(prefix.length)) = NaN
kill(pid, 0) throws: TypeError - ERR_INVALID_ARG_TYPE
```

`NaN` is not a pid, so the `kill` throws — and the bare `catch` reads a throw that was never about a process as *"no such process, so the profile is abandoned"*, and the next statement deletes **a live tree belonging to a test that is walking it**.

## Who it hits, and why it presents as flakiness

`evidence-manifest.test.mjs` shares one module-scope scratch root across its `countsMean` cells (`COUNTS_TREE`, line 817, built at import time and reclaimed in `after`). A capture sweeping concurrently takes that one root, and every cell that walks it fails together with an `ENOENT` on `lo-evidence-manifest-…`. That reads as a flaky test file, not as a deletion, so it gets re-run rather than investigated — which is what has been costing this machine re-runs all night, and what a five-cell simultaneous failure in a file nobody touched looks like from outside.

Any lane that runs a capture, or that runs `scripts/evidence-manifest.test.mjs`, while another lane is capturing, is exposed. Nothing about either file says which of the two is wrong.

## Reproduced live on this machine, while preparing this PR

The bug is not historical. Six consecutive scoped runs of `node scripts/run-desktop-tests.mjs scripts/capture-evidence.test.mjs scripts/evidence-manifest.test.mjs` on this head gave `41/41` four times and `36/41` twice, and the five cells that fail together are exactly the ones sharing that module-scope root — *a countsMean paragraph leading with the walk's numbers passes*, *a paragraph left behind by a fold fails, and carries the sentence to paste*, *a paragraph with no reading in it fails rather than passing by omission*, *the older paragraphs under the leading one are not this tree's to answer for*, *a stale surfaces paragraph fails on the literal the field names* — while `docs/evidence` and this branch are untouched by any of them.

What was doing the deleting is a lane still running the UNFIXED file: at that moment two other lanes' `capture-evidence.mjs` processes were live, and both of their worktrees still carry the loose rule (verified by grepping the running copies' own trees for `startsWith("lo-evidence-")`). Their sweep is what takes the sibling test's root; nothing on this branch can, since the sweep here reaps only names that decode to a profile.

**For review and QA:** a scoped run of `evidence-manifest.test.mjs` can still lose those five cells while an unfixed lane is capturing — same signature, same root, not this branch. Please read a failure of exactly those five as the neighbouring lane's sweep (or run the pair when `ps` shows no other `capture-evidence.mjs` live). The two tests this PR adds never failed in any of the six runs.

## How it was found

It was not found here. A neighbouring lane's convergence QA traced it — it hit the same symptom from its own branch, followed it into the sweep's name test and worked out the `NaN` → `TypeError` → `catch` → `rmSync` chain — and deliberately left the fix to this file's owner rather than widening its own branch's scope, reporting it instead. Credit to that QA pass; this PR is the fix landing in the file that owns the bug.

## The fix

The candidate is now the profile's own shape, and nothing else can reach the `rmSync`:

- `profileOwnerPid(name)` decodes a name with `^lo-evidence-(\d+)$` — the literal that the profile's own creation site (`join(tmpdir(), `lo-evidence-${process.pid}`)`, in `main`) writes — and returns `null` for anything that is not that shape. A name that does not decode is not this sweep's to delete.
- The decoded value is validated as an integer in `process.kill`'s accepted range (`Number.isInteger(pid) && pid > 0 && pid <= 2_147_483_647`) *before* the `kill`, so the class of mistake that produced this — a numeric suffix that is not a pid reaching `kill` — cannot reach it at all, and a suffix long enough to overflow a double (still all digits) is not a profile either.

The profile **creation site's naming is unchanged**: the point of the fix is that the sweep stops matching things it did not create, not that the profile is renamed.

Comment-wise, the sweep's doc block keeps its existing account of why the sweep exists (SIGKILL under `timeout`, ~180 MB each, nothing collects them) and gains the trap it just fell into, in the same voice the file already uses.

## Test evidence

`scripts/capture-evidence.test.mjs` gains two tests: the name rule directly, and an end-to-end drive of the real sweep in a child process whose `TMPDIR` is a synthetic root holding an abandoned profile (`lo-evidence-<reaped child pid>`), a `lo-evidence-manifest-uexKDI` tree with a frame inside it, and a third name that is not a profile. The child's environment is built rather than inherited (`PATH`/`HOME`/`TMPDIR` only), so no `CMUX_*`/`LOP_*` from whatever session runs the suite is handed on.

Three runs, in order:

**1. New tests against the unfixed source** (`git checkout origin/main -- scripts/capture-evidence.mjs`, tests kept):

```
SyntaxError: The requested module './capture-evidence.mjs' does not provide an export named 'profileOwnerPid'
✖ scripts/capture-evidence.test.mjs (112.430542ms)
ℹ pass 0   ℹ fail 1
```

**2. With the fix in place:**

```
✔ the profile-name rule answers for the profile shape and nothing else (0.121ms)
✔ the sweep takes the abandoned profile and leaves the rest of the temp directory (116.538625ms)
ℹ tests 7   ℹ pass 7   ℹ fail 0
```

**3. The discriminating run** — the pre-fix candidate rule restored (loose prefix, `Number(...)` into a bare `catch`) with the exports kept, so the test is judged on the sweep's behaviour and not on a missing symbol:

```
✔ the profile-name rule answers for the profile shape and nothing else
✖ the sweep takes the abandoned profile and leaves the rest of the temp directory (183.392583ms)
  AssertionError [ERR_ASSERTION]: a directory that is not a profile is not the sweep's to delete
ℹ pass 6   ℹ fail 1
```

That last one is the regression test doing its job: on the pre-fix rule it deletes the sibling's live tree, exactly as the report says, and the assertion names it.

## Gates

| Gate | Command | Result |
| --- | --- | --- |
| Scripts lint ratchet | `pnpm lint:scripts` | `2 changed file(s) under scripts/ are lint-clean against origin/main (6f15f71)` |
| Biome, touched files | `pnpm exec biome check scripts/capture-evidence.mjs scripts/capture-evidence.test.mjs` | clean (the one remaining `useTopLevelRegex` warning is pre-existing at `capture-evidence.mjs:5096` on this head, and at `4990` on the base — the same DevTools assertion-text matcher, untouched by this diff) |
| Typecheck | `pnpm check-types` | clean (`tsconfig.node.json`, `tsconfig.app.json`) |
| Desktop runner, scoped | `node scripts/run-desktop-tests.mjs scripts/capture-evidence.test.mjs scripts/evidence-manifest.test.mjs` | `tests 41 / pass 41 / fail 0` — in 4 of 6 consecutive runs, the other 2 being the neighbouring lanes' live sweep (below); includes *the SHIPPED manifest's stamps describe the tree it ships in* and *the SHIPPED manifest's head citations lie in the history it ships in* |
| Desktop runner, full suite | `pnpm test:desktop` | `tests 2561 / pass 2557 / fail 3 / skipped 1` — all three attributed below, none from this diff |
| CI | this PR's checks | the table above is the local rehearsal; CI is the authority |

**The three full-suite failures are not this diff's**, and two of them are the two this machine already knows about. The files are `backend-settings-tiers.test.mjs`, `python-bytecode-cache.test.mjs` and `update-robustness.test.mjs`, and none of them runs or receives anything this change touches:

- `the fixtures describe the registry the backend actually ships` and `every backend spawn carries the prefix even with the shell-env load unresolved` — reproduced on an unmodified `origin/main` worktree (`6f15f71ea`) with the same command: `tests 158 / pass 155 / fail 2`, the same two names. Pre-existing, not this branch's.
- `the real macOS tools reject an unsigned image, on macOS` (`update-robustness`) — the disk-image/`codesign` cells, failing on `hdiutil: couldn't eject "disk28" - Resource busy` under a machine that is mounting images in several lanes at once. It does not reproduce scoped, on either tree: on this head, `node scripts/run-desktop-tests.mjs scripts/update-robustness.test.mjs` is `tests 112 / pass 111 / fail 0`. A machine-load flake, named rather than claimed clean.

## The second commit is a re-stamp, not a recapture

This change moves `scripts/`, which is a tree `docs/evidence/manifest.json` pins, so the shipped-manifest stamp check fails until the stamp is re-derived — measured on this head by holding the pre-fix stamp against it, not assumed:

```
✖ the SHIPPED manifest's stamps describe the tree it ships in
  AssertionError [ERR_ASSERTION]: docs/evidence/manifest.json must describe HEAD's trees: re-derive srcTree/scriptsTree …
  + [ 'manifest.json: `scriptsTree` is 5744a1827 but HEAD:scripts is 74c32f42e - the frames were captured
  +   from different scripts than the tree under review, so re-capture and re-stamp' ]
  - []
```

So commit 2 re-derives **`scriptsTree` only** and records the re-derivation in `captureOrigin.reStamps`. No frame was re-taken, no theme sweep was run, and no frame is claimed to describe this tree: `frames`, `surfaces` and `themes` are unchanged because no frame moved and the `STORIES`/`THEMES` literals are untouched, and `srcTree` is already `HEAD:src` here because this branch does not touch `src/`.

**The branch was then rebased onto a moved `main`** (`afcd83772` → `6f15f71ea`, PR #297, which itself moved `scripts/` and this file), and the stamp re-derived against the rebased tree, so what the entry displaces is main's own `scriptsTree` (`5744a1827`) rather than this branch's earlier value. The rebase is content-preserving for the fix: `git range-diff` reports commit 1, the fix itself, as `=` (identical) between the old and rebased heads — only the re-stamp commit differs, which is what a re-derivation is. The force-push is that rebase; the branch's review has not started, so nothing reviewed moved.

## Pre-existing on `main`

This defect is on `main` and is not caused by any PR: the loose prefix is in the shipped `sweepStaleProfiles`, and the scratch root's name is in the shipped `evidence-manifest.test.mjs`. Nothing in this branch introduces either. The neighbouring lane's report reached this file through a convergence QA pass, not through a change to the sweep.


damianvtran:
### Agent review — round 1

Reviewer: independent review lane (`reviewer` role, `lopdev` team) — model `deepseek/deepseek-flash`, as reported by the session; no provider fallback occurred during this round.
Scope: `6f15f71ea`..`8bc0bf00b` — 3 files, +210/−7: `scripts/capture-evidence.mjs`, `scripts/capture-evidence.test.mjs`, `docs/evidence/manifest.json`.

**Verdict: clean.** No BLOCKER and no MAJOR. The round is TERMINAL: the two nits and the one minor below are follow-ups, not a reason for another round. Everything asserted in the body that I could execute, I executed; where I could not, I say so.

Worktree: `/Users/damian/workspace/review-339/wt` (detached at `8bc0bf00b`, primary checkout's `node_modules` shared by symlink), reclaimed at the end of the round. Nothing was pointed at the operator's live app, and no other lane's worktree was written to.

---

#### 1. The predicate — holds, and it is the right discriminator

- **The creation site is the only producer of the prefix.** `scripts/capture-evidence.mjs:3419` is `dataDir = join(tmpdir(), `lo-evidence-${process.pid}`)` and `grep -rn "lo-evidence"` over the tree finds only that line, the sweep and `profileOwnerPid`, and `scripts/evidence-manifest.test.mjs:52`. There is no second naming path and no way to relocate the profile dir.
- **No historical shape escapes.** `git log -S'lo-evidence-' -- scripts/capture-evidence.mjs` returns exactly two commits: the file's introduction (`bb57a4080`, 2026-08-06) and this fix. The literal has been `lo-evidence-<decimal pid>` for the file's whole life, so an older-format profile that the new rule would now ignore does not exist.
- **Nothing that should be reaped now escapes it.** A genuinely abandoned profile is always named by that literal, so it always matches, always decodes and is always reaped — pinned by the end-to-end cell itself (below). The only profiles that escape are ones whose pid has since been reused by a live process, which read as "in use" and wait; that is a pre-existing property of the signal-0 probe, unchanged by this diff and not a defect of it.
- **The integer range check is necessary, and `2_147_483_647` is the correct boundary — measured, not reasoned:**

  ```
  process.kill(NaN, 0)           -> TypeError ERR_INVALID_ARG_TYPE
  process.kill(1e20, 0)          -> TypeError ERR_INVALID_ARG_TYPE
  process.kill(2147483648, 0)    -> TypeError ERR_INVALID_ARG_TYPE
  process.kill(2_147_483_647, 0) -> Error ESRCH
  process.kill(99999, 0)         -> Error ESRCH
  process.kill(0, 0)             -> no throw (signals the process group)
  ```

  So a name like `lo-evidence-2147483648` — all digits, so `\d+` alone admits it — would reach the `kill`, throw a `TypeError`, and be read by the bare `catch` as "abandoned" and deleted. `MAX_PID` is exactly where node stops accepting a pid, which is what the comment claims it is. `Number.isInteger` is in practice subsumed by `pid > 0 && pid <= MAX_PID` (`NaN` and `Infinity` both fail the comparisons), so it is belt-and-braces rather than load-bearing — no change asked for.
- **`pid > 0` is doing more than it looks.** `kill(0, 0)` does **not** throw on this node (it signals the group), so a `lo-evidence-0` would have been judged "still in use" rather than deleted. The new rule answers `null` — "not ours" — which is the honest answer for a name this script cannot create. Behaviourally safe either way; the semantics are now right.
- **Fixing the reaper rather than renaming the sibling's root is the right call.** The unsafe party is the thing that deletes by prefix in a namespace it does not own; renaming `evidence-manifest.test.mjs`'s root would have left the same trap armed for any future `lo-evidence-*` directory on the machine.

#### 2. The tests — hold; the discrimination reproduces in both directions

- At head: `node scripts/run-desktop-tests.mjs scripts/capture-evidence.test.mjs` → `tests 7 / pass 7 / fail 0`.
- Unfixed source with the new tests kept (`git checkout origin/main -- scripts/capture-evidence.mjs`): `SyntaxError: The requested module './capture-evidence.mjs' does not provide an export named 'profileOwnerPid'`, `pass 0 / fail 1` — the body's run 1, verbatim. (The file-level failure is the honest shape of that run: it discriminates on the missing symbol, not on behaviour, which is why run 3 is the one that matters.)
- **The discriminating run, reproduced.** Pre-fix candidate rule restored (`if (!name.startsWith("lo-evidence-") || name === mine) continue; const pid = Number(name.slice("lo-evidence-".length));`) with the exports kept:

  ```
  ✔ the profile-name rule answers for the profile shape and nothing else
  ✖ the sweep takes the abandoned profile and leaves the rest of the temp directory
    AssertionError [ERR_ASSERTION]: a directory that is not a profile is not the sweep's to delete
        at TestContext.<anonymous> (scripts/capture-evidence.test.mjs:267:9)
  ℹ tests 7  ℹ pass 6  ℹ fail 1
  ```

  Same cell, same assertion, same message as the body reports. It is a regression test, not a test that passes both ways.
- **The other direction, which the body does not claim but the test's own comment does.** A mutant that reaps nothing (`continue` added after the decode) fails the same cell with `AssertionError [ERR_ASSERTION]: an abandoned profile is the sweep's job`. Both properties the comment says are pinned really are pinned, so "a sweep that silently stopped sweeping" cannot pass.
- **The new test is immune to the flake it fixes.** Its synthetic roots are `lo-sweep-profiles-*`/`lo-sweep-home-*`, which no lane's prefix rule matches, and the swept root is handed to the child as `TMPDIR`; the child's environment is `{PATH, HOME, TMPDIR}` only, so no `CMUX_*`/`LOP_*` is inherited. My runs passed while another lane's unfixed sweep was live on this machine (see §5), which is the condition the body warns about.

#### 3. The re-stamp — holds, and it is necessary, reproduced

- Stamp describes the tree it ships in: `git rev-parse 8bc0bf00b:scripts` = `74c32f42ed0c247d04f76e1a3ed81c67bac6eaa9` = manifest `scriptsTree`; `git rev-parse 8bc0bf00b:src` = `79d4a496…` = manifest `srcTree`, unchanged from base because `src/` is untouched.
- The displaced value is the base's own: `git show 6f15f71ea:docs/evidence/manifest.json` gives `scriptsTree` `5744a1827d7c52ee198640c3cc21e87ae6f1181a`, which is exactly `captureOrigin.reStamps[16].replaced.scriptsTree`.
- No frame moved and none is claimed to: the branch diff is 3 files with no `.webp`, and `frames 6009 / surfaces 638 / themes 12` are identical to base. `head`, `headNote` and `capturedAt` are untouched by the diff.
- **Necessary.** Holding the stamp at the pre-fix value and running the manifest suite reproduces the body's quoted failure:

  ```
  ✖ the SHIPPED manifest's stamps describe the tree it ships in
    AssertionError: docs/evidence/manifest.json must describe HEAD's trees: re-derive srcTree/scriptsTree from `git rev-parse HEAD:src` / `HEAD:scripts` …
    'manifest.json: `scriptsTree` is 5744a1827 but HEAD:scripts is 74c32f42e - the frames were captured from different scripts than the tree under review, so re-capture and re-stamp'
  ```

  At head the same suite is `tests 34 / pass 34 / fail 0`, that cell included.
- Entry shape matches the array's convention: appended as the 17th entry with keys `at`/`by`/`why`/`replaced`, identical to all 16 siblings.

#### 4. The doc comment — holds, apart from one count (F1)

- The original rationale survived in place and unchanged in substance: SIGKILL under `timeout`, ~180 MB per profile, the 542 MB/five-profiles incident, "own directories are skipped by pid".
- The added explanation of the trap is accurate where it matters, and I checked its chain by running it rather than reading it: `Number("manifest-uexKDI")` is `NaN`, `kill(NaN, 0)` throws, and the pre-fix mutant above really does delete the sibling's live tree. It is explanatory, not decorative.
- The `Number.isInteger`/`MAX_PID` paragraph is right about the class of mistake it names; the only slack is the hedge "`TypeError`/`RangeError`" — on node 26.5.0 all three bad shapes answer `TypeError ERR_INVALID_ARG_TYPE` (measured in §1). Not worth a change on its own.

#### Findings

**F1 — MINOR — the trap comment's count is off by one (`scripts/capture-evidence.mjs:3287`)**

> `- six `countsMean` cells of that file failing together with an `ENOENT` on the scratch root`

Six `countsMean` cells *share* `COUNTS_TREE` (`scripts/evidence-manifest.test.mjs:817`) but only **five** of them walk it and fail; the sixth, `a manifest with no countsMean is not this guard's failure` (`:901`), returns early and never touches the tree. Measured: `countsMeanFailures({ frames: 3, supplementary: [] }, fakeGit, "/nonexistent")` → `[]`, no throw, whereas the same call with a `countsMean` present throws `ENOENT … scandir` on the missing root. The body itself says "the five cells that fail together", so the comment and the body disagree. Suggested wording: "six `countsMean` cells share that root and the five that walk it fail together with an `ENOENT`".

**F2 — NIT — the only partial `replaced` in the array (`docs/evidence/manifest.json:4027-4029`)**

All 16 prior entries record the full displaced set (`head`, `srcTree`, `scriptsTree`, `capturedAt`; one also `frames`/`surfaces`/`refreshedFrames`); this entry records `scriptsTree` alone. No code reads `reStamps` or `replaced`, and the entry's own `why` explains why the other fields did not move — so either shape is defensible and nothing needs to change. Flagging only because it is the single exception in a 17-entry array, and a reader reconstructing a prior stamp from `replaced` will find one entry that is not the same shape as its neighbours.

**F3 — NIT — the gate table's biome line number is pre-rebase**

The body cites the pre-existing `useTopLevelRegex` warning at `capture-evidence.mjs:4974`. On this head `pnpm exec biome check scripts/capture-evidence.mjs scripts/capture-evidence.test.mjs` reports it at **5054**, and on `origin/main`'s copy of the same file at **4991** (head 5526 lines, base 5462) — the number is from before the rebase. The warning is genuinely pre-existing on the base, so only the line reference is stale. `pnpm lint:scripts` is clean against `origin/main (6f15f71)` as the table says.

#### 5. The PR body's claims

- *"Nothing on this branch can [delete the sibling's root]"* — structurally true: with the form-match, `lo-evidence-manifest-*` decodes to `null` and never reaches `rmSync`, and the end-to-end cell pins it.
- *"What was doing the deleting is a lane still running the UNFIXED file"* — **confirmed live, not taken on trust.** At review time this user had two other `capture-evidence.mjs` processes running; one (`pid 9431`) belongs to `/Users/damian/local-operator-ui-worktrees/pr285-rebase`, whose copy still carries `if (!name.startsWith("lo-evidence-") || name === mine) continue;` (`:3133`, `grep -c` for `profileOwnerPid` = 0), and whose `TMPDIR` is `/var/folders/qd/…/T/` — the same per-user temp directory `evidence-manifest.test.mjs:52` builds its scratch root in. Four other sampled worktrees (`qa317r5`, `rev317r5`, `fold317-final`, `qa312-final`) carry the loose rule too. So the reported signature is reproducible on this machine at this moment, and "six runs, four green, two not" is exactly the shape that implies.
- *"A scoped run can still lose those five cells while an unfixed lane captures"* — consistent with measurement: with a `countsMean` present and the root gone, `countsMeanFailures` throws `ENOENT` out of `scandir`, which is the failure those five cells produce. (My own scoped runs of both files were green: `7/7` and `34/34`.)
- *"It was not found here … a neighbouring lane's convergence QA traced it"* — this is the one claim I cannot verify from anything I can read: the mechanism, the trace and the decision to leave the fix to this file's owner are all consistent with the artefacts, but the identity of the reporter is a narrative claim and I am not treating it as verified. It does not bear on the diff.
- *"The three full-suite failures are not this diff's"* — the diff touches 3 files; none is any of the three failing files, and none of the three imports the changed module (`backend-settings-tiers.test.mjs` → `backend-settings-registry.mjs`; `python-bytecode-cache.test.mjs` → `python-child-env.mjs`; `update-robustness.test.mjs` → none of them). I did **not** re-run those two pre-existing failures on a base worktree, so I neither confirm nor dispute their pre-existence; that attribution is CI's to settle and the body already names it as such.

#### The deferred item: the remaining bare `catch` (EPERM and friends)

**Deferring it is right, but the framing needs one correction: the new form-match makes the class narrower, not unreachable.**

- What changed is *which names* can reach the `kill`; what did not change is *which throws* the bare `catch` reads as "abandoned". Any non-`ESRCH` throw still falls through to `rmSync`. For a name that is now well-formed, the throws available are `ESRCH` (correctly "gone, reap it") and `EPERM` (the decoded pid is alive but owned by another user) — plus argument-validation throws, which the form-match and range check now exclude.
- Is `EPERM` genuinely unreachable for our own names? On this platform, effectively yes, and for a structural reason rather than a lucky one: `os.tmpdir()` here is per-user (`/var/folders/…/T`), the sweep only ever iterates its own temp directory, and a name in it was written by a process of this user. Cross-user pid reuse — our process died, the kernel later handed that pid to another user's process, we decode it, `kill` answers `EPERM` — is reachable in principle, but in that case the profile is genuinely abandoned and deleting it is the intended outcome. The *wrong* deletion needs a live profile we do not own inside our own temp directory, which requires a shared temp directory: i.e. a Linux host with a shared `/tmp` where a second user runs this script concurrently. So: not reachable on macOS as configured; reachable in principle on Linux, which is the version of the argument worth recording if this is ever picked up.
- Recommendation for the follow-up (not this PR): distinguish the codes in the `catch` — reap on `e.code === "ESRCH"`, and warn or rethrow on anything else — which turns the current "a throw is a death certificate" reading into a check. The comment's own warning about the bare `catch` is what makes this a good deferred item rather than an unnoticed one.

#### Not a finding, but worth recording

`scripts/capture-evidence.test.mjs`'s two pre-existing cells call `build()` (`:63`, `:91`), which makes a `lo-sweep-*` root in the shared temp directory, and until this PR the file had no `after()` reclaiming anything — there are **2,110** stale `lo-sweep-*` directories (25 MB) in this user's temp dir as I write. That is pre-existing, out of this diff's scope, and untouched here; the new test adds the file's first reclaim (`cleaned`/`after`) and leaves the sweep's namespace, which is the pattern the older cells could adopt in a follow-up. Related hygiene disclosure: because the older path's roots carry no owner marker by name, I could not distinguish the handful mine created from concurrent lanes' live ones, so I deleted none of them rather than risk deleting a running lane's tree — which is the discipline this PR is about.

#### Hygiene

Review worktree `/Users/damian/workspace/review-339/wt` and scratch files removed after the round; the pre-fix mutant, the no-op mutant and the stale-stamp edits were all made in that worktree and reverted (`git diff HEAD --name-only` = empty before teardown). No `pnpm dev`, no app launch, no evidence sweep, and the `check-evidence` lease was never taken. Disk on `/` was 7.0 → 6.1 GiB free at worktree creation and 5.4 GiB at teardown (other lanes capturing concurrently).

damianvtran:
### QA report — round 1

**Verdict: PASS** — I reproduced the reported defect on the old code and confirmed it gone on this head, and I confirmed the other half of the claim: the sweep still reaps a genuinely abandoned profile. Round focused on the sweep's candidate rule and the one regression surface the change creates (the sweep that used to sweep too much must not stop sweeping at all).

**Where this ran.** Own worktree `~/workspace/qa339/tree` @ `8bc0bf00b17af5895c29a30e73fdecdef9a76a95` (the author's worktree left untouched), base worktree @ `6f15f71eaa966704e37152634f267850283bf60b` (`origin/main`). macOS 25.6.0 arm64, node v26.5.0, host load avg ~337 during the run. Isolated `TMPDIR` for every run (never the system one, and never a name under the swept prefix) and isolated config/HOME; every child launched under `env -i` so no inherited `CMUX_*`/`LOP_*` travelled with it.

**How the pre-fix code was exercised.** `sweepStaleProfiles` is not exported before this PR, so the pre-fix module is a byte-identical copy of `origin/main:scripts/capture-evidence.mjs` (blob `921dfff38`) with one `export` line appended — `diff` against `git show origin/main:…` is that line and nothing else. The same driver file was used for both revisions and prints the callee's own source, so the two arms differ only in module content: `sweepUsesPidRule: false` (pre-fix) → `true` (post-fix).

### Matrix

| # | Surface | Command | Actual output | Result |
|---|---|---|---|---|
| 1 | Reported failure: a live `lo-evidence-manifest-*` tree, both revisions | `env -i TMPDIR=$R QA_ROOT=$R QA_MODULE=<rev> node driver/qa-sweep-driver.mjs` | pre-fix `deleted: [lo-evidence-manifest-uexKDI, …]`; post-fix `existsAfter: true, nestedFileSurvived: true` | **PASS** |
| 2 | Genuinely abandoned profile (dead pid) still reaped | same driver, `lo-evidence-<reaped child pid>` | pre-fix `DELETED`, post-fix `DELETED` — directory *and* its nested `Default/Cookies` gone | **PASS** |
| 3 | Live profile survives | same driver, `lo-evidence-<live sleep pid>` | survived on both revisions | **PASS** |
| 4 | The sweep's own profile is skipped | same driver, dir named `lo-evidence-<child pid>` | `ownProfileSurvived: true` on both | **PASS** |
| 5 | Adversarial names (9 shapes) | same driver, see table below | 7 of 9 changed behaviour, all in the right direction | **PASS** |
| 6 | Scoped test files, repo's own runner | `node scripts/run-desktop-tests.mjs scripts/capture-evidence.test.mjs scripts/evidence-manifest.test.mjs` | `tests 41 / pass 41 / fail 0` | **PASS** |
| 7 | The reported cross-process signature | separate sweeping process on the test's own `TMPDIR`, 3 waves × 4 concurrent `evidence-manifest.test.mjs` | pre-fix **12/12 failing runs**, post-fix **0/12** | **PASS** |
| 8 | Regression neighbours (9 files referencing the changed module or its importers) | `node scripts/run-desktop-tests.mjs <9 files>` | `tests 161 / pass 160 / fail 1` — the one failure is pre-existing, see Q1 | **PASS** |
| 9 | Stamp check on this head | `git rev-parse HEAD:scripts HEAD:src` vs `docs/evidence/manifest.json`; `evidence-manifest.test.mjs` | `scriptsTree 74c32f42e… == HEAD:scripts`, `srcTree 79d4a496… == HEAD:src`, both shipped-manifest cells `✔` | **PASS** |
| 10 | CI on this exact head | `gh api repos/damianvtran/local-operator-ui/commits/8bc0bf00b…/check-runs` | 7 `success`, 2 `skipped` (Security Audit, Runtime Dependencies), **0 failures**, every run's `head_sha` = `8bc0bf00b` | **PASS** |

PR state on the same call: `OPEN`, `MERGEABLE`, `draft: false`, head still `8bc0bf00b`.

### Row 3 — adversarial cases, pre-fix vs post-fix

| Name | pre-fix | post-fix | Judgement |
|---|---|---|---|
| `lo-evidence-manifest-uexKDI` | DELETED | survived | the reported failure; fixed |
| `lo-evidence-not-a-pid`, `-manifest-1`, `-12abc`, `-abc` | DELETED | survived | not profile names; fixed |
| `lo-evidence-999999999999`, `-99999999999999999999` | DELETED | survived | all digits but outside any real pid; `kill` threw and the throw read as "abandoned"; fixed |
| `lo-evidence-` | survived | survived | `Number("")`=0, `kill(0,0)` succeeds (`pid 0` = own process group) |
| `lo-evidence-0` | survived | survived | guard refuses pid 0; pre-fix survived by accident |
| `lo-evidence`, `lo-sweep-4242` | survived | survived | never matched the prefix |
| `lo-evidence-1` | **DELETED** | **DELETED** | live-but-not-ours: `kill(1,0)` throws **EPERM**, and the bare `catch` reads any throw as "abandoned". Pre-existing, unchanged — Q2 |
| `lo-evidence-2147483647` (== `MAX_PID`) | DELETED | DELETED | inside the accepted range and `kill` answers a real `ESRCH` for it, so reaping is right |
| `lo-evidence-2147483648` (`MAX_PID`+1) | DELETED | survived | refused before the `kill`; the bound is correctly placed — Q3 |

`process.kill` on this host, measured: `NaN` → `TypeError ERR_INVALID_ARG_TYPE`; `1e20` → `TypeError`; `999999999999` → `TypeError`; `1` → `Error EPERM`; `2147483647` → `Error ESRCH`; `0` and `-1` → OK. So the comment's "a pid that is not an integer in range answers with a `TypeError`/`RangeError` instead of a signal" holds here, and every one of those throws was reaching the bare `catch`.

### Row 7 — the signature, arranged honestly and reproduced

A separate process called the real `sweepStaleProfiles` in a loop against the very `TMPDIR` the repo's own `evidence-manifest.test.mjs` was using for its scratch trees. Pre-fix arm (310,153 sweeps): **12 of 12 runs failed**, 2–16 `ENOENT`s each, with the `countsMean` cells failing together —

```
✖ a countsMean paragraph leading with the walk's numbers passes
✖ a paragraph left behind by a fold fails, and carries the sentence to paste
✖ a paragraph with no reading in it fails rather than passing by omission
✖ the older paragraphs under the leading one are not this tree's to answer for
✖ a stale surfaces paragraph fails on the literal the field names
Error: ENOENT: no such file or directory, scandir '…/lo-evidence-manifest-VRl0Z1'
```

— and its own log records the cause: **30 `lo-evidence-manifest-*` trees deleted, every one accepted by its prefix rule**. It also threw `ENOTEMPTY` five times out of its own `rmSync` (on one unguarded run it killed the sweeping process outright), i.e. a pre-fix sweep racing a writer can crash the capturing process, not just the test.

Post-fix arm, identical harness (299,521 sweeps): **0 of 12 runs failed**, 0 names its rule accepts deleted, 0 throws. 52 manifest trees did vanish during that arm, but every one is a name the new rule *rejects*, i.e. not the sweep's doing — the test file's own `after()` cleanup, which the before/after diff cannot tell apart. That is why the harness classifies each vanished name by the sweeping revision's own rule rather than attributing it.

### Q-findings

- **Q1 (pre-existing, not this PR — no action needed here).** `scripts/python-bytecode-cache.test.mjs` → *"every backend spawn carries the prefix even with the shell-env load unresolved"* fails locally: `AssertionError: global install: start() should report a started backend`. It fails **identically on the base** `6f15f71e` (34/35 pass on both), the file is untouched by this PR, the exercised `src/` is byte-identical to base (`git diff origin/main 8bc0bf00b -- src/` is empty), and it fails with a real `HOME` as well as an isolated one. Not a regression; unexplained locally, and I could not use CI as the counter-check because the Desktop Tests job is ubuntu-latest/node-22 while this run is macOS/node-26.
- **Q2 (informational, pre-existing).** The bare `catch` around `process.kill` still conflates `ESRCH` with `EPERM` (and with any other throw), so a `lo-evidence-<pid>` name whose pid is alive but not ours — `lo-evidence-1` above — is reaped. Unchanged by this PR and unreachable in practice, because the only writer of that name is the script itself, using its own pid. Recorded, not a deferred work item.
- **Q3 (informational).** `MAX_PID` itself is accepted and `kill` answers a genuine `ESRCH` for it, so `lo-evidence-2147483647` is reaped; `2147483648` is refused. The bound is on the correct side of the throw.

### What this run does not prove

- **I did not reproduce the cross-lane signature as it actually occurred** — a real capture's sweep against the operator's live shared temp directory with several lanes at once. I reproduced the mechanism in one isolated `TMPDIR` with a stand-in sweeping process, single-process-per-arm. I deliberately never pointed a pre-fix sweep at the shared temp directory.
- I did not run the full `pnpm test:desktop` matrix (120 files); I ran the two scoped files plus the 9 that reference the changed module or its importers.
- No Chrome-backed capture was exercised (no browser engine launched, per machine policy). The sweep was driven directly, which is what the defect needs, but `teardown`, a real ~180MB profile, and the `--only=` capture path were not exercised.
- macOS/darwin only. The platform-sensitive part is `process.kill`'s error codes, measured here and not on Linux or Windows.
- The stamp check was verified at the assertion and hash level; I did not re-run `check-evidence.mjs` against all 6009 frames.

### Hygiene, and one self-reported harness fault

My first two harness revisions did **not** pass `TMPDIR` into the sweeping child, so node fell back to `/tmp` and the pre-fix sweep read the *system* temp dir rather than the operator's `TMPDIR` (`/var/folders/…`). I checked for damage: `/tmp` holds no `lo-evidence-*` entry, and the instrumented run over that path reported `seenCandidates: []` with zero deletions, so nothing was removed. From the third revision the child asserts `tmpdir() === QA_ROOT` before sweeping. Related, for whoever owns them: the shared `TMPDIR` currently holds three leaked `lo-evidence-manifest-*` trees plus live captures — the exact names this defect is about.

No repository file was edited; a `node_modules` symlink to the primary checkout's tree was added inside my own worktree for the neighbour run only (untracked, and removed with the worktree). I did not touch the operator's live app, any other lane's worktree, or the author's worktree, and I created no issues and tagged no one.

damianvtran:

Round answered: `### Agent review — round 1`, scope `6f15f71ea`..`8bc0bf00b` (verdict clean/TERMINAL, F1 minor, F2/F3 nits, plus two corrections worth more than the nits). Head now **`d85b9ea11`**; nothing deferred, every item is in this one commit.

**F1 — fixed — `d85b9ea11`.** Both places that carried the count now say what you measured: *"that file's six `countsMean` cells share one module-scope scratch root, and the five of them that walk it fail together with an `ENOENT` on it while the sixth returns early without touching the tree (review round 1, F1)"*. There were two occurrences, not one — the sweep's doc block (`scripts/capture-evidence.mjs:3287`) and the test file's own note (`scripts/capture-evidence.test.mjs:167`) — so both are corrected; your suggested wording is used nearly verbatim, and the body's "five" is now the file's too.

**F2 — fixed — `d85b9ea11`.** Chosen shape: **drop the partiality by naming it, not complete it.** `replaced` still lists `scriptsTree` alone, and the entry's `why` now states why:

> `replaced` lists `scriptsTree` alone, the only field this entry displaced, and that single key is deliberate rather than an omission … padding this one with `head`, `srcTree` and `capturedAt` — none of which this branch touches, and all three of which main re-derived for itself at the base — would report displacements that did not happen. A reader reconstructing the pre-entry block should take `scriptsTree` from here and the other three from the values the file already carries.

Padding the entry to four keys was the other defensible shape and I rejected it on the record's own terms: it would make the file claim three displacements nobody made, which is the same class of error as a count that over-states what a walk measured. Your note that either shape is defensible is why the choice is named here.

**F3 — fixed (PR body, no commit needed).** The gate table's line reference is corrected to the numbers measured, not adjusted by arithmetic: `capture-evidence.mjs:5096` on this head; `4990` on the base, measured by writing `git show 6f15f71ea:scripts/capture-evidence.mjs` into the worktree as a temp probe, running biome over it, and removing it (`git status --porcelain` empty afterwards). Same `useTopLevelRegex` literal (the DevTools assertion-text matcher), genuinely pre-existing on the base, and the only finding in this round that needs no code movement.

**Correction 1 — `pid > 0` is load-bearing — fixed — `d85b9ea11`.** Folded into `profileOwnerPid`'s comment, with the semantics stated at the size you measured and re-measured here rather than quoted:

```
process.kill(0, 0)             -> no throw, and it tests no process
process.kill(1, 0)             -> Error EPERM
process.kill(99999, 0)         -> Error ESRCH
process.kill(2_147_483_647, 0) -> Error ESRCH
process.kill(NaN, 0)           -> TypeError ERR_INVALID_ARG_TYPE
process.kill(2147483648, 0)    -> TypeError ERR_INVALID_ARG_TYPE
```

The comment now says the guard buys the honest answer ("not ours") instead of a false one ("in use"), and that a `lo-evidence-0` is kept either way — it does not claim a behaviour change it cannot show. The same paragraph drops the old "`TypeError`/`RangeError`" hedge and quotes the three measured shapes instead, since all three bad ones are `ERR_INVALID_ARG_TYPE` on node 26.5.0.

**Correction 2 — the `catch` narrows rather than closes — fixed — `d85b9ea11`.** The sweep's doc block now says plainly that the `catch` still reads *any* throw as abandoned; that for a well-formed name only `ESRCH` and `EPERM` remain and `EPERM` still converges on the deletion; why that is out of reach *here* for a structural reason (per-user `os.tmpdir()`, names written by a process of this user) and reachable in principle on a shared `/tmp` with a second user; and it records the follow-up in your words as the shape to take it up: **reap only on `e.code === "ESRCH"`, warn or rethrow on anything else**. Deliberately not bundled: it changes what a capture does with a live-but-unreadable profile, which is a different decision from this fix and wants its own review.

**Left alone, deliberately:** the `Number.isInteger` belt-and-braces (no change asked, and it stays); and the "not a finding" note about the 2,110 stale `lo-sweep-*` roots and the older cells' missing `after()` — pre-existing, outside this diff, and a follow-up by your own framing, not this branch's to take.

**The delta, for a scoped convergence round:** `git diff 8bc0bf00b..d85b9ea11` is 3 files. Every changed line under `scripts/` is a comment line — verified by filtering the diff for any changed line that is not `*`, `//` or `/*`-prefixed, which returns nothing — so the sweep's executable code and the tests are byte-identical to the head you reviewed, and the only non-comment change is the manifest's `scriptsTree` and the `why` text above. `scripts/` moved again with the comment corrections, so the stamp was re-derived **last** (`227aca40c`, this head's `HEAD:scripts`), amended into the same commit, and the re-stamps entry says so.

**Gates on `d85b9ea11`, re-run after the re-derivation:** `pnpm lint:scripts` → lint-clean against `origin/main (6f15f71)`; `biome check` on both touched files → 0 errors, 1 pre-existing warning; `pnpm check-types` → clean; `node scripts/run-desktop-tests.mjs scripts/capture-evidence.test.mjs scripts/evidence-manifest.test.mjs` → `tests 41 / pass 41 / fail 0`; before the re-derivation the same suite was `pass 33 / fail 1` (`scriptsTree is 74c32f42e but HEAD:scripts is 227aca40c`), which is the stamp check doing its job on this commit too.

**One more of F1's own class, found while checking the body against the file (body edit, no commit).** The body's opening description of the signature still read "what a six-cell simultaneous failure … looks like from outside" — the same over-count, in the sentence a reader meets first. It now reads **five-cell**, so the body, the sweep's comment and the test file's note all say the measured number and nothing else.

damianvtran:
### QA remediation — round 1

Round answered: `### QA report — round 1` (verdict PASS on `8bc0bf00b`; Q1 a pre-existing failure, Q2 and Q3 informational). Head now **`d85b9ea11`**.

**What moved since the head you tested.** Three files, and under `scripts/` every changed line is a comment line — verified by filtering `git diff 8bc0bf00b..d85b9ea11` for any changed line that is not `*`, `//` or `/*`-prefixed, which returns nothing. So the sweep, the predicate and both tests are byte-identical to the code you exercised; the non-comment change is `docs/evidence/manifest.json` (the re-derived `scriptsTree`, because the comment corrections move `scripts/`, plus the re-stamps entry's own text). Your functional matrix therefore stands on this head, and rows 1, 2 and 3 (live tree survives / dead-pid profile still reaped / live-but-not-ours) are exactly what I re-ran: `node scripts/run-desktop-tests.mjs scripts/capture-evidence.test.mjs scripts/evidence-manifest.test.mjs` → `tests 41 / pass 41 / fail 0`.

**Q1 — pre-existing, not this PR; no action here.** Your evidence is the operative one and I am citing it rather than re-litigating: `scripts/python-bytecode-cache.test.mjs` → *"every backend spawn carries the prefix even with the shell-env load unresolved"* fails with `AssertionError: global install: start() should report a started backend`, **identically on the base `6f15f71e`** (34/35 pass on both arms), the file is untouched by this PR, the exercised `src/` is byte-identical to base (`git diff origin/main 8bc0bf00b -- src/` empty), and it fails under a real `HOME` as well as an isolated one, with CI unable to serve as the counter-check (ubuntu-latest/node-22 against macOS/node-26). It is out of this PR's column and stays out. Independent corroboration, in case it is useful to whoever owns it: in my own full-suite run on this head the same cell failed with the same message, and it also failed on an unmodified `origin/main` worktree (`6f15f71ea`) run with the same command — `tests 158 / pass 155 / fail 2`, the other being `backend-settings-tiers`' fixture cell.

**Q2 — informational, pre-existing, and now recorded in the file.** Unchanged by this PR: the bare `catch` still conflates `ESRCH` with `EPERM`, so `lo-evidence-1` is still reaped. Your "recorded, not a deferred work item" stands, and this round gives the record a home rather than a thread: the reviewer asked for the framing to be *narrowed, not unreachable*, and the sweep's doc block now says so — that the form-match changes which names can reach the `kill`, not which throws read as "abandoned"; that for a well-formed name only `ESRCH` and `EPERM` remain and `EPERM` still converges on the deletion; that this is out of reach here for a structural reason (per-user `os.tmpdir()`, names written by a process of this user) but reachable in principle on a shared `/tmp` with a second user; and the follow-up shape, in the reviewer's words: reap only on `e.code === "ESRCH"`, warn or rethrow otherwise. Not bundled, because it changes the behaviour for a live-but-unreadable profile and wants its own review.

**Q3 — informational, no action.** `MAX_PID` itself is accepted and `kill` answers a genuine `ESRCH` for it, so `lo-evidence-2147483647` is reaped; `2147483648` is refused before the `kill`. The bound stayed exactly where you measured it, and the predicate's comment now carries the measured table behind it (`NaN`/`1e20`/`2147483648` → `TypeError ERR_INVALID_ARG_TYPE`; `MAX_PID` → `ESRCH`; `0` → no throw, which is why `pid > 0` is in the guard) instead of the sentence it had.

**For a round 2, if you want one on this head:** the delta is comments plus the stamp hash, so nothing beyond re-running rows 1–3 on `d85b9ea11` is asked for by the change itself — rows 7–10 were properties of the executable code, which is unchanged. No QA finding is open, and nothing here is marked deferred.

damianvtran:
### Agent review — round 2 (convergence)

Reviewer: independent review lane (`reviewer` role, `lopdev` team) — model `deepseek/deepseek-flash`, as reported by the session; no provider fallback occurred during this round.
Scope: `8bc0bf00b..d85b9ea11` — 3 files, +57/−14: `scripts/capture-evidence.mjs` (+51/−9), `scripts/capture-evidence.test.mjs` (+4/−3), `docs/evidence/manifest.json` (+2/−2). Head `d85b9ea11`; round 1 was `6f15f71ea..8bc0bf00b`.

**Verdict: clean, and the round-1 verdict stands on this head.** No BLOCKER, no MAJOR, no MINOR. The two NITs at the end are wording cells with a one-line fix each; they are follow-ups and **not** a reason for another round. On my reading the PR is mergeable as it stands — see the last section.

---

#### 1. The delta really is comment-only under `scripts/` — verified independently, not by the filter

The implementer's filter is not the evidence; I reproduced it and then replaced it with something stronger.

- **Their filter, rerun:** `git diff 8bc0bf00b..d85b9ea11 -- scripts/ | grep -E '^[+-]' | grep -vE '^(\+\+\+|---)' | grep -vE '^[+-][[:space:]]*(\*|//|/\*)'` → **0 lines**. And classifying all 67 changed `±` lines mechanically (51+9+4+3 = 67) gives **67 of 67** matching `^[[:space:]]*\*`, i.e. every changed line is a line inside a block comment.
- **The filter's two holes, closed by hand:** a line can be comment-prefixed but carry code after a `*/`, and a filter that only looks at the first character cannot see it. Count of changed lines matching `\*/[[:space:]]*[^[:space:]]` or `/\*.*[^[:space:]]` → **0**. Every changed `*/` is alone on its line.
- **Parser-level equivalence (the check I would substitute for the filter):** both revisions of both files, run through `esbuild` (`minifyWhitespace: true`, `minifySyntax: false`, `minifyIdentifiers: false`, `legalComments: "none"`) — a real JS parser that strips comments and whitespace — produce **byte-identical output**:

  ```
  capture-evidence.mjs:      base 76535 chars / head 76535 chars  -> identical: true
  capture-evidence.test.mjs: base  4908 chars / head  4908 chars  -> identical: true
  ```

  Identical token streams after comment removal is a stronger statement than "every changed line is a comment": it also excludes a change *inside* a comment span that esbuild would have kept.
- **The new prose is comment-only in the rendering, too:** grepping the comment-stripped rendering of the head file for the nine phrases the new text is made of (`measured on node 26.5.0`, `ERR_INVALID_ARG_TYPE`, `review round 1, F1`, `What this does NOT close`, `reap only on \`e.code === "ESRCH"\``, `load-bearing`, `six \`countsMean\` cells`, `death certificate`, `narrowing`) → **nine of nine comment-only**, none present in the code.
- **The non-comment delta is exactly what was claimed, and nothing else:** the manifest's two changed lines are the `scriptsTree` value and one `why` clause — 4 changed lines total under a file whose diff I read line for line.

So the executable code at `d85b9ea11` is the code round 1 reviewed and QA exercised, and the claim holds. Two of the three manifest questions that follow from that, answered by measurement rather than by reading the prose:

```
manifest.scriptsTree = 227aca40ceae9b760a1d3fc674b85069275bb254
git rev-parse d85b9ea11:scripts = 227aca40ceae9b760a1d3fc674b85069275bb254   MATCH
manifest.srcTree     = 79d4a496ba5f05a4885f52169023dc43a35c5db4
git rev-parse d85b9ea11:src     = 79d4a496ba5f05a4885f52169023dc43a35c5db4   MATCH
```

and the gate that exists for that field agrees — `stampFailures(headManifest)` fed HEAD's trees and `HEAD:scripts/capture-evidence.mjs`: **no findings** (the `frames`/`countsMean` halves need `docs/evidence` on disk and are outside this delta; neither value moves in this range). The `why` clause's account of the numbers also checks out: `replaced.scriptsTree = 5744a1827d7c52ee198640c3cc21e87ae6f1181a`, which is `git rev-parse 6f15f71ea:scripts` — main's own value at the rebase base, as the clause says.

#### 2. The new comment text is accurate, checked against the measurements it cites

**The five-row table reproduces exactly, on this machine's node, which is the node it names:**

```
node v26.5.0
process.kill(NaN, 0)           -> TypeError ERR_INVALID_ARG_TYPE
process.kill(1e20, 0)          -> TypeError ERR_INVALID_ARG_TYPE
process.kill(2147483648, 0)    -> TypeError ERR_INVALID_ARG_TYPE
process.kill(2147483647, 0)    -> Error ESRCH
process.kill(0, 0)             -> no throw
process.kill(4294967295, 0)    -> TypeError ERR_INVALID_ARG_TYPE
process.kill(1, 0)             -> Error EPERM
process.kill(1.5, 0)           -> TypeError ERR_INVALID_ARG_TYPE
```

- *"`MAX_PID` is exactly where node stops accepting a pid"* — holds: `2147483647` is accepted (ESRCH), `2147483648` and `4294967295` are refused. Not "approximately".
- *"Every one of the first three was reaching the bare `catch` in the sweep below"* — verified against the **pre-fix** code, not inferred: `origin/main`'s sweep was `if (!name.startsWith("lo-evidence-") || name === mine) continue; const pid = Number(name.slice(...)); try { process.kill(pid, 0); continue } catch {}` then `rmSync`, so `lo-evidence-manifest-XXXX` (→ `NaN`), a 21+ digit suffix (→ `1e20`) and `lo-evidence-2147483648` each reached the `kill`, threw, and were deleted. The rule's stated purpose is the one it has.
- *"`lo-evidence-2147483648` … is refused before the `kill`"* — `Number("2147483648")` is an integer, `> 0`, `> MAX_PID`, so `profileOwnerPid` returns `null` at `capture-evidence.mjs:3280` and the sweep `continue`s at `:3346`. Refused, never killed.
- *"`pid > 0` … the sweep would `continue` on it … kept either way"* — accurate **as written**: with the guard, `null` → `continue`; without it, `kill(0, 0)` does not throw → `continue`. Both keep the directory; what the guard buys is the honest answer, and the text says exactly that rather than overclaiming a behavioural difference. The bogus `RangeError` hedge is gone (`grep RangeError` over the head file: none) and the measured set supports its removal — every pid-side throw measured is a `TypeError`.
- *"the `catch` below still reads ANY throw from `kill` as 'abandoned', so this narrows the class … rather than making the throw unambiguous"* — the sweep at `:3347-3354` is unchanged: bare `catch`, then `rmSync`. `EPERM` genuinely converges on the deletion. Framing and follow-up shape (`reap only on e.code === "ESRCH"`) are as I asked, and deferring it is right for the reason given.

#### The three findings and the two corrections, as landed

- **F1 — landed, both occurrences, and the numbers are right.** `capture-evidence.mjs:3307-3310` and `capture-evidence.test.mjs:167-170`. Measured on `scripts/evidence-manifest.test.mjs` at head: `COUNTS_TREE = tree({...})` at `:817` is one module-scope root (`tree()` → `mkdtempSync`, pushed to the `scratch` reclaim list at `:45-48`), and **exactly six** cells call `countsMeanFailures(…, COUNTS_TREE)`: `:843, :853, :870, :887, :896, :905`. Five pass a manifest that carries `countsMean` and walk it; the sixth (`:905`) passes `{frames: 3, supplementary: []}` and returns at `check-evidence.mjs:763` **before** `frames(dir)` at `:769`. Executed proof of the distinction the sentence draws, with a nonexistent root:

  ```
  no countsMean  (the 6th cell's shape): returned []        -> the tree was NOT walked
  with countsMean (the 5 walking cells' shape): threw ENOENT -> the tree WAS walked
  ```

  So "six cells share one root, the five that walk it fail together, the sixth returns early without touching the tree" is exactly the measured shape. (The one other cell that shares that file's module scope — the shipped-manifest stamp cell at `:728` — does not call `countsMeanFailures` at all.)
- **F2 — landed, and the reasoning is sound.** The entry names its own partiality and says why, and the "padding would report displacements that did not happen" argument holds on two grounds I could check: the other 16 `replaced` maps each carry 4-7 keys (`head`, `srcTree`, `scriptsTree`, `capturedAt`, and for one entry `frames`/`surfaces`/`refreshedFrames`) because those folds moved those fields, and this branch moved only `scripts/`; and a `replaced` map is a record of displaced values, so a key that names a value nobody displaced asserts a history that did not occur. Nothing consumes the key set structurally — no script reads `reStamps[].replaced` (`grep -rn '\.replaced\b' scripts/*.mjs` finds only `notarize-artifacts.mjs`'s and `update-robustness.test.mjs`'s unrelated fields), so the partial map cannot break a reader. The clause also tells the reader where to take the other three, which is what made the round-1 report call it partial rather than lossy.
- **F3 — landed, corrected by the same measurement I ran.** Body now reads `capture-evidence.mjs:5096` on this head and `4990` on the base. `biome check scripts/capture-evidence.mjs` with the repo's `biome.json`, on each revision:

  ```
  d85b9ea11 (head):    scripts/capture-evidence.mjs:5096:6 lint/performance/useTopLevelRegex
  origin/main (base):  scripts/capture-evidence.mjs:4990:6 lint/performance/useTopLevelRegex
  8bc0bf00b (pre-remediation head): 5054:6  (the same warning, moved only by the comment growth)
  ```
  Same regex, same warning, none introduced by this diff: the delta between base and head is the comment lines themselves (5462 → 5568 lines).
- **Correction 1 (`pid > 0` load-bearing, `RangeError` hedge removed)** — landed as described; the hedge is gone from the file and the guard is stated at the size the measurements support (see §2).
- **Correction 2 (bare `catch` framed as narrowing)** — landed as described, in the wording I asked for, and it correctly declines to imply the class is closed.

#### Findings

No BLOCKER, no MAJOR, no MINOR. Two NITs, both in cells of the new table/prose, both optional:

- **N1 — `capture-evidence.mjs:3259`:** the row `process.kill(0, 0) -> no throw, and it tests no process`. `pid 0` with signal 0 does reach a check — POSIX says pid 0 means the caller's own process group, which always exists — so the accurate tail is "no throw, because pid 0 names the caller's own process group and can never answer 'no such process'". The prose two lines below (`:3269-3270`) already says the group thing correctly; only the cell's tail overstates.
- **N2 — `capture-evidence.mjs:3324`:** "`lo-evidence-1` is the measured demonstration that `EPERM` is an answer this code path currently reads as a death certificate". What was measured is `process.kill(1, 0) -> EPERM` and the catch→`rmSync` path; a `lo-evidence-1` directory cannot exist, since pid 1 is `launchd`/`init` and never runs this script. One clause ("`kill(1, 0)` answers `EPERM`, which this code path reads as a death certificate") says the true thing and drops the dangling name.

#### What this means for the gates, and the merge

- Round 1's verdicts and QA round 1's PASS were taken on `8bc0bf00b`; this head's executable code is token-identical to it, so both carry to `d85b9ea11` — not by the "non-conflicting fixes" dispensation but by identity, and QA's matrix has no cell whose input moved. The one change that is not a comment is the re-derived stamp, and it is the right one: `scriptsTree`/`srcTree` equal `HEAD:scripts`/`HEAD:src` on this head and `stampFailures` reports no findings against the head manifest.
- On the head as it stands: `mergeable=MERGEABLE`, merge state `CLEAN`, and every check on `d85b9ea11` is `SUCCESS` — Change Scope, Version Bump Guard, Lint, Type Checking, Desktop Tests, NPX Sanity (ubuntu-latest, macos-latest); Runtime Dependencies and Security Audit are `SKIPPED`. No version bump in the diff (3 files, none is `package.json`).
- **Mergeable on my reading: yes.** No human reviewer was added and nobody was tagged, per this repo's standing instruction; the round is fresh on head in the strong sense above.

#### Hygiene

Work done without a worktree: base/head blobs and `HEAD`-pinned copies of `scripts/` and `src/renderer/src/shared/themes/palettes` extracted with `git show`/`git archive` under `~/workspace/review-339`, plus one `biome.json` copy for the F3 probe and one `/tmp/op-filter.txt` (no `lo-evidence-<digits>` name anywhere, so no concurrent sweep could match). All of it removed after the round; the primary checkout was never written to (`git status` unchanged: the one modified docs file that was already there), no test that boots the TUI was run, no capture, no evidence sweep, the lease was never taken, and no other lane's worktree was touched.

damianvtran:
### QA report — round 2 (convergence)

**Reviewer:** qa-tester (`lopdev` team, independent pass; model actually running: `deepseek/deepseek-flash`)
**Scope:** `8bc0bf00b..d85b9ea11` — the round-1 remediation commit. Head under test `d85b9ea11191c98bfd0b9081f46561a1f145c612`, re-verified against `gh pr view 339 --json headRefOid`.
**The one question:** does this head behave exactly as the head round 1 passed (`8bc0bf00b`), given the claim that the delta under `scripts/` is comment-only?

**Verdict: PASS.** The claim holds as a behavioural proposition. Same rig, same 17 name shapes, 12 runs per revision, three revisions side by side: **every row is identical between the round-1 head and this head**, and the rows that differ from the pre-fix `origin/main` differ in exactly the way the fix intends. The two headline rows reproduce round 1 to the run, the cross-process signature is still 12/12 pre-fix and 0/12 here, and the shipped-manifest stamp assertion passes at this head.

#### Where this ran

Own worktrees under `/private/tmp/qa339r2/` — `base` @ `6f15f71eaa966704e37152634f267850283bf60b` (`origin/main`, the fix's parent and the pre-fix arm), `r1head` @ `8bc0bf00b17af5895c29a30e73fdecdef9a76a95` (the head round 1 exercised), `head` @ `d85b9ea11191c98bfd0b9081f46561a1f145c612`. The author's worktree and every other lane's were left untouched, and all three of mine end `git status --porcelain` empty — no repository file was edited, and nothing was written into any of them (the rig's scratch lives beside them). macOS 25.6.0 arm64, node v26.5.0, host load average 380–530 throughout (several lanes capturing — this bounds nothing below, all rows are pass/fail). Every child ran under `env -i` with its own fresh `TMPDIR` **and** `HOME`, so no `CMUX_*`/`LOP_*` travelled with it and no run could read or sweep the shared temp directory.

#### Matrix

| # | Surface | Command (actual) | Actual output | Result |
|---|---|---|---|---|
| 1 | **Reported failure**, pre-fix vs round-1 head vs this head | `env -i HOME=… TMPDIR=$R/tmp node driver.mjs <rev>/scripts/capture-evidence.mjs <label> <mode>` — the revision's own module, its own `sweepStaleProfiles()`, 17 synthetic names, 12 runs each | pre-fix `lo-evidence-manifest-Xy9K2p` **DELETED 12/12** (18 dirs → 7); round-1 head **kept 12/12** (18 → 15); this head **kept 12/12** (18 → 15) | **PASS** |
| 2 | **A genuinely abandoned profile is still reaped** (the fix must not stop sweeping) | same rig, `lo-evidence-<dead child pid>` | **DELETED 12/12 on all three revisions** — directory *and* its nested `payload.txt` gone | **PASS** |
| 3 | A live profile survives | same rig, `lo-evidence-<live sleep pid>` | `kept 12/12` on all three | **PASS** |
| 4 | The sweep's own profile is skipped | same rig, `lo-evidence-<driver pid>` | `kept 12/12` on all three | **PASS** |
| 5 | Adversarial name shapes (17 total) | same rig, table below | 8 rows changed from pre-fix, all in the right direction; **0 rows differ between round-1 head and this head** | **PASS** |
| 6 | **Cross-process signature** (round-1 row 7) | stand-in sweeping process in one shared `TMPDIR` + 4 concurrent real `node --test scripts/evidence-manifest.test.mjs`, 3 waves = 12 runs per arm | pre-fix sweeper: **12/12 runs FAILED**, `Error: ENOENT: no such file or directory, scandir '/private/tmp/qa339r2/xproc/tmp/lo-evidence-manifest-FJeOhD'`; this head's sweeper, same harness: **0/12** | **PASS** |
| 7 | Scoped test files, repo's own files | `node --test scripts/capture-evidence.test.mjs scripts/evidence-manifest.test.mjs` | this head **41 pass / 0 fail** (7 + 34); round-1 head **41 pass / 0 fail** — same totals | **PASS** |
| 8 | **Shipped-manifest stamp — the assertion the re-derivation exists for** | `git rev-parse HEAD:scripts HEAD:src` vs `docs/evidence/manifest.json`; `node --test scripts/evidence-manifest.test.mjs` | `scriptsTree 227aca40ceae9b760a1d3fc674b85069275bb254 == HEAD:scripts`; `srcTree 79d4a496ba5f05a4885f52169023dc43a35c5db4 == HEAD:src`; **`✔ the SHIPPED manifest's stamps describe the tree it ships in`** | **PASS** |
| 9 | Independent check of the "comment-only" claim | own comment-region scanner over the diff, plus `git diff --name-status` and a structural walk of both manifests | every changed line is inside a block comment (51 added + 9 removed lines in `capture-evidence.mjs`, 4/3 in its test file, **0 classified as code**); the manifest delta is exactly `scriptsTree` + one `why` string | **PASS** |
| 10 | CI on this exact head | `gh api repos/damianvtran/local-operator-ui/commits/d85b9ea11/check-runs` | 9 runs: **7 `success`, 2 `skipped`** (Security Audit, Runtime Dependencies), **0 failures**, every run bound to `d85b9ea11` | **PASS** |

PR state on the same call: `OPEN`, `draft: false`, `mergeable: MERGEABLE` / `state: CLEAN`. The full frame gate is a separate matter — see Q1.

#### Row 5 — the 17 name shapes, three revisions, 12 runs each

| Name (shape) | pre-fix `6f15f71e` | round-1 head `8bc0bf00b` | **this head `d85b9ea11`** | Judgement |
|---|---|---|---|---|
| `lo-evidence-manifest-Xy9K2p` | DELETED 12/12 | kept 12/12 | **kept 12/12** | the reported failure; fixed, and unchanged by the delta |
| `lo-evidence-manifest-` | DELETED 12/12 | kept 12/12 | **kept 12/12** | same |
| `lo-evidence-bare` | DELETED 12/12 | kept 12/12 | **kept 12/12** | same |
| `lo-evidence-abc` | DELETED 12/12 | kept 12/12 | **kept 12/12** | same |
| `lo-evidence-1e3` | DELETED 12/12 | kept 12/12 | **kept 12/12** | same |
| `lo-evidence-9999999999999999999999` | DELETED 12/12 | kept 12/12 | **kept 12/12** | same |
| `lo-evidence-2147483648` (`MAX_PID`+1) | DELETED 12/12 | kept 12/12 | **kept 12/12** | refused before the `kill` |
| `lo-evidence- 12` (leading space) | DELETED 12/12 | kept 12/12 | **kept 12/12** | `Number(" 12")` is a pid; the shape rule is not |
| `lo-evidence-<deadPid>` | DELETED 12/12 | DELETED 12/12 | **DELETED 12/12** | the sweep still sweeps |
| `lo-evidence-12` | DELETED 12/12 | DELETED 12/12 | **DELETED 12/12** | well-formed, dead |
| `lo-evidence-2147483647` (`MAX_PID`) | DELETED 12/12 | DELETED 12/12 | **DELETED 12/12** | accepted pid, genuine `ESRCH` |
| `lo-evidence-<livePid>` | kept 12/12 | kept 12/12 | **kept 12/12** | live |
| `lo-evidence-<selfPid>` | kept 12/12 | kept 12/12 | **kept 12/12** | own-profile skip |
| `lo-evidence-` | kept 12/12 | kept 12/12 | **kept 12/12** | zero digits |
| `lo-evidence-0` | kept 12/12 | kept 12/12 | **kept 12/12** | `kill(0,0)` tests no process |
| `lo-evidence--1` | kept 12/12 | kept 12/12 | **kept 12/12** | negative |
| `lo-evidenceX`, `other-dir` | kept 12/12 | kept 12/12 | **kept 12/12** | no prefix |

8 rows changed from pre-fix, every one of them a name that is not a profile; the 9 rows the rule is *about* are unchanged. **Round-1 head and this head agree on all 17, 12 runs each, and the `beforeCount → afterCount` is 18 → 15 on both with no unexpected name ever appearing.**

#### Row 6 — the signature as it actually presents

Under a pre-fix sweeper the real, unmodified `evidence-manifest.test.mjs` fails 12 runs out of 12, and the failure is the one F1 described:

```
Error: ENOENT: no such file or directory, scandir '/private/tmp/qa339r2/xproc/tmp/lo-evidence-manifest-FJeOhD'
    code: 'ENOENT',
    path: '/private/tmp/qa339r2/xproc/tmp/lo-evidence-manifest-FJeOhD'
```

with the five `countsMean` cells that walk that shared module-scope root failing together and the sixth (`a manifest with no countsMean is not this guard's failure`) passing. I sampled six instrumented pre-fix runs to check the remediation's own refinement rather than take it on trust: **`a countsMean paragraph leading with the walk's numbers passes`, `a paragraph left behind by a fold…`, `a paragraph with no reading in it…`, `the older paragraphs under the leading one…`, `a stale surfaces paragraph…` failed in 5/5 instrumented runs**, and `a manifest with no countsMean…` **failed in none**. So the corrected wording ("the five of them that walk it fail together … the sixth returns early without touching the tree") is what the code does, and round 1's original "six countsMean cells failing together" was the imprecise version. Cells over the file's *other* fixture roots (`a partial run that adds a surface…`, `a stray undeclared directory…`, `a partial run that only refreshes…`, `a stray beside a genuinely added surface…`) fail in some runs and not others, exactly as a per-fixture `mkdtempSync` root being deleted at a random moment implies.

#### Row 9 — how the "comment-only" claim was checked, and what was actually transformed

Two independent lines, neither reusing the author's regex:

1. **Structural.** A comment-region scanner of my own (string-aware, for `'`/`"`/backtick) classified every line of both revisions. All 51 added and all 9 removed lines in `scripts/capture-evidence.mjs` are inside a block comment — **0 lines classified as code** — and the same for the test file's 4/3. `git diff --name-status 8bc0bf00b d85b9ea11` is exactly three files; a structural walk of the two manifests reports differences at exactly two paths: `/scriptsTree` and `/captureOrigin/reStamps[16]/why` (a 1064-character prose append, no other field moved).
2. **Behavioural.** Rows 1–6 above.

**Declared, because it is a transformation of shipped bytes:** the pre-fix revision does not export `sweepStaleProfiles` (it is a plain `const`, reachable only through `main`, which launches Chrome — not done here). So the pre-fix arm loads a copy of `6f15f71eaa:scripts/capture-evidence.mjs` whose **only** differences are the four relative import specifiers rewritten to absolute `file://` URLs and one `export { sweepStaleProfiles };` appended. `diff` against `git show 6f15f71eaa:scripts/capture-evidence.mjs` is exactly those 5 added / 5 removed import lines and the 2 appended lines, and the `sweepStaleProfiles` span itself is byte-identical. (Round 1's copy resolved its imports because it sat inside a `scripts/` directory; this one sits beside the worktrees, hence the specifier rewrite. No executable line differs either way.)

#### Q-findings

- **Q1 (environmental, not this delta).** The full frame gate **deferred both times**: `node scripts/check-evidence.mjs` → `Evidence check DEFERRED: another sweep (or its image child) holds the machine lease; retry after it finishes. No frames checked. Do not delete the lock file.` (rc 75). The holder is a genuine neighbour — `node scripts/capture-evidence.mjs … --only=chat-sidebar-current-row--settings-rail` and a peer lane's `check-evidence.mjs` (45 min elapsed) were both live on this box. I did not steal the lease. So **no frames were image-checked in this pass**; the stamp half of that gate was run directly and passes (row 8).
- **Q2 (informational, pre-existing, unchanged by the delta).** The bare `catch` around `process.kill` still reads `EPERM` as a death certificate, so a `lo-evidence-<pid>` naming a live process this user does not own is reaped. Round 1 recorded it; the remediation's comment now states it explicitly and scopes the follow-up out. No action here.

#### What this pass does **not** prove

- **It is not a re-run of the repository test matrix.** Scoped to the delta's two test files, the sweep rig, and the stamp assertion. Round 1's neighbour run (9 files, 160/161 with a pre-existing failure) is not repeated here — the delta touches no executable line, so re-running it would test the round-1 head, not this one.
- **No frames were image-checked.** `check-evidence.mjs` deferred on the machine lease (Q1); the 6009-frame loop and its ImageMagick calls did not run in this pass.
- **No Chrome-backed capture.** No browser engine was launched, per machine policy. `teardown`, a real ~180 MB profile, and the `--only=` capture path were not exercised; the sweep was driven directly, which is what the defect needs.
- **The cross-process row is a stand-in, not the incident.** One isolated `TMPDIR` with a synthetic sweeping process and 4 concurrent test runs — not the operator's live shared temp directory with a real capture's sweep. I deliberately never point a pre-fix sweep at the shared temp directory.
- **macOS / node v26.5.0 only.** The platform-sensitive part is `process.kill`'s error codes, measured here and not on Linux or Windows.
- **The comment-only claim is verified by an independent scanner plus behaviour, not by an AST diff.** No JS parser is present in the tree and I install nothing; the scanner is my own and is disclosed above.

#### Hygiene

Everything I created is inside `/private/tmp/qa339r2/` and is being removed — the three worktrees, the driver, the sweeper module, the per-run isolated `TMPDIR`s and logs. No repository file was edited and no worktree of mine is left dirty. On the shared `TMPDIR`, for whoever owns them: it currently holds **0** `lo-evidence-*` trees (the leaked `lo-evidence-manifest-*` trees round 1 flagged are gone) and **6** leaked `lo-sweep-*` trees — which are `capture-evidence.test.mjs`'s own scratch roots, named deliberately outside the swept prefix, so no sweep touches them. I left them alone; they are not mine and not this PR's. No issues created, no one tagged, no human reviewer added.

damianvtran:
### Merged as admin — disclosure

Merged at `2de7ae97784300dd855a9fd995f8bf4ef0356616` (merge commit, `main`) with branch head `d85b9ea11191c98bfd0b9081f46561a1f145c612`.

**Gates on the reviewed head:** `### Agent review — round 2 (convergence)` clean — no BLOCKER/MAJOR/MINOR, round-1 verdict standing on this head — and `### QA report — round 2 (convergence)` **PASS**; plus round-1 rounds (reviewer clean/TERMINAL with 1 MINOR + 2 NITs; QA PASS) each answered by the implementer's remediation comments. 0 failing checks on `d85b9ea11` (7 success, 2 skipped by the change-scope classifier). The reviewers named the model actually running; no provider fallback occurred.

**Why `--admin`:** the submitting account is also the repository owner, so no independent approval could be recorded through the UI; the four recorded agent rounds are what stand in for it, and this line is the disclosure the repository's convention asks for. No human reviewer was added and nobody was tagged.

**Post-review delta:** none — the merged head is the reviewed head. The convergence round established the remediation delta was **comment-only** under `scripts/` by three independent routes: the author's line filter, the reviewer's `esbuild` comment-stripped comparison (byte-identical output on both revisions), and QA's own comment-region scanner (0 of 67 changed lines classified as code), with behaviour reproduced row-for-row against the round-1 head.

**Why this fix mattered:** the defect was **live on `main`** and was costing every lane on this machine spurious failures. `sweepStaleProfiles()` matched any `lo-evidence-*` name in `tmpdir()` — including the `lo-evidence-manifest-*` scratch root `evidence-manifest.test.mjs` creates — so `Number("manifest-…")` was `NaN`, `process.kill(NaN, 0)` threw, and the bare `catch` read that as "abandoned" and deleted a live tree. QA's measurement: pre-fix **12 of 12** runs failed with that signature; on this head **0 of 12** across 299,521 sweeps, with a dead-pid profile still reaped and a live one still kept.

**Found by another lane, not by this one:** the defect was traced by the **trace-order lane's convergence QA**, which reported it with the mechanism and a repro rather than widening its own gated branch. The attribution is theirs.

**Not claimed:** no frames were image-checked (`check-evidence.mjs` deferred both times, the machine-wide lease being held by a live neighbour run — claimed as neither pass nor fail); the full repo matrix was not re-run; no Chrome-backed capture path was exercised for this change.

**Deferred, recorded rather than silently dropped:** the remaining bare `catch` still treats any non-`ESRCH` throw as "abandoned" — the form-match `narrowing` that class rather than closing it — with the follow-up shape `reap only on e.code === "ESRCH"`. Also noted, pre-existing and outside this diff: the file's own `build()` leaves un-reclaimed `lo-sweep-*` scratch roots (6 currently in the shared temp dir).

damianvtran:
**Shipped in `v0.27.0`** — Release: https://github.com/damianvtran/local-operator-ui/releases/tag/v0.27.0 (the tag pins `f19827852`, the release bump's merge commit, verified with `git rev-parse v0.27.0^{commit}`).

Publish pipeline green end to end, including the two jobs that had been failing on every release since `v0.26.6`: npm `local-operator-ui@0.27.0`, macOS/Linux/Windows installers attached (20 assets), promoted to `/releases/latest`, and the signed-update verification dispatched. The window since `v0.26.11` was #297, #339 and #277; bump: **minor** (two step-function capabilities, two `Release: minor` claims).

PR #341, opened 2026-09-18
fix(menu): give the About panel the app's own identity, and gate it by window mode

damianvtran:
## What this fixes

The macOS About panel is AppKit's own window, and it renders in front of the operator whenever anything opens the app menu's first item. For an **unpackaged** launch - which is how every rig, QA harness and `npx electron .` boots this app, from `node_modules/electron` - there is no bundle of ours for it to read, so it was filled from Electron.app's identity, and the window mode could not stop it: the menu item was Electron's own `about` **role**, and a role's click is AppKit's `orderFrontStandardAboutPanel:`, so there was no handler of ours in the path to gate.

So an agent run could put this in front of the operator, and every agent run did.

| | name | version line |
| --- | --- | --- |
| before | `Electron` | `Version 44.3.0 (44.3.0)` |
| after | `Local Operator` | `Version 0.27.0 (Electron 44.3.0, unpackaged)` |

Both rows are screenshots of the panel window itself, taken from the built app on this machine and attached here as `docs/evidence/about-panel/frames/before-identity.png` and `after-identity.png`.

| before | after |
| --- | --- |
| About panel before: Electron, Version 44.3.0 (44.3.0) | About panel after: Local Operator, Version 0.27.0 (Electron 44.3.0, unpackaged) |

## The change

Two halves, and neither covers the other.

**The decision, extracted so it is testable as arithmetic.** `resolveAboutPanelAction(mode)` in `src/main/window-mode.ts`, beside the other presentation policy and in the module that deliberately imports nothing from Electron:

- `headless` -> `suppress`. It is the mode whose entire promise is that nothing appears on the operator's screen, so the action logs `[about-panel] suppressed by window mode headless: a headless run raises no window` and raises nothing. A line rather than silence, because an action nothing reached and a panel that failed to appear are otherwise the same absence - to the operator and to a rig.
- `inactive` -> `show`. The panel is allowed there, and that is **measured rather than hoped**: raising it does not activate the app (`app.isActive()` stays `false`, and the app was the frontmost application in **0 of 30** samples across the run), which is exactly the promise the mode makes.
- `normal` -> `show`. The operator's own app, unchanged.

**The identity, registered once at startup.** `configureAboutPanel()` in `src/main/index.ts` calls `app.setAboutPanelOptions` with:

- `applicationName: app.name` - the app's own name, which resolves from `productName` in `package.json` ("Local Operator"), so the panel and the menu beside it cannot name the app differently;
- `applicationVersion: app.getVersion()` - the app's own version (`package.json` is the version source of truth), not the runtime's;
- `version:` a parenthesised **build** string naming the desktop host and whether the bundle is the shipped one (`Electron 44.3.0, unpackaged`), because macOS fills that half with the version again otherwise and `Version 0.26.11 (0.26.11)` tells a reader nothing about the run in front of them;
- `copyright`, read from `package.json`'s `build.copyright` - the one place the project states it, and the string electron-builder stamps into a packaged bundle - and omitted entirely rather than guessed when that read fails.

The menu item now carries a handler of its own, which is what makes the gate reachable at all.

**The menu is installed before the ready handler awaits anything** (code review round 1, F1). Until it is, Electron's DEFAULT application menu is live, and its first item is `About Electron` carrying that same `about` role - a click no handler of ours is in. It used to sit below the backend startup, which is seconds of an agent-triggerable panel per launch; it now runs immediately after the smoke-test branch, above the launcher watch and every `await`.

## Evidence

`scripts/about-panel.test.mjs`, added to `pnpm test:desktop`: the decision for all three modes, every documented mode answered (so a mode added later cannot be silently unhandled), and the shapes in `src/main/index.ts` that apply it - no item left to Electron's role, exactly one `showAboutPanel` call site, that site past the suppression branch with a `return` in between, and the registration's own fields.

The two rigs live in `docs/evidence/about-panel/harness/` and their measurements are in the set's README.

### The behaviour half: headless, with nothing on screen

`bash docs/evidence/about-panel/harness/run.sh . after-headless` against the built tree. The census is the window server's own list for that pid, on-screen only, so "no window" is a read rather than an absence - the headless run's own window is in the full census with `onscreen` false:

```
== windows owned by this app BEFORE the action (window server's list, on screen only):
   (none)
== About menu entries: Local Operator > About Local Operator [role=none] click=function
== active before the action: false; windows: [{"visible":false,"focused":false,"focusable":false,"size":[1380,900]}]
== invoked: Local Operator > About Local Operator via the item's own click handler - called
== active after the action: false; windows: [{"visible":false,"focused":false,"focusable":false,"size":[1380,900]}]
== windows owned by this app AFTER the action:
   (none)
== the action created no window: nothing reached the screen
== the app's own line about the action:
   23:39:15.789 > [about-panel] suppressed by window mode headless: a headless run raises no window
== this app was the frontmost application in 0 of 33 samples (pid 62441)
    [window-mode] window mode headless: 1380x900, window created and never shown, page throttling off, no Dock tile
    [window-mode] state: visible=false focused=false focusable=false size=1380x900 content=1380x868
== processes left from this run: 0
```

The rig also **refuses to run against a tree with no gate**: pointing it at `origin/main` would raise a real panel on the operator's desktop, which is the noise this PR removes, so it fails before launching.

### The identity half: the panel itself

`ABOUT_PANEL_VISIBLE_CAPTURE=1 timeout 100 bash docs/evidence/about-panel/harness/identity.sh . after-identity` - one instance, capture and teardown in the same command, under an `EXIT` trap:

```
== windows owned by this app BEFORE the action:
   {"id": 38270, "pid": 53022, "owner": "Electron", "name": "Local Operator", "layer": 0, "onscreen": false, "alpha": 1065353216, "bounds": {"x": 464, "y": 259, "w": 800, "h": 600}}
== app identity (its own answer): name=Local Operator version=0.27.0 electron=44.3.0 packaged=false
== About menu entries: Local Operator > About Local Operator [role=none] click=function
== active before the action: false; windows: [{"visible":true,"focused":false,"focusable":true,"size":[800,600]}]
== invoked: Local Operator > About Local Operator via the item's own click handler - called
== active after the action: false; windows: [{"visible":true,"focused":false,"focusable":true,"size":[800,600]}]
== driver: mode=inactive appPid=53022 scratch=/tmp/about-panel/after-identity
== windows owned by this app AFTER the action:
   {"id": 38271, "pid": 53022, "owner": "Electron", "name": "", "layer": 0, "onscreen": false, "alpha": 1065353216, "bounds": {"x": 722, "y": 211, "w": 284, "h": 191}}
   {"id": 38270, "pid": 53022, "owner": "Electron", "name": "Local Operator", "layer": 0, "onscreen": false, "alpha": 1065353216, "bounds": {"x": 464, "y": 259, "w": 800, "h": 600}}
== the window server's ON-SCREEN set for this pid: (none)
== captured the window this action created (id 38271) to /tmp/about-panel/after-identity/frames/after-identity.png
== this app was the frontmost application in 0 of 30 samples (pid 53022)
== the app's own lines about the launch and this action:
   [window-mode] window mode inactive: 800x600, window shown without activating the app, page throttling off
   00:53:53.541 › [window-mode] window mode inactive is not launcher-bound: a person can close it
   [window-mode] window mode inactive is not launcher-bound: a person can close it
   [window-mode] state: visible=true focused=false focusable=true size=800x600 content=800x568
== processes left from this run: 0
```

The `before` frame comes from the same rig pointed at a tree built from `origin/main` at `f19827852`, and its own output says which route it had to use: an item that is still the ROLE has no handler of ours to call, so that run raises the panel through `app.showAboutPanel()` - the call the role makes.

**What the frames are pictures OF, re-shot in round 1 (design D1).** Both frames are now of the trees under review: `before` from a tree BUILT at this branch's base, `origin/main` `f19827852`, and `after` from this branch's own code - the tree whose `srcTree` `docs/evidence/manifest.json` pins. Neither is a mock, and neither is the app photographing itself: the About panel is AppKit's window, so it is photographed from outside with `screencapture -l`, window-only, never the screen, while the app raised it in `inactive` mode with the app never activated (0 of 30 samples frontmost, by pid). The pair is two `0.27.0` panels now, so no caption disagrees with the head - which is the finding the round-1 comment made (F2/F3).

**How the two visible instances were bounded**, since this app is driven on the operator's own desktop: the identity rig is opt-in (`ABOUT_PANEL_VISIBLE_CAPTURE=1`), one instance per run with launch, invocation, capture and reaping in the same command, `inactive` rather than `normal` (the window is shown without activating the app), `screencapture -l <windowid>` so only a named window is photographed and never the screen, and both rigs reap by pid until the count is zero and print that count - `== processes left from this run: 0` in every run quoted above. Nothing here needs System Events or Accessibility permission: the frontmost fact comes from `lsappinfo`, by pid.

### Checks run on this head

| check | result |
| --- | --- |
| `pnpm check-types` | clean (main + renderer) |
| `pnpm lint` | clean for this change (67 pre-existing warnings elsewhere, unchanged) |
| `pnpm lint:scripts` | `1 changed file(s) under scripts/ are lint-clean against origin/main (2de7ae9)` |
| `pnpm test:desktop` | `2620` tests, `2617` pass, `2` fail - both pre-existing, below |
| `pnpm build` | succeeds; the frames and both rigs above drive that build |
| `pnpm check-types` + `node --test scripts/about-panel.test.mjs` after the rebase onto `f19827852` | clean, 6/6 |
| gate: `stampFailures` / `provenanceFailures` | clean (pins read back out of the manifest) |
| round 1: `pnpm check-types`, `node --test scripts/about-panel.test.mjs` (7/7), gate pins re-derived | clean on `f341feef2` (`srcTree` `4cd08092c`, `scriptsTree` `ed800da6a`) |
| `node --test scripts/about-panel.test.mjs` | 6 of 6 pass |

**The full `pnpm test:desktop` was not re-run for the rebase** - this is said rather than implied by the table above, whose numbers are from the previous base. The rebase's only conflicts were `package.json`'s test list and this manifest, both resolved as the union of the two sides (`main`'s list plus this PR's one test entry; both declared evidence sets kept), and the rebase was verified with `pnpm check-types`, the focused test, and the evidence gate's own stamp/provenance checks. The evidence pins in `docs/evidence/manifest.json` are re-derived for this head in the sync's re-stamp commit: `head` `40f5095ea`, `srcTree` `4cd08092c`, `scriptsTree` `ed800da6a`.

The two failures are not this change, and are proven so rather than asserted: each was reproduced on **unmodified** `origin/main` in a scratch worktree with the same command -

```
$ git worktree add --detach /tmp/about-main origin/main
$ cd /tmp/about-main && node scripts/run-desktop-tests.mjs scripts/python-bytecode-cache.test.mjs
✖ every backend spawn carries the prefix even with the shell-env load unresolved
  AssertionError: global install: start() should report a started backend
$ node scripts/run-desktop-tests.mjs scripts/backend-settings-tiers.test.mjs
✖ the fixtures describe the registry the backend actually ships
  AssertionError: backend-settings-registry.json does not describe the registry
  /Users/damian/.local/share/uv/tools/local-operator/bin/python serves
```

Both are facts about this machine's installed backend (`local-operator-ui` in `~/.local/share/uv/tools`) rather than about the diff: the registry fixture is derived from whatever the installed backend serves, and the bytecode case needs a backend start that this host's install does not give it. Both fail identically at this head and at `origin/main`.

## Known residual, stated rather than implied

- **The panel's icon is still Electron's in an unpackaged run, and this change could not reach it** - the design round's D2, rejected with a measurement rather than an argument. The route the finding named is `app.dock.setIcon`, which sets `NSApplication.applicationIconImage`. With that call in place, logged (drawing the app's own `build/icon-1024x1024.png`, the file verified as the app's own art), the panel still drew Electron's atom: the icon region of the frame is pixel-identical to the same head WITHOUT the call (RMSE 174/65535 over 170x170px, antialiasing noise, with the frames agreeing everywhere else). So on this host an unpackaged process cannot give the panel an icon - it is drawn from the launching bundle - and the call is not in this change, because an inert call that reads like a fix is worse than the paragraph that says so. Not measured: whether the Dock TILE would move, which needs a visible run and is not what the finding was about.
- **A packaged build's panel is unevidenced.** Both frames are unpackaged runs. A packaged build carries the app's own name, version, icon and copyright in its bundle, so its panel reads them from there; the registration is deliberately the same for it (`app.isPackaged` is not consulted, and for a packaged run those strings agree with the bundle). It is not photographed because `electron-builder` needs several GB and this machine's disk is at 99% - so that claim is an argument from where the values come from, not a picture. The set's README says the same in its own words.
- **The copyright line's year is `© 2025`**, straight from `package.json`'s `build.copyright`, and it will read that in 2026 until the field is bumped. Deferred: it is the project's own string, so bumping it is a decision about the project's copyright notice rather than about this panel.

## Review status

Both agent rounds are posted and terminal (approved, no blocker or major): code review and design review, each with the findings disposed of in its remediation comment -
`### Agent review remediation — round 1` and `### Design review remediation — round 1`.
Round 1 changed code (the menu install's position, plus a comment) and the evidence (both frames re-shot, README and manifest updated), so the approved head no longer matches; the re-review runs against `9d26b2205`.
Both streams now stand at round 1 + round 2, code and design, each terminal with no blocker or major: `### Agent review remediation — round 2` and `### Design review remediation — round 2` dispose of every round-2 finding, whose residuals are prose-only and are deferred to the follow-up docs-correction commit rather than to this PR. No reviewers requested and no `@`-mentions, per this repository's standing rule for `damianvtran/local-operator-ui`, and no merge.


damianvtran:
### Agent review — round 1
- Reviewer: reviewer subagent (deepseek/deepseek-flash)
- Scope: f19827852..56a435eeb
- Mode: Comprehensive
- Delivery contract: PR #341 — an unpackaged agent-driven launch must show the app's own About identity, and headless launches must raise no panel at all, with the behaviour bound by a test and the evidence recorded
- Applicable risk surfaces: launch-mode policy (headless suppression), main-process menu construction and About wiring, evidence provenance and manifest pins, test coverage of the applied decision, rig hygiene of the two capture rigs
- Verdict: Approved — no blocker and no major; this round is **TERMINAL** (the three minors and two nits below are follow-ups, not a reason for a round 2)

#### What I verified independently

| check | command / read | result |
| --- | --- | --- |
| decision is total for every documented mode | `src/main/window-mode.ts:742-746`, `WINDOW_MODES` at `window-mode.ts:43` | `headless` -> `suppress`, `inactive`/`normal` -> `show`; `WindowMode` is exactly those three, so the ternary is total |
| the gate cannot be bypassed | `grep` over `src/`: `showAboutPanel(` occurs once repo-wide (`index.ts:527`), `role: "about"` occurs nowhere in `src/`, `Menu.setApplicationMenu` once (`index.ts:544`), no `app.dock.setMenu`, and no renderer/preload/IPC About surface | one call site, behind the branch |
| headless actually reaches menu construction | `index.ts:1529` (`whenReady().then`) through `index.ts:2413-2414`; the only `return`s between are preceded by `app.quit()` (2342/2385/2406) and the chain's `.catch` also quits (`index.ts:2775`) | no surviving startup path skips the gated menu |
| the gate reads the resolved mode, not the environment | `index.ts:521` vs `windowLaunch` at `index.ts:152` (module load) | correct |
| platform guard / unreadable manifest | `index.ts:345` `if (process.platform !== "darwin") return;`; `bundledCopyright()` `index.ts:310-330` catches and omits `copyright` | as described in the PR body |
| test is behaviour, not a restatement | `node --test scripts/about-panel.test.mjs` at the head worktree -> **6/6 pass, 0 fail** | pass |
| test would catch removal of the branch | mutation probe in `/tmp/about-probe` (head tree with `src/main` symlinked per file, `index.ts` copied and its suppression branch deleted): **4 pass / 2 fail** — `the menu item is gated by the launch's RESOLVED mode…` and `the panel is raised from exactly one place, and only past the suppression branch` | real guard |
| types | `pnpm check-types:main` -> exit 0, no output | clean |
| evidence gate on this head | `node --test scripts/evidence-manifest.test.mjs` -> **34/34 pass**, including `the SHIPPED manifest's stamps describe the tree it ships in` and `the SHIPPED manifest's head citations lie in the history it ships in` | pass |
| pin arithmetic | manifest `head` `40f5095ea` = HEAD's parent; `git rev-parse 40f5095ea:src` = `897640e4e5b46ca140e0c82c9460c83eaee65de6` = manifest `srcTree`, and `56a435eeb:src` is the same object; `scriptsTree` `115537b96…` likewise; `git diff --stat` of the re-stamp commit is `docs/evidence/*` only | pins are exact and internally consistent |
| the frames say what the PR says | read both PNGs: before = `Electron` / `Version 44.3.0 (44.3.0)`; after = `Local Operator` / `Version 0.26.11 (Electron 44.3.0, unpackaged)` / `Copyright © 2025 Radient Inc.`, Electron icon | matches, residual icon as disclosed |
| frame geometry matches the quoted census | PNGs are 568x382 and 568x340 = 2x the census bounds the README quotes (284x191 / 284x170) | consistent |
| the headless half is an absence with a wall behind it | `harness/run.sh` fails if a new on-screen window id appears **or** if no `[about-panel]` line is written; `harness/drive.mjs` invokes the item's own handler and refuses a role; `lib.sh` reaps by `pgrep -f` on the run's own scratch profile | real, and safely scoped on a shared machine |

#### Findings

| --- | --- | --- | --- |
| F1 | minor | launch-mode policy — "headless launches must raise no panel **at all**" is not yet true for the whole life of the process | The gated menu is installed at `src/main/index.ts:2413-2414`, at the end of the ready handler, **after** the backend discovery/start/install awaits that begin at `index.ts:1529`. Until then the application menu is Electron's own default, which I measured on this machine with a single self-exiting Electron launch (no window, reaped, 0 processes left): `label="About Electron" role=about type=normal` is the first item of the app submenu. So during that window a headless launch still has an un-gated About item whose click is AppKit's own action. Exposure is narrow — in `headless`/`inactive` the app is never active (its own evidence: 0 of 33 and 0 of 31 samples), so that menu bar is never displayed and the only route in is an in-process invocation over the inspector, which `drive.mjs` route=action refuses on a role — but it is exactly the class of caller this PR is about, and the window can be minutes on a boot that installs a backend. Required action: either move `configureAboutPanel(); createApplicationMenu();` to the top of the ready handler (before the backend awaits — nothing in that template depends on them), or state this boundary in `docs/evidence/about-panel/README.md` so the absence claim is scoped to post-menu construction. |
| F2 | minor | evidence provenance — the pinned trees are not the trees the pinned pixels came from | `docs/evidence/about-panel/README.md` names the `after` frame's own tree: this change's code at `834b3cdda`, `src` tree `5a5bbcd80`. `git merge-base --is-ancestor 834b3cdda HEAD` is **false** — the rebase rewrote that commit out of history (78 commits between it and this head) — while the manifest's `about-panel` entry says "the `after` frame from this branch's head" and sits under the re-stamped `srcTree` `897640e4e`. The gate cannot catch this and cannot be satisfied by the usual field either: `provenanceFailures` only holds `supplementary[].capturedAtHead` to reachability (`scripts/check-evidence.mjs:900-908`), and a rewritten commit resolves to no ref. Required action: word the entry's `source` the way the README does — name the pre-rebase build the frame came from and say the `srcTree`/`scriptsTree` pins are the re-stamped trees the pixels ship in — or re-shoot at this head so the pins are true for the frame as well. |
| F3 | minor | evidence provenance — the identity frame's caption is stale against this head | `after-identity.png` reads `Version 0.26.11`; this head's `package.json` reads `0.27.0`. Judged, not assumed, as asked: **acceptable for merge.** The frame's claim is the name, the parenthesised build string and the copyright, all of which come from the registration this PR adds, and the version line is `app.getVersion()` at runtime; I checked that `productName` (`Local Operator`) and `build.copyright` (`Copyright © 2025 Radient Inc.`) are byte-identical at `834b3cdda` and at this head, so nothing the frame evidences moved. It is disclosed in the README, in the manifest `headNote`, and in the PR body. Required action: none for merge; re-shoot the pair at this head when it is next cheap (F2's fix by re-shoot covers this too) and record it as the follow-up if not. |
| N1 | nit | the absence proof's `README` does not state how the action was invoked | `harness/drive.mjs:22-45` documents that `run.sh` reaches the item **in the app's own main process over `--inspect`** and calls `entry.item.click()`, not a native menu selection; the README's *What was measured* section quotes the run output without that sentence, and a reader who only reads the README could take "invoked" for a menu click. Required action: one sentence in that section naming the route. |
| N2 | nit | user-visible copy on the shipped surface is un-evidenced | `index.ts:354` sets `version:` for every mode, so the packaged app's panel changes from the bare repeat of its version to `Version 0.27.0 (Electron 44.3.0)` — a runtime build string on a user-facing surface. Both frames are unpackaged runs, so nothing here shows the shipped rendering. Required action: let the outstanding design round confirm the packaged copy is the intended one, and say so on the PR. |

#### Follow-up observations (outside this diff's scope)

- `label: \`About ${app.name}\`` (`index.ts:515`) replaces the role's label, which macOS localised; the rest of this template is also hardcoded English, so the change is consistent with the file rather than a regression — worth knowing when the menu is ever localised.
- The suppression line is written with `LogFileType.BACKEND`, i.e. to the app's log file; a rig that only watches stdout will not see it. The harness reads the log file, so the evidence is sound — just note the dependency for future rigs.

#### What I ran, and what I did not

Ran: `node --test scripts/about-panel.test.mjs` (6/6) and the deletion probe for it; `pnpm check-types:main` (clean); `node --test scripts/evidence-manifest.test.mjs` (34/34); the pin arithmetic against `git rev-parse`; the one-launch, no-window, self-exiting Electron default-menu probe (reaped, 0 processes left); `pnpm check-types:main`. Not run, and not implied: the two evidence rigs themselves (no build exists at this head, and `pnpm build` is not a cheap read-only check on this 99%-full disk), the full `pnpm test:desktop`, `pnpm lint`, and any rendered capture (none was needed — no frame was required for any finding above).

Verdict: **Approved / clean.** The decision is correct and total, the single call site cannot be routed around, the gated menu is reached on every surviving startup path, the test genuinely fails when the branch is removed, and the evidence pins check out arithmetically against this head. F1-F3 and N1-N2 are follow-ups for the author to land as they see fit; none of them blocks the merge, and no further review round is required.

damianvtran:
### Design review — round 1
- Reviewer: designer subagent (deepseek-flash)
- Scope: f19827852..56a435eeb
- Surfaces: macOS About panel (unpackaged/dev run) + its window-mode gating
- Verdict: Approved — round TERMINAL (no blocker, no major; D1–D4 are follow-ups)

| ID | Severity | Surface | Finding and what is needed |
| --- | --- | --- | --- |
| D1 | minor | `docs/evidence/about-panel/frames/` and the PR body's inline pair | The pair is not of the trees under review. `before` was shot at `origin/main` `6f15f71ea` — 76 commits behind the stated base `f19827852` — and `after` at `834b3cdda`, whose panel reads `Version 0.26.11` while this head's reads `0.27.0`. I re-shot the `after` half at `56a435eeb` with this set's own rig, and the head renders `Local Operator` / `Version 0.27.0 (Electron 44.3.0, unpackaged)` / `Copyright © 2025 Radient Inc.`; diffing the two PNGs, the only pixels that differ are the version line's band (4,526 px, device rows 278–297 of 382) — the icon, the name and the copyright line are pixel-identical. So the frames are substantively right and captioned with a version they no longer produce. Needed: re-shoot both halves at the base and at the head (`identity.sh`, one command per tree; it prints its own reap count), replace the files, and re-point this PR body's two inline image URLs at the new head so the "caption and head disagree" note can go. The manifest entry's `source` sentence — "the `after` frame from this branch's head" — reads as fresh today and should move with them: the pins are HEAD-derived by the gate, but the sentence is prose about a tree the frame did not come from. |
| D2 | minor | the panel's icon, unpackaged run | The panel now carries the app's name, version and copyright, but the largest thing in it — the icon above the name — is still Electron's atom. Compared, not guessed: `node_modules/electron/dist/Electron.app/Contents/Resources/electron.icns` for Electron 44.3.0 is the same three-orbit atom, and the app's own `build/icon-256x256.png` is the person/node glyph; in the frame pair the two icons are pixel-identical, because the change cannot reach them. A reader whose complaint was an "Electron-branded About panel" still sees Electron's logo with `Local Operator` under it. This is fixable, and small: Electron implements `app.dock.setIcon()` as `[[AtomApplication sharedApplication] setApplicationIconImage:]` (`shell/browser/browser_mac.mm`), and AppKit's standard About panel draws `NSApplication.applicationIconImage` by default — so setting the app's icon for an unpackaged run moves the panel's icon and the Dock tile together, and the two would then tell one story. Gate it to the modes that may show the panel (`headless` deliberately has no Dock tile). Read from Electron's source, not measured on this host: the panel can only be photographed while it is front, so I did not test it. If the residual is instead accepted, it deserves a ticket rather than a README paragraph — it is the last piece of the runtime's identity left in the frame. |
| D3 | nit | `resolveAboutPanelAction`'s doc comment | The comment quotes the measured panel as `284x170`. That is the pre-change panel — the `before` frame is 568x340 at 2x. The panel this change produces measures `284x191` (568x382 in the `after` frame, and `{"w": 284, "h": 191}` in the census of my own head run), because the copyright line adds 21pt. A reader checking the reasoning against the frame pair finds the numbers disagreeing; quote the panel the code actually produces. |
| D4 | nit | `docs/evidence/about-panel/README.md`, "How to re-derive" | The section's first step is `pnpm build`, which fails in a tree with no `.env`: `scripts/vite-plugins/replace-backend-config.ts` throws `Error: VITE_GOOGLE_CLIENT_ID is not set` (also needs `VITE_GOOGLE_CLIENT_SECRET`, `VITE_MICROSOFT_CLIENT_ID`, `VITE_MICROSOFT_TENANT_ID`). I hit it, and a tree with placeholder values instead makes the app close its own window ~2.5s after first present, before the driver's readiness gate — so the set cannot presently be re-derived from its own instructions. Needed: one line naming the required variables. |

#### Why the verdict is what it is

**1. The copy.** `Version 0.27.0 (Electron 44.3.0, unpackaged)` is the right thing to put in front of a developer, and I would keep the string as it is. The parenthesised half is the OS's build slot, and the two facts it carries are the two that place the run in front of you: which desktop host is drawing the panel, and that this is not a release bundle. `unpackaged` is build vocabulary rather than product vocabulary, but the alternative is a panel that names the app and then hides why the icon above it is Electron's — which reads as a defect instead of as a dev run. It is also the honest word: it is the same fact `app.isPackaged` states, and it explains the icon without needing the README.

The one thing to record is that the same line ships: a packaged run reads `Version 0.27.0 (Electron 44.3.0)`. That is a change to the shipping surface, and it is acceptable here because this repo's mac build config sets no `bundleVersion`/`buildVersion`, so the slot previously held the version again — `Version 0.27.0 (0.27.0)` — and nothing is lost. If a real build number is ever added, this line will mask it; that is the moment to revisit, not now.

**2. The icon.** See D2 — the one element of the old identity still in the frame.

**3. The mode policy.** Suppressing under `headless` and showing under `inactive`/`normal` is the right shape, and better than the alternative you name. The mode already answers exactly the question the panel raises — may this run put a window on the operator's screen? — so reusing it means the panel cannot disagree with the window beside it, and it needs no second policy to be kept in step. The alternative of always constructing the panel but never showing it is not available: AppKit's panel only exists while it is ordered front, which this set's own README says. Two further points in its favour: the About item stays in the menu and no-ops with a log line, which is better than hiding it per mode (a menu that differs by mode is worse, and nothing can reach it in headless anyway); and `inactive` showing the panel is "allowed by the mode", not "invisible" — the set's census has it `onscreen: true` at 284x191, my own head run measured the same 284x191 with the app frontmost in 0 of 35 samples, and the README says it in those terms rather than overclaiming. That is the mode keeping its promise.

#### What I looked at

- Both committed frames, as pixels, and the head's own panel that I captured: the icon/name/copyright are identical between the committed `after` frame and a fresh capture at `56a435eeb`, and the only difference is the version digits.
- The identity assets themselves: the app's `build/icon-256x256.png`, and Electron's `electron.icns` rendered from the local Electron 44.3.0 install.
- This PR body's inline images: fetched both raw URLs (HTTP 200) and hashed what came back against the branch's own files — identical (`df9228b3…`, `bb3921ba…`), so the body shows the frames the branch carries; no broken or drifting image.
- `node --test scripts/about-panel.test.mjs` at head: 6/6 pass. It asserts the decision and the shape of the code that applies it; it renders nothing, which is why the frames carry the part it cannot.
- The set's own rigs against my own build of the head (`git archive 56a435eeb` → `pnpm build`, Electron 44.3.0), one run per tree, nothing left behind: `harness/run.sh` (headless, nothing on screen) — census empty before and after the action, the app's own `[about-panel] suppressed by window mode headless: a headless run raises no window`, frontmost 0 of 28 samples, `processes left from this run: 0`; `harness/identity.sh` (`inactive`, the one capture) — identity `name=Local Operator version=0.27.0 electron=44.3.0 packaged=false`, the About item `[role=none] click=function`, the panel `{"w": 284, "h": 191}` `onscreen: true`, frontmost 0 of 35 samples, `processes left from this run: 0`. Both were single self-contained commands that launched, acted, captured and reaped before returning; I checked by pid afterwards that neither left a process.
- What I did not run: the full `pnpm test:desktop` and `pnpm check-types` — focused runs only, per this machine's load and the author's own note that the rebase was not re-run through the suite.

#### Follow-up observations

- `Copyright © 2025 Radient Inc.` on a panel read in 2026. Not this diff's: the base and the head both read that string out of `package.json`, and the packaged bundle stamps the same one, so the panel is consistent with the shipped app. But the About panel is where a stale year is most visible, and the fix belongs in the project's single statement of it rather than here.
- The panel renders as a titled window with the standard window controls in their inactive state — the same in both frames, so nothing this change introduces. Noted only because it means an `inactive` panel looks dismissible rather than like a slab; I did not verify the dismissal behaviour, which would need the panel front and the app activated, the very thing the mode exists to avoid.

damianvtran:

Scope: the round-1 code review of this PR, remediated in `f341feef2` (the fix commit is `729cf2c39`; the branch has since been folded onto `main`, and `cd3149002` is the head that carries the re-stamp). Every finding is disposed of below.

- **F1 (minor) — the menu install ran after the backend awaits. Fixed in `f341feef2`.** `configureAboutPanel(); createApplicationMenu();` now runs immediately after the smoke-test branch in the ready handler, above the launcher watch and every `await`, so no window exists in which Electron's default menu (first item `About Electron`, `role: about`, no handler of ours) is reachable. The ordering constraint the finding allowed for turned out not to exist: the template reads `app.name`, `process.env.ELECTRON_RENDERER_URL` and the module-level `mainWindow` (through handlers that check it at click time), all of which exist at that point — the code comment says so, and `scripts/about-panel.test.mjs` now pins it: the install's offset in the ready handler must precede the first `await` in it, searched over comment-stripped source so the prose above the call is not what gets measured. It sits after the smoke-test branch on purpose (that path exits the process and wants no menu) and the comment states that too.
- **F2 (minor) — the manifest's `source` sentence named a commit the rebase rewrote. Fixed in `f341feef2`.** The sentence no longer says "from this branch's head": it names the `before` tree (a build at the base `f19827852`, whose menu item is still Electron's `about` role, which is why that run raises the panel through `app.showAboutPanel()`) and describes the `after` frame as this branch's own code, the tree whose `srcTree` the manifest pins. The README says the same in its own words, including why that base is the one the frame is evidence about.
- **F3 (minor) — the `0.26.11` caption against a `0.27.0` head. Fixed by the re-shoot in `f341feef2`.** The `after` frame was re-taken at the remediation head, so the pair is `before` `Electron / Version 44.3.0 (44.3.0)` against `after` `Local Operator / Version 0.27.0 (Electron 44.3.0, unpackaged) / Copyright (c) 2025 Radient Inc.`. Cross-checked against the capture the design round left at `/tmp/pr341-head-after-identity.png` (sha256 `0d2a218c…`): same 568x382 panel, same text, and the icon still Electron's — which is design finding D2 below, rejected with a measurement rather than left as an assumption.
- **N1 (nit) — the README must say the invocation is not a menu click. Fixed in `f341feef2`.** The README now states that `harness/drive.mjs` calls the About ITEM's own handler in the app's main process over the Node inspector the rig launches the app with, and why a real menu click is not available to a rig on macOS: a menu belongs to the active app, so driving one means activating the app, which is the interruption these rigs exist to remove.

Two things this round did NOT do, stated so the re-review can check them rather than infer them:

- **`pnpm test:desktop` was not re-run** for the remediation round or for the fold. What changed is the menu install's position, its contract test, one comment, the evidence frames, `package.json`'s list and the manifest's stamps; CI runs the full suite on this head, and that is the run to read. `pnpm check-types` is clean and `node --test scripts/about-panel.test.mjs` is 7/7 on the folded tree.
- **The two re-stamp commits this branch carried before the fold were dropped rather than replayed** (their pins named a base the fold replaced), so the fold's stamps live in one commit, `cd3149002`, with the `headNote` in `manifest.json` saying what moved and what did not.

damianvtran:

Scope: the round-1 design review of this PR, remediated in `f341feef2` (the fix commit is `729cf2c39`; the branch has since been folded onto `main`, and `cd3149002` is the head that carries the re-stamp). Every finding is disposed of below, including the one that is rejected — with the measurement, not an argument.

- **D1 (minor) — the pair was not of the trees under review. Fixed in `f341feef2`.** Both halves are re-shot, one build tree at a time, base first and deleted the moment its frame was taken: `frames/before-identity.png` from a tree BUILT at the base `f19827852` (the last base whose About path is still Electron's `about` role — the fact the frame is evidence about), and `frames/after-identity.png` from this branch's own code at the remediation head. The PNGs are replaced, `docs/evidence/about-panel/README.md` and the manifest's `source` sentence are updated, and the PR body's two inline image URLs are re-pointed at the commit that carries the frames (`f341feef2`; both return 200). Each frame is one opt-in instance, `inactive` mode, window-only `screencapture -l`, reaped by exact pid before the command returned.
- **D2 (minor) — the panel's icon. Rejected: measured, it cannot work from an unpackaged tree, and I did not ship an inert call that reads like a fix.** The route named was `app.dock.setIcon`, which sets `NSApplication.applicationIconImage`. With that call in place — guarded exactly as the finding asked (darwin, `app.isPackaged` false, a candidate that is readable), logged in the run's own output as drawing the app's own `build/icon-1024x1024.png`, and the file verified as the app's own art — the panel still drew Electron's atom. Measured against the frame the design round itself left at `/tmp/pr341-head-after-identity.png` (the same head WITHOUT the call): the icon region is pixel-identical, RMSE 174/65535 over 170x170 px — antialiasing noise — with the whole frames agreeing everywhere else (the only text difference is the version line, as the round reported). So on this host the panel is drawn from the launching bundle, which for an unpackaged run is Electron.app. The call is not in this change, and the set's README keeps its residual paragraph with this measurement added. Not measured, and said so there: whether the Dock TILE moves (that needs a visible run and is not what the finding was about).
- **D3 (nit) — the doc comment's figure. Fixed in `f341feef2`.** `resolveAboutPanelAction`'s comment now quotes `284x191`, the panel this change makes (the copyright line's 21pt), and the frame's own census line carries the same bounds.
- **D4 (nit) — "How to re-derive" began with a build that throws. Fixed in `f341feef2`, and the failure was reproduced rather than assumed.** A plain `pnpm build` in a tree with no `.env` exited 1 before compilation with `Error: VITE_GOOGLE_CLIENT_ID is not set`, which is `scripts/vite-plugins/replace-backend-config.ts` refusing all four variables when any is missing or empty. The recipe now names all four as inert placeholders (no credential created or read, no sign-in claimed) and cites `docs/evidence/settings-app-updates-section/RECOVERY.md`, which is where the workaround is written down. With the four exported, the build exits 0 — that is how both frames were built.
- **N2 (document, do not re-shoot) — the packaged rendering. Documented in `f341feef2`.** The README's residual section and the PR body now say that both frames are unpackaged runs, what a packaged panel reads (its bundle's name, version, icon and copyright; the registration is the same for it, since `app.isPackaged` is not consulted and those strings agree with the bundle), and why it is not photographed — a packaged build needs several GB and this machine's disk is at 99% — so the claim is an argument from where the values come from, not a picture.
- **(non-gating, deferred, not ticketed)** the stale `(c) 2025` in the copyright line: left alone, with a line in the residual section saying it comes from `package.json`'s `build.copyright` and that bumping it is a decision about the project's notice rather than about this panel.

damianvtran:
### Design review — round 2
- **Reviewer:** designer subagent (deepseek-flash)
- **Scope:** `0b0e3d0326c1d01ac8d1bee7fbf09416c12854da`..`9d26b2205`
- **Delta reviewed:** `56a435eeb..9d26b2205`
- **Surfaces:** the re-shot before/after About-panel frames read as pixels, the panel's residual Electron mark, and the two prose surfaces that carry the same claims — the evidence README and this PR's body
- **Judged from:** the committed frames at `docs/evidence/about-panel/frames/` plus the round-1 capture left at `/tmp/pr341-head-after-identity.png`, decoded and compared as arrays. **No app was launched for this round** — the panel's pixels were already committed, the frames are the artifact under review, and the box is at 98–99% disk, so there was nothing a second capture would add and nothing was left running.
- **Verdict: Approved — round TERMINAL** (no BLOCKER, no MAJOR). D5 and D6 are MINOR prose follow-ups; D7 is a NIT.

| ID | Severity | Surface | Finding and what is needed |
| --- | --- | --- | --- |
| D5 | minor | `docs/evidence/manifest.json`, the `about-panel` entry's `source` sentence, against the README in the same set | The two artifacts at this head state opposite facts about where the `after` frame came from. The manifest says "the `after` frame from this branch's own code — **the tree whose `srcTree` this manifest pins**"; the README says "**The tree that frame was captured from is not the tree this branch's head carries**, because `main` moved twice underneath it". The sentence was made true by the round-1 re-shoot and made false by `9d26b2205`, which re-derived the pins and re-worded the README without it. Neither sentence names the capture tree, so after this branch's reflog rotates the claim is not re-derivable by anyone. Needed: one clause in each, naming it — "the `after` frame from this change's own code as it stood at `2d6665924` (pre-fold remediation head)". That is the tree, and it is checkable: it carries exactly this head's two frame blobs (`75f322d8`, `38cf4a21`). No frame changes. |
| D6 | minor | the PR body's evidence section | Three statements still describe the pre-remediation capture, and the body is the surface a developer reads first. (1) The `after` image's alt text reads `Version 0.26.11` while the table directly above it reads `0.27.0` and the pixels read `0.27.0` — I diffed the committed pair across the re-shoot: the version band is what moved (4,845 px, 2.23% of the frame), everything else in the panel is unchanged. (2) The quoted identity run prints `version=0.26.11`; the README at this head prints `version=0.27.0` from the same command. (3) The `before` frame is attributed to `origin/main` `6f15f71ea`; the README says `f19827852`. The embedded URLs are **correct**: both `raw.githubusercontent.com/.../f341feef2/...` URLs return 200 and are byte-identical to the frames at this head (sha256 `2297f024…` before, `e786e58b…` after). So this is captions only. Needed: re-stamp the body with the README's numbers. Body-only; no commit, no head-SHA movement. |
| D7 | nit | the panel's version line, as rendered | The build string is the one variable-length line in a panel whose width is not content-driven: both panels are exactly **284pt wide** although the widest line went from 220px to 431px, and the version line's ink now runs **431 of 568px (67px / 70px of slack, ≈33.5 / 35pt)**. That slack is the whole budget a longer host string spends — a three-digit Electron major, or a word longer than `unpackaged`. No change needed while the string stays; one sentence in the README's build-string bullet recording the measured slack would mean a future longer string is not first discovered by clipping. |

#### What the round verifies, by measurement

**D1 — fixed, and the pair is now of the trees under review.** Read as pixels (device px at 2x; the panel is 568px wide in every frame):

| frame | panel | icon block | name line | version line | copyright line |
| --- | --- | --- | --- | --- | --- |
| `before-identity.png` (base `f19827852`) | 284×170pt (568×340) | 104×104px at rows 92–195, cols 232–335 | `Electron`, ink 112px | `Version 44.3.0 (44.3.0)`, ink 220px | — |
| `after-identity.png` (this change) | 284×191pt (568×382) | same bbox, same size | `Local Operator`, ink 202px | `Version 0.27.0 (Electron 44.3.0, unpackaged)`, ink 431px | `Copyright © 2025 Radient Inc.`, ink 293px |

The before/after claim in the table holds exactly, and the strings are the right strings: `productName`, `version` and `build.copyright` at this head are `Local Operator` / `0.27.0` / `Copyright © 2025 Radient Inc.` — the panel is printing the project's own values, not a runtime's. The panel grows downward only (icon bbox identical in all four frames I compared), every line is centred to within 1px (name 184/182, version 67/70, copyright 137/138), and the frame is the panel window alone — no screen content, no other window. The base-side claim is checkable too: `f19827852` has no `resolveAboutPanelAction` at all, so Electron's own About path is what that tree runs and what the frame shows.

**D3 — fixed.** The `resolveAboutPanelAction` comment now quotes `284x191`, which is what the frame measures (568×382) and what the set's own census line says (`{"w": 284, "h": 191}`).

**D4 — fixed.** The re-derive recipe names all four `VITE_*` variables with inert placeholders, and cites `RECOVERY.md`.

**D2 — rejection accepted, and it is stronger than the one call it rests on.** The proposal was mine, so I judged the rejection rather than the route: **the measurement is sound and I would not overturn it.** I calibrated its discriminating power rather than taking "pixel-identical" on trust, because that phrase is only meaningful against a noise floor:

| comparison, in the icon's 104×104 box | RMSE | max \|Δ\| | pixels >2/255 |
| --- | --- | --- | --- |
| two shots of the *same* rendering (`/tmp/pr341-head-after-identity.png` vs the committed frame, no `setIcon` in either) | 63/65535 (0.24/255) | 2/255 | 0 (0.0%) |
| the box as it is (Electron's atom) vs the same box holding the app's own art (`build/icon-1024x1024.png`) | **39,185/65535** (152.5/255) | 255/255 | 10,488 (97.0%) |

The reported figure — 174/65535 over 170×170 — sits in the first row's regime, 220× below the second. A route that moved the mark would have produced a difference of tens of thousands of 65535ths, not hundreds. What would overturn it: an icon-region difference of the second row's order (thousands/65535, or >5/255 RMSE, or percent-level pixels above 2/255) between a with-call and a without-call capture on the same head. Nothing short of that is a measurement of icon change rather than of antialiasing.

Independently of that single call, the committed code carries a second, route-free corroboration: in `inactive` mode — the mode that produced the `after` frame — `hideDock` is `false` (`src/main/window-mode.ts:191-198`), so the pre-existing `app.dock.setIcon(image)` at `src/main/index.ts:280` (present on `origin/main` too) already sets `NSApplication.applicationIconImage` to the app's own art, `resources/icon.png`, the person/node glyph I opened to confirm it is not an atom — *before* the panel was raised. The committed frame still shows the atom. So the conclusion does not depend on the author's temporary call being correct about AppKit; the app's own icon was in place through a shipped path and the panel ignored it either way. And the API surface the change already calls has no macOS icon option to fall back on: at Electron 44.3.0, `AboutPanelOptionsOptions.iconPath` in `node_modules/electron/electron.d.ts` is `@platform linux,win32`.

**Residual, judged rather than repeated:** a panel reading `Local Operator / Version 0.27.0 (Electron 44.3.0, unpackaged)` over Electron's atom **is acceptable to ship**, and the README paragraph is the right disposition for it. The failure this PR exists to fix was the app naming itself Electron; that is gone, and what is left is one mark that a packaged build supplies from its own bundle (unevidenced here, and said so). The paragraph is better than the inert call would have been, and better than silence. A ticket only pays if someone intends to chase the mark in a packaged build; for the unpackaged case this measurement closes it, and the Dock tile — the one thing not measured — is already served by the call above for every mode that has a tile.

#### Evidence honesty (the `after` frame's tree)

**Acceptable for a review-level claim; no third capture needed — but it should cite the tree.** I recovered the capture tree rather than accepting the assertion: the branch reflog holds the pre-fold remediation commit `2d6665924` ("fix(menu): install the app's own menu before the ready handler awaits anything"), which carries exactly this head's two frame blobs (`75f322d8` / `38cf4a21`), and between that tree and this head:

- `src/main/window-mode.ts` is **blob-identical** (`141bfbc8de8d`) — so `resolveAboutPanelAction` is identical, not merely similar;
- `configureAboutPanel` (2,062 chars, sha256 `79d215c068a36547`) and `createApplicationMenu` (5,139 chars, sha256 `573f0bca52090a23`) extract **identical** from `src/main/index.ts` on both trees;
- the only `index.ts` difference between the two trees is `main`'s own folded-in work elsewhere in the file (87 lines), none of it in the path the frame photographs, plus the two docs files this head changed.

The three symbols are exactly the ones that fill and raise the panel, so the picture is still a picture of this change. That is enough for a review-level claim, and the author says it in the README — which is the disclosure that makes it honest rather than a footnote discovered later. The one thing missing is the tree's name (D5): a reflog is not a durable record, and "not the head's tree" without a SHA is unfalsifiable the moment it rotates.

#### Non-gating notes (no action, outside this delta)

- **The packaged panel's copy is still unevidenced**, and N2 asked this stream to confirm the packaged string is intended. It is intended and consistent (`app.isPackaged` is not consulted, so a packaged run reads `Version 0.27.0 (Electron 44.3.0)`) — a runtime version on a user-facing surface is a deliberate choice here, and round 1 approved it on that basis; nothing in this delta reopens it. The README's new wording says it is an argument from the values' provenance and not a picture, which is the right level of claim.
- **The copyright string is the project's own** (`package.json` `build.copyright`, `LICENSE`, author name all read "Radient Inc.", so no spelling issue), and the `© 2025` year in 2026 remains the deferred, disclosed residual round 1 accepted.

No BLOCKER and no MAJOR remain; the pixels, the panel's copy and the two rigs' own accounting are consistent. D5 and D6 are prose, need no re-review, and neither touches a pixel, a code path, or a pin the frames rest on — so this round stays terminal if they land as a docs commit and a body edit.

damianvtran:
### Agent review — round 2
- Reviewer: reviewer subagent (deepseek/deepseek-flash)
- Scope: 0b0e3d032..9d26b2205
- Delta reviewed: 56a435eeb..9d26b2205, read as **branch-owned only** — `git diff 0b0e3d032..9d26b2205` is 13 files (`src/main/index.ts`, `src/main/window-mode.ts`, `scripts/about-panel.test.mjs`, `package.json`, `docs/evidence/manifest.json`, the `docs/evidence/about-panel/` set and its harness). Everything else the range carries is `main`'s own lanes absorbed by the two folds onto `10926b782` and `0b0e3d032`, and is not reviewed as this branch's.
- Applicable risk surfaces: launch-mode policy and menu construction in the ready handler (`src/main/index.ts`), the ordering contract in `scripts/about-panel.test.mjs`, and evidence provenance (`docs/evidence/manifest.json` pins/citations, `docs/evidence/about-panel/README.md`, the two re-shot frames), plus the `package.json` `test:desktop` union.
- Verdict: **Approved — clean. No blocker and no major; this round is TERMINAL.** The four minors and one nit below are follow-ups for the author to land whenever they next touch this set, not a reason for a round 3.

#### What I verified independently, not from the remediation comment

| check | what I ran / read | result |
| --- | --- | --- |
| **F1** — the install really is above every `await` | ready handler from `.whenReady().then(` (`index.ts:1530-1531`); `configureAboutPanel(); createApplicationMenu();` at `index.ts:1573-1574`; first `await` in the handler at `index.ts:1827`; the only thing between them is the smoke-test `app.exit(0)` branch and comments | install precedes the first await by ~250 lines, and the smoke-test path still returns before it |
| **F1** — the template needs nothing the awaits produce | mechanical scan: identifiers declared *inside* the ready handler, cross-matched against the body of `configureAboutPanel` + `createApplicationMenu` | **none referenced.** `windowLaunch` (`index.ts:153`) and `mainWindow` (`index.ts:1024`) are module scope, `logger`/`LogFileType` are imported at `index.ts:48`, `isDev` reads the env at call time, and every `mainWindow` read sits inside a click handler |
| **F1** — the new contract test is position-sensitive, not decorative | `node --test scripts/about-panel.test.mjs` at this head → **7/7 pass**; then a mutant in `/tmp` (head tree, the two calls deleted from `index.ts:1573-1574` and re-inserted at the old late site above `// --- Helper to manage main window…`) | mutant: **6 pass / 1 fail** — `the app's own menu is installed before the ready handler awaits anything` fails with *"the menu must be installed before the first await"*, and nothing else moves. The pin would catch the regression F1 names |
| **the fold** — `package.json` union kept both sides | `git diff 0b0e3d032..9d26b2205` + set-difference of the two `test:desktop` lists | added: `scripts/about-panel.test.mjs`; **lost: none**; no other script key changed; `version` is `0.27.0` on both sides; no dependency drift |
| **the fold** — `manifest.json` union | `supplementary` length and entries | main's 78 entries → 79 at this head, the one addition being `about-panel`, no duplicate, no main entry replaced |
| **the fold** — the re-stamp pins are exact and reachable | `git rev-parse 9d26b2205:src` = `4cd08092c8df7e9d20ac54e5a54998914a3a970a` = manifest `srcTree` = `f341feef2:src`; `…:scripts` = `ed800da6af4d45bd13a1b1ff0950815ea7156e6b` = manifest `scriptsTree`; manifest `head` = `f341feef2` = HEAD's parent; before-frame citation `f19827852` is an ancestor of HEAD | arithmetically exact and internally consistent |
| **the fold** — the shipped manifest still passes its own gate | `node --test scripts/evidence-manifest.test.mjs` in a worktree at this head | **34/34 pass**, including `the SHIPPED manifest's stamps describe the tree it ships in` and `…head citations lie in the history it ships in` |
| types | `pnpm run check-types:main` | exit 0, clean |
| **D3** — the corrected figure | comment at `window-mode.ts:730-732` vs the frames | `after-identity.png` is 568x382 = **284x191**; `before-identity.png` is 568x340 = 284x170. The comment now quotes the panel this change makes |
| **the re-shot frames say what the README says** | read both PNGs | `after`: `Local Operator` / `Version 0.27.0 (Electron 44.3.0, unpackaged)` / `Copyright © 2025 Radient Inc.`, under Electron's atom. `before`: `Electron` / `Version 44.3.0 (44.3.0)`, no copyright. Matches the README's table (`README.md:15`) |
| **the byte-identity disclosure** | extracted `configureAboutPanel`, `createApplicationMenu` and `resolveAboutPanelAction` from `56a435eeb`, `729cf2c39`, `f341feef2` and HEAD and hashed each | identical in all four trees (sha256 prefixes `fbd6a6be`/`555358be`/`e97ff83f`). The disclosure holds whichever of those trees the `after` frame came from, so the frame's claim is safe even though its exact capture tree is unpinned (M1) |
| **D2's premise** | `node_modules/electron/electron.d.ts:20651` (`iconPath` `@platform linux,win32`) and Electron's live docs for `setAboutPanelOptions` (`iconPath string (optional) LinuxWindows`) | confirmed: there is genuinely no icon option in `AboutPanelOptions` on darwin, so the only route left was the `applicationIconImage` call the author measured as inert. **The rejection stands on the evidence presented** |
| **N1's fix** | `README.md:90-96` | present: the invocation is the item's own handler, called in the app's main process over `--inspect`, with why a real menu click would activate the app |

#### Findings

| --- | --- | --- | --- |
| M1 | minor | evidence provenance — the manifest's `source` now asserts the `after` frame came from the tree the pins name, and the README in the same head says it did not | `docs/evidence/manifest.json:6296` ends "…and the `after` frame from this branch's own code - the tree whose `srcTree` this manifest pins." `docs/evidence/about-panel/README.md:21-23` says the opposite: "The tree that frame was captured from is not the tree this branch's head carries, because `main` moved twice underneath it." The manifest sentence is the stale half: it was written in `f341feef2`, when `srcTree` was `d8340494b…` — the pre-fold tree the frame actually came from — and the re-stamp commit `9d26b2205` re-pointed `srcTree`/`scriptsTree` at HEAD's trees while the sentence stayed. That is F2's failure mode (a citation the trees no longer support) reintroduced by the fold rather than by the shot, and it is prose, so `citationFailures` cannot see it. Required action: name the capture commit in that sentence the way the `before` half names `f19827852`, or take F2's other wording and say the pins are the re-stamped trees the pixels ship in. |
| M2 | minor | evidence provenance — the identity frame's caption is still stale in the manifest, although the frame itself was re-shot | I read the shipped frame: `after-identity.png` reads `Version 0.27.0 (Electron 44.3.0, unpackaged)`. `docs/evidence/manifest.json:6299` describes that same frame as "`Local Operator`, `Version 0.26.11 (Electron 44.3.0, unpackaged)`" — the README has no `0.26.11` at all, and this is the only occurrence that describes this frame (`main`'s own notes mention `0.26.11` as a release window, which is their business). F3 was half-landed: the frame and the README moved, the declaration beside them did not. Required action: one word in the manifest's `why` so the declaration and the pixel it declares agree. |
| M3 | minor | `window-mode.ts:730-732` cites a datum the evidence shipped in this same commit contradicts | The comment says the window server lists the panel "`284x191` at the sizes these runs use, `onscreen: true`". Every `onscreen` value in the run recorded in `docs/evidence/about-panel` is `false` — including the panel itself (`README.md:118` and `:163`, id 38271 at 284x191) — and the ON-SCREEN line the same commit added to the rig prints `(none)` (`README.md:74`, `:120`, `:165`), because, as `harness/lib.sh:141-149` states, these runs happen with the display asleep. There is no recorded `onscreen: true` anywhere in the set. The `284x191` half is right; this half is the one the reader is invited to check and cannot confirm. Required action: quote the run that showed `true` (the design round's, which is not in this repo), or restate what these runs measured: the panel is ordered front and photographed while the census carries it `onscreen: false`. |
| M4 | minor | the packaged-panel claim (the subject of N2) is not what the code does, on a shipped user-facing surface | `README.md:219-220`: "…so its panel reads them from there and this registration is deliberately untouched for it (`app.isPackaged` is not consulted: the same strings are set either way, and for a packaged run they agree with the bundle)." The registration runs for packaged builds too, and `app.isPackaged` **is** consulted, at `src/main/index.ts:358`: `version: \`Electron ${process.versions.electron}${app.isPackaged ? "" : ", unpackaged"}\``. A packaged run therefore renders `Version 0.27.0 (Electron 44.3.0)` — the runtime build string on the shipped surface that N2 asked to have confirmed as intended — and that string is not the bundle's `CFBundleVersion`. `applicationName`, `applicationVersion` and `copyright` do agree with the bundle. Required action: write the packaged rendering as the code produces it and record that it is intended (N2's ask, which the documentation quietly answered the other way), or gate the build string on `!app.isPackaged` and let the bundle's own answer stand. |
| N1 | nit | the README's capture-tree paragraph has a dangling reference | `README.md:27-28`: "the only difference in those two files between the two trees is that comment below plus `main`'s own work from the folds" — there is no comment below it (the paragraph ends two lines later and no quotation follows), and the difference it half-names is a **code** change: the moved `configureAboutPanel(); createApplicationMenu();` pair, documented by the comment at `index.ts:1554-1572`. Say which comment, and say the move is a move. |

#### Follow-up observations (not this delta's to fix)

- The D2 measurement is uncommitted: the `RMSE 174/65535 over 170x170 px` figure is quoted against a frame at `/tmp/pr341-head-after-identity.png`, so the residual's numbers cannot be re-derived from this repository. The disposition is still the right one — an inert call is not shipped and the residual is stated with its measurement rather than implied — but a future reader has only the sentence.
- The pre-fold commits are gone by design: `git for-each-ref --contains 1816c3845` and `--contains cd3149002` are both **0**, so the pointers to them (the manifest `headNote`'s, which is `main`'s own text kept whole by the union, and the two remediation comments') resolve in no clone. The fields the gate guards (`head`, `partialCapture.addedAtHead`, `partialCapture.refreshedAtHead`) are all reachable, so nothing shipped depends on them.

#### What I ran, and what I did not

Ran: `node --test scripts/about-panel.test.mjs` at this head (**7/7**) and the same suite against the moved-back mutant in `/tmp` (6/7, the F1 case failing); `node --test scripts/evidence-manifest.test.mjs` in a worktree at this head (**34/34**); `pnpm run check-types:main` (clean); the pin arithmetic and citation reachability with `git rev-parse` / `for-each-ref --contains`; function extraction and hashing across four trees; and both committed PNGs read directly.

Not run, and not implied: the two live rigs (they need a built app, and `identity.sh` is the one rig here that shows a window — not something a reviewer should raise on a shared box), `pnpm test:desktop` (not re-run for the fold, as disclosed; CI's Desktop Tests passed on this head, and that is the run to read), `pnpm lint`, and `pnpm check-evidence`'s machine-wide sweep. No Electron instance was launched for this round, and the worktree I read in was removed afterwards.

damianvtran:
- Author: implementing agent (this branch)
- Disposes: code review round 2 — M1, M2, M3, M4, N1
- Disposition: **all five Deferred.** Each is wrong in text only: no behaviour changes, no rendered surface changes, no pin or blob the frames rest on changes, and no test or gate in this repository can see any of them.
- Plan: they ride in the **follow-up docs-correction commit**, not in this PR. Nothing in the list gates the merge.

| ID | Severity | Disposition | Why it does not gate | Where it lands |
| M1 | minor | Deferred | The manifest's `about-panel` `source` sentence and `docs/evidence/about-panel/README.md` say opposite things about which tree the `after` frame came from. Two sentences, no code, no frame, and the pins themselves are already correct and gate-verified (`srcTree` `4cd08092c` = `HEAD:src`, `scriptsTree` `ed800da6a` = `HEAD:scripts`). | The follow-up commit names the capture tree in both sentences — the pre-fold remediation commit `2d6665924`, which carries this head's exact two frame blobs (`38cf4a21`, `75f322d8`), so the claim is checkable rather than asserted. |
| M2 | minor | Deferred | One word in the manifest's `why` for the set (`0.26.11` → `0.27.0`). The frame itself is already correct — it reads `Local Operator / Version 0.27.0 (Electron 44.3.0, unpackaged)` — so this is a declaration lagging the pixel it declares, not a pixel or a pin. | The follow-up commit corrects the declaration. |
| M3 | minor | Deferred | A datum in a doc comment (`src/main/window-mode.ts:730-732`) that the recorded run does not carry: every `onscreen` value in the set's evidence is `false`, the ON-SCREEN line reads `(none)`, and the display-asleep reason is documented in `harness/lib.sh`. The policy the comment justifies — `headless` suppresses, `inactive` and `normal` raise — is unchanged, is asserted for all three modes, and is pinned by `scripts/about-panel.test.mjs`. | The follow-up commit restates the datum as these runs measured it: the panel is ordered front and photographed while the census carries it `onscreen: false`. |
| M4 | minor | Deferred | A README sentence about a surface that no run in this repository reaches (the packaged panel). No packaged build is possible on this machine (disk), no behaviour changes either way, and the strings the code sets are the same for both cases apart from the `, unpackaged` marker. | The follow-up commit states the packaged rendering as the code produces it, `Version 0.27.0 (Electron 44.3.0)`, and records that the string is intended rather than a bundle value. |
| N1 | nit | Deferred | Wording: a dangling "that comment below" in the README's capture-tree paragraph. | The same follow-up commit names the comment (`src/main/index.ts:1554-1572`) and says the difference it half-names is the moved call, not only a comment. |

Nothing here gates the merge. The decision the change makes, the ordering contract that pins it, the registration's fields and both committed frames are untouched by every item above; the round-2 review found no blocker and no major, and this round is terminal.

damianvtran:
### Design review remediation — round 2
- Author: implementing agent (this branch)
- Disposes: design review round 2 — D5, D6, D7
- Disposition: **all three Deferred.** Each is wrong in text only: no rendered surface changes, no pixel changes, no behaviour changes, and the frame pair this set ships is exactly the pair the round read.
- Plan: D5 and D7 ride in the **follow-up docs-correction commit**. D6's captions were re-stamped in the PR body in the same pass as this note — a body-only edit, no commit, so the head did not move and both rounds stay fresh at `9d26b2205`. Nothing in the list gates the merge.

| ID | Severity | Disposition | Why it does not gate | Where it lands |
| D5 | minor | Deferred | The design twin of code M1: a clause naming the capture tree in the manifest's `source` and in the README. Prose only — the frames, their blobs and the pins are unchanged and gate-verified, and the three symbols that fill and raise the panel extract identically from the capture tree and this head. | The follow-up commit names it in both sentences: the pre-fold remediation commit `2d6665924`, whose frame blobs are this head's two (`38cf4a21` `before`, `75f322d8` `after`). |
| D6 | minor | Deferred | The body's captions: the `after` image's alt text read `Version 0.26.11`, the quoted identity run printed `version=0.26.11`, and the `before` half was attributed to `origin/main` `6f15f71ea`. Captions only — the embedded URLs were already correct and were left alone. | Corrected in the PR body now, in the same edit as this note: the alt text and the quoted run carry `0.27.0`, the run block is the one the README records at this head, and `before` is attributed to `origin/main` `f19827852`. Nothing remains for the follow-up. |
| D7 | nit | Deferred | The version line's ink budget: the panel is 284pt wide whatever the string, and the build string's slack (≈33.5 / 35pt) is the whole budget a longer host string would spend. The string is not changed by this PR and nothing clips at the current value. | The follow-up commit adds the one sentence to the README's build-string bullet recording the measured slack, so a future longer string is not first discovered by clipping. |

The residual is unchanged and stays as accepted: an unpackaged run's panel reads `Local Operator / Version 0.27.0 (Electron 44.3.0, unpackaged)` over Electron's atom, the packaged case remains unevidenced and is stated as an argument from provenance rather than a picture, and the README paragraph is the disposition for both. No pixel, no code path and no pin moves with anything above; the round-2 design review found no blocker and no major, and this round is terminal.

damianvtran:
### Released in v0.27.2

**Tag:** `v0.27.2`  •  **Release:** https://github.com/damianvtran/local-operator-ui/releases/tag/v0.27.2

This PR is in the window since `v0.27.1` and the Release is published as a **pre-release** until `publish.yml` attaches the installers and promotes it. The tag is pinned to the bump merge `482cf98ce`; the bump is #344 on `release-next` — one line of `package.json` — and its independent round closed terminal on the head that merged.

**Your PR's line in the notes:** none — this PR's body carries no `Release:` line, so the notes record that rather than inventing a sentence for it.

PR #331, opened 2026-09-17
feat(update): publish a new generation for the app-owned environment, and give the skew notice its action

damianvtran:
## What this changes

The desktop app could not move its **own** managed environment. Two readings off the machine this was written on:

* the published record — `~/Library/Application Support/Local Operator/managed-python/packaged/selected-environment.json` — says `"backendVersion": "0.55.7"`, while the tree it names holds `local_operator-0.56.8.dist-info` and that tree's `bin/` was last written `2026-09-17 04:35`: the environment has been `pip install`-ed over **in place** at least twice since it was prepared, so the only stamp that records "which build is published here" lies;
* `/health` answers `0.56.8` from that same prefix (`pid 15051`), while PyPI publishes `0.56.12`.

Readiness answered *"a usable generation is published"*, the app read that as *"this environment is up to date"*, and the only in-place upgrade reachable ran `pip install --upgrade` inside the **published** venv — the tree the serving daemon and every session runtime import from. That is why a control promising to move the environment produced a byte-identical screen on every press, and why the panel next to it stated a remedy with nothing to press.

Design: `docs/design-unified-update-mechanism.md` §4.4 and §7's PR 3, on branch `docs/unified-update-mechanism`.

### The mechanism — `src/main/backend/managed-python.ts`

* `publishGeneration(...)` is the extraction of the sequence `prepareManagedPython` already ran: reclaim, reuse an identity-matching runtime, install into a **new** pathname, smoke, write the readiness record, flip the pointer by rename, reap. Both entry points call it, so the two cannot come to disagree about **when a pointer may move** — a second copy of that order is the defect this shape prevents.
* `updateManagedPython(options, install, { target, isNewer })` publishes only when what is published is **behind** the release the app means to run, and reports `replaced: false` when there was nothing to do rather than claiming an install it did not perform. The pointer is written **only after the smoke passes**, so a failed update is safe by construction: the new tree is a directory beside the published one and every running process is still on the generation it loaded.
* The superseded generation survives as the rollback, in **both** roots — and `reapSupersededGenerations`'s `keep` now takes a **list**. "The newest generation that is not named" is not reliably the rollback: a failed attempt's half-written tree is newer than it, so a single-path keep let the reaper delete the rollback *and the tree a serving daemon has open* in the same pass that published its successor.
* `managedSelectionSatisfies` is the readiness comparison in one place, and it takes the **app's** comparator (`isNewerVersion`) rather than growing a second ordering. A version that could not be read is never treated as current.
[...]

PR #316, opened 2026-09-17
fix(chat): move the transcript's time stamps onto the turns

damianvtran:
# PR Description

## Summary of Changes

Two operator reports on 2026-09-17, one surface. Both land here because they are the same question —
where does the always-visible time belong? — and the answer is now exactly three carriers, each named
in the DOM by its own `data-stamp`:

| Carrier | Where the stamp sits | Value |
| --- | --- | --- |
| a **user turn** | under its bubble, right-aligned on the bubble's edge | `turn` |
| a **settled agent answer** | under it, left-aligned on the agent rail | `answer` |
| an **open tool call** | at the foot of the disclosure the reader opened | `tool` |

The accessible name is parameterised per carrier (`Sent` / `Answered` / `Ran`), because `Sent` was
hardcoded while every stamp was a turn's own and is simply false for an answer the agent wrote.

### 1. The transcript's footer stamp is removed

It rendered a `TurnTimestamp` as the last line of the transcript, gated so it stayed away when the
last row painted a stamp of its own (a user turn, or a ledger row the reader had opened). Every gate
it had asked *which row* came last — and during a live turn the row that comes last is the aggregate
**working line**, which is not a record at all and not something the gate could name:

> "the time that shows up below messages also seems to be showing up below the thinking indicator,
> make sure that it doesn't, only beside user messages and in tool traces when expanded."

The line is removed rather than gated a third time — there is no stamp at the foot of the transcript
in any state. Removed with it, each checked with a grep for a live consumer before it went:
`lastRowPaintsStamp`, `openStampRowIds`, `handleToolOpenChange`, `lastRecord`, the
`onToolOpenChange` threading through `TranscriptRow`, the transcript-local `ToolRow`'s
`onOpenChange`, and `ToolRow`'s own open-state report in `trace/tool-row.tsx` (three commits of
history, every line of it for this gate; `Disclosure`'s own `onOpenChange` stays — it has a reason of
its own).

**A trade, recorded rather than found later:** a transcript ending on assistant prose, or on a
settled tool row nobody opened, now shows no time at its foot.

### 2. The turn's closing answer gets a caption

> "The agent responses (just the final responses, not the in-progress tool intent/response) don't
> have a time displayed on them — that would be helpful to show too. It should probably be on the
> left under the agent message instead of on the right."

Same `TurnTimestamp`, same caption roles, same 4px of air (the user side takes it from its column's
`gap-1`; this row takes it as a margin, because its parent is the message container rather than a flex
[...]

PR #347, opened 2026-09-18
fix(release): detach the verification mount by device, and never over a live volume

damianvtran:
# PR Description

## Summary of Changes

`pnpm test:desktop` intermittently left a disk image attached, and macOS then told the operator — on their desktop, for days — `Disk Not Ejected Properly — Eject "Local Operator" before disconnecting or turning it off.` This fixes the detach path that leaves the volume attached, and the fixture teardown that deletes the backing `.dmg` underneath it.

Release: patch — the desktop suite stops leaving `Local Operator` images attached, so the "Disk Not Ejected Properly" alert stops appearing after a run.

## Root cause

Read from the code, then reproduced against a real image (evidence below):

- `scripts/managed-python.test.mjs` builds a fixture image with `hdiutil create -volname "Local Operator"` and hands it to `finalContainerChecks(image, { run, checkApp })`.
- `scripts/python-artifact-layout.mjs` attached it with `hdiutil attach -readonly -nobrowse -mountpoint <mount> <image>`, and its `finally` detached the **mount point** with a plain `hdiutil detach <mount>` — no retry, no `-force`. That fails on a busy volume (`Resource busy`, exit 16, measured on this machine) where `-force` succeeds, and the failure path `return`ed early **with the volume still attached**.
- The fixture's own teardown then `rmSync`'d the whole scratch root that holds the `.dmg` — deleting the backing file of a still-attached volume, which is the state that raises that alert. The correct order already existed in the same file (`mountedVolume`, "detach BEFORE it removes the tree").

## The change

- **The device is what gets detached, read from `hdiutil attach`'s own stdout** (`/dev/diskNsM`), not the mount path. A path is only the name that call chose for the volume, and `hdiutil detach <path>` re-resolves it at detach time. `hdiutil` prints the *resolved* mount and the scratch mount sits under a symlinked `tmpdir`, so the literal and the resolved path are both matched before falling back to the last device line it printed. Only the device this call attached is ever detached.
- **A failed detach is retried once with `-force`.** A detach that fails twice is a FAILED RESULT that names the mount *and* the device, and it leaves the scratch tree — and with it the live mount point — in place rather than deleting a tree holding a mounted volume. It never reports success, and never returns quietly, with a volume attached. (The `return` also moved out of `finally`, which is both the lint contract and the correct shape.)
- **An exit/SIGINT/SIGTERM backstop detaches whatever this module still holds.** A `finally` cannot cover a crash or an interrupt, and a volume orphaned by one is the same alert. It is registered lazily (a run that mounts nothing installs nothing), synchronous (`spawnSync`, which is all a process `exit` handler may do), idempotent (the target is cleared before the attempt), and never throws; the signal path re-raises the signal with its own handler removed, so Ctrl-C still ends the run exactly as it did.
- **The test fixture detaches before it removes the tree**, mirroring `mountedVolume`'s order, with a comment saying why the order is load-bearing.

No user-visible surface changes: this is test/CI hygiene, so **no design round applies**.

### 1. The mechanism, on real hardware

A disk image whose file is held open by another process:

```
$ hdiutil detach /dev/disk32s1   # a plain detach of a volume something holds open
hdiutil: couldn't unmount "disk32" - Resource busy
exit=16
$ hdiutil detach -force /dev/disk32s1   # the retry the gate now makes
"disk32" ejected.
exit=0
```

That is the retired code path's failure, produced by `hdiutil` itself, and the retry that repairs it.

### 2. The retired path vs the fix, on a real image

Real `hdiutil attach`, real image, with the plain detach refused (exit 16, `Resource busy`) so the path is driven deterministically:

```
--- PRE-FIX (origin/main's python-artifact-layout.mjs)
  hdiutil calls:      ["attach -readonly -nobrowse -mountpoint .../local-operator-artifact-DRr6Be/mount .../local-operator-ui-0.0.0-arm64.dmg",
                       "detach .../local-operator-artifact-DRr6Be/mount"]
  result rows:        ["final-container-app=FAIL","artifact-unmount=FAIL"]
  unmount row:        ["hdiutil: couldn't unmount "disk32" - Resource busy\n"]
  hdiutil info still lists this image (attached): true
  mount still populated after return: true
  -> teardown would rmSync the image of a STILL-ATTACHED volume (this is the alert)
--- POST-FIX (this branch)
  hdiutil calls:      ["attach -readonly -nobrowse -mountpoint .../local-operator-artifact-sjsB5N/mount .../local-operator-ui-0.0.0-arm64.dmg",
                       "detach /dev/disk33s1",
                       "detach -force /dev/disk33s1"]
  result rows:        ["final-container-app=FAIL"]
  unmount row:        []
  hdiutil info still lists this image (attached): false
  mount still populated after return: false
  -> nothing attached at teardown; rmSync of the tree is safe
```

(`final-container-app=FAIL` on both sides is the harness's own fixture: the image carries no `.app`, and the harness stubs `checkApp`. The detach contract is the row under test.)

The pre-fix run was **detached by hand before the tree was removed** — deliberately not deleted, because deleting the backing file of a live volume is the alert this PR exists to remove.

### 3. Abrupt exits (the backstop)

Both detach attempts refused, so the module returns still holding a live volume, and the process then ends abruptly:

```
--- PRE-FIX / exit      volume attached after the process ended: YES
--- POST-FIX / exit     volume attached after the process ended: no
--- PRE-FIX / sigint    exit=130   volume attached after the process ended: YES
--- POST-FIX / sigint   exit=130   volume attached after the process ended: no
```

Exit status 130 on SIGINT both before and after, so Ctrl-C is unchanged; the volume is what changes.

### 4. The machine's own state

Before and after a real run of the touched file (`node --test --test-concurrency=1 scripts/managed-python.test.mjs`, **26 tests, 26 pass, 0 skipped**):

```
$ hdiutil info | grep -i "Local Operator"     before: (no match)     after: (no match)
$ ls /Volumes                                   before: Macintosh HD   after: Macintosh HD
$ hdiutil info | grep -c '^image-path'          before: 15            after: 15
```

Nothing from this work is left attached, and the count did not move.

Two notes, because the obvious check is weaker than it looks:

- `hdiutil info | grep -i "Local Operator"` **cannot see this defect at all**: the volume *label* is not in that output, only the image *filename*, and these fixtures ship as `local-operator-ui-0.0.0-<arch>.dmg`. The real check is the image path, which is what the evidence above and the tests use. The label was what macOS showed the operator, which is how this was tracked to `Local Operator` in the first place.
- The 15 images that *are* attached on this machine belong to other sessions and earlier runs (devices `/dev/disk4`-`/dev/disk30`, ages 3.8h to 51h, and 14 of them with their **backing file already deleted** — the exact state that raises the alert). This work's volumes were `/dev/disk32`-`/dev/disk35` and all were detached; this PR deliberately does not touch volumes it did not create.

### 5. The forced failure, end to end

The run above is the happy path. The failure path was driven deliberately, on a real image, with the plain detach refused — and the suite's own file run afterwards (section 4) shows the count unchanged.

Added to `scripts/managed-python.test.mjs` (the file that already owns these contract tests, driven through the injected `run` — no new runner, and they run on Linux CI too):

- the device is read from `hdiutil`'s own output rather than assumed (resolved path, literal path, last-device-line fallback, and no-output-is-`null`);
- a busy mount is detached **by device** and repaired with exactly one `-force` retry, and a repaired detach still removes the scratch tree;
- a doubly-failed detach does not proceed to delete the scratch tree, reports a failure naming both the mount and the device, and leaves the backing image standing.

The real-mount test also now asserts the detach names `/dev/disk*`, and its fixture teardown detaches before removing the tree.

## Testing Details

- `node --test --test-concurrency=1 scripts/managed-python.test.mjs` — **26 tests, 26 pass, 0 skipped** (runs the real `hdiutil` create/attach/detach path plus the two full-volume fixtures).
- `pnpm test:desktop` on the rebased head — **2,775 tests: 2,772 pass, 2 fail, 1 skipped** (exit 1). Both failures are pre-existing and environmental, and both are in files this PR does not touch:
  - `backend-settings-tiers.test.mjs` — `backend-settings-registry.json does not describe the registry /Users/damian/.local/share/uv/tools/local-operator/bin/python serves` (the committed fixture vs the operator's installed backend, which carries `display.narration`);
  - `python-bytecode-cache.test.mjs` — `global install: start() should report a started backend`.

  Isolated rather than assumed: with this branch's two files reverted to `origin/main` in the same tree, the same two tests fail the same way (`46 tests, 44 pass, 2 fail`), so neither is this change.
- `pnpm lint` — 0 errors, 74 warnings; `pnpm lint:scripts` — **2 changed files under `scripts/` are lint-clean**. Both files carried the tree's pre-existing backlog (`format`, `organizeImports`, `noUnusedTemplateLiteral`), and the gate is a ratchet over the diff, so that cleanup is in this PR as its own commit. Verified behaviour-preserving by formatting the fix commit's own copies and diffing: identical apart from the import ordering and the one literal.
- `pnpm check-types` — passes.
- The manifest's `scriptsTree` is re-derived after the fold onto `origin/main` (`6d3bcc2d0` -> `d0e88daf1`); `srcTree` is unchanged because no `src/` file moves. Verified with the gate's own exported `stampFailures` rather than a full `pnpm check-evidence` sweep: a sweep is the machine-wide lease and this change moves no frame.

## Checklist

- [x] The gates pass: `pnpm lint` (0 errors), `pnpm check-types`, `pnpm check-themes`, `pnpm test:desktop` (2,772 of 2,775 pass; the 2 failures are pre-existing and reproduce with this branch's files reverted — see above).
- [x] I have updated the documentation when required (comments carry the why and the constraint).
- [x] Security considerations are not affected (no new input path; the backstop runs a fixed `hdiutil detach` on a device this process attached).
- [ ] No related issue to link: the report is the operator's desktop, not a tracker item.


damianvtran:
### Agent review — round 1
- Reviewer: reviewer subagent (deepseek-flash)
- Scope: ab4953a2c8f6211dba5bf1018881bcc3bb7cf214..bfe4ffbf3
- Mode: Comprehensive
- Delivery contract: PR #347 — a test run must never leave the fixture's disk image attached, and the alert state (an attached volume whose backing file is deleted) must be impossible to reach through this code path
- Applicable risk surfaces: test-suite hygiene, on-disk state outside the repo, signal handling
- Verdict: Changes requested

No design round applies (test/CI hygiene, no user-visible surface), so this code round is the only gate.

| --- | --- | --- | --- |
| R1 | major | The backstop detaches a device THIS process never attached — the shipped fake contract test leaves a fabricated `/dev/diskN sM` recorded until process exit, where the real `hdiutil` force-detaches whatever holds that name | `python-artifact-layout.mjs:157-169` runs the real `/usr/bin/hdiutil detach -force -quiet <target>` on whatever `attachedTarget` holds; `:176-192` records it from the **injected** runner's stdout (`:277-279`); `managed-python.test.mjs:638` fabricates `/dev/disk9s2` and `:700-722` drives it with a detach that always fails, so the record is never cleared, and the module exports no reset (`:52,65,211,251,366`). Reproduced on this machine with my own image: the probe really attached an image → `/dev/disk10s1`; a runner that only *claimed* that device, with detach always failing, left it recorded; the process exited normally and the exit handler force-ejected the real volume — `hdiutil info \| grep -c lo-collide` went `1 → 0` and `/dev/disk10`,`10s1` disappeared. So the PR body's claim that the backstop runs "on a device this process attached" does not hold, and on a machine where `/dev/disk9s2` is a live volume owned by something else (another session's run on this shared box), every `pnpm test:desktop` force-ejects it at exit. **Required:** the fabricated name must never reach the real backstop — export a reset the fake fixture's `t.after` calls, or have the backstop act only on a target the module attached for real. |
| R2 | major | The touched file's own fixture teardowns delete a tree that holds an attached volume's backing file: the detach status is discarded and the `rmSync` is unconditional | `managed-python.test.mjs:507-517` (new: `hdiutil detach -force -quiet <mount>` then `rmSync(scratch)`), and `:1109-1112` (`mountedVolume`, same shape, unchanged). The `.dmg` lives in `scratch` (`:506-513`) while the mount point lives in `finalContainerChecks`'s own scratch, so when that forced detach fails, `rmSync` deletes the backing file of a still-attached volume — the exact alert this PR exists to remove, reached through the suite, which is where the alert was reported from. Today's `local-operator-artifact-*` module path is careful; its own caller is not. **Required:** check the teardown detach's status and skip the `rmSync` (or reuse the module's recorded target) in both teardowns. |
| R3 | minor | A second attachment silently drops the record of a first, undetached one — the backstop then cannot retry it | `attachedTarget` is one module-global slot (`python-artifact-layout.mjs:147`) that `recordAttachedTarget` overwrites (`:177`), while `verify-macos-artifacts.mjs:723-732` loops `finalContainerChecks` over every DMG in one process. With a doubly-failed detach on the first DMG, the second attachment replaces the target and the exit backstop no longer sees it, contradicting "detaches whatever this module still holds". **Required:** hold a set of targets, or refuse to overwrite a non-null one. |
| R4 | minor | An already-ejected volume is reported as a failure and leaks the scratch tree — the outcome is not honest for the "mount is already gone" case | Measured on this machine: `hdiutil detach /dev/disk99s2` → exit 1 `detach failed - No such file or directory`; `-force` → exit 1; `-force -quiet` → exit 1. `python-artifact-layout.mjs:332-346` turns any non-zero status into `artifact-unmount=false` and (via `:349-357`) keeps the scratch tree, so a volume someone else ejected first fails the run and leaks a tree although nothing is attached. **Required:** treat an absent device as detached (or confirm with `hdiutil info` before declaring failure). |
| R5 | minor | `mountedDevice`'s fallback can name a device this call did not mount, and detaching a slice ejects the whole image | `python-artifact-layout.mjs:220-228`: when no row's last column matches the literal or resolved mount path, it returns the LAST `/dev/` row. Both real attaches I ran here (and both fixture shapes) do put the mounted volume last, so this is not a live failure — but a shape that no longer does, or a mount path echoed differently, hands `finally` a slice belonging to another volume of the same image. **Required:** return `null` when no row names the requested mount; the caller's fallback (its own private mount path) is the safe one. |
| R6 | nit | The backstop's doc contradicts its behaviour | `python-artifact-layout.mjs:150-156` says a second call is "a no-op even when the first attempt failed" — the one case a retry would help is the one it cannot retry. Say it is single-shot, or retry once. |
| R7 | nit | Guard implies a state the comment denies | `:337` `if (attachedTarget === target) attachedTarget = null;` guards against exactly the overwrite R3 describes, while `:144-145` states "only one attachment is live at a time". Drop the guard or state the real invariant. |

### Verified clean (independently, on the reviewed head)
- **Device parsing** is from `hdiutil attach`'s own stdout, and both the literal and the `realpathSync`-resolved mount are matched before the fallback; `null` (no device) falls back to the caller's private mount path rather than inventing one.
- **Retry semantics**: the `-force` attempt runs only after a non-zero plain detach, and a failed `-force` is surfaced as an `artifact-unmount=false` row naming both mount and device, never swallowed; the tree is left standing on a doubly-failed detach and `rmSync` is the only deletion site in the module.
- **Mutation probe** (in `/tmp`, three mutants against the contract tests, run with the tests' own names): reverting the detach to the mount path fails `a busy mount is detached by device…`; deleting the `-force` retry fails the same test; deleting the early `return` (deleting the tree anyway) fails `a mount that will not detach…`. Each mutant is caught by exactly one test, and the unmutated copy passes 3/3 — the tests are not vacuous.
- **Signal/exit backstop**, on real images I attached myself: with both in-function detaches forced to fail, the volume was still detached when the process exited (`hdiutil info` count 1 → 0 for my image); under SIGINT the process still dies with **exit 130** and the volume was detached — the author's §3 claim reproduces, and Ctrl-C still interrupts.
- `node --test --test-concurrency=1 scripts/managed-python.test.mjs` → **26 tests, 26 pass, 0 skipped**; `scripts/evidence-manifest.test.mjs` → **34 pass**; the manifest's `scriptsTree` matches `git rev-parse HEAD:scripts` (`d0e88daf1…`) via the gate's own `stampFailures` (returns `[]`), so the re-derived stamp is correct.
- **Linux**: the three new tests run on ubuntu-latest in CI — `gh run view --job 105500939830` shows `ok 1236/1237/1238` for them, and the real-mount test skips by name. CI on `bfe4ffbf3`: Desktop Tests, Lint, Type Checking, Change Scope all pass.
- **Evidence claims** (§1-§5) are supported as far as they go: §2's pre/post rows match the code path, §4's own caveat that `hdiutil info | grep -i "Local Operator"` cannot see this defect is correct, and its "count unchanged" check reproduces here — I ran the touched file and the attached-device set was identical before and after. §3 is the claim I checked hardest and it holds. Nothing material is asserted without being shown.

Skipped by choice: the full `pnpm test:desktop` locally — this disk is at ~99% (6 GiB free) on a machine shared with ~25 sessions and the suite is a heavy, machine-wide job; the ubuntu CI job that runs exactly that suite is green on this head SHA, which is the stronger check for the CI question above.

Machine state: I attached only my own images and detached them by device before returning; the pre-existing `/dev/disk4`-`/dev/disk27` volumes were not touched. `hdiutil info` before and after my work lists the same **10** device lines — `/dev/disk4,disk4s1,disk5,disk5s1,disk19,disk19s1,disk22,disk22s1,disk27,disk27s1` (the leftover `Local Operator`/`volume.dmg` fixtures from earlier runs, backing files intact). Devices that appeared during my session under other prefixes (`lo-artifact-*`, `lo-managed-python-test-BFWS97`, `disk12/13`) were another session's and were left alone; one of them detached itself while I watched. My probes left no attached image and no visible window.

damianvtran:

- **Scope:** `ab4953a2c`..`25007b7b6`. The reviewed head is `bfe4ffbf3`; the branch was then folded onto `main` (`fc7bc8c5c`), so R1–R7 are in `8c0f6462b` on top of that fold, with the manifest re-stamp in `25007b7b6`.
- **Format:** one batched remediation commit for all seven findings, then the re-stamp commit — the repo's rule, and a re-stamp whose message says what moved.
- **What did not change:** no new reviewers, no mentions, no tickets; the two commits this branch already had are unreworked apart from the fold.

| ID | Disposition | Evidence |
| --- | --- | --- |
| R1 | **Fixed in `8c0f6462b`** | The record can no longer be armed from a claim; bound by two tests, one of them your out-of-process repro |
| R2 | **Fixed in `8c0f6462b`** | Both teardowns route through `removeVolumeTree` (detach for real, delete only once the volume is gone) |
| R3 | **Fixed in `8c0f6462b`** | `Set` instead of a slot; a second attachment no longer drops the first |
| R4 | **Fixed in `8c0f6462b`** | `hdiutil info` decides, not the exit code; an already-gone volume is success |
| R5 | **Fixed in `8c0f6462b`** | The last-device fallback is gone: `null` means "this output did not name my volume" |
| R6 | **Fixed in `8c0f6462b`** | Doc states the real contract: single-shot per target, set cleared before the attempts |
| R7 | **Fixed in `8c0f6462b`** | The guard drops out with the slot — `releaseAttachedTarget` is a `Set.delete` |

#### R1 (major) — the backstop may detach only what this process is shown to have attached

The record is now armed by `recordConfirmedAttachment(claimed, mount)`, which requires `hdiutil info` to report **that same device mounted at the private mount point this call created**. A fabricated name, a name belonging to some other process's volume, and a misparsed output all fail that confirmation, so the injected runner is separated from "we attached": the runner's stdout no longer decides anything the backstop acts on. `heldAttachTargets()` and `releaseAttachTargets()` are exported for fixtures.

Your repro, encoded as a test and also run by hand on the previous head vs this one. A real volume is attached, a child is handed its device, the child's runner only *claims* it and every detach it drives fails, and the child exits normally:

```
--- OLD module (bfe4ffbf3), child claims /dev/disk6s1
  child exit=0
  image still attached: NO - the child's exit EJECTED it
--- NEW module (8c0f6462b), child claims /dev/disk6s1
  child exit=0
  image still attached: YES (survived the child's exit)
```

Two tests bind it, and each is shown to be non-vacuous:

- `a runner that only claims a device cannot arm the attach backstop` — samples `heldAttachTargets()` **from inside the detach**, the first moment after the attach would be recorded. (Asserting only at the end was answered by R4's release of a target that is not attached, so that version passed against the old code — the sampling is why this test now catches it.)
- `a claimed device cannot eject another process's live volume at exit` (macOS) — your shape: the probe attaches a real image, hands the device to a child process, and asserts the volume survives that child's exit.

#### R2 (major) — the caller's teardowns

`removeVolumeTree(t, root, target)` detaches for real (`-force`, by device) and removes the tree **only when `hdiutil info` no longer lists the target**; otherwise the tree stays, with a `t.diagnostic` naming it. Both call sites go through it: the disk-image fixture's teardown (which now detaches by the device the module itself parsed from the attach) and `mountedVolume`, which you correctly flagged as untouched.

One disclosure: the helper's *refusal* branch (tree kept, volume still up) is reached only when a real `hdiutil detach -force` fails, which a test here cannot arrange — `-force` wins on this machine for a busy volume, and I will not detach somebody else's volume to manufacture one. The branch is four lines and documented; every other path through the helper runs in the suite.

#### R3/R7 — the collection

`attachedTargets` is a `Set`: `verify-macos-artifacts.mjs` looping DMGs in one process can no longer have a second attachment replace the first's record. With the slot gone, the `attachedTarget === target` guard has nothing to guard — `releaseAttachedTarget(target)` is a `Set.delete`, and the comment that denied overwrite ("only one attachment is live at a time") is gone.

#### R4 — the state decides, not the exit code

Measured here, for the record: detaching an absent device exits 1 with `detach failed - No such file or directory` — plain, `-force` and `-force -quiet` alike. A target that is still listed by `hdiutil info` after both attempts is the failure; anything else is a mount that is gone, its record released, and its scratch tree removed. Test: `a mount that is already gone is not reported as a failure`.

Related consequence, stated because it changed a test: the "will not detach" case now uses a **real** volume, since the failure row is owed only when `hdiutil info` still lists it — a fabricated device name cannot produce one any more, which is the point.

#### R5 — no fallback device

`mountedDevice` returns `null` when no row names the requested mount; the caller's own private mount path is the fallback. Test: the "other mount" row must yield `null`, not that row's device.

#### Round-1 verification (unmutation and mutation)

- `node --test --test-concurrency=1 scripts/managed-python.test.mjs` → **29 tests, 29 pass, 0 skipped** (was 26; three added).
- `scripts/evidence-manifest.test.mjs` → **34 pass**; `update-robustness` + `prune-python-resource` → **138 tests, 137 pass, 0 fail**.
- Mutation probe, each reverted fix caught by a test:
  - arming the record from the runner's claim (R1) → fails **both** claim tests;
  - dropping the `hdiutil info` re-check (R4) → fails `a mount that is already gone…`;
  - restoring the last-device fallback (R5) → fails `the device is read from hdiutil's own output…`.
- `pnpm lint:scripts` → 2 changed files lint-clean; `stampFailures` → `none`; CI Lint, Type Checking, Desktop Tests, NPX, Change Scope, Version Bump Guard green on `25007b7b6`.

#### Machine state

`hdiutil info` listed the same **3** images before and after (`lo-managed-python-test-Fy7NSw`, `-Dc6Br9`, `-volume-blqHce` — the ones with intact backing files you asked me to leave, and I did); nothing of mine is attached, `ls /Volumes` unchanged (`Macintosh HD`). Every volume I created was detached by device, and the probes above never named a device this process had not attached.

damianvtran:
### Fold note — the remediation SHAs after the rebase

`main` re-stamped `docs/evidence/manifest.json` twice while this round was in flight (its own fold, then the theme-legibility PR landing `scripts/` commits), so the branch was folded onto `fd6ec6086` and the stamp re-derived from the merged tree. **The round-1 remediation is unchanged** — only the SHAs moved, and every `Fixed in …` in the comment above maps like this:

| Round-1 reference | Now |
| --- | --- |
| reviewed head `bfe4ffbf3` | (unchanged, still the reviewed head) |
| R1–R7 fixes in `8c0f6462b` | `385e41396` |
| head after the round `25007b7b6` | `9c63a53c2` |

Verified on the new head: `stampFailures` empty (`scriptsTree` `1de3d7025`), `pnpm lint:scripts` reporting the two changed files lint-clean, `node --test --test-concurrency=1 scripts/managed-python.test.mjs` 29/29, and the mutation probe still catches all three reverts. CI is re-running on `9c63a53c2`; Desktop Tests was green on the previous head and the only difference is the fold.

damianvtran:
### Agent review — round 2
- Reviewer: reviewer subagent (deepseek-flash)
- Scope: fd6ec6086e9c29ffc64f458473d19384afd67f38..9c63a53c2
- Reviewed head: 9c63a53c2b4be2e7cdd7911fcff567a0ac11fc07
- Delta reviewed: bfe4ffbf3..9c63a53c2
- Applicable risk surfaces: test-suite hygiene, on-disk state outside the repo, signal handling
- Verdict: **Approved** — no BLOCKER and no MAJOR remains. This round is TERMINAL: the four findings below are follow-ups, not a reason for another round.

No design round applies (test/CI hygiene, no user-visible surface), so this code round is the only gate.

| --- | --- | --- | --- |
| R2-1 | minor | Nothing in the suite pins that a genuine attachment ARMS the backstop, so the R1 guard can be deleted outright and every attach/backstop test stays green | Mutation `MUT-3` (`/tmp/mut3-r2`: a copy of the module whose `recordConfirmedAttachment` returns immediately, i.e. the backstop can never arm) passes the whole attach/backstop set — **8/8**, including both new claim tests, the refused-detach test and the real-mount test. The new assertions are one-directional: they pin that a claim does NOT arm the record, and nothing pins that a real mount DOES, so the eject-on-exit behaviour the record exists for is never observed by a test. Required action: one test in the existing fixture shape — attach a real image through `finalContainerChecks` with a wrapper `run` that reads `heldAttachTargets()` on the first `ditto` call (the first call after the record is armed), assert the device is there, then assert it is gone after the module's own detach. I ran exactly that as a probe: `HELD_WHILE_MOUNTED=["/dev/disk6s1"] HELD_AFTER=[] STILL_ATTACHED=false FAILURES=[]`, so the positive case is stable enough to assert. |
| R2-2 | minor | R3's set is implemented but unasserted, and the shipped shape is exactly two attachments in one process | `attachedTargets.add` does what R3 asked, and the real gate is the loop R3 was written for: `verify-macos-artifacts.mjs:721-729` runs `finalContainerChecks` over every DMG (arm64 and x64) in ONE process. No test drives two attachments in one process, so a regression back to a single slot would not be caught by any assertion here. Required action: extend the R2-1 test to two images and assert both devices are held once the second attach has been recorded. |
| R2-3 | minor | A mount this process really made is left unrecorded whenever `hdiutil attach`'s stdout is not the shape `mountedDevice` parses — the backstop is then inert for that mount, which is the opposite direction of failure from R1 and the one the module's own doc does not cover | `recordConfirmedAttachment` returns early on `claimed === null` (:218), so the record is armed only when the runner's attach output named a device. The confirmation already has a state-based answer sitting unused: `mountedDevice(attachedVolumeOutput(), mount)` would name the device straight off `hdiutil info` whenever the private mount is live. Required action: drop the `claimed === null` early return and record the device `hdiutil info` reports at `mount`; the confirmation stays exactly as strong, because the row still has to name this call's own private mount point. |
| R2-4 | nit | The success verdict can short-circuit the state check, while the doc below it states an absolute | `:442` is `if (detached.status === 0 \|\| !isVolumeAttached(target))`, so a detach reporting success skips the check, whereas `:343-347` says "No path of this function reports success over a volume that is still attached". The state is the only real evidence either way and it is already computed; deriving the verdict from `isVolumeAttached` alone would match the doc and fail safe. |

### R1's fix, adversarially (the round-1 blocker)
The claim is that the record is armed only by `recordConfirmedAttachment`, which requires `hdiutil info` to report the device mounted at the private mount point the call created. I attacked it four ways, all with images I attached myself.

1. **The fabricated-claim route (the author's in-process test).** Controlled mutation `MUT-1` (`/tmp/mut-r2`: `recordConfirmedAttachment` arms from the runner's own output — `bfe4ffbf3`'s behaviour) run against the new test file: **FAILS**, `actual: [ '/dev/disk9s2' ]` against `expected: []`, with the test's own message "a claim must not be recorded, or the exit handler force-detaches that name". The unmutated copy of the same harness passes. The test is not vacuous, and the sampling-from-inside-the-detach concern is answered by the mutant.
2. **The out-of-process route (the author's version of my round-1 repro).** Mutation `MUT-2` (`MUT-1` plus `bfe4ffbf3`'s release-only-on-exit-status-0, i.e. the round-1 module's arm-and-never-clear) run against that test alone: **FAILS** — "the child's exit must not eject a device it only claimed", `actual: false, expected: true`. The round-1 bug reproduces inside the new test's own harness, and the fixed module passes it. This assertion is the load-bearing one, and it is real: it measures disk state, not the record.
3. **A live device at a DIFFERENT mount point** (my own route, not the author's): a 16 MB image of mine attached at its own mount, then a child whose runner claims that same device as mounted at the child's private mount. Real module: `heldAtDetach=[]`, `stillAttachedAfterChildExit=true`. `MUT-1` mutant: `heldAtDetach=["/dev/disk6s1"]`. So "a row exists in `hdiutil info`, but not at the mount this call made" is exactly what the confirmation rejects — the mutant arms it and the child's exit would force-eject a live volume it never mounted. Finding 3 of round 1 is fixed on the real code.
4. **Every other arming route I could find.** A fabricated attach output is route 1; an injected runner cannot arm anything else, because the module's only exported mutators are `heldAttachTargets()` (read) and `releaseAttachTargets()` (clear) — there is no exported way to add a target; a `hdiutil info` row that exists but is not at the private mount is route 2/3; and a second attach of the same image lands in the set (R2-2 is that it is unasserted, not that it is broken). Positive control: for a genuine attach the record IS armed while the volume is up and released after the module's own detach, so the gate did not disarm the backstop — this is the behaviour R2-1 asks to be pinned.

### R2's fix
- Both sites named in round 1 now route through `removeVolumeTree`: the disk-image test's own root (`:539`) and `mountedVolume` (`:1373`); so do the other three teardown sites in the file (`:654-655`, `:769`, `:944`).
- No unconditional `rmSync` of a volume-bearing tree is left. Of the 11 remaining `rmSync` calls, `:69` and `:995` are fixture/non-image roots, `:189` removes a planted escape entry, `:1102`, `:1249`, `:1299`, `:1527`, `:1555`, `:1566`, `:1567` remove venv/runtime trees — `:1527` deletes a directory INSIDE the mounted volume (`join(mount, "support")`), which touches neither the backing image nor the mount point.
- The refusal branch reads the state and reports: `:88-95` force-detaches, re-checks, and on a volume still up calls `t.diagnostic(...)` and returns before the `rmSync`. The module's own refusal (`:457-465`) IS covered — "a mount that will not detach is reported and its tree is left standing" asserts the `artifact-unmount` row names the mount and the device, and that both the mount point and the backing image survive; it passes here in 37 s against a real image with every detach refused.
- **On the author's disclosure that `removeVolumeTree`'s refusal branch is unreachable by test, I agree with the disclosure and I do not think the guarantee rests on that branch.** The guarantee is the `isVolumeAttached` state check, which is exercised in both directions by tests that do run (attached — the refused-detach test's own assertions; absent — "a mount that is already gone is not reported as a failure"), and it is the same predicate the module's covered refusal path rests on. What is untested is the diagnostic, not the safety property — and the branch cannot be driven without holding another process's volume open, which is not a thing a test should do on this box.

### R3–R7
- **R3** — set, not slot: `const attachedTargets = new Set()` (`:159`), `add` (`:192`), `delete` (`:226`), no overwrite path, and `verify-macos-artifacts.mjs:721-729` is the multi-DMG loop that made this matter. Unasserted (R2-2).
- **R4** — an absent device is success: the verdict is the state, not the exit code (`:442`), and "a mount that is already gone is not reported as a failure" passes with a fabricated `detach failed - No such file or directory`.
- **R5** — no last-device fallback: `mountedDevice` returns `null` when no row names the requested mount (`:325-328`), pinned by `:1018-1031` (a row naming another mount, and empty output, both `null`).
- **R6** — the backstop doc now says "SINGLE-SHOT PER TARGET" (`:164`) and no longer claims a retry it cannot make.
- **R7** — the overwrite guard is gone; `releaseAttachedTarget` is a plain `delete` (`:225-227`), so the code no longer implies an invariant the prose denied.

### The delta as a whole (fold onto a moved `main` + re-stamp)
- The three content commits are **patch-identical** to the round-1 remediation the author posted, so the fold re-spelled the SHAs rather than re-working or reverting anything: `a27cce4c1→30a9949cc`, `0e80cf048→6cddef128`, `8c0f6462b→385e41396` (`git show <sha> | git patch-id --stable` equal in all three pairs).
- The fourth commit re-derives only the stamp: `scriptsTree 7feb600c1… → 1de3d7025…`, and `1de3d7025890bd8390747a18368fff34ccdf8300` is exactly `git rev-parse HEAD:scripts`. `srcTree` is untouched and equals `git rev-parse HEAD:src` (`0e72b5c6b…`). The diff of the manifest against the merge-base is that one line.
- Union kept both sides: against the merge-base the branch touches only `scripts/python-artifact-layout.mjs`, `scripts/managed-python.test.mjs` and that manifest line, so no `main` file is reverted by the fold, and every R1–R7 behaviour is present at this head (read item by item on the current files, not inferred from the remediation comment).

### Verified by running
- `node --test --test-concurrency=1 scripts/managed-python.test.mjs` → **29 tests, 29 pass, 0 fail, 0 skipped, 137 s** (the real mounts ran: 36.6 s, 37.3 s, 28.4 s — nothing skipped on this platform).
- `node --test --test-concurrency=1 scripts/evidence-manifest.test.mjs` → **34 pass**, so the re-derived stamp passes the gate's own check.
- `node scripts/check-scripts-lint.mjs` → "2 changed file(s) under scripts/ are lint-clean against origin/main (fd6ec60)".
- Mutants and probes above, in `/tmp` only; the worktree was read, never written (`git status` clean, and the head is still `9c63a53c2`).
- Skipped: the full `pnpm test:desktop` (disk is at ~99 %, ≈6 GiB free; it runs ~600 files and the changed surface is these two files plus the manifest gate). Nothing in the changed files is exercised only by that suite.

### Follow-up observations
- Machine state around this review: another session on this box is still running the unfixed path. Mid-review a fourth image appeared that no run of mine created — `lo-managed-python-volume-jkjvZ4`, attached with its scratch tree ALREADY DELETED, which is the "Disk Not Ejected" state this PR removes — and its owner reclaimed it a minute later. I did not touch it, and I touched none of the three pre-existing orphans either. Every device I attached (fixtures and probes) was created by me in `/tmp` and detached by device before I finished; `hdiutil info`'s device table is identical to the one I recorded at session start.
- The delta's second commit is labelled `style(lint)` but carries a whole-file reformat of both files (~950 changed lines) — a large share of the delta is formatting interleaved with the fix, so the fold note's SHA mapping is more useful than the commit sequence for reading it. Nothing wrong with the result (the scope gate is green); worth folding the reformat into the functional commits if this branch is squashed.

damianvtran:

- **Scope:** the four findings are against the head they were written on, `9c63a53c2b4be2e7cdd7911fcff567a0ac11fc07`. Nothing in this branch has changed since, and nothing was pushed for this round — the round-2 review is the record of that head.
- **Format:** no remediation commit for this round. Round 2 returned **no blocker and no major** and was marked terminal; all four findings are dispositions, and the code the review cleared is the code that is here.
- **What did not change:** no push, no reviewers, no mentions, no tickets. The head is still `9c63a53c2`.

| ID | Disposition | One-line reason and plan |
| --- | --- | --- |
| R2-1 | **Deferred** | An assertion is missing, not a guard: the arming path is exercised and correct at this head (probe on a real mount: `HELD_WHILE_MOUNTED=["/dev/disk6s1"]`, `HELD_AFTER=[]`), so what does not exist is the test that pins it. Plan: a `scripts/` test-coverage follow-up adds one real attach through `finalContainerChecks` whose wrapper `run` reads `heldAttachTargets()` on the first `ditto` call (the first call after the record is armed) and asserts the device is present, then that the set is empty after the module's own detach — which is the reviewer's probe, encoded. |
| R2-2 | **Deferred** | Same class, same file, same follow-up: `Set` semantics across two attachments in one process is the shipped loop (`verify-macos-artifacts.mjs:721-729`), and the delta changed the tracking, not the loop. Plan: the same follow-up drives two images and asserts both devices are held once the second attach is recorded. |
| R2-3 | **Deferred** | This is the one finding that would move shipped code, and the review established it is a coverage gap rather than a failure — the release gate's own attach is not `-quiet`, so `mountedDevice` names the device today. It is deferred rather than folded in here because it changes the arming rule, and a change to the arming rule is precisely what round 1 asked to have pinned by tests before it ships. Plan: it lands after R2-1's positive control, in the same follow-up — drop the `claimed === null` early return and record the device `hdiutil info` reports at `mount`, keeping the confirmation's requirement that the row name this call's private mount point. |
| R2-4 | **Deferred** | Nit, and a disagreement the doc and the code reach only in a case that cannot occur: `hdiutil detach` reports success only after ejecting, and the line right after the disjunct already asks the state what actually happened. Plan: fold it into the same follow-up as a tightened condition (state as the sole decider) plus the comment that matches it. |

None of the four gates the merge, and none of them is a reason for another round on this PR. The round-2 review closed with no blocker and no major, it was marked TERMINAL, and its Scope line names the reviewed head that is still the head here, so the approval is fresh. The four items ride as a `scripts/` test-coverage follow-up rather than another round.

damianvtran:
### Released — `v0.28.1`

The window since `v0.28.0` shipped as **`0.28.1`** (patch), published as a pre-release from the bump commit `e06336aaff74efa54209002c71a99a4a1765033c`, so the updater's `/releases/latest` feed keeps answering the previous complete Release until this one's installers exist. The pipeline owns promotion; there is no manual promote step.

- Release: https://github.com/damianvtran/local-operator-ui/releases/tag/v0.28.1
- Tag: `v0.28.1` -> `e06336aaff74efa54209002c71a99a4a1765033c` (`git rev-parse "v0.28.1^{commit}"`)
- Window: #331, #316, #347, #351, #352 - one bump for all five, by materiality; all five bodies claimed `patch`, so the cut agrees with every PR's own `Release:` line.
- This PR's line, as quoted in the notes: ``Release: patch — the desktop suite stops leaving `Local Operator` images attached, so the "Disk Not Ejected Properly" alert stops appearing after a run.``

PR #352, opened 2026-09-18
docs(evidence): cite a before-set head a reader can actually resolve

damianvtran:
# PR Description

## Summary of Changes

One field in `docs/evidence/manifest.json`, and the sentence beside it. The shipped manifest names a
commit a reader cannot resolve, and `check-evidence`'s own gate reports it — this is the fix, plus the
audit that shows it is the only one.

- **What was wrong.** `supplementary[turn-answer-stamp-before].capturedAtHead` named `f6d73dc1e`, a
  spelling of this lane's before-half commit that an earlier fold produced and a later one dropped:
  `git merge-base --is-ancestor f6d73dc1e 6e6a08b86` answers **not-an-ancestor** and
  `git for-each-ref --contains f6d73dc1e` answers **0 refs**, so it dies at the next gc.
  `provenanceFailures` and `citationFailures` — the two `check-evidence`'s `main()` calls *first*, and
  the two the desktop suite does not ask — report it on the merged `main`.
- **The correction of my own record.** #316 shipped this citation, and my `### Agent review remediation
  — round 4` marked it "fixed by the merge", reasoning that merging makes the branch's commits
  ancestors of the merge commit. That reasoning is wrong for an orphan from an abandoned lineage, and
  the erratum on #316 says so with the two commands and their output. This PR is the fix that comment
  said was not needed.
- **The fix.** `capturedAtHead` now names `955444dfa`, which **is** an ancestor of `origin/main` (four
  refs resolve it). Substitution verified rather than assumed: for all four paths in the set,
  `git rev-parse f6d73dc1e:<path>` equals `git rev-parse 955444dfa:<path>` — same blobs, same commit
  subject, different lineage spelling.
- **The sentence beside it, which had drifted further.** The entry's `source` still named `1d94e4a33`
  while the field said `f6d73dc1e` — two claims about one capture that disagreed. It now names the hash
  the field names, records the chain the folds produced (`742bf439d` → `1d94e4a33` → `f6d73dc1e` →
  `955444dfa`), and says which link the guards caught.
- **The audit, so this is not the first of a series.** Every structured citation in the manifest was
  checked with `git for-each-ref --contains`: `head`, `partialCapture.refreshedAtHead`,
  `partialCapture.addedAtHead`, and all 42 `supplementary[].capturedAtHead` values — **45 fields, one
  unreachable citation, this one.** The other 40-odd hash-shaped strings a broader sweep finds live in
  prose *notes* describing past states, not in citation fields.

[...]

PR #351, opened 2026-09-18
docs(evidence): correct the About-panel record's residual prose and the comment it cites

damianvtran:
# PR Description

## Summary of Changes

The deferred findings from PR #341's two review rounds — code review **M1–M4**, design **D5** and **D7**, nit **N1**. Every one of them is prose or a comment: the committed evidence for the About panel must not state things the code and the pixels do not show. Nothing here changes behaviour, and no frame is re-captured.

Three commits, because one of them moves `src/`:

1. `99eb2f180` — the six corrections (five findings in `docs/evidence/`, one comment in `src/main/window-mode.ts`).
2. `85e65309b` — re-stamp only: the manifest's `srcTree` pin is `git rev-parse HEAD:src`, so the first commit makes it stale by construction. `0e72b5c6b` → `cc7c2d793`, derived from the first commit's own tree and read back out of the file. `scriptsTree` (`1de3d7025`), `frames`, `surfaces` and `themes` are unchanged, because this range moves no `scripts/` file and adds no frame.
3. `8c48726e5` — whitespace only: the first commit's build-string bullet left one 101-character prose line, against a file that wraps prose at 80 (its only long lines before this PR are the recorded run transcripts inside their fence). Wrapped, no word of the claim changed. It is a separate commit rather than an amend because AGENTS.md forbids force-pushing, and a rewrap is not worth rewriting a pushed branch for.

**No design round applies, and PR #341's design sign-off is unaffected.** The delta is comments and prose in three files; none of it is reachable from a rendered surface — the `src/` change is a comment block inside `resolveAboutPanelAction`'s doc comment, and the other two files are the evidence record itself. A comment cannot repaint a panel, so the frames the round-2 design review approved (`9d26b2205`) stay pictures of the trees they name.

Release: patch — the About panel's committed evidence record now states what the code and the pixels actually show (no runtime change)

## Related Issue(s)

No issue.

- Related: #341 — the change this record describes; its review rounds raised and deferred these findings

## Impact

- Prose and one comment only. No runtime, renderer, preload or test code changes; no dependency updates; no performance or security implications.
- `version` in `package.json` is untouched.
- `docs/evidence/about-panel/frames/`, `harness/` and `scripts/` are untouched, so no committed pixel and no gate input moves.

## The findings, and what was done

| ID | Disposition | Where |
| --- | --- | --- |
[...]

PR #346, opened 2026-09-18
feat(themes): every ground to a legibility floor, and a second accent

damianvtran:
# PR Description

## Summary of Changes

Every theme's page grounds are lifted to a legibility floor, the ink floors are raised, the sidebar's current-row mark is re-derived so it outranks a hovered neighbour, the app's shared selection ground stops being a barely-visible tint, and the palette carries a second decorative accent for the themes that have always had two signature hues.

This is the answer to four things the operator reported on the running app, in his words:

1. "some are a bit hard on the eyes, particularly around darker mode themes where the background behind the main text is too dark/black. Ideally everything should be off-black closer to grey to be legible"
2. "for some the text is a little too grey on grey backdrop"
3. "review the light themes as well to make sure that they're well contrasted and follow proper legibility guidelines for light mode themes"
4. "selections are generally not very visually appealing and look very off, and are often not well contrasted" (screenshot: the command palette, active row)

and one he asked for as an improvement: "proper support for more than 1 accent color in a tasteful way for themes that typically have multiple accent colors".

### What changed

**Grounds.** 20 of the 41 dark-ground themes had a `canvas` below L\* 10 and 8 below L\* 5 — `obsidian` 2.5, `arcade` 2.8, `tron` 2.9, `matrix` 3.0, `cyberpunk` 3.3, `neon` 3.4, `synth` 3.5, `dune` 3.7. Every dark theme's `canvas` is now at or above a floor with its ladder spread preserved (`obsidian` #09090B to #202021, `cyberpunk` #0E0A16 to #211F28), and the light ramps are capped so they stop being glaring (17 light themes; `solarized-light` #FDF6E3 to a capped value). 39 palettes moved a ground, 20 were already inside the band.

**Ink.** `inkDim` sat exactly on the AA floor for body text (4.50-4.62:1 on `elevated`, 4.53-4.62:1 on `sunken`) while being used at 11-13px meta sizes — the "grey on grey" half of the report. All three weights are raised, and `borderControl` moves with the lifted `elevated` because it failed 3:1 in 8 themes once `elevated` moved. The old pinned exceptions and pinned ink steps are gone: the set now solves the contract with zero exemptions.

[...]

PR #345, opened 2026-09-18
fix(chat): remove the per-row browser mark from the chat sidebar

damianvtran:
Removes the per-row browser mark from the chats sidebar — the dim Globe (with its tab count, its loading spinner and its approval badge) drawn at the right of every conversation row — and retires the apparatus behind it. The operator asked for it out, and for the slot it was sitting in:

> "I don't think we need a browser button for each chat row in local-operator-ui, can you remove that? Especially since we're going to have a pin icon that should show up on hover, it takes up too much space for a button that would be infrequently used."

**This reverses the ask the control was built for, and this PR says so rather than leaving the design record contradicting the code.** Design R2 is titled "a corner affordance on every conversation row"; `3996c3749`, which is what made it unconditional, records the earlier ask as "a corner tab on EACH conversation" (review round 1's D2/A4 is that change). `docs/design/sidebar-conversation-browser.md` now opens with a superseded note quoting BOTH asks, and R2 — including its states table — carries a heading marker saying so, so a reader landing there from a search cannot take it for the contract. The body is left as the record of a decision that was taken and then reversed.

Release: patch — the sidebar's per-row browser button is gone, every title keeps the width it held, and the browser is opened from the chat header's Globe.

## What the user gets

| where | before | after |
|---|---|---|
| every conversation row | a Globe: dim when quiet, `n` beside it when the conversation has tabs, a spinner while a tab loads, an attention badge with that conversation's waiting approvals | nothing; every title keeps the **28px** the control held (`size-6` 24px + the wrapper's `gap-1` 4px) |
| the conversation you are in | open its browser from the row's Globe, or the header's Globe | the header's Globe (unchanged, `chat-header.tsx`) |
| a conversation you are NOT in | one press on its row's mark | select the conversation first, then the header's Globe |
| another conversation's waiting approval, at a glance | the mark's badge | nothing in the chrome without opening the browser — see the capability cost below |

## Impact — the capability cost, inventoried

The first draft of this section named one loss and implied a second. Four review streams independently found the same under-listing (code review R3, design D3, UX U1, QA Q17/Q18), so here is the whole inventory, with the surviving surface named for each:

- **A non-current conversation can no longer have its browser opened from the sidebar** without being selected first. What survives: select the row, then the header's Globe — the pane follows the selection, so nothing about another conversation is shown by accident (UX round 1 walked it).
[...]

PR #353, opened 2026-09-18
chore(release): bump version to 0.28.1

damianvtran:
# chore(release): bump version to 0.28.1

## What this is

The version bump for the window since **`v0.28.0`** (published 07:15Z today, `Latest`, non-prerelease). `package.json` only, one line, on `main`'s own tree at `b1baf7450`: `0.28.0` -> `0.28.1`.

## The bump, by materiality — and a reversal, recorded

**`patch`**, and the reversal is stated first because the argument was made the other way round first.

The window owner's **initial call was `minor`**, on #331's capability reading: before it the app could not move its own managed backend environment at all, and after it the app publishes a new generation of that environment, smoke-tests it and flips a pointer with the previous generation kept as rollback. That sentence was written **after** #331 merged, into its merge-disclosure comment (`### Merge note — completion`), and it carries `Release: minor`. It is not in the PR's body.

Read against the PR's own body, **#331 claims `patch`**: `Release: patch — the app can finally move its own backend environment without writing into a tree a running daemon has open.` This repository's rule is that the body is the PR's claim and the notes quote it, and that a divergence is recorded rather than silently overridden — #344 is the model for how that is written, and it argues the same shape in the opposite direction. Both positions are on the record on #331's own thread. **Where they disagree the cut follows the body**, and the repository's stated preference on the direction of the error — under-calling is corrected by the next release, while an over-called minor permanently misreports how much changed — points the same way. So the honest shape of this window is a rendering fix, a mount fix, two evidence-doc PRs and one update-path capability; that is a patch.

The other four PRs argue no class either way.

## The window (`git log --first-parent --oneline v0.28.0..origin/main`)

- **#331** `feat(update): publish a new generation for the app-owned environment, and give the skew notice its action` - `b1baf7450`, squash from `1517abd45`, merged as admin and disclosed on the PR (no human reviewer requested, nobody tagged). Its body's line: `Release: patch — the app can finally move its own backend environment without writing into a tree a running daemon has open.`
- **#316** `fix(chat): move the transcript's time stamps onto the turns` - `1db64584f`. Its body's line: `Release: patch — the always-visible time now appears in exactly two places: under a user turn's bubble, and inside an opened tool call. The stamp at the foot of the conversation is gone (under the working line mid-turn, and everywhere else), and the agent side carries one caption per turn, under the answer the turn ends on, left-aligned on the prose's own rail.`
- **#347** `fix(release): detach the verification mount by device, and never over a live volume` - `fad6f4c91`. Its body's line: `Release: patch — the desktop suite stops leaving `Local Operator` images attached, so the "Disk Not Ejected Properly" alert stops appearing after a run.`
- **#351** `docs(evidence): correct the About-panel record's residual prose and the comment it cites` - `6e14e8d2e`. Its body's line: `Release: patch — the About panel's committed evidence record now states what the code and the pixels actually show (no runtime change)`
- **#352** `docs(evidence): cite a before-set head a reader can actually resolve` - `904d576d5`. Its body's line: `Release: patch — the evidence manifest no longer cites a commit that cannot be resolved, so the evidence gate's own provenance check stops reporting it`

**All five carry a `Release:` line**, so the Release notes for this window have five quoted claims and no authored absences — worth stating because a `docs:`-only window is exactly where one would expect an absence, and neither #351 nor #352 is one.

## Precondition, checked rather than assumed

`git diff v0.28.0..origin/main -- package.json` is **empty**, so no merged PR carried its own bump and no number was spent. `0.28.1` is neither an existing tag (`git ls-remote --tags origin v0.28.1` returns nothing) nor an existing Release (`gh release view v0.28.1` -> `release not found`).

## Carried debt, not smoothed

#331's body discloses it and its merge note repeats it: **the committed evidence frames photograph the pre-lift palette**. #346 lifted every page ground and the inks after those frames were taken - 160 of the 384 colour variables the twelve themes declare differ between the tree they photograph and this head - so the swept set is a mixture of ages, which is the repo-wide debt `docs/evidence/manifest.json`'s `themeLegibilityCapture.owed` records. It reaches #316's and #347's frames for the same reason they are themed surfaces. It is carried into the Release notes rather than presented as fresh evidence.

## What rides the next window

Open at this cut, by number: #345, #343, #337, #313, #296, #288, #287, #273. Nothing here is held back deliberately; they are simply not merged.

Release: patch — the window since v0.28.0: the app can move its own backend environment by publishing a new generation with the previous one kept as rollback, the transcript's time stamps move onto the turns, the desktop suite stops leaving verification images attached, and two evidence records are corrected.


damianvtran:
### Agent review — round 1 — second pair of eyes (read-only)

**Reviewer:** session `Take over stale PR 246` (pid 35679) — independent of the PR's author and of the `### Agent review` round the author is running in parallel; the model running this pass is the session's own (`deepseek/deepseek-flash`), no fallback.
**Scope:** `b1baf7450..6a2d0376f` (the bump head) — verified read-only: no merge, no tag, no claim, no edit.

Everything the bump is *for* checks out, measured rather than accepted. One defect found, in the rendering of a quote rather than in its content.

**Verified**

- **The diff is the bump and nothing else.** `git diff origin/main 6a2d0376f` is `package.json | 2 +-`, `+ "version": "0.28.1"` / `- "version": "0.28.0"`, one file, one insertion, one deletion. Its parent is `b1baf7450…` and **`origin/main` is still that same SHA at this reading**, so the window has *not* grown under the PR — the check you asked for most is the one that passes.
- **The window is exactly five.** `git log --first-parent --oneline v0.28.0..origin/main` → `b1baf7450` (#331), `904d576d5` (#352), `6e14e8d2e` (#351), `1db64584f` (#316), `fad6f4c91` (#347). No sixth, no listed PR that did not merge.
- **Nothing spent.** `git diff v0.28.0..origin/main -- package.json` is empty, and `v0.28.1` is neither a tag (`git ls-remote --tags origin 'v0.28.1*'` → 0) nor a Release.
- **The verbatim check you asked for, done mechanically rather than by eye.** For all five PRs I extracted the `Release:` paragraph from each **body** (joining soft wraps) and compared it to every quoted block in this PR's body after collapsing whitespace: **#331, #352, #351 and #316 match byte-for-byte**, and all five bodies do carry a line — so "five quoted claims and no authored absences" is true as stated. **#347's content also matches** — the defect below is its delimiter, not its words.

#### R1 (MINOR) — #347's quote closes its own code span early, so it renders broken

Line 21 of this body quotes:

```
`Release: patch — the desktop suite stops leaving `Local Operator` images attached, so the "Disk Not Ejected Properly" alert stops appearing after a run.`
```

The span is opened with a single backtick and the sentence itself contains backticks around ``Local Operator``, so markdown closes the code span at that point: a reader of the **Release notes** sees `Release: patch — the desktop suite stops leaving` as code, ``Local Operator`` as a second code span, and the remaining clause as loose text — the same sentence, visually cut in three, in the one document that is quoted permanently. Fix by wrapping in a double-backtick fence (`` … ``) or by dropping the inner backticks; either keeps the words byte-identical. Worth catching now rather than at tag time because `scripts/release-state.mjs` reads the *body* it is given and neither refuses nor annotates a broken span, so nothing downstream would tell you.

**My class read, on the record as you invited.** **`patch`, and the reversal was right.** I argued `minor` earlier on the sentence you had composed into #331's merge comment; against the PRs' own bodies, #331's claim is `patch — the app can finally move its own backend environment without writing into a tree a running daemon has open.`, and the materiality rule this repository writes down says the body is the claim and under-calling is the preferred direction. So my earlier read is withdrawn as an opinion that the body overrules — not a disagreement I want standing between you and the tag. The capability argument was not silly (the app *can* now move its own environment), but a claim is weighed as the PR makes it, and the earlier claim was never the PR's.

Nothing else needs a round from me: with R1's delimiter fixed the artifact and its quotes are clean as far as these four checks reach. Merge, tag and promotion are yours — I hold no lock and will not touch this PR.

damianvtran:
### Agent review — round 1

**Reviewer:** reviewer subagent (deepseek/deepseek-flash)
**Scope:** `v0.28.0..6a2d0376f` (the window since the previous tag; this PR's own diff is `b1baf7450..6a2d0376f` — one file, one line)

The diff is `package.json` `0.28.0` -> `0.28.1`, and the PR head's only parent is `b1baf7450791bd6137a5039214a458926afffe97`, which is `origin/main`'s head. So the diff is not the review; the release it authorises is. Everything below is checked against primary sources — git, the forge API, the code on the released head, and a re-run of the notes gate — not against this PR's prose.

**1. The window is complete and is the right window — verified.** `git log --first-parent --oneline v0.28.0..origin/main` on `b1baf7450` returns exactly five merges, no more and no fewer: `b1baf7450` (#331), `904d576d5` (#352), `6e14e8d2e` (#351), `1db64584f` (#316), `fad6f4c91` (#347). `v0.28.0^{commit}` is `8fec800a4`. `main` has not moved since the cut: re-fetched immediately before this comment, still `b1baf7450`, so the PR head is `main` + one commit and the bump sits on the tree the release would publish.

**2. Precondition — verified.** `git diff v0.28.0..origin/main -- package.json` is empty, so no merged PR carried its own bump and no number was spent. `git ls-remote --tags origin v0.28.1` returns nothing, and `gh release view v0.28.1` returns `release not found`; `gh release list` shows `0.28.0` (Latest, non-prerelease) as the newest. The deleted `assertVersionSurface`'s intent checked by hand: `origin/main:package.json` reads `0.28.0` and `v0.28.0` exists, so the version on the released head is itself a released tag.

**3. The class — R1, judged, and I agree with `patch`.** The reversal is recorded honestly: the `minor` reading was #331's *merge-disclosure comment*, the PR's *body* carries `Release: patch`, the notes quote the body, and the comment nowhere claims otherwise — I re-read #331's thread to confirm the body line and the comment sentence are as this PR describes them. On this repository's bar ("a step-function capability in its own right — a new thing the app can do"), #331 is the window's only candidate and it is a close one: the app could not move its own managed backend environment before, and now it can. What decides `patch` for me is that the change supplies the action behind an offer the app already made (the skew notice) rather than opening a new surface, that the other four PRs are fixes and evidence records, and that this repository's stated direction of the error makes an under-call the correctable mistake. Had the comment's `minor` sentence been collected as the claim, the window would have been over-called on a divergence the rule says to follow the body on. No action; the disagreement is on this record deliberately.

**4. The notes — verified.** `/tmp/loui-rel-0281.B26rxS/release-notes.md`:
- All five quoted `Release:` lines compared programmatically (whitespace-normalised) against each PR's **current body** via `gh pr view <n> --json body`: identical for #331, #316, #347, #351 and #352. Nothing here is a message's version of a body line.
- The composed `minor` sentence from #331's merge comment appears nowhere: `grep -n "Release: minor"` and a grep for its distinctive clause both return empty. The mistake this round exists to catch is not repeated.
- The gate itself, re-run rather than assumed: I imported `releaseNotesVerdict` from `scripts/release-state.mjs` and called it on the file — `{ refusal: null, warnings: [] }`. The body is neither empty nor GitHub's `## What's Changed` draft, and it carries both template shapes: `## What's New` at line 1 (the gate's match is case-insensitive, so it passes as written) and `**Full Changelog**: https://github.com/damianvtran/local-operator-ui/compare/v0.28.0...v0.28.1`.
- Assertions about state, checked rather than taken:
  - **Retained, not restorable:** `origin/main:src/main/update-service.ts:6094-6100` states it in those terms ("the superseded generation is RETAINED … nothing flips the pointer back to it … a restore path is a user-facing capability with its own story"). Independently, `git grep -rn -i "restoreGeneration\|rollbackGeneration\|previousGeneration" origin/main -- src` is empty — there is no restore path on this head, so "not restorable" is true of the code, not just of the prose. #331's review records it as R4 "recorded, not changed", and the interrupted-update residual as U3.
  - **The debt:** `themeLegibilityCapture` is present on the released head (`origin/main:docs/evidence/manifest.json:29174`) with `owed`, `owedCommand` and `restampNote`. I reproduced its central number rather than quoting it: over the twelve swept themes (`partialCapture.refreshedThemes`) the pre-lift tree `12b52ff57` (the #346 parent) declares exactly **384** colour variables, and exactly **160** of those 384 differ at this head — "160 of the 384 colour variables the twelve themes declare" is exact. The head declares 408, the 24-variable delta being the added `accentAlt`/`accentAltWash` roles, which the notes' "no breaking changes" framing does not contradict. #346 (`fd6ec6086`) is inside the `v0.27.4..v0.28.0` window, so the notes' attribution of the lift to `v0.28.0` is correct, and the 11-52 h deferral arithmetic ("about 0.37 s/frame … 7-33 s/frame … 11-52 hours") is `.owed`'s own sentence reproduced faithfully.
  - Read from `origin/main`, not from this checkout's working tree: this checkout is on an unrelated branch, where the file differs.
- I did not run the application or the desktop suite; the diff cannot change runtime behaviour (one version string) and the scope classifier skips those jobs, so there is nothing here for them to exercise. The one thing that could differ is the version string, and I read it at the PR head: `0.28.1`.

#### Findings

| ID | Severity | Finding |
| --- | --- | --- |
| R1 | judgement, no action | The class: `patch` is the right call on this bar, as argued above. Recorded so the reversal is answered rather than left implicit. |
| R2 | NIT, no change required | #352's body `Release:` line is hard-wrapped across two physical lines ("…so the" / "evidence gate's own provenance check stops reporting it") and the notes join it with a space. Word-for-word verbatim, not byte-for-byte; the notes' rendering is the correct one. Stated so "verbatim" is read precisely, given this round's history. |
| R3 | NIT, no change required | The notes say the no-restore follow-up is "listed as a follow-up"; it is listed in #331's merge-note/remediation comment and in `update-service.ts`, not in that PR's body's "Not addressed" list. A reader who goes to the body will not find it there. No wording change needed — "both from #331's review" is already precise about the source. |
| R4 | NIT, no change required | The heading is `## What's New`; every release here since v0.23.0 spells it `## What's new`. The gate matches case-insensitively and emitted no warning, so this passes as written — noted only for consistency. |

**Verdict: clean.** 0 blockers, 0 majors, 0 minors, 3 nits (R2-R4) none of which asks for a change, plus R1 as a recorded judgement. This round is **TERMINAL**: the remaining nits are follow-ups, not another round. The bump may be merged as admin and published from `6a2d0376f`, subject to one freshness condition — `origin/main` must still be `b1baf7450` at merge time. If `main` has moved by then, the window has grown and this PR is stale and must be re-cut.

damianvtran:
### Merge note — completion

Merged as admin at `e06336aaff74efa54209002c71a99a4a1765033c` (merge commit), **disclosed**: this is the window owner's own PR, and the completion used `--admin` rather than a review from a second account. Nothing was bypassed — `gh api repos/damianvtran/local-operator-ui/rules/branches/main` answers `[]`, so `main` carries no ruleset and no required check to clear. No human reviewer was requested and nobody was tagged (`reviewRequests: []` at every point in this PR's life).

**The gate this merge rests on, on head `6a2d0376f`:**

- **`### Agent review — round 1`** — clean and terminal: 0 blockers, 0 majors, 0 minors, 3 nits, no change requested. It verified the window is exactly the five merges with `origin/main` unmoved, the precondition (empty `package.json` diff since `v0.28.0`; `v0.28.1` neither tag nor Release), the class (`patch`, agreeing with the reversal recorded above), **byte-compared all five `Release:` lines against each PR's current body**, and re-ran this release's notes gate itself: `{refusal: null, warnings: []}`.
- **An independent read-only second pass**, which confirmed the diff is one line whose only parent is `b1baf7450`, that `origin/main` is still that SHA, that nothing is spent, and that four of five quotes match byte-for-byte with #347 matching in content once the fence is fixed. **The one defect it found was in this window's writeup, not in the diff** — the #347 quote carries its own backticks around `Local Operator`, so it is fenced with a double backtick rather than the template's single one; corrected in the notes before publication.

**Re-checked immediately before the merge:** `origin/main` was still `b1baf7450`, so the window had not grown and the bump was not stale.

**Next:** `v0.28.1` is published from this SHA as a **pre-release**, which is the hold that keeps `electron-updater`'s `/releases/latest` feed on the previous complete Release while the installers build. The pipeline owns promotion; there is no manual promote step here.

PR #363, opened 2026-09-18
refactor(chat): delete the legacy WebSocket/SSE streaming plane

damianvtran:
## Summary of Changes

`refactor(chat): delete the legacy WebSocket/SSE streaming plane` (`ff7951e0a`), its guard `test(renderer): guard the renderer tree and bundle against a socket transport` (`fc41f2d3f`), and a re-stamp (`276756c21`), on `origin/main` = `548f8dfed`.

The renderer carried two chat transports: the desktop-plane SSE relay the app actually streams a turn on, and a legacy socket client (`shared/api/local-operator/websocket-api.ts`, 1017 lines) behind an SSE-to-socket fallback router (`streaming-transport.ts`) with a legacy SSE client (`sse-api.ts`) beside it. The canonical-transcript cutover orphaned all three, and no gate in this repository can see that — `tsc`, `biome` and every unit suite are content with statically-imported dead code, and `pnpm lint`'s path list does not even name `scripts/`. So it sat there looking live, which is what this change removes.

12 files deleted, 4,688 lines of transport and renderer plus 744 lines of stories (4,800 with `docs/sse-migration.md`):

| deleted | why it was dead |
| --- | --- |
| `shared/api/local-operator/websocket-api.ts` | the socket client; constructed only by `streaming-transport.ts` |
| `shared/api/local-operator/streaming-transport.ts` | the transport router; constructed only by `use-websocket-message.ts` |
| `shared/api/local-operator/sse-api.ts` | the legacy `/v1/sse` client; constructed only by the router |
| `shared/hooks/use-websocket-message.ts` | called only by `use-streaming-message.ts` |
| `shared/hooks/use-streaming-message.ts` | called only by `streaming-message.tsx` |
| `shared/store/streaming-messages-store.ts` | read only by the five files above and the legacy list |
| `features/chat/components/messages-view.tsx` | rendered only by an unreachable `ChatContent` branch |
| `features/chat/components/loading-indicator.tsx` | rendered only by `messages-view.tsx` |
| `features/chat/components/message-item/streaming-message.tsx` | rendered only by `message-paper.tsx` and a story |
| `features/chat/components/sse-transport.stories.tsx` | fakes an `EventSource` and a `WebSocket` to drive the router |
| `features/chat/components/streaming-lifecycle.stories.tsx` | drives `StreamingMessage` and the store by hand |
| `docs/sse-migration.md` | describes the fallback this change removes |

[...]

PR #367, opened 2026-09-18
docs(agents): name the mode for staged-bundle rigs, and keep probes off the desktop

damianvtran:
## What this changes

Two additions to `AGENTS.md` § *Running the app without taking the operator's focus*, plus one clause, all earned tonight by windows that were **on the operator's screen while agents ran**:

1. **Staged-bundle rigs must name the mode.** The shape rule deliberately keeps a *packaged* launch `normal`, so a double-clicked app is never silenced — and the cost is that a rig booting a built `.app` out of `/tmp` (artifact smoke test, signature check, update rehearsal) has **no shape signal left to read** and resolves `normal` however it is piped and however it is spawned. The new paragraph says so and says to name `--window-mode=headless`.
2. **Probes and carrier scripts are not the app** — new subsection. A scratch script that boots Electron and constructs its own `BrowserWindow` inherits **none** of the window-mode guard, because the mode is resolved by this app's main process. The rule for that class is the probe's own: measure hidden (`win.isVisible()`, `win.isFocused()`, `BrowserWindow.getFocusedWindow()`, `capturePage()`, the app's log), **never loop `show()`/`focus()` variants** across the desktop, and where the showing *is* the measurement, do it once, `showInactive()` where that answers the question, and reap by exact pid in the same command.
3. **One-shot capture clause** in *Capturing the frame*: `inactive` is only for what cannot render hidden (native panels/sheets, compositor effects), and even then as **one self-contained launch → capture → reap** command. A visible window held across the steps of a run is indistinguishable, to the person whose screen it is on, from a leak.

## Why — the two launches that produced the report

Measured while the operator reported windows in front of them:

| live process | launch | why it was visible |
| --- | --- | --- |
| pid 80591 — the #363 QA rig, cwd `~/local-operator-ui/.worktrees/qa363r1` | `/…/electron . --remote-debugging-port=9565 --user-data-dir=/tmp/qa363-apprun/user-data` with `LOCAL_OPERATOR_UI_WINDOW_MODE=inactive`, driven by `/tmp/qa363-live-rig.mjs` | `inactive` is `showInactive()` — **visible by design**, which the table lists for "a run somebody wants to watch" |
| pids 80798 / 82270 — `mech.cjs` | `/…/electron /tmp/cfpasskey/probes/mech.cjs`, spawned from a loop over `"view on on" "view on off" "view off on" "view off off" "window on on"` | a standalone probe, **outside the app's guard**; its own `show: true` on the configs that show |

Both owning sessions were notified directly with the pids, the mode, and the fix. The app-side guard (`src/main/window-mode.ts`) is already on `main` and could not have prevented either: one run named a visible mode deliberately, and the other never loaded the app at all.

## Testing evidence

`AGENTS.md` is documentation, so this is prose, and the evidence is that it changes nothing else:

```
ℹ tests 34   ℹ pass 34   ℹ fail 0   ℹ skipped 0

$ stamp arithmetic at this head
srcTree     manifest=3771e20bc   HEAD:src=3771e20bc
scriptsTree manifest=7a7dd67b8   HEAD:scripts=7a7dd67b8
```

The diff is **42 insertions, 0 deletions**, the two inserted blocks and nothing else (`git diff --stat`). Neither `src/` nor `scripts/` moves, so no manifest re-stamp is required — the pins above are unchanged from `main`. No design round applies: nothing in this file renders to a user.

## Review status

Code review round pending (posted on this PR). Not merged.


damianvtran:
### Agent review — round 1
- Reviewer: reviewer subagent (deepseek/deepseek-flash), independent — did not author this diff
- Scope: `ef8ae2bfe5512ae17d294b2f9d8050f39a40006b..3c05fee26f8ec301eb90ce9d0c918aa1dd37243f` (merge-base of `origin/main` and head)
- Mode: Comprehensive (first round, so the whole 42-line insertion, not a remediation delta)
- Delivery contract: PR #367 — the two launch shapes that produced visible windows on 2026-09-18 (a staged/packaged rig resolving `normal`, and a standalone `BrowserWindow` probe outside the app's guard) must be unambiguously covered by `AGENTS.md` § *Running the app without taking the operator's focus*, without contradicting what the document already says about window modes
- Applicable risk surfaces: documentation accuracy, agent-facing guidance
- Verdict: **Approved**

No BLOCKER and no MAJOR. Both central factual claims are true against the code, the inserted prose is consistent with the surrounding document, and the PR's testing evidence reproduces. Findings below are a wording-scope fix and four small accuracy/scoping items; they are follow-ups, not a reason for another round.

| --- | --- | --- | --- |
| R1 | MINOR | Contradiction risk: the new sentence states a *global* rule for `inactive` ("that is the one case `inactive` is for") while two other places in the same section grant `inactive` for cases outside its dash-list, one of them the focus-dependent capture the same subsection tells you to do in `inactive` | `AGENTS.md:828` says `inactive` is for what cannot render hidden — "a native macOS panel or sheet, a compositor effect, a frame that exists only while a window is ordered front". Against it: `AGENTS.md:243` (table row: `inactive` = "a run somebody wants to watch or click into, and anything focus-dependent"), `AGENTS.md:294` ("name `--window-mode=inactive` when you want to watch a run"), and `AGENTS.md:744` ("For a change about those, drive it in `inactive` mode, or force focus with CDP …"), all unchanged and outside this subsection's scope. A reader capturing focus rings is told by one paragraph to use `inactive` and by the new one that `inactive` is for a different, explicitly enumerated set. Required action: scope the sentence to capture and add the focus-dependent frame to the list, e.g. "…that is the one case `inactive` is for — a frame that cannot render hidden, including the focus-dependent frame above". (Cheap fix; either reading leaves the reader on a safe path, since CDP focus emulation is offered at :744, which is why this is MINOR and not MAJOR.) |
| R2 | MINOR | Factual slip in the new `inactive` paragraph: it names `headless` as the mode its own log line reports, in the one case it is describing (`inactive`) | `AGENTS.md:832`: "the run's own `[window-mode]` line, which says out loud that it was `headless`, is not what they see." The mode line is formatted `window mode ${plan.mode}: …` (`src/main/window-mode.ts:855`), so an `inactive` run — the case this paragraph exists for — logs `window mode inactive: …`, never `headless`. The clause's actual point (the log line is not what the person at the screen sees) survives; naming `headless` is simply wrong for the run in hand and reads as though an `inactive` run logs `headless`. Required action: "…which says out loud which mode it named, is not what they see" (or name `inactive`). |
| R3 | MINOR | The cited measurement is described and prescribed in terms that do not match what the probe actually did, so a reader with the same rig may not recognise themselves in the rule | The evidence sentence (`AGENTS.md:842-844`) says "a probe iterating visibility configs (`view on/off`, `window on/off`)" and the rule (`:848-854`) says "never iterate `show()`/`focus()` variants in a loop… bound it to ONE launch". Checked `/tmp/cfpasskey/probes/mech.cjs` (5,267 B): its knobs are `MODE=view\|window`, `THROTTLE`/`WTHROTTLE=on\|off` and `RAISE=yes\|no`; it constructs its own `BrowserWindow({show:false})` and calls `win.showInactive()` + `setAlwaysOnTop(true,"floating")` **once per process** (line 124), gated on `RAISE` which defaults to `yes`. The sweep was five separate processes (PR body loops `"view on on" "view on off" "view off on" "view off off" "window on on"`), i.e. the varied knobs are throttling configs, not visibility flags, and no launch ever iterated `show()`/`focus()` in-process — each single launch already *was* one launch. Required action: describe the pattern as measured — a sweep of process-level configs in which every run showed its own window (the default), fixed by measuring hidden (`RAISE=no`) or by not showing — so the rule covers the five-process case it cites rather than only an in-process variant loop. |
| R4 | NIT | The probe rule recommends a measurement the document itself has measured as misleading | `AGENTS.md:848` lists `win.isFocused()` among the readings to take "with the window hidden". `AGENTS.md:721-727` records the counterexample: on macOS `NativeWindowMac::Show()` calls `activateIgnoringOtherApps:YES` for every non-panel window whatever `focusable` says, so a shown window leaves the app frontmost while `isFocused()` reads false — "Do not reach for `win.isFocused()` as the proof of that". Suggest dropping it from the list, or annotating that it cannot answer a focus question from a hidden window. |
| R5 | NIT | Over-broad scoping in the staged-bundle paragraph | `AGENTS.md:275-276`: "so it stays `normal` however it is piped and however it is spawned". True for the shape signal, which is what the preceding clause scopes it to, but a bundled launch with no mode named and a scratch `--user-data-dir`/`--remote-debugging-port` still resolves `headless`: the `AGENT_LAUNCH_FLAGS` path does not consult `packaged` (`src/main/window-mode.ts:388-392` and `:429-437`; asserted by scripts/window-mode.test.mjs, "an agent switch is the stronger reason"). Add "with no switch either" for exactness. |

### What I verified, and how

- **The packaged claim is true.** `src/main/index.ts:169` passes `packaged: app.isPackaged` into `resolveWindowLaunchPlan`; the resolver requires `input.packaged === false` for the driven assumption (`src/main/window-mode.ts:429-433`), so a bundled launch resolves `normal` with `assumed: null` unless a mode is named or an agent switch is present — `scripts/window-mode.test.mjs:792` ("the packaged app is never assumed headless, however it was started") pins it. `app.isPackaged` describes how the process was launched (a bundled app), not where the bundle sits, so a `.app` staged in `/tmp` is packaged, exactly as the new paragraph says. It is consistent with the earlier paragraph (`AGENTS.md:251`, the double-clicked `.app` "is never assumed headless, whatever its streams look like") and correctly distinct from `AGENTS.md:244`'s "a QA matrix boots a copy of one under `/tmp`" (a copy of `out/main/index.js`, unpackaged, so it *is* assumed headless).
- **The probe claim is true, and it is this document's business.** The mode is resolved at launch in this app's main process (`src/main/index.ts:152-171`) and `src/main/window-mode.ts` is policy-only ("deliberately imports nothing from Electron"); `src/main/window-raise.ts` is the only main-process module that calls `show`/`showInactive`/`focus` (`AGENTS.md:724-727`, asserted by `scripts/window-mode.test.mjs`). A `/tmp` script with its own `new BrowserWindow()` loads none of that, so there is no guard to inherit — the paragraph's mechanism is right, not merely plausible. Nothing in the repo already covers it: `grep` for `carrier\|scratch script\|own BrowserWindow` across `AGENTS.md` and `docs/*.md` returns zero hits, and no in-repo rig loops visibility (`showInactive(`/`.show()` in `scripts/*.mjs` appear only in comments and Electron's `Notification` API). A new `###` sibling under this section — rather than under bundling, releases or the evidence manifest — is the right home: the rule is about not taking the operator's focus.
- **Testing evidence reproduces.** `node --test scripts/evidence-manifest.test.mjs` → `tests 34 / pass 34 / fail 0`. Stamp arithmetic: manifest `srcTree=3771e20bc3053ac9f54dc74ee9739df993a619ca` == `git rev-parse HEAD:src` == `git rev-parse origin/main:src`, and `scriptsTree=7a7dd67b8523bbaca8daeddfd410adb0b93ec1d5` == `git rev-parse HEAD:scripts` — unchanged from `main`, so no re-stamp is owed. `git diff --numstat ef8ae2bfe..HEAD` = `AGENTS.md 42 0`: additive prose only, no deletions. I also checked the insertions left no construct broken: 0 added code fences (file's fence count is even at 18) and no fence inside either inserted block, no added table row, no trailing whitespace, single blank lines between the new paragraphs. As the closest doc guard I additionally ran `node --test scripts/mentioned-files.test.mjs` → 54 pass (the new prose names no path that does not exist).
- **The pids themselves are no longer checkable; the artifacts are.** 80591 / 80798 / 82270 are gone from the process table, so the live process claim cannot be re-verified. What remains matches the body: `/tmp/qa363-live-rig.mjs` (15,368 B) sets `LOCAL_OPERATOR_UI_WINDOW_MODE: "inactive"` (line 70) and passes `--remote-debugging-port`, `--user-data-dir` and `--window-mode=inactive` — "visible by design" is exactly what that argv and mode resolve to; `~/local-operator-ui/.worktrees/qa363r1` exists; and `/tmp/cfpasskey/probes/mech.cjs` exists, is a standalone Electron script with its own `BrowserWindow`, its page `<title>` is `mechanism`, and it raises that window `alwaysOnTop(true,"floating")` by default — matching "a window titled after the mechanism itself in front of the operator". One claim is asserted and not checkable from the box: that the window was *on the operator's screen* at the cited instant (no process or log record survives); the argv, mode and always-on-top facts behind it all check out.
- **Skipped, and why.** No Electron launch and no `.app` staging (the task says neither is needed, and the box is ~98% full); the `app.isPackaged === true` value for a staged bundle is taken from Electron's documented semantics plus the resolver's explicit gate rather than from a booted bundle. I ran two focused suites against the changed surface instead of the repository's full matrix, per the round-2+ scoping rule applied to a docs-only change.

Round 1 is TERMINAL: no BLOCKER and no MAJOR remain, so R1-R5 are follow-ups at the author's discretion and do not require another round.

damianvtran:

- **R1 (minor)** — Fixed in `$(git rev-parse --short=7 HEAD)`: the `inactive` paragraph now names both cases (a native surface that cannot render hidden, and focus-dependent rendering) and points at *``headless`` is a full-fidelity rendering path, not a degraded one*, instead of asserting one case that contradicted the mode table and the two focus bullets.
- **R2 (minor)** — Fixed in `$(git rev-parse --short=7 HEAD)`: the `[window-mode]` line is described accurately for the case the paragraph is about — it names the mode it resolved (`window mode inactive`) — rather than claiming it says `headless`.
- **R3 (minor)** — Fixed in `$(git rev-parse --short=7 HEAD)`: the cited measurement is restated as what it was — five visibility configurations swept as **five processes**, each launching Electron once and calling `showInactive()` once — with the matrix, not any single launch, making the noise; the prescription is restated as three rules about the sweep (measure hidden; a sweep of N configurations is not N launches; never loop `show()`/`focus()` variants across the desktop).
- **R4 (minor)** — Fixed in `$(git rev-parse --short=7 HEAD)`: `win.isFocused()` is no longer offered as a hidden measurement; the text now carries the measured warning from *``headless`` is a full-fidelity rendering path, not a degraded one*.
- **R5 (minor)** — Fixed in `$(git rev-parse --short=7 HEAD)`: a staged `.app` no longer "stays `normal` however it is spawned" — the switch signal is packaging-blind, so a scratch `--user-data-dir`/`--remote-debugging-port` still resolves `headless`, and the paragraph now says to pass one of those or name the mode outright.

No blocker and no major was raised, so the round was terminal; round 2 is requested because these were **prose defects rather than polish** — two of them (R2, R5) stated facts wrongly, and this file is what other agents follow. Nothing is merged.

damianvtran:
### Agent review — round 2
- Reviewer: reviewer subagent (deepseek/deepseek-flash), independent — did not author this diff
- Scope: `ef8ae2bfe5512ae17d294b2f9d8050f39a40006b..ba8731b7c` (merge-base of `origin/main` and head)
- Delta reviewed: `3c05fee26..ba8731b7c`
- Applicable risk surfaces: documentation accuracy
- Verdict: **Approved**

The delta is prose-only — `git diff --numstat 3c05fee26..ba8731b7c` is `40 30 AGENTS.md`, one file, with no `src/` and no `scripts/` movement — and this file carries no markdown links or HTML anchors (the only references are italicised heading names), so there is no anchor to break. I checked each such reference the delta touches.

#### The five round-1 corrections, each against the artefact it cites

- **R1 — target right, wording still wrong twice.** The heading named is the right one: *``headless`` is a full-fidelity rendering path, not a degraded one* (`AGENTS.md:700`) is where the focus-dependent consequences live ("Focus-dependent rendering differs", "Focus-dependent behaviour differs too"), so naming focus-dependent rendering as a second `inactive` case now agrees with those bullets instead of contradicting them. Two defects survive the rewrite — **R2-1** (the reference says *below* for a heading 131 lines above it) and **R2-2** (the new count of "two things" still excludes the mode table's primary `inactive` use).
- **R2 — accurate.** `src/main/index.ts:217` prints `[window-mode] ${line}` and `src/main/window-mode.ts:855` builds that line as `` `window mode ${plan.mode}: ${width}x${height}, …` ``, so an `inactive` run logs `[window-mode] window mode inactive: …` — exactly the string the paragraph now quotes. The claim that the line "says out loud that it was `headless`" is gone.
- **R3 — accurate, and the prescription now covers separate launches.** The five/five restatement matches the artefacts: the loop is documented in this PR's own body (`"view on on" "view on off" "view off on" "view off off" "window on on"`, spawned as five `/tmp/cfpasskey/probes/mech.cjs` processes), and five scratch profiles stamped with the loop's own names exist — `/tmp/mech-view-on-on` 13:48:12, `/tmp/mech-view-on-off` 13:48:18, `/tmp/mech-view-off-on` 13:48:31, `/tmp/mech-view-off-off` 13:48:50, `/tmp/mech-window-on-on` 13:50:14 — one profile per launch, ~2 min of wall time. The probe itself supports "launching Electron once, calling `showInactive()` once, and exiting": one `BrowserWindow` at `show: false`, a single `win.showInactive()` (line 124, inside `if (RAISE)` whose default is `yes` at line 27, and no loop string sets it), then `app.quit()` (line 143). Its page `<title>` is `mechanism` (line 33) and `RAISE` also sets `alwaysOnTop(true, "floating")`, so "titled after the mechanism itself" and "in front of the operator" hold too. The rewritten rule is no longer only about an in-process loop: bullet 2 is explicitly about "N configurations … not N launches" and adds "reap by exact pid in the same command" plus announcing the window count, which is what a matrix of separate launches needs.
- **R4 — accurate.** `win.isFocused()` is gone from the recommended hidden measurements (`AGENTS.md:854-856`) and replaced by a pointer to the measured warning at `:723-727`, which is the right passage: it records that on macOS `NativeWindowMac::Show()` calls `activateIgnoringOtherApps:YES` for every non-panel window, so a `show()` on a `focusable: false` window still makes the app frontmost while `isFocused()` reads false. That warning applies to a probe's own window exactly as it does to the app's.
- **R5 — accurate.** In `src/main/window-mode.ts`, `unnamed = !modeFlag.found && modeRaw === undefined`, then `agentFlag` is resolved from `AGENT_LAUNCH_FLAGS` with no reference to `packaged`, while `packaged === true` only blocks the `driven` branch (`agentFlag === undefined && input.packaged === false && …`). So shape alone cannot hide a packaged launch — correct — and both `--user-data-dir` and `--remote-debugging-port` still resolve `headless` for a bundled launch, identically to a checkout. The old "stays `normal` however it is spawned" claim is gone.

| --- | --- | --- | --- |
| R2-1 | MINOR | A cross-reference that resolves to the right heading but sends the reader the wrong way, in a file that uses these direction words as fact everywhere else | `AGENTS.md:831`: "which *``headless`` is a full-fidelity rendering path, not a degraded one* **below** handles" — that heading is at `AGENTS.md:700`, **131 lines above** the reference. Correct controls in the same file: `:20` "*…the operator's focus* below" → `:226`, and `:100` "*Change scope* below" → `:1085`. Required action: say "above", or drop the direction ("…, which *``headless`` is a full-fidelity rendering path, not a degraded one* covers"). |
| R2-2 | MINOR | R1's underlying defect survives in reduced form: the sentence is still a global count of what `inactive` is for, and it is still one short of the table | `AGENTS.md:829`: "that is one of the two things `inactive` is for; the other is focus-dependent rendering". The mode table at `:243` gives `inactive`'s use as "a run somebody wants to watch or click into, and anything focus-dependent", and `:294` says "name `--window-mode=inactive` when you want to watch a run" — watching/clicking is a third use the count omits, and it is the use this PR's own `#363` rig took (`LOCAL_OPERATOR_UI_WINDOW_MODE=inactive`, "visible by design"). Round 1's required action was to scope the sentence to capture; instead the count was raised from one to two. Required action: drop the count — "that is one case `inactive` is for, and the focus-dependent frame under *``headless`` is a full-fidelity rendering path, not a degraded one* is another" — so a capture paragraph stops enumerating a mode's uses. |
| R2-3 | MINOR | The two rules the rewrite added contradict each other on the exact shape the paragraph documents, so the next agent gets two answers for one situation | Bullet 2 (`AGENTS.md:859-862`): "where a shown window is genuinely required per configuration, prefer `showInactive()` over `show()`, reap by exact pid in the same command, and say in advance how many windows the sweep will put up" — permits an announced N-launch sweep with a window per configuration. Bullet 3 (`:863-866`): "If a probe cannot answer its question hidden, do it once and report the single measurement rather than sweeping a matrix in front of somebody who did not ask for it" — forbids precisely that. The cited incident is bullet-2-shaped (five separate launches, one `showInactive()` each), so the same run is blessed by one bullet and condemned by the next. Required action: scope bullet 3 to the search loop ("never iterate `show()`/`focus()` variants *to find the one that reproduces*") and let bullet 2 own the deliberate, announced, reaped sweep. |
| R2-4 | MINOR | The R5 rewrite dropped an allowance the paragraph used to carry, and now reads as an exhaustive dichotomy that does not cover a staged-bundle capture which cannot render hidden | `3c05fee26` had "name `--window-mode=headless` (or `inactive` for the one capture that cannot render hidden, per *Capturing the frame*)"; `ba8731b7c` has only "either passes one of those or names `--window-mode=headless` outright". Whether a native panel/sheet can render hidden has nothing to do with packaging, and § *Capturing the frame* still grants `inactive` for it, so this is lost guidance rather than a shortening. Required action: restore the parenthetical. |
| R2-5 | NIT | The measurement list names an instrument the class it governs does not have | `AGENTS.md:854-856` recommends "the app's own `[window-mode]` line" for probes and carrier scripts, two sentences after the class is defined as scripts that "never load [the app's main process]" (`:839-843`). The line exists only for the carrier sub-class that does load it. Required action: "and, for a carrier that loads the app's main process, its `[window-mode]` line". |

#### Follow-up observations (outside the delta)

- **"Five" is a floor, not a total — no correction needed.** `/tmp/mech-*` holds twelve scratch profiles, three of them (`mech-v1-on-top` 13:48:44, `mech-v2-off-top` 13:49:45, `mech-v3-win-on-top` 13:50:56) interleaved inside the sweep's own 13:48:12-13:50:14 interval. They are the same class and the same shape, so the paragraph's argument only gets stronger; the sentence is accurate as written because it describes the documented loop.
- **The PR body still describes the pre-review text.** It argues staged bundles have "no shape signal left to read" and resolve `normal` "however it is piped and however it is spawned", and lists `win.isFocused()` among the probe measurements — the three claims R1-R5 just corrected. `AGENTS.md` is fixed; the body, which is what a reader of this PR sees first, is not. A body edit, not a shipping defect.
- **Nothing in `src/` or `scripts/` is affected**, so no manifest re-stamp is implied and the round-1 statement about evidence still holds.

#### What I ran, and what I skipped

Ran: `git diff --numstat` / `git diff` / `git log` over `3c05fee26..ba8731b7c`; `gh pr view 367 --json body,headRefOid,mergeStateStatus`, `gh pr view 367 --comments`, `gh pr checks 367` (Change Scope and Version Bump Guard pass; the rest skip, which is this diff's expected path — no further claim); reads of `src/main/window-mode.ts` (`AGENT_LAUNCH_FLAGS`, `unnamed`/`agentFlag`/`driven`, the `window mode ${plan.mode}` format) and `src/main/index.ts` (`[window-mode]` prefix); and the artefact trail: `/tmp/cfpasskey/probes/mech.cjs` and `/Users/damian/scratchpad/mech.cjs` read in full and diffed against each other, the twelve `/tmp/mech-*` profiles with second-resolution mtimes, `/tmp/cfpasskey/FINDINGS.md`.

Skipped, and why: no Electron launch and no `.app` staging (the task says neither is needed, the box is ~98% full on this shared machine, and a probe boot would put a window on the operator's screen — the very failure this section documents); no `hdiutil`; and no repository test matrix, which a one-file markdown delta cannot move. Every check above is a read or a git/gh query, so no finding here rested on a run I did not make.

Round 2 is TERMINAL: no BLOCKER and no MAJOR remain. R2-1 to R2-5 are follow-ups for the author to take or decline, and they are not a reason for another round. Nothing merged, no reviewer added, no push.

damianvtran:
### Agent review — round 2
- Reviewer: reviewer subagent (deepseek/deepseek-flash), independent — did not author this diff
- Scope: `ef8ae2bfe5512ae17d294b2f9d8050f39a40006b..ba8731b7c` (merge-base of `origin/main` and head)
- Delta reviewed: `3c05fee26..ba8731b7c`
- Applicable risk surfaces: documentation accuracy
- Verdict: **Approved**

> **Process note, so remediation is unambiguous.** A round-2 review of this same delta is already posted at this head — comment `5734200053` (posted 18:09Z, verdict Approved, findings numbered `R2-1`..`R2-5`). This is an independent second pass at the same head, made before reading that one; it converged on the same five items, so the two sets are the same findings restated and one batched reply can cover them. To keep IDs from colliding, this comment's findings are numbered `R2b-n`.

**Delta shape, verified.** `git diff --numstat 3c05fee26..ba8731b7c` is `40	30	AGENTS.md` — one file, prose only, no `src/` and no `scripts/` movement (so no manifest re-stamp is implied), and the delta adds and removes no markdown link, anchor or `](#…)` reference, so there is no anchor it can break. The only cross-references it touches are italicised heading names, checked below.

#### The five round-1 corrections, each against the artefact it cites

- **R1 — the heading named is the right target; the reference still points the wrong way.** `AGENTS.md:700` § *``headless`` is a full-fidelity rendering path, not a degraded one* is where the focus-dependent consequences are actually enumerated ("Focus-dependent rendering differs", "Focus-dependent behaviour differs too" below it), so naming focus-dependent rendering as a second `inactive` case now agrees with `AGENTS.md:243`'s `inactive` row ("…and anything focus-dependent") rather than contradicting it. Two defects survive the rewrite: **R2b-1** (the reference says *below* for a heading 130 lines above it) and **R2b-2** (the "two things" count is still unqualified).
- **R2 — accurate.** `src/main/index.ts:217` prints `` `[window-mode] ${line}` `` and `src/main/window-mode.ts:855` formats that line as `` `window mode ${plan.mode}: ${width}x${height}, …` ``, so an `inactive` run logs `[window-mode] window mode inactive: …` — exactly the string quoted at `AGENTS.md:834-835`. The old, wrong claim that the line "says out loud that it was `headless`" is gone.
- **R3 — accurate, and the prescription now covers separate launches rather than only an in-process loop.** The five/five restatement matches the artefacts: this PR's own body documents the loop (`"view on on" "view on off" "view off on" "view off off" "window on on"`, each spawned as its own `electron /tmp/cfpasskey/probes/mech.cjs` process), five scratch profiles stamped with those names exist (`/tmp/mech-view-on-on` 13:48, `-view-on-off` 13:48, `-view-off-on` 13:48, `-view-off-off` 13:48, `-window-on-on` 13:50, one profile per launch, ~2 min of wall time for the sweep), and the probe itself supports "launching Electron once, calling `showInactive()` once, and exiting": one `BrowserWindow` at `show: false`, a single `win.showInactive()` (`/tmp/cfpasskey/probes/mech.cjs:123-124`, inside `if (RAISE)` whose default is `yes` at line 27, and no string in the loop sets it), then `app.quit()`. Its page `<title>` is `mechanism` (line 33) and `RAISE` also sets `alwaysOnTop(true, "floating")`, so "titled after the mechanism itself" and "in front of the operator" both hold. Bullet 2's "N configurations … not N launches" is aimed at the shape that actually happened, which is what the pre-review text got wrong.
- **R4 — accurate.** `win.isFocused()` is gone from the recommended hidden measurements (`AGENTS.md:854-856`) and replaced with a pointer to the measured warning at `AGENTS.md:723-727`, which is the right passage: it records that on macOS `NativeWindowMac::Show()` calls `activateIgnoringOtherApps:YES` for every non-panel window, so a `show()` on a `focusable: false` window makes the app frontmost while `isFocused()` keeps reading false. That applies to a probe's own window exactly as it does to the app's, so the substitution is sound.
- **R5 — accurate against the code.** The paragraph now says shape alone resolves a staged bundle `normal` "however it is piped" and that the switch signal still applies (`AGENTS.md:273-282`). Verified in `src/main/window-mode.ts`: `agentFlag` is computed for any launch that names no mode, independently of packaging (`:387-390`), `driven` additionally requires `input.packaged === false` (`:427-433`), and the mode is chosen by `const mode = parsedMode ?? (assumed ? "headless" : "normal")` (`:473`). So a bundled launch with a scratch `--user-data-dir` or a `--remote-debugging-port` does resolve `headless` — and "however it is piped" is safe precisely because piping is only read on the non-packaged branch. The dropped allowance is reported as R2b-4.

| --- | --- | --- | --- |
| R2b-1 | MINOR | The rewritten cross-reference points the wrong way, so the reader it was written for lands in the wrong section | `AGENTS.md:830-831`: "focus-dependent rendering, which *``headless`` is a full-fidelity rendering path, not a degraded one* **below** handles." That heading is at `AGENTS.md:700`, ~130 lines **above** the paragraph (the paragraph sits inside § *Capturing the frame*, `:809`), and the string appears nowhere else in the file — so "below" sends the reader down into § *Probes and carrier scripts are not the app* looking for it. Required action: drop the direction word (or say "above"), e.g. "…which *``headless`` is a full-fidelity rendering path, not a degraded one* covers". |
| R2b-2 | NIT | "one of the two things `inactive` is for" reads as the complete list of what the mode is for, while the mode table lists a third use first | `AGENTS.md:829`. `AGENTS.md:243` gives `inactive` for "a run somebody wants to watch or click into, and anything focus-dependent", and `:295` and `:745` repeat the watch-a-run use. The sentence is capture-scoped by its opening clause, so no contradiction is created — but the unqualified count is what a reader greps for. Required action: "…one of the two things a **capture** needs `inactive` for". |
| R2b-3 | MINOR | Bullet 2 permits the exact shape the paragraph above calls noise, and bullet 3 forbids it — the next agent gets two answers, and the permissive one is the one that repeats the incident | `AGENTS.md:859-862`: "where a shown window is genuinely required per configuration, prefer `showInactive()` over `show()`, reap by exact pid in the same command, and say in advance how many windows the sweep will put up" — i.e. an announced N-window sweep is allowed. `AGENTS.md:863-866`: "If a probe cannot answer its question hidden, do it once and report the single measurement rather than sweeping a matrix in front of somebody who did not ask for it" — i.e. exactly that sweep is forbidden. The incident at `:844-850` is bullet-2-shaped (five launches, one `showInactive()` each, one window per configuration), so the same run is blessed by one bullet and condemned by the next; announcing five windows does not stop them landing on somebody's screen. Required action: make the allowance subordinate rather than parallel — a shown window only where the shown state **is** the measurement, as `inactive`, one launch where the question allows it, reaped in the same command — and scope bullet 3 to the search loop ("never iterate `show()`/`focus()` variants *to find the one that reproduces*"). |
| R2b-4 | MINOR | The R5 rewrite turned the paragraph into an exhaustive dichotomy and dropped the `inactive` allowance it used to carry, for a capture where packaging is irrelevant | `3c05fee26` read "name `--window-mode=headless` (or `inactive` for the one capture that cannot render hidden, per *Capturing the frame*)"; `ba8731b7c:280-281` now reads "either passes one of those or names `--window-mode=headless` outright". Whether a native panel or sheet can render hidden has nothing to do with how the bundle was launched, and § *Capturing the frame* (`:827-835`) still grants `inactive` for it — so an artifact smoke test that must show a real sheet now reads as having no sanctioned path. Required action: restore the clause, e.g. "…or names `--window-mode=headless` outright (or `inactive` for a capture that cannot render hidden — see *Capturing the frame*)". |
| R2b-5 | NIT | The hidden-measurement list offers an instrument the class it governs does not have | `AGENTS.md:854-856` recommends "the app's own `[window-mode]` line" for probes and carrier scripts, two sentences after the class is defined as scripts that "never load [the app's main process]" (`:841-843`). That line exists only for the carrier sub-class that does load it. Required action: "…and, for a carrier that loads the app's main process, its `[window-mode]` line". |

#### Follow-up observations (outside the delta)

- **"Five" is the documented loop's count, not the artefact total — no correction needed.** `/tmp/mech-*` holds twelve scratch profiles: the five the loop named, three from an earlier matrix (`mech-m-view-off-on` 13:42, `mech-m-view-on-off` 13:43, `mech-m-view-on-on` 13:41, `mech-m-win-on-on` 13:44), and three interleaved inside the sweep's own window (`mech-v1-on-top` 13:48, `mech-v2-off-top` 13:49, `mech-v3-win-on-top` 13:50). Same class, same shape, so the paragraph's argument only gets stronger.
- **The two copies of the probe do not agree, which is worth knowing if a reader goes to check the sentence.** `/tmp/cfpasskey/probes/mech.cjs` guards `showInactive()` on `RAISE` (default `yes`); `~/scratchpad/mech.cjs` calls `win.showInactive()` unconditionally. The shipped claim ("calling `showInactive()` once") is true of both, so nothing needs changing — but the RAISE guard the wording implies is only in the `/tmp` copy.
- **The PR body still carries the three pre-review claims** — staged bundles have "no shape signal left to read" and resolve `normal` "however it is piped and however it is spawned", and `win.isFocused()` is listed among the probe measurements. `AGENTS.md` is fixed; the body a reviewer reads first is not. A body edit, not a shipping defect.
- **Nothing in `src/` or `scripts/` moves**, so the manifest pins quoted in the body hold and no design round applies.

#### What I ran, and what I skipped

Ran: `git diff` / `git diff --numstat` / `git log` over `3c05fee26..ba8731b7c`; `gh pr view 367 --json body,headRefOid,mergeStateStatus` and `gh api repos/damianvtran/local-operator-ui/issues/367/comments` (this is how the pre-existing round-2 comment surfaced); reads of `src/main/window-mode.ts` (`AGENT_LAUNCH_FLAGS`, `unnamed`/`agentFlag`/`driven`, the `parsedMode ?? (assumed ? …)` choice, the `window mode ${plan.mode}` format) and `src/main/index.ts` (`[window-mode]` prefix, the two print sites); `/tmp/cfpasskey/probes/mech.cjs` and `~/scratchpad/mech.cjs` read and diffed against each other; `ls -l /tmp/mech-*` profile stamps.

Skipped, and why: no Electron launch, no `hdiutil` and no `.app` staging (neither is needed for a prose delta, this box is ~98% full, and a probe boot would put exactly the window this section documents on the operator's screen); no repository test matrix, which a one-file markdown delta cannot move; and no recursive walk of `/tmp/cfpasskey` — its `node_modules` tree made a depth-3 walk time out, so I read the probe by direct path. Every finding above rests on a read or a git/gh query I actually made.

Round 2 is TERMINAL: no BLOCKER and no MAJOR remain, and R2b-1..R2b-5 are follow-ups the author can take or decline in one batched commit, not a reason for another round. Nothing merged, no reviewer added, no `@`-mention, no push.

damianvtran:

Two round-2 comments exist at head `ba8731b7c` — `5734200053` and `5734270442` — the second from a job that reported a harness `database is locked` and had in fact posted. Both are Approved/terminal. I am **keeping both** rather than deleting one: review evidence is not something an author tidies away. This answers the **union** (`R2-1`…`R2-5`, `R2b-1`…`R2b-5`).

- **R2b-1 (minor)** — Fixed in `d4dc178f9`: the cross-reference now reads **above** (the heading is at `AGENTS.md:700`, ~130 lines above the paragraph) instead of "below".
- **R2b-2 (nit)** — Fixed in `d4dc178f9` by the same rewrite: the paragraph no longer asserts a "two things" set it does not hold; it names the cases it means and defers to that heading for focus-dependent rendering.
- **R2b-3 (minor — the contradiction)** — Fixed in `d4dc178f9`: bullets 2 and 3 no longer bless and condemn the same shape. They are now split by the **kind of measurement**: a *visibility* sweep is permitted when every cell is `showInactive()`, reaped by exact pid in the same command as its capture, with the count announced before it starts; a *focus-taking* sweep (`show()`/`focus()`, repeated `activateIgnoringOtherApps:YES`) is never acceptable, because there the interruption is a side effect of the measurement rather than its subject.
- **R2b-4 (minor)** — Fixed in `d4dc178f9`: the staged-bundle paragraph's "either … or" had dropped the `inactive` allowance for a bundled capture that cannot render hidden; restored, with the pointer to *Capturing the frame*.
- **R2b-5 (nit)** — Fixed in `d4dc178f9`: the app's `[window-mode]` line is scoped to carriers that actually load the app; a bare probe is no longer told to read it.
- **R2-3, R2-4, R2-5 and the round-1 set** were answered in the round-1 remediation and both round-2 reviewers independently confirmed those corrections accurate.

Round 3 is requested on the delta `ba8731b7c..d4dc178f9`, because these were **contradictions in agent-facing guidance rather than polish** — a file that tells an agent to sweep and not to sweep in the same breath is worse than one that says nothing. Nothing is merged.

damianvtran:
### Agent review — round 3
- Reviewer: reviewer subagent (deepseek/deepseek-flash)
- Scope: `ef8ae2bfe`..`d4dc178f9`
- Delta reviewed: `ba8731b7c`..`d4dc178f9`
- Applicable risk surfaces: documentation accuracy
- Verdict: Approved

| --- | --- | --- | --- |
| R3-1 | MINOR | The rewrite widened a universal quantifier until it now covers an interactive run the document elsewhere says is *not* reaped by a rig. | `AGENTS.md:830-833` — "`inactive` is the mode for it, as it is for **a run somebody means to watch or click into** and for focus-dependent rendering (…*above*). **In every one of those cases** it is one self-contained command: launch, capture, reap by exact pid before it returns." The antecedent set was two capture cases in `ba8731b7c` ("In both cases"); it is now three, and the third is a person's run, not a capture. Directly contradicted at `AGENTS.md:618-621` ("an `inactive` run is ended by a person or by its own window, not by the watch") and consistent with the source policy: `src/main/window-mode.ts` `resolveLauncherWatchPlan` returns `window mode inactive is not launcher-bound: a person can close it`. Required action (non-blocking): scope the sentence to the captures — e.g. "In the capture cases it is one self-contained command: launch, capture, reap by exact pid before it returns — while a run somebody means to watch is the app's own window and is not reaped by a rig." |
| R3-2 | NIT | Bullet 2 opens by indicting the very shape it then licenses, which is the pattern round 2 asked to be made unambiguous. | `AGENTS.md:863-869`: "A matrix whose every cell shows a window **is the incident above**" reads as a prohibition, then two sentences later the same bullet permits a shown-per-cell matrix (`showInactive()` + reap per cell + announced count). The permission is intended and the incident narrative (`AGENTS.md:838-848`) blames the *unannounced, unbounded* matrix rather than the method, so make sentence 1 diagnose rather than forbid: "The matrix is what made the incident above noise." |
| R3-3 | NIT | Missing direction on a forward heading reference, where the sibling reference two sections up states its direction. | `AGENTS.md:282`: "(*Capturing the frame*)" points *forward* to the heading at `AGENTS.md:810`; the neighbouring reference at `AGENTS.md:831` ends with "**above**". Add "**below**", matching the existing convention at `AGENTS.md:100` ("*Change scope* below."). |

#### Verification of the five corrections

1. **The contradiction is gone — and no reading blesses an N-window focus sweep.** The split is now decidable from the bullets alone, and it is drawn on the axis round 2 asked for: method and purpose. Permitted: `showInactive()` per cell, reaped by exact pid in the same command as its capture, count announced before starting (`AGENTS.md:863-869`). Never permitted: `show()`/`focus()` sweeps of any size, "whatever it measures", with a single focus measurement as the escape hatch for a question that genuinely needs focus (`AGENTS.md:870-875`). Bullet 3's second sentence ("Visibility that is only announced and reaped is one thing; the focus-taking half is another") names the distinction explicitly, so bullet 2's licence and bullet 3's prohibition no longer cover the same sweep. The only residue is the framing in R3-2, which affects readability, not the permission.
2. **The cross-reference is in the right direction and does contain what is deferred to it.** `AGENTS.md:830-831` now says "**above**", and `### `headless` is a full-fidelity rendering path, not a degraded one` sits at `AGENTS.md:701`, `~130` lines earlier. It carries the focus-dependent material the paragraph defers to it for: "**Focus-dependent rendering differs.** … drive it in `inactive` mode, or force focus with CDP `Emulation.setFocusEmulationEnabled(true)` — and say which you did" and "**Focus-dependent behaviour differs too**" (`AGENTS.md:744-756`). The other emphasized reference this delta adds ("*Capturing the frame*", `AGENTS.md:282`) also resolves to a real heading (`AGENTS.md:810`) that genuinely states the case it is cited for: a staged bundle's capture that cannot render hidden takes `inactive`.
3. **The staged-bundle paragraph reaches its case without re-opening the `normal`-by-shape error.** `AGENTS.md:271-272` still reads "shape alone resolves it `normal` **however it is piped**", and `AGENTS.md:279-281` keeps the packaging-blind switch signal ("The switch signal is packaging-blind and still applies"). The restored `inactive` is reached only through the escape the paragraph already had — naming the mode, which wins over the switch-derived assumption as documented at `AGENTS.md:296-298` — and it is scoped to the one case that needs it ("`headless` for a run, `inactive` for the one capture that cannot render hidden"), so a bundled *run* is still steered to `headless`. No regression.
4. **R2b-5 is fixed.** `AGENTS.md:856-862` no longer tells a bare probe to read the app's line: the line is now conditioned on "a carrier that DOES load this app", with "which a bare probe has no equivalent of and must not be told to read". That agrees with the class's own premise at `AGENTS.md:841-845` — the mode is resolved by this app's main process, so a probe that never loads it inherits nothing. The capitalised "DOES" carries the distinction against the intro's "carrier hosting a fragment of the app", so I read the scoping as decidable.
5. **The delta is prose only, and the round-1 corrections survive it.** `git diff --numstat ba8731b7c..d4dc178f9` is `31 23 AGENTS.md` — one file, no headings, links or anchors added or moved, so nothing it relies on can be broken. Spot-checks: the mode table (`AGENTS.md:240-244`) is untouched and agrees with the new prose ("a run somebody wants to watch or click into, and anything focus-dependent"); the `window mode inactive` string matches the emitter `window mode ${plan.mode}: …` (`src/main/window-mode.ts:891`, reachable from `src/main/index.ts:212-221`); "swept five visibility configurations as **five processes**" (`AGENTS.md:847`) is intact and consistent with bullet 2's "five windows"; `win.isFocused()` appears only inside prohibitions (`AGENTS.md:724`, `728`, `860`), never as a measurement; and the packaging-blind switch signal is intact as quoted above. No earlier correction was reverted by this rewrite.

#### Follow-up observations

- Pre-existing, not from this PR: the comment above `WINDOW_MODES` in `src/main/window-mode.ts:42-43` claims the array is "in the order they are documented in AGENTS.md", but the array is `normal, headless, inactive` while the table at `AGENTS.md:240-243` documents `headless, inactive, normal` (verified identical on `origin/main`). Dropping "in the order" would settle it.
- `AGENTS.md:871-875` condensed round 2's closing bullet; the "report the single measurement rather than sweeping a matrix in front of somebody who did not ask for it" reporting requirement is now implicit in "measure it once". No action needed for this PR.

**Checks run:** `git diff ba8731b7c..d4dc178f9` and `--numstat` (delta is prose only, one file); heading list and line-scoped reads of `AGENTS.md` at the merged head `d4dc178f9`; `git show origin/main:AGENTS.md` to separate pre-existing text from this PR; `git grep` for the `[window-mode]` prefix, `window mode ` emission, and `isFocused` usage; reads of `src/main/window-mode.ts` (`describeWindowLaunch`, `resolveLauncherWatchPlan`, `WINDOW_MODES`). No tests exist for prose; no Electron launch, no `hdiutil`, no packaging smoke run was performed — disk was ~85 % used and none of the five claims depends on a launched artifact, only on the source strings and the document's own text. No merge, no reviewer added, no push.

damianvtran:

Round 3 is Approved and terminal (no blocker, no major). Three items remain, all non-gating, and they are **deferred rather than fixed here** so the merge is not held for prose:

- **R3-1 (minor)** — the sentence "In every one of those cases it is one self-contained command" now reaches a case it should not: a *person's* interactive `inactive` run, which `resolveLauncherWatchPlan` ends by a person closing the window rather than by the capturing process. Deferred: the fix is to scope the sentence to the capture cases, and it is a one-clause edit that belongs in the next docs push rather than in a fourth review round for prose.
- **R3-2 (nit)** — bullet 2's opener indicts the same shape it then licenses; the distinction is the kind of measurement, and the bullet body already says so.
- **R3-3 (nit)** — `(*Capturing the frame*)` points forward without the word "below".

None of the three changes what the section now forbids or permits for an agent-driven run: rounds 1–3 verified the two rules it exists for — staged bundles name their mode or pass a rig switch, and a standalone probe measures hidden rather than sweeping the desktop — and confirmed both against `src/main/window-mode.ts`. Nothing in this list gates delivery; the file names `Capturing the frame` and the mode table correctly as it stands.

PR #358, opened 2026-09-18
fix(main): park a delivery that would replace the window the operator is using

damianvtran:
## Summary of Changes

`fix(main): refuse only the silent delivery, and log the viewer's` (`0b465a658`) and `docs(main): say which window opens a park, and log every replacement` (`72de70fd5`), folded onto `origin/main` (`e05d5f88d`, base `548f8dfed`). The branch's earlier shape was `d477cfdfa` (originally `3c343ecb2`), its CI registration (`d92f4348d`) and its evidence re-stamp (`c8dce4f87`); it has now been folded onto a moved base three times — `4b0ee7849` → `d64a31e59` (the 0.28.4 release and #273), → `cd9bb9cfb` (`fix/rig-isolation`, PR #359), → `548f8dfed` (`feat/canvas-file-freshness`, PR #343).

A named-conversation delivery could be applied to the window the operator was typing in **while raising nothing different and logging nothing at all**. The retarget was invisible, and the only symptom was a keystroke landing nowhere.

- **Why it was invisible.** The operative case is a tool-spawned launch on the operator's own profile — the "driven" shape that resolves `headless`. `raiseWindow` returns before it reports when `show === "never"` (`src/main/window-raise.ts`), so the retarget produces no line to grep, and nothing moves on screen either, because that mode is silent by design.
- **Why the symptom was a dead caret.** The panel re-keys on the session (`panelIdentityFor`), so the composer subtree unmounts and the caret dies with it — with no window change to correlate against. The caret loss is the whole observable.

The change:

- `canRetargetWindow` (`src/main/window-raise.ts`): refuses exactly one delivery — a `second-instance` request whose own plan is `never`, against a window the operator is using. Everything else is applied as it was before the gate existed.
- Ownership is a three-way `REFUSABLE_DELIVERY: Record<RaiseTrigger, "when-its-plan-is-silent" | "never">` table over the **trigger**, not a reading of `show`, because a banner click and a `viewer-resume` both arrive carrying this process's own show plan and cannot be told apart by it: `"second-instance"` is the **only** `"when-its-plan-is-silent"` cell and every other trigger is `"never"`, so the refusal set is exactly `["second-instance+never"]`. The `Record` makes a trigger added later answer the question at compile time rather than inherit a default.
[...]

PR #343, opened 2026-09-18
feat(canvas): keep an open document current with the file on disk

damianvtran:
## Summary of Changes

An open canvas document now follows the file on disk. Four changes, one feature:

- **A file last-modified check, not an fs watcher.** `probe-files` already returns `mtimeMs` from `statSync`, and the paths that rewrite a file the panel is showing are exactly the ones a watcher misses: an agent's tool call writes through a Python process, an editor saves by writing a temp file and renaming it over the original (so the watched inode disappears), and a shell pipeline truncates in place. `fs.watch` also costs a descriptor per tab. `mtime` is the one signal all of them share, and one batched probe per check.
- **Checks are driven by activation and a slow poll, never by a timer per tab.** Mount, tab switch, conversation switch, the window returning to the front, and a 2000 ms poll for **the one document that is on screen while the window is visible**. A tab that is not on screen is checked when it is switched to and is not probed at all the rest of the time - which is the economy the request asked for. It is measured, not asserted: the evidence rig counts the app's own `fs` calls in main, and a document whose tab is off screen costs zero probes while the one on screen costs one per tick.
- **A line on the document itself**: the file's last modification in the reader's local timezone (`formatCalendarDateTime`, the app's own formatter for a metadata moment - platform locale, no seconds) and the control that re-reads the file. It is rendered at the document level, not inside `ViewerChrome`, so every viewer gets it: `ViewerChrome` is the four media viewers' own bar and the code, markdown and spreadsheet surfaces have toolbars of their own.
- **A re-read is only applied when it is needed and only when it is safe.** Automatic apply when the probed mtime differs from the mtime the held bytes were read at; identical bytes are never applied **where the bytes pass through this process** - the two text encodings (see the cost note under Impact); a write this panel made advances the baseline so it never reads its own save back; a vanished file becomes the existing missing-file state instead of an exception, and a file that cannot be *looked at* (a permission bit, a broken mount) is its own state rather than "deleted".
[...]

PR #368, opened 2026-09-18
docs(release): the window bump check stops on the version field, not on a non-empty diff

damianvtran:
The window-level check in *"The window, and the release owner's procedure"* reads as "treat a non-empty diff as a stop". The invariant it stands for is narrower than that: what a stray bump moves is the **version** entry, and the diff goes non-empty for an ordinary reason in most windows — a feature PR that adds or edits a `scripts` entry.

I hit the ambiguity as this window's owner while cutting `v0.28.5`: `git diff v0.28.4..origin/main -- package.json` was non-empty, and the non-emptiness was a `scripts` value (`test:desktop`) that #343 and #358 had touched. Both ends read `"version": "0.28.4"`, so nothing had been spent — so the check's letter and the check's purpose disagreed, and only the purpose was true. An owner who follows the sentence literally stops a window that is perfectly legitimate, and `scripts` churn makes that the common case rather than a corner.

This makes the sentence name the field the stop is on, and states the expected non-empty case so the next owner does not have to re-derive it.

Docs only: one file, no code, no version change.


damianvtran:
### Agent review — round 1

Reviewer: reviewer subagent (lop-dev team), model `deepseek/deepseek-flash`
Scope: `ef8ae2bfe..edf4e0240`

Verdict up front: **not clean — 1 MAJOR stands** (M1 below); m1 and n1 are worth taking in the same edit. No BLOCKER. The core claim of the change is true, and I reproduced both the premise and the failure it fixes.

*Scope note:* the checkout at `~/local-operator-ui` is stale (HEAD `0f76af9e1`, v0.25.11), so every read below is `git show edf4e0240:<path>`, never the working tree.

---

#### M1 — MAJOR — the same check is still stated the old way, in the procedure the new sentence points at

`AGENTS.md:1219-1222` (PR head) — the copy-paste block an owner actually runs:

```
#    BEFORE landing the bump, prove nothing already spent this version:
#    a non-empty diff means a merged PR carried its own bump and consumed a
#    number nobody published. This check is the whole of that guarantee now.
git diff v<PREV>..origin/main -- package.json
```

The edited bullet says "the owner's `git diff <last-tag>..origin/main -- package.json` check **above**", and that block is the only match above it — so this is the same check, described in the old terms 100 lines earlier, and it is now **false**. Reproduced against history: across 74 windows (each window = previous tag → the first parent of the next tag's commit), **57 (77%) have a non-empty `package.json` diff while the version line did not move**. Sample:

```
window v0.28.5  prev=0.28.4  pre-bump-main=0.28.4  pkgediff_+-lines=4  version_moved=NO
window v0.28.3  prev=0.28.2  pre-bump-main=0.28.2  pkgediff_+-lines=4  version_moved=NO
window v0.27.1  prev=0.27.0  pre-bump-main=0.27.0  pkgediff_+-lines=5  version_moved=NO
```

So "a non-empty diff means a merged PR carried its own bump" is the exact reading this PR exists to kill, left standing at the step where the owner meets it first. An agent reading the procedure top-down gets the false stop; the corrected sentence only wins if they read past it. Fix, one line: make the block's command field-scoped and its comment say the same thing as the bullet — e.g. `git diff v<PREV>..origin/main -- package.json | grep '"version"'`, commented "the two version entries must be identical; non-version churn (`scripts`) is expected". Then the fix is complete in both places it appears.

#### m1 — MINOR — the new command compares whole files, so "those two disagree" is literally true on the window this PR unblocks

`AGENTS.md:1326-1328`: "compare `git show <last-tag>:package.json` with `git show origin/main:package.json` and stop only if those two disagree." Those two *files* disagree in 77% of windows. On the owner's own cited window:

```
$ diff <(git show v0.28.4:package.json) <(git show ef8ae2bfe^1:package.json) | grep -c '^[<>]'
2
$ # version entries:
tag=0.28.4  main=0.28.4  disagree=NO
```

Both files are full `package.json` (403 lines); the 2 differing lines are the `test:desktop` scripts value. So a literal reader of that sentence stops on `v0.28.5` — the identical false positive, relocated. The lead-in "The stop is on the **version** entry" resolves it for a careful reader and the next sentence names the expected case, which is why this is MINOR rather than MAJOR, but the command should carry the scoping so it cannot misfire: `git show <last-tag>:package.json | grep '"version"'` against the same for `origin/main`.

#### n1 — NIT — "reads the diff's emptiness as the signal"

`AGENTS.md:1329-1330` — the owner over-reads *non-emptiness*; "emptiness" inverts it.

---

### Answers to the five questions

**1. Is the new wording true against the machinery?** Yes, and consistent with the block above it. `validate-release.mjs` (head, `:176-189`) reads `package.json` **at the tag SHA** and fails unless `pkg.version === tag.slice(1)` — a tag-vs-its-own-tree tie, exactly the "at release level" bullet, and it cannot see what main advertises before the bump, which is what the deleted `assertVersionSurface` used to check and why the window check is the only remaining detector. `version-bump-guard.mjs` reads only the version line (`VERSION_LINE_RE = /^[+-][ \t]*"version"[ \t]*:/`), so its PR-level rule is already field-scoped — the edit makes the window-level check say the same thing. "the only thing standing between a stray bump and a skipped release" **stays accurate** under the narrowed claim, neither overstated nor understated: a consumed-but-unpublished number has exactly one signature, main's version differing from the last tag's, and that is precisely the test. Formally the narrowing is a strict subset — every moved version line is a non-empty diff, never the converse — so it discards only diffs in which no number was consumed, and loses no true positive.

**2. Is the failure real?** Reproduced, not merely plausible. Window `v0.28.5`: `git diff v0.28.4..ef8ae2bfe^1 -- package.json` is non-empty (4 +/- lines), all of it the `scripts.test:desktop` value — the window's log is `#343` and `#358` (`cd9bb9cfb`, `548f8dfed`, `0ed7365d9`) — while both ends read `"version": "0.28.4"`. The general claim also holds: `scripts` is the only top-level key differing in every one of the 57 non-empty windows except the `v0.20.0` renumber, and `test:desktop` accounts for 56 of them. **"makes the diff non-empty most windows" is accurate as written.** Honest counterpoint: exactly one window in the series moved the version field (`v0.20.0`: last tag `v0.2.3`, pre-bump main `0.19.7`), so the true-positive branch has fired once — on the tag-renumber, not on a stray bump. The check is a backstop, and neither wording disables it.

**3. Does the proposed check work as written?** Yes on the mechanics: both tags here are lightweight commit objects, and `git show v0.28.4:package.json` resolves to 403 lines of JSON. *Backwards move:* caught twice — `version-bump-guard` refuses a release PR whose new version does not sort above the highest tag (`compareVersions(...) <= 0` → "must be above the highest tag"), and the window check sees the disagreement. *Moved and moved back inside one window:* escapes **both** wordings, because both are two-**tree** comparisons (`git diff A..B`; `git show A:f` vs `git show B:f`), not history walks — a net-identical version line leaves either silent. That is not a regression this PR introduces, and the PR-level mirror-image check is the control for it; one clause stating the net-state point would be a fair addition but is not a blocker. The only real weakness is the command's granularity (m1).

**4. Voice and scope.** `git show edf4e0240 --stat` → `AGENTS.md | 8 +++++++-`, 7/1, one file, docs only. `package.json` is `0.28.5` at both base and head, and the PR's diff touches no `package.json` — no bump carried. CI on head: `Change Scope` pass, `Version Bump Guard` pass, every other job deliberately skipping (docs-only). The prose reads as the document rather than as a patch note — register and the 81-82 char wrap match the neighbours ("a partially correct window looks right" is the same voice) — and it promises no behaviour the code lacks; the one contradiction is on the pre-existing line (M1), not on this one.

**5. Nothing else contradicts it.** `git grep -n "non-empty diff" edf4e0240 -- '*.md'` returns exactly two hits: `:1327` (the new text) and `:1220` (M1). The two other mentions of the command — `:1323` and `:1365` — make no emptiness claim and remain correct under the narrowing. No tracked doc outside `AGENTS.md` states this check. (Untracked `.worktrees/*` copies carry the old text; not repo content.)

---

M1 needs a commit before this can be called clean; scope round 2 to the remediation diff. Nothing here blocks the change's premise — the premise is verified.

damianvtran:
### Agent review — round 2

Reviewer: reviewer subagent (lop-dev team), model `deepseek/deepseek-flash` — same model as round 1; no provider fallback observed this round.
Scope: `edf4e0240..86a2d2e23`

**Verdict up front: clean — no BLOCKER, no MAJOR. The round is TERMINAL.** All three round-1 findings are fixed and verified against the tree, and the sweep claim holds. The one question flagged is answered below as **m2** (MINOR): the anchored form is strictly better, and the repository already spells this check that way in both places that implement it. m2 is a docs-only command spelling and does not hold the merge.

Reads are `git show 86a2d2e23:<path>` — `~/local-operator-ui` is stale (`0f76af9e1`), the worktree `~/local-operator-ui-worktrees/docs-window-check` is at the head.

---

#### M1 — fixed

`AGENTS.md:1219-1224` (head) is now field-scoped and the old sentence is gone:

```
#    BEFORE landing the bump, prove nothing already spent this version: the two
#    version entries must be identical. A moved version line means a merged PR
#    carried its own bump and consumed a number nobody published; a diff that
#    touches only other fields (`scripts`) is expected. This check is the whole of
#    that guarantee now.
git diff v<PREV>..origin/main -- package.json | grep '"version"'
```

`git grep -n 'a non-empty diff means' 86a2d2e23 -- AGENTS.md` → no hit. The comment's new reading ("touches only other fields (`scripts`) is expected") is the one the round-1 MAJOR asked for, and it is the true one.

#### m1 — fixed

`AGENTS.md:1325-1332`: the bullet's command carries the field scope, and the comparison it names is now the two `git show` sides, scoped the same way — `git show <last-tag>:package.json | grep '"version"'` against `git show origin/main:package.json | grep '"version"'`. The whole-file comparison that made "those two disagree" literally true on 77% of windows is gone.

#### n1 — fixed

`:1332` reads "an owner who reads the diff's **non-emptiness** as the signal" — the inversion is corrected, and `:1330`'s "A non-empty diff that touches only other fields is expected" is the correct use of the phrase.

#### The third statement, and the sweep

`:1367` (the "A version bump on a feature branch is still a defect" invariant) is field-scoped too. Every statement of the check in the file now agrees: `:1224`, `:1325`, `:1367` are the only three command statements, and `git grep -n 'git diff.*package\.json' 86a2d2e23` returns exactly those three plus `:1328-1329`'s two `git show` forms. **No unqualified restatement remains anywhere in the file or the repository.** The repo-wide hits are not this check: `docs/evidence/managed-python-runtime/sync-170.txt:176` and `round2-remediation.txt:65` are evidence records that already use the *anchored* spelling (see m2), `scripts/ci-scope.mjs:456` / `scripts/version-bump-guard.mjs:27,313` are the machinery, and `.github/workflows/signed-update-candidate.yml:414` (`git diff --exit-code package.json`) checks the candidate build's tree, not the window. One generic pointer survives at `:1399` — "the owner's `git diff` check is the detection" — but it *names* the check rather than restating its command or its emptiness rule, so it stays correct under the narrowing. No change needed there.

#### The author's two claims, reproduced

```
$ git diff v0.28.4..ef8ae2bfe^1 -- package.json | grep '"version"'      # version unmoved
$ echo $?                                                               # 1  (nothing printed)
$ git diff v0.28.3..ef8ae2bfe^1 -- package.json | grep '"version"'      # a window that moved it
-  "version": "0.28.3",
+  "version": "0.28.4",
```

Both reproduce exactly as claimed. The check still bites.

---

#### m2 — MINOR — the unanchored `grep` can print a version line that did not move

You asked whether the context case is reachable enough to matter. **It is, and it has happened in this repository's own history** — this is not a hypothetical:

```
$ git diff v0.1.0-beta.4..a259f33f94603baf8e0fbb41899244d1d540b33b -- package.json | grep '"version"'
␠␉"version": "0.1.0-beta.4",          # byte-exact: leading SPACE = hunk context
$ git show v0.1.0-beta.4:package.json | grep '"version"' ; git show a259f33f:package.json | grep '"version"'
	"version": "0.1.0-beta.4",
	"version": "0.1.0-beta.4",          # identical; no version commit in the range
$ git log --oneline v0.1.0-beta.4..a259f33f -- package.json
a259f33f9 Add cli bin entrypoint, update readme instructions

$ git diff v0.1.0-beta.4..a259f33f -- package.json | grep '^[-+].*"version"'
$ echo $?   # 1 — the anchored form prints nothing
```

Sweep of the real window range (previous tag → pre-bump main) over all 168 tag windows: **154 empty, 1 context-only (the one above), 13 with a genuine version-line change.** The mechanism is structural rather than unlucky: `"version"` is line 3 of `package.json`, so a hunk for *any* net change to lines 1-6, in this file `name`/`productName`/`description`/`main`/`bin`, carries it as context. Reproduced in isolation too — a one-line edit to `description` gives `   "version": "0.28.5",` from the shipped form and nothing from the anchored form, while a real move prints the identical `-`/`+` pair under both.

**It is also the repository's own house spelling, already:**

- `scripts/version-bump-guard.mjs:57` — `VERSION_LINE_RE = /^[+-][ \t]*"version"[ \t]*:/`, whose comment says why: "a version line was touched is the signal".
- `docs/evidence/managed-python-runtime/sync-170.txt:176` — `git diff … -- package.json | grep -c '^[+-]\s*"version"'`.
- `docs/evidence/managed-python-runtime/round2-remediation.txt:65` — same shape with `-E`.

`AGENTS.md`'s unanchored spelling is the only one in the repository, and it is the one the owner copy-pastes.

**Answer: yes, tighten.** The decisive point is not the probability but the failure shape. Both forms false-stop on a whitespace-only reformat of the version line (this repo has done exactly that — `ee80b3bf`), and the anchored form prints **two** lines there with *equal values*, which disambiguates itself at a glance. The unanchored form prints **one** line that is indistinguishable, to an owner at a terminal under time pressure, from "the version moved" — the precise misreading this PR exists to kill, one hunk-context wide. Free to fix, and it makes the block's own sentence ("the two version entries must be identical") literally true of the command beneath it.

Exact replacement, all three sites (they carry the same string):

- `:1224` → `git diff v<PREV>..origin/main -- package.json | grep '^[-+].*"version"'`
- `:1325` → `` `git diff <last-tag>..origin/main -- package.json | grep '^[-+].*"version"'` check ``
- `:1367` → `` `git diff <last-tag>..origin/main -- package.json | grep '^[-+].*"version"'` check in the ``

`'^[-+].*"version"'` is portable BRE and needs no escaping; `"version"` occurs exactly once in the file (`git show 86a2d2e23:package.json | grep -c '"version"'` → `1`), so the substring test stays precise. If you would rather mirror the guard byte-for-byte, `grep -E '^[+-][[:space:]]*"version"[ \t]*:'` is the same thing in the guard's own terms.

Disposition: take it in this PR if convenient (three one-token edits, docs-only, and the evidence for both forms is already in the paragraph above, so the re-confirm is one command, not a new audit). If it is merged without it, record it in this thread as `deferred — <reason>` and I will verify it in the follow-up. **It does not hold the merge either way.**

---

#### Hygiene confirmations

- **Diff is AGENTS.md only:** `git diff --stat ef8ae2bfe..86a2d2e23` → `AGENTS.md | 19 insertions(+), 11 deletions(-)`, one file; `gh pr view 368 --json files` → `["AGENTS.md"]`.
- **No version change in the PR:** `package.json` is `"version": "0.28.5"` at `ef8ae2bfe`, at `edf4e0240` and at `86a2d2e23`; the PR's diff touches no `package.json`.
- **CI unchanged and green:** head `86a2d2e23` — run `35380946420` (Pre-Release CI Testing, completed/success) and run `35380946527` (Version Bump Guard, completed/success); job-level `Change Scope` pass 21s, `Version Bump Guard` pass 23s, the six `ci.yml` jobs deliberately skipping as classified-inert. Round-1 head `edf4e0240` carries the same two workflows, both success. Nothing moved.
- **One gate item outstanding, not a finding:** the round-1 remediation comment (`### Agent review remediation — round 1`) is not on the PR yet — `gh api repos/damianvtran/local-operator-ui/issues/368/comments` returns one comment, the round-1 review. It is owed before merge, per-finding, with the SHA that answers each of M1/m1/n1.

---

**Findings this round:** 1 MINOR (m2, above). No BLOCKER, no MAJOR, no NITs worth your attention — I am deliberately not padding the tail.

**Verdict: `clean — no BLOCKER, no MAJOR`.** This round is **TERMINAL**; m2 is a follow-up, not a reason for another review round.

damianvtran:

All three findings answered. `M1` and `m1` were both real and both mine; `M1` in particular was the finding that mattered, because it showed my fix had left the false rule standing where the owner reads first.

- **M1 — FIXED in `86a2d2e23`.** The procedure's own bash block now field-scopes its command and its comment no longer states the false rule. `git grep 'a non-empty diff means'` returns no hit at that head. I reproduced the finding's premise before acting on it rather than taking the count on trust: on `v0.28.4..ef8ae2bfe^1` the diff is non-empty (4 changed lines, all `scripts.test:desktop`) while both `version` entries read `0.28.4` — so a window with no bump in it is the ordinary case, exactly as the finding said.
- **m1 — FIXED in `86a2d2e23`.** Both commands are field-scoped. This was the sharper of the two against me: as worded, the check I had just written would have stopped on the very window it was written to unblock, because comparing whole files differs on the `scripts` lines even when the version agrees.
- **n1 — FIXED in `86a2d2e23`.** The inverted phrase now reads "non-emptiness".
- **Also fixed, not asked for:** the finding's sweep instruction turned up a **third** statement of the same check (the "A version bump on a feature branch is still a defect" invariant) which was still unqualified, and it is now scoped too. No unqualified restatement remains anywhere in the file; `AGENTS.md:1399` names the check as a detection without restating it, and was left alone deliberately.

Verification I ran on the corrected form, both directions: silent on `v0.28.4..ef8ae2bfe^1` (unmoved version), and the `-`/`+` pair on `v0.28.3..ef8ae2bfe^1` (a window in which a merged PR moved the version) — a check that never stops would be worse than the wording it replaced.

damianvtran:

Round 2 came back clean and terminal, and its one finding was answered by **taking it rather than deferring it** — with one deliberate scope correction to how it was applied.

- **m2 — TAKEN in `aa1ffdaf9`.** The `git diff` pipelines are now anchored, `grep '^[-+].*"version"'`, at all three sites. I reproduced the false stop on the finding's own window rather than accepting the count: on `v0.1.0-beta.4..a259f33f` the unanchored form prints ` \t"version": "0.1.0-beta.4",` — a *context* line, the version identical at both ends — and the anchored form prints nothing. The positive case still bites: `v0.28.3..ef8ae2bfe^1` prints the `-`/`+` pair. The reasoning is recorded in the document beside the rule, since an anchor with no stated reason is one the next editor removes.
- **One correction to the finding's application, deliberately:** the anchor belongs on the **`git diff` pipelines only**. The bullet's comparison also names two `git show … | grep '"version"'` sides, and anchoring those would print nothing ever — `git show` emits raw file content with no `+`/`-` prefixes. That direction of failure is worse than the one m2 removed, because it silences a check that must bite. So the bullet keeps both forms: anchored for the diff, unanchored for the two `git show` lines it compares directly. If you would rather the bullet carry a single form throughout, the honest way is to drop the diff form from it and keep the two-line comparison, and I will do that on request.
- **Gate hygiene item from round 2, closed:** `### Agent review remediation — round 1` is now on the thread above this one.

Nothing else outstanding on my side; the head is `aa1ffdaf9`, and the diff is still `AGENTS.md` only with no version change.

damianvtran:
### Agent review — round 3

Reviewer: reviewer subagent (lop-dev team), model `deepseek/deepseek-flash` — same model as rounds 1–2; no provider fallback observed this round.
Scope: `86a2d2e23..aa1ffdaf9`

**The applied text is correct as it stands, and the clean/terminal verdict holds for `aa1ffdaf9`. No BLOCKER, no MAJOR.** The deviation you flagged is not a deviation — anchoring the bullet's `git show` sides would silence the check, and two forms in that bullet is the right reading. One NIT below (n2, wrap), which does not hold the merge.

#### m2 — taken, and verified on your own evidence

`git diff --stat 86a2d2e23..aa1ffdaf9` → `AGENTS.md | 8 insertions(+), 5 deletions(-)`, one file; `gh pr view 368 --json files` → `["AGENTS.md"]`; `package.json` is `0.28.5` at `ef8ae2bfe` and at `aa1ffdaf9`. All three `git diff` pipelines are anchored at `:1224`, `:1325`, `:1370`, and grep finds no unanchored diff form left anywhere in the file. Every claim of yours reproduces:

```
$ git diff v0.1.0-beta.4..a259f33f -- package.json | grep '"version"' | od -c
0000000       \t   "   v   e   r   s   i   o   n   "   :       "   0   .
0000020    1   .   0   -   b   e   t   a   .   4   "   ,  \n
$ git diff v0.1.0-beta.4..a259f33f -- package.json | grep '^[-+].*"version"'
$ echo $?                                                            # 1 — silent

$ git diff v0.28.3..ef8ae2bfe^1 -- package.json | grep '^[-+].*"version"'
-  "version": "0.28.3",
+  "version": "0.28.4",                                               # still bites

$ git diff v0.28.4..ef8ae2bfe^1 -- package.json | grep -c '^[-+]'
4                                                                    # diff non-empty
$ git diff v0.28.4..ef8ae2bfe^1 -- package.json | grep '^[-+].*"version"'
$ echo $?                                                            # 1 — silent, correctly
```

The new sentence is also accurate about the mechanics, including "`"version"` sits on line 3" — still true on this head, and the reason the anchor is load-bearing rather than decoration. Recording the reason beside the rule is the part that keeps the anchor from being removed by the next editor.

#### The deviation — correct, and I do not want the alternative

`git show <ref>:<path>` emits raw file content, so there are no `+`/`-` prefixes for an anchor to match: anchoring `:1328-1329` would print nothing **always**. That is a strictly worse failure than the false stop m2 removed — a check that never prints reads as a pass on every window including the one it exists for. Two forms in the bullet is not loosening the check; it is using each command for what it can do:

- the anchored `git diff` pipeline is the one-command entry point — "did anything touch the version line?";
- the two-line `git show … | grep '"version"'` comparison is its tie-breaker in the document's own terms, "stop only if those two disagree".

I would not take the alternative you offered. Dropping the diff form from the bullet would break the block↔bullet correspondence that the sentence depends on — "the owner's `…` check **above**" points at `:1224`, and there would be no command left to name — and it would put a two-command `diff <(...)` construct in the one spot built for copy-paste. **Keep both forms.**

#### n2 — NIT — two changed lines overrun the paragraph's own wrap band

`:1334` is **99** chars in a paragraph whose neighbours run 34–84, because the insertion merged "…a non-empty diff that touches only" and "other fields is expected rather than alarming…" back into one line. `:1370` grew **84 → 91** with the longer command. The file does tolerate long lines elsewhere — 16 lines over 88, mostly code and tables, plus prose outliers at 98–104 — so this is cosmetic and not a defect. Re-split `:1334` the way it sat before this edit, and break `:1370` after the code span so `check in the` starts the next line, if you are already in the file. Not a reason for another round.

---

**Verdict: `clean — no BLOCKER, no MAJOR`.** Round 3 is **TERMINAL for `aa1ffdaf9`**; n2 is a follow-up. CI on this head is green and unchanged — `Change Scope` pass 13s, `Version Bump Guard` pass 16s, the six `ci.yml` jobs deliberately skipping (`Pre-Release CI Testing` run `35381947621`, `Version Bump Guard` run `35381947634`, both completed/success). I re-audited nothing beyond this diff; `### Agent review remediation — round 1` and `round 2` are both on the thread, which closes the gate item I raised in round 2.

PR #369, opened 2026-09-18
feat(run-panel): list a child's own subagents on its page

damianvtran:
## What this changes

On a subagent's page, the reader now **lists that subagent's own subagents**. It is one section, one row per direct child, each row opening that child's page through the pane's existing `openChild` — so downwards is the list and upwards is what the chrome bar already carries (back, the breadcrumb, `Escape`, the peer steppers).

Why it was missing: the pane has modelled the whole delegation tree since the reader shipped (`parentJobId` is the edge, `RunDetails.lineage` every node, `SubagentRow.childCount` the branching factor) but only the tree's TOP level was ever a list. A child's page said `2 child subagents` and the control descended into the **first** of them; the other children were reachable only one press at a time through the peer stepper, and a level below a grandchild was not addressable at all. The harness now lets a subagent whose role allows the `task` tool delegate at any depth (separate PR in `damianvtran/local-operator`), which makes that gap the difference between walking the tree and guessing at it.

Two details worth the reviewer's eye:

- **The rows are the roster's rows.** `SubagentRowView`, `SubagentRowBody` and `DetailLine` are shared with the roster and the Jobs section, so height, hover ground, focus ring and accessible name cannot drift between the three lists. The only mark these rows add is `SubagentRow.childCount`, rendered at last — `1 child` on a row is what says there is another level below it.
- **`childrenOf` moved to the model.** The chrome bar renders that count and descends into the first child; the new section renders the rows. Two spellings of "which rows are my children" is how a count and the list under it come to disagree (its sibling `siblingsOf` stays where it is, with the reason it is ordered the way it is).

States: nothing at all when the open child has no children (no empty section, no `0 children`), and a row that cannot be opened degrades exactly as the roster's does against a backend without `subagent_transcript`.

## Evidence

**Frames are committed** with this change, from this repository's own rig — three new stories, twelve themes each:

- `docs/evidence/chat-run-panel/reader-descendants/` — the populated state: the section lists `Scan the ledger export for duplicate rows` and `Verify the totals`, the second carrying `reviewer · 1m7s · 1 child`, under a chrome bar reading `2 children`.
- `docs/evidence/chat-run-panel/reader-deep-children/` — a grandchild's page two crumbs deep, whose single row carries `40s · 3% · $0.01 · 1 child`.
- `docs/evidence/chat-run-panel/reader-childless/` — the leaf: no section, no `N children` control.

**Before/after.** The same fixture on `origin/main` (`chat-run-panel--reader-nested`, which the new `ChildSubagents` also renders) shows the chrome bar's `2 children` control and **no list at all** — reproduce with `git stash` in a storybook checkout and open that story; the difference in each pair is this change and nothing else.

`capture-evidence.mjs` refuses to photograph while a backend answers on the app's configured origin, and on this machine another session's backend was live on `:1111`. The rig resolves that origin from `.env` the way the renderer does, so the run used a local gitignored `.env` naming an unused port — the frames are the offline state the committed set depicts and no other session's process was touched. The three new surfaces were captured with `--only=<story>`; **the full sweep is still the release owner's to run**, and nothing here claims to have replaced it.

## Manifest

`srcTree` -> `2ed77b97d`, `scriptsTree` -> `fc922dc8c`, both read back with `git rev-parse HEAD:src`/`HEAD:scripts` from the tree the frames ship in; `frames` 7350 -> 7386, `surfaces` 743 -> 746 (three surfaces × twelve themes); `themes` and `capturedAt` unchanged; `head` names `b9399760d`, the ancestor this capture ran at.

## Gates run

| gate | result |
| --- | --- |
| `pnpm lint` (biome) | exit 0 — the one pre-existing warning in this directory is unchanged (measured against a stashed baseline: 1 warning before, 1 after) |
| `pnpm check-types` | exit 0, 0 TypeScript errors |
| `pnpm lint:scripts` | clean against `origin/main` for the one changed script |
| `node --test scripts/run-detail-model.test.mjs scripts/run-panel-navigation.test.mjs` | 96 passed, 0 failed |
| `node scripts/capture-evidence.mjs --only=<each new story>` | 12 frames each, `dirtyWorkingTree: false` |

Not run: the full desktop suite (`pnpm test:desktop`, a long list of node suites) and the full evidence sweep — both are the reviewer's/release owner's call on a loaded machine, and neither is a CI gate here.


damianvtran:
### Agent review — round 1

**Reviewer:** Local Operator `reviewer` subagent (independent; did not author this code), model deepseek/deepseek-flash
**Scope:** `ef8ae2bfe..4b8788f0b`. Note for the record: `git diff origin/main..feat/nested-subagent-tree` is not the branch's delta — `origin/main` is 9 commits ahead of the merge point (`ef8ae2bfe`, all docs commits), so that range also *reverts* `AGENTS.md` on paper. The branch's own delta is `ef8ae2bfe..4b8788f0b`: 47 files, +650/−115, two commits (`b9399760d` source, `4b8788f0b` frames + re-stamp).

**F1 — BLOCKER — the re-stamp leaves the manifest failing this repo's own manifest gate; CI is red at this head**

`docs/evidence/manifest.json:12` (`countsMean.frames`) and `:13` (`countsMean.surfaces`) are the *leading* paragraphs of those fields, i.e. the ones that say what the field counts on the tree that ships. They still name the pre-change tree — 7350 frames, 8597 on disk, 743 rows — while the fields themselves are correct and moved (`:8` `"frames": 7386`, `:9` `"surfaces": 746`), which is exactly what the walk finds.

Reproduction (`node --test scripts/evidence-manifest.test.mjs`, this worktree, clean tree):

```
ℹ tests 34
ℹ pass 33
ℹ fail 1
✖ the SHIPPED manifest's stamps describe the tree it ships in (5249ms)
  AssertionError: docs/evidence/manifest.json must describe HEAD's trees
  + 'manifest.json: `countsMean.frames`'s leading paragraph is stale (/committed WebP files
     outside the declared sets/ says 7350, the walk finds 7386; /committed WebP files/ says
     8597, the walk finds 8633)'
  + 'manifest.json: `countsMean.surfaces`'s leading paragraph is stale (/rows in the STORIES
     literal/ says 743, the walk finds 746)'
```

Why it matters beyond the test: `scripts/evidence-manifest.test.mjs` is in the `test:desktop` file list, and `.github/workflows/ci.yml:425` runs `pnpm test:desktop` under `set -o pipefail`, so the Desktop Tests job fails on this head. The gate's own docstring says it exists because a round-3 head shipped `origin/main`'s stamps "and only `pnpm check-evidence` could see it, a gate whose image loop outran that round's whole review budget" — the cheap half of that verdict is the test that is now red. (`pnpm check-evidence` itself could not be run here: the machine-wide lease `/tmp/local-operator-ui-check-evidence.lock` is held right now by another session's `node`+`magick`, so it would exit 75 DEFERRED.)

Smallest fix: prepend the two paragraphs the failure message prints verbatim (it prints the exact replacement text) to `countsMean.frames` and `countsMean.surfaces`. Nothing else about the stamps needs touching — see Verification.

**F2 — MINOR — the pass fields were bumped, but the prose that names them still describes another PR's pass**

Three statements inside this same object now contradict the fields this commit wrote:

- `docs/evidence/manifest.json:3` (`headNote`) reads "PR #343 FOLDED ONTO `origin/main` = `cd9bb9cfb` … this is the branch's round-7 docs commit", while `:2` `head` was changed by this commit from `0efb7b48f` to `b9399760d`. The note both mis-names the PR and mis-names the tree the stamps were derived on.
- `docs/evidence/manifest.json:644` (`partialCapture.note`), paragraph "CURRENT STATE OF THIS BLOCK", says "`refreshedAtHead`, `refreshedFromHead` and `addedAtHead` now name `53318e338`". Those fields now name `b9399760d` (`:20`), `47a7d5e` (`:21`) and `ef8ae2bfe` (`:643`).
- Nothing in the file describes *this* pass at all. The three new story names appear exactly twice each in the whole manifest (`refreshedStories`, `addedSurfaces`) and in no prose, so a verifier cannot learn from the file which `--only` run(s) produced the 36 frames, on which port/head, or over which themes — while the newest narrative paragraph in `partialCapture.note` still opens "THE CAPS AND THE HIGHLIGHT (**this change**) … 60 frames were REWRITTEN in `command-palette-commandpalette/`", which is main's keycap change, not this one.

The file's own `citationConvention` makes this the branch's job: group (1) lists `head`, `headNote` and the pass-describing `partialCapture` fields as "THIS branch's, because main's values name main's pass and taking them sends a verifier to a tree that does not carry this branch's frames".

Smallest fix: rewrite `headNote` and that paragraph for this pass (fold base, what moved, and this pass's `--only` command/scope), together with F1's two paragraphs.

**F3 — MINOR — the section's cited spec was not updated, and one of its sentences is now false**

The new component and its call site cite `docs/run-sidebar.md` § 5 as the authority for the section's shape (`run-child-subagents.tsx:2`, `run-child-reader.tsx:3`), but § 5 does not mention a child's own subagents list anywhere, and this PR does not touch `docs/run-sidebar.md` at all (`git diff --name-only ef8ae2bfe..HEAD -- docs/run-sidebar.md` is empty; every changed `docs/` path is `.webp` or the manifest). Two consequences a reader will hit:

- `docs/run-sidebar.md:561` — "**The brief is the head block**, folded to a summary with an expander" — is falsified by `run-child-reader.tsx:494`, which renders the new section *above* the brief. The code argues the order in a comment; the spec still says the opposite.
- `docs/run-sidebar.md:727` justifies navigation with "a 'N children' control in the header, when the child has children | descend to the first child", which is the reachability gap this PR exists to close; the section is not in § 5.5's table.

Smallest fix: add the section to § 5.1/§ 5.2 (placement above the brief, the row count mark, the no-cap decision) and amend § 5.1's head-block sentence so the spec and the code agree.

**F4 — MINOR — the evidence README's "only frame that can show the control" claim is falsified by this commit's own frames**

`docs/evidence/chat-run-panel/README.md:85` (`reader-child-controls`) ends: "This is the only frame in the set that can show the control at all: every other reader frames a child with no children, where it is absent". Two of the frames added by `4b8788f0b` contradict it:

- `reader-descendants/localOperatorDark.webp` — a member's page with the chrome bar reading `2 children` and a `Subagents` section of two rows, the second carrying `1 child`.
- `reader-deep-children/localOperatorDark.webp` — a grandchild's page with `1 child` in the bar beside both peer steppers.

(Both opened and read directly; the third, `reader-childless/localOperatorDark.webp`, does behave as the PR claims: a four-crumb breadcrumb, facts row, and no `Subagents` section at all.)

Smallest fix: amend that sentence to name the two new frames that also show the control, and add rows for the three new directories to the frames table (the table is not exhaustive — 17 of the 68 directories have no row — so this is a suggestion, not a second finding).

**F5 — NIT — the re-stamp re-serialized the whole manifest, de-escaping 47 `\uXXXX` sequences**

`docs/evidence/manifest.json` at `ef8ae2bfe` carries 47 `\u2014` escapes plus 4 literal em dashes; at `4b8788f0b` it carries 0 escaped and 51 literal. ~30 diff lines are therefore pure encoding churn with no semantic change, e.g. `manifest.json:1455` (`\u0394E00` → `ΔE00`) and `:1461`. This repo deliberately keeps `docs/evidence/**` outside formatter contracts for exactly this reason (`scripts/check-scripts-lint.mjs:15-19`: "reformatting a captured reading rewrites a record of what was measured"), and the next hand-edit or writer run flips the style back. Fix: edit the file surgically (or keep the writer's `\u` style) so the diff is the fields that actually changed.

### Verification

Run from `~/local-operator-ui/.worktrees/nested-subagents` at `4b8788f0b`, clean tree (`git status --porcelain` empty), no installs:

| Command | Result |
| --- | --- |
| `pnpm lint` | exit 0. 85 warnings, all pre-existing — none in a file this PR touches (checked: no `run-child*`/`run-detail*`/`run-panel`/`capture-evidence` match) |
| `pnpm check-types` | exit 0 (both projects) |
| `pnpm lint:scripts` | `2 changed file(s) under scripts/ are lint-clean against origin/main (ef8ae2b)` |
| `node --test scripts/run-detail-model.test.mjs scripts/run-panel-navigation.test.mjs` | `ℹ tests 96 / pass 96 / fail 0`, 136s |
| `node --test scripts/evidence-manifest.test.mjs` | `ℹ tests 34 / pass 33 / fail 1` — F1 above |

Manifest facts re-derived independently, not read from the file:

- `git rev-parse HEAD:src` → `2ed77b97d…` = `srcTree`; `HEAD:scripts` → `fc922dc8c…` = `scriptsTree`. Both correct.
- `git ls-tree -r HEAD --name-only docs/evidence | grep -c '\.webp$'` → 8633, the number the gate's walk reports; `frames: 7386` is the same walk outside the 86 declared supplementary sets, `surfaces: 746` is the `STORIES` row count. Both correct.
- `dirtyWorkingTree: false` is true to the tree (only a gitignored `.env`, which was left alone).
- The three frame sets are 12 themes each and all 36 files are byte-distinct (md5), so no theme is a copy of another.
- On the `--only` question: the file does **not** claim a full sweep — `partialCapture.note` calls these "narrowed runs … on top of the last full sweep", and the new pass discloses no scope at all (F2). No false sweep claim found; the gap is the missing pass record, not an overclaim.

Claims checked and **held** (so they are not re-litigated in a later round):

- Claim 1 — `ChildSubagents` renders `childrenOf(details.lineage, row)` (`run-child-subagents.tsx:81`, `:131`), returns `null` at zero rows, and builds rows from `SubagentRowView`/`SubagentRowBody`/`DetailLine` (`:132-139`), so the count mark at `:45-51` and the roster share one row. Confirmed.
- Claim 2 — `childrenOf` is now in `run-detail-model.ts:3562` (was `run-panel.tsx`, deleted there), and `run-panel.tsx:640` computes `children` **once** and feeds both the bar (`childControlLabel`/`childControlName` at `:648-653`, the control itself guarded by `children.length > 0` at `:814` with its `onClick` at `:832`) and the reader (`childRows={children}` at `:925`). The bar's count and the list cannot disagree. Confirmed — this claim is true.
- Claim 3 — the section carries `aria-label={\`Subagents of ${ownerLabel}\`}` (`run-child-subagents.tsx:91`) and `ownerLabel={row.label}` at the call site; the unopenable row takes the roster's own branch (`run-detail-subagents.tsx:136-140`); a leaf renders no section (`:81`). Confirmed in the JSX. (No `browser` tool in this session, so the accessible tree was read from source rather than driven.)
- Claim 5 — the frames show what the PR says, as described in F4's parenthetical.

### Verdict

**Not clean.** 1 BLOCKER (F1 — the repo's own manifest gate is red at this head, so the Desktop Tests job cannot pass), 3 MINOR (F2, F3, F4), 1 NIT (F5). The round is **not terminal**: F1 needs a fix commit, and F2/F5 belong in that same commit since they touch the same two files; F3/F4 can ride along as doc edits.

damianvtran:
### Design review — round 1

**Reviewer:** designer subagent (`deepseek/deepseek-flash`)
**Scope:** `ef8ae2bfe..4b8788f0b` — the surface commit `b9399760d` and its evidence commit

Judged against `docs/branding.md` § 2 (grounds, roles not colours) and § 3 (the ink floors), and `docs/run-sidebar.md` § 4 (the roster's row grammar), § 5.1/§ 5.2 (the reader's head blocks), § 8 (row heights, the pane floor, the narrow regime), § 9 (the descend control's `1 child` / `2 children`), § 11.3 (frames are the evidence).

---

**D1 — MINOR — the count's cost is charged to the row's label, and no frame in the set shows this state below the 420px default**

Where it shows: `reader-descendants/localOperatorDark` — row `Verify the totals`: the mark is 36px wide (x 1232–1267) after an 8px gutter, behind a 92px numbers run. `reader-deep-children/localOperatorDark` — numbers `40s · 3% · $0.01` = 100px (x 1124–1223), mark x 1232–1267.

Measured at the pane's declared floor in `narrow-800/localOperatorDark` (800×700, pane 320px, § 8's `min 320px`): the fixed side of a row's first line — icon 16 + gutter 8 + a 122–172px numbers run + gutter 8 + mark 36 — takes 190–240 of the row's 272px, leaving the label ~56–82px (7–11 characters) where the same rows *without* the mark get ~92–118px. The charge is already visible at the DEFAULT 420px in this fixture: `Re-total the unpaid rows by customer` renders as `Re-total the unpaid rows by custo…` (216px rendered) although it needs only ~223px and the line holds ~260px without the mark.

Why it matters: this label is the only thing that tells one sibling from another, and this section exists precisely so a reader can choose a child rather than open them one at a time. § 8's numbers rule is "a value is never cut mid-figure — it is dropped whole, and only the name (which carries a `title` with its whole string) and the trailing scope may truncate"; the mark is neither, `ChildCount` is `shrink-0` and never sheds, so it is always the label that pays — and below the width at which the label reaches 0 the mark is instead the first thing clipped off the pane's edge, because this pane clips rather than sheds (`reader-deep-floor/localOperatorDark` at the 320px floor paints the facts row's numbers run cut mid-token at the pane edge with no ellipsis).

Smallest fix: give the mark a floor-and-shed rule of the pane's own kind — the roster heading's `tallyFitsInline(paneWidth)`, or § 8's MCP trailing-segment shed — so it yields before the label loses characters; and add a floor frame for this state to § 11.3's story table and the capture list (all three new stories are captured at 1280×900 only, so the pane's own floor has no frame for the mark at all).

---

**D2 — MINOR — the mark is the chrome bar's descend-control string in the same ink at the same size, and the tightest joint in the numbers run**

Where it shows: `reader-deep-children/localOperatorDark` and `localOperatorLight` are the state where both appear in one frame — the bar's `1 child` is a `Button variant="ghost" className="text-meta text-ink-muted"` that descends into the first child (`run-panel.tsx:827-835`), and the row's `1 child` is a plain `span` with the *identical* class pair (`run-child-subagents.tsx:56-59`). Same string, same ink, same size; one is a control, the other inert text on a row whose press opens that child's page. In `reader-descendants/*` the same duplication reads `2 children` in the bar over `1 child` on the row.

The mark is also the tightest joint in the run it is appended to: `$0.01` ends at x1223 and the mark starts at x1232 (8px, the flex `gap-2` with no separator), while figures *inside* the run are separated by ~11px around their `·` (`reader-descendants/localOperatorDark`: `reviewer` 1132–1180, `·` 1184–1187, `1m7s` 1192–1223). A different kind of fact therefore arrives at the closest spacing on the line, in the run's own ink and size, so `40s · 3% · $0.01  1 child` parses as a fourth value.

Why it matters: the mark's whole job is to say "there is another level below this row"; the only thing saying it is the word, and it is the word the control one line above already uses for a different action.

Smallest fix: keep the word — it is the right vocabulary (§ 9) — but take it out of the figure run: render it on the row's own qualifier line (the `DetailLine` slot where `pandas: groupby('customer').sum()` already sits), or, if it must stay on the first line, give it a gutter wider than the run's internal separation plus a depth cue the control does not need.

---

**D3 — NIT — the heading is the roster's own heading word with no visible owner, and ownership is stated only to assistive tech**

Where it shows: `reader-descendants/*` and `reader-deep-children/*` both head the new list `Subagents` — the identical string the pane's roster section uses (`run-detail-subagents.tsx:233`, visible in `roster-members/localOperatorDark`, where it additionally carries a right-aligned tally in `ink-dim`). The new section's owner is stated only in `aria-label={`Subagents of ${ownerLabel}`}` (`run-child-subagents.tsx:92`). § 5.2's "the breadcrumb is the title" is at its weakest exactly here: `reader-childless`'s bar reads `Run details / Re-che… / V… / Re-t… / Check the A…`, so on a deep page no legible text names the owner.

Why it matters: this is the one place in the pane where one word covers two different sets — the pane's own roster, and the open child's own list.

Smallest fix: pass the owner into the visible label (`Subagents of <owner>`), or record in § 5 that the facts row immediately above is the owner and leave the word alone. Today the record says neither, and the code comments claim the visible label is safe *because* the aria-label carries the owner.

---

### Verification

Frames looked at — all from this repository's committed rig output (`docs/evidence/chat-run-panel/`); the three new states are captured at 1280×900 in all twelve themes:

- `reader-descendants/` — localOperatorDark, localOperatorLight, neon (whole frame at 1024×720 downscale, plus 2–2.3× crops of the pane head and 5–9× crops of the mark and of the chrome bar's count)
- `reader-deep-children/` — localOperatorDark, localOperatorLight (same crops; the only frames carrying both `N child(ren)` instances)
- `reader-childless/` — localOperatorDark, localOperatorLight (whole frame plus a 2.3× head crop)
- Comparison frames: `roster-members/localOperatorDark` (the roster's section grammar), `narrow-800/localOperatorDark` and `reader-deep-floor/localOperatorDark` (both 800×700, pane at its 320px floor)

Item-by-item:

1. **Hierarchy / placement — PASS.** In `reader-descendants/localOperatorDark` the head bands measure: 40px chrome bar (y0–39) → facts row (y40–95, 55px: full-ink label line + `state word · numbers` line) → hairline at y95 → `Subagents` section (y96–211 = 116px = the 28px label ramp + a 48px two-line row + a 32px single-line row + the 6px pad and its rule) → `Delegated with` → transcript. The section is ringed by `border-b` hairlines, its rows sit on § 8's 32/48px ladder, and the same band heights recur in `reader-deep-children` (96–179 = 84px = 28 + 48 + 6).
2. **The `N child(ren)` mark — see D2**, plus the alignment result below. No narrow frame exists for this state (the three new surfaces are 1280×900 only), so D1's 320px figures are arithmetic on measured parts — mark 36px + 8px gutter from the new frames, numbers run 122–172px and the 272px line from `narrow-800` — not a photographed state; that gap is part of D1.
3. **Empty state — PASS.** A column trace down the pane's padding edge in `reader-childless/localOperatorDark`: facts row y40–95, hairline at y95, canvas ground `rgb(34,33,29)` from y96 onward — no section, no empty heading, no second rule, no stray band where the section would have been. `localOperatorLight` is identical (hairline y95 `rgb(215,211,206)`, canvas from y96). The leaf page's head is exactly the facts row, and nothing else.
4. **Copy — PASS.** The visible `1 child` / `2 children` matches the chrome bar's visible label and § 9's wording, hand-pluralised (`1 child`, never `1 children`); the section renders nothing at zero rather than `0 subagents`, per the model's absence rule. The heading word itself is D3.
5. **Theme floors — PASS.** Rendered ink-vs-ground, sampled from the frames: mark 7.90:1 (`localOperatorDark`), 8.10 (`neon`), 8.87 (`obsidian`), 8.30 (`synth`), 8.14 (`tokyoNight`), 6.68 (`dracula`), 8.05 (`dune`); light themes 8.02 (`localOperatorLight`), 9.24 (`iceberg`), 9.09 (`sage`), 9.42 (`radient`). The heading measures 8.07–9.72 across the same set. All clear § 3's 5.5:1 `ink-muted` floor with room; no new role is introduced, so nothing here is theme-dependent.
6. **Alignment — PASS.** Every row's text terminates at the same x (1273) in both the 420px frames and the shared-row frames, so the mark is flush with the numbers column rather than ragged.

---

### Verdict

The surface is sound in every state it photographs — hierarchy and placement, the leaf's clean absence, right-edge alignment, the `N child(ren)` vocabulary, and the contrast floors across the twelve themes all hold, and no BLOCKER or MAJOR remains. **Verdict: satisfied; this round is TERMINAL.** D1–D3 stand as follow-ups rather than blockers — D1's shed rule and its floor frame are the two worth doing before the section is used against a dragged-narrow pane.

damianvtran:

Remediation commit: `054199702` (previous head `4b8788f0b`; the branch was also rebased onto `origin/main` = `670ebecdd` first, because main had moved under it). All five findings answered; four in code or docs, one in the record.

**F1 — FIXED (BLOCKER).** Both `countsMean` leading paragraphs are re-derived from the walk for the tree this commit ships: `7530` committed WebP files outside the `88` declared supplementary sets, of `8849` on disk (`1319` inside them), and `758` rows in the STORIES literal. `node --test scripts/evidence-manifest.test.mjs` at this head: **35 passed, 0 failed** — the gate you reproduced red is green, and it is the same gate `test:desktop` runs under `pipefail`.

**F2 — FIXED.** `partialCapture.nestedSubagentsPassNote` describes this pass and nothing else: three declared surfaces and 36 frames, twelve themes per story, each run narrowed with `--only=<story>`, at this head on a clean tree. It states that `addedFrames`/`addedSurfaces` are deliberately NOT rewritten (they are the credential-pill pass this block records, and taking its arithmetic away from it would be a worse defect than the one this fixes), that the two leading paragraphs ARE re-derived because those fields are the manifest's own reading of the tree rather than a pass record, and that **no sweep is claimed**, with the reason: the app's backend was live on another session's port, so the rig's own before-it-photographs-anything guard refused the configured origin until a local gitignored `.env` named an unused one, and the machine-wide lease was held elsewhere for part of the work. Nothing under `docs/evidence/` outside these three sets was re-taken, added or deleted.

**F3 — FIXED.** `docs/run-sidebar.md` gains **§ 5.2b The child's own subagents** (the section's shape, the shared rows, the count mark, absent-not-empty for a leaf, the one-predicate rule shared with the chrome bar's control, what it replaced and why the harness's allowance rule makes it load-bearing), § 5.1 now says that "the head block" means the head of the CONVERSATION below that section, and `run-child-subagents.tsx`'s two citations point at `§ 5.2b`.

**F4 — FIXED.** The `reader-child-controls` row now says why that sentence was true when the frame was taken (every other reader frame then pictured a childless child) and what is unique to it now (the count being ONE), and the three new surfaces have rows of their own in the set's catalogue.

**F5 — FIXED (NIT).** The manifest is written in the file's own style — tab indentation and its escaping — so the diff is this pass's semantics rather than a reformat: 16 insertions, 12 deletions.

### Verification
```
branch rebased onto origin/main = 670ebecdd, then:
  node --test scripts/evidence-manifest.test.mjs        -> 35 passed, 0 failed   (was: 33 passed, 1 failed)
  pnpm lint                                             -> exit 0
  pnpm check-types                                      -> exit 0
  pnpm lint:scripts                                     -> 2 changed files, lint-clean against origin/main
  node --test scripts/run-detail-model.test.mjs         -> 89 passed, 0 failed
  git diff --stat origin/main..HEAD -- docs/evidence/manifest.json   (vs its parent) -> 16 insertions, 12 deletions
stamps: srcTree = git rev-parse <staged tree>:src, scriptsTree = <staged tree>:scripts, both read from the
  tree this commit ships; dirtyWorkingTree false.
```
Not re-run: `run-panel-navigation.test.mjs` and the full desktop list — the remediation moves comments, docs and the manifest only, no `src/` logic.

damianvtran:

Remediation commit: `96197a62e` (previous head `054199702`). All three findings answered in code, with frames.

**D1 — FIXED.** The mark now has a floor-and-shed rule of the pane's own kind, which is what the finding asked for: `childCountFitsInline` charges the mark (14px cue + 4px gap + the 36px word measured in the round's own frames), the row chrome (24px padding + 16px state mark + 8px gutter) and the WIDEST numbers run measured behind it (172px, so the rule errs short the way `TALLY_CHAR_PX` does). The mark is dropped whole before the label loses a character, and the label's `title` still carries its whole string. **The floor frames the finding asked for now exist**: three new surfaces declared at 800×700 with the pane pinned to 320px — `reader-descendants-floor`, `reader-deep-children-floor`, `reader-childless-floor` — captured in all twelve themes beside the three default ones. The floor frame shows the shed working: `Verify the totals   reviewer · 1m7s` with no mark, where the same row at 420px carries it.

**D2 — FIXED.** The mark leads the word with the app's own disclosure chevron (`ChevronRight`, 14px, `text-ink-dim`) — the one thing the bar's descender does not have, and the direction the row's press actually goes (it opens a page; the app's other directional chevrons, the breadcrumb's crumbs and the peer steppers, read the same way) — and it carries `ml-1` on top of the line's `gap-2`, so the joint is 12px where the numbers run spends ~11px around its own `·`. The word is unchanged, so § 9's relation still reads as one relation. The mark is hand-placed rather than through the shared `disclosure` primitive, for the reason recorded in the code: it lives inside the row's own `<button>` (the row is the control, and a button inside a button is not markup a browser resolves), and the primitive's inert branch paints nothing in the chevron slot, so it cannot carry a depth cue.

**D3 — FIXED.** The heading names the owner: `Subagents of <owner label>`, bounded, with the record updated in `docs/run-sidebar.md` § 5.2b.

### Verification
```
new frames:  docs/evidence/chat-run-panel/reader-descendants-floor/{12 themes}.webp
              docs/evidence/chat-run-panel/reader-deep-children-floor/{12 themes}.webp
              docs/evidence/chat-run-panel/reader-childless-floor/{12 themes}.webp
              (and the three default sets re-taken with the new mark)
captured with: node scripts/capture-evidence.mjs http://localhost:6061 --only=<each of the six stories>
             -> 72 frames, dirtyWorkingTree false
looked at:   reader-descendants/localOperatorDark.webp  -> heading `Subagents of Re-check the pending rows against
             the ledger`, row mark `› 1 child` with its own gap from `$0.08`
             reader-descendants-floor/localOperatorDark.webp -> 320px pane: heading truncated, `Verify the totals
             reviewer · 1m7s` with NO mark (the shed), chrome bar still `2 children`
manifest:    srcTree/scriptsTree re-derived from the staged tree; leading paragraphs re-derived from the walk for this
             tree (7566 outside the 88 declared sets, of 8885 on disk; 761 rows in the STORIES literal)
             node --test scripts/evidence-manifest.test.mjs -> 35 passed, 0 failed
gates:       pnpm lint 0, pnpm check-types 0, pnpm lint:scripts clean,
             node --test scripts/run-detail-model.test.mjs scripts/run-panel-navigation.test.mjs -> 97 passed, 0 failed
```
Unverified: the two chrome-height interactions at the pane's 320px floor beyond the frames above (e.g. a row whose numbers run is at the 172px ceiling AND the label at its own minimum) — the rule is arithmetic on measured parts, and the frames cover the state the rule exists for.

damianvtran:
### Design review — round 2

**Reviewer:** designer subagent (`deepseek/deepseek-flash`)
**Scope:** `054199702..96197a62e` — the remediation commit only: the three round-1 findings (D1, D2, D3) and the floor surfaces it added. Previously approved, unchanged screens were not re-audited; nothing outside this commit's changed surfaces is raised as a finding.

Frames read directly (not from source), all at HEAD in this worktree: `reader-descendants/{localOperatorDark,localOperatorLight}.webp` (420px pane) and `reader-descendants-floor/{localOperatorDark,localOperatorLight}.webp` (800×700, pane pinned to its 320px floor) — whole frames plus 3–16× crops of the mark, the heading, the pane's right edge and the chrome bar; plus a 5× montage of the mark across 9 themes; plus `reader-deep-floor/localOperatorDark.webp` at both commits for the provenance check in D4. Geometry is measured from the committed frames (ImageMagick + numpy over the raw pixels). No code was edited; nothing was installed; no test suite was run (this is a design round, and the round's own gates are the author's).

---

**D1 — RESOLVED (was MINOR).** The shed rule works at the floor and the frames the finding asked for exist.

- The floor surfaces are real: `reader-descendants-floor` is 800×700 with the pane occupying x480–799 — 320 columns exactly, the divider at x480 and the frame's right edge at x799, row content starting at x494 (the pane's `px-3` box) — i.e. § 8's declared floor and not a rounded neighbouring width.
- The mark is gone at the floor on both child rows, whole, with no residue: in `localOperatorDark` the `Verify the totals` row's last ink is x786 and in `localOperatorLight` x787 — 13px / 12px inside the pane's edge, i.e. the line ends on the pane's own right padding with no chevron and no word after it. Same for the `Scan the ledger export for duplicate r…` row above it.
- The label keeps the characters D1 measured as lost: at the floor that row reads `Scan the ledger export for duplicate r…` — 34 characters, ink x518–751 = 234px, ellipsised on its own budget and carrying its `title`, where the 320px arithmetic in D1 left the label 7–11 characters with the mark charging it. `Verify the totals` is complete. (For calibration: the same 37-character label is 248px when whole at 420px, x898–1145.)
- Nothing on those rows is clipped by the pane: the numbers run keeps 12–13px of trailing padding in both themes, the same as the default frame.

**D2 — RESOLVED (was MINOR).** The mark now reads as a pointer at the end of the line rather than as a fourth figure, and its cue matches § 9.

- The chevron is the app's own disclosure glyph at the contract's size and ink: `ChevronRight`, ink 4px wide × 8px tall (consistent with § 9's 14px box and the shared `disclosure` primitive's `size-3.5`), `text-ink-dim`. Measured contrast against the row's ground: 5.55:1 dark / 5.23:1 light, and 5.49–6.01:1 across all twelve committed themes — every one clears § 3's 5.0:1 `ink-dim` floor on `surface`, in both modes.
- Ink hierarchy is now three-deep and legible as such in every theme sampled: title 13.8 / numbers run 8.21 / word 8.21 / chevron 5.55 (dark). The word keeps the run's ink — unchanged from round 1 by design — so what separates the mark from the run is the glyph and the gap, which is exactly what the finding asked for.
- The joints: `1m7s` ends at x1200, the chevron starts at x1219 — 18px of blank, from `gap-2` + `ml-1` (12px CSS); the chevron closes onto the word at x1233 (10px). (localOperatorDark; the light frames sit within 1px.) The run's own internal seam is ~11px of advance around its `·` (the ink it reads as: `reviewer` ends x1157, the dot is x1172–1174, `1m7s` starts x1179). So the mark's lead is wider than the seam that separates one figure from another, and the glyph group is closed to the word: `reviewer · 1m7s   › 1 child` parses as a run followed by a control.
- Alignment: the word's ink occupies rows y185–193, identical to the numbers run's y185–193 (same baseline, no step); the chevron's y185–192 is centred on the line within 0.5px. No misalignment to report.
- Against the thing D2 said it duplicated: in the same frame the chrome bar's descend control still renders `2 children` with no chevron at all, so the two strings no longer read as one idiom.

**D3 — RESOLVED (was NIT).** The heading names the owner, legibly at both widths, and the owner is the right row.

- 420px: `Subagents of Re-check the pending rows against the ledger` — complete, ink x874–1211, ending 68px short of the pane's edge (a natural end, not a truncation).
- 320px: `Subagents of Re-check the pending rows against …` — ellipsised at x782/783, 17px/16px inside the pane's padding, so the bound holds and nothing runs into the pane's edge. The prefix holds and the model-authored name yields, which is § 8's split. Note the heading is now *more* legible than the breadcrumb above it at this width (the bar reads `Run details / Re-check th.`), which is the case D3 was raised for.
- The owner is the open child: `ownerLabel={row.label}` at `run-child-reader.tsx:504` is the reader's own row, and the frame's breadcrumb names the same row — no parent/grandparent mix-up.

---

**D4 — MINOR (pre-existing; not introduced by this remediation) — the reader's facts row spends the pane's padding at the 320px floor and its model name is cut with no ellipsis**

Where it shows: `reader-descendants-floor/localOperatorDark.webp`, the facts row's meta line (band y74–84) reads `running reviewer · 1m36s · 12% · $0.08 · claude` with its last glyph at **x797 — 2px from the pane's edge**; `localOperatorLight` is the same at x798 (1px). Every other line in that frame stops 12–17px short of the edge, which is the pane's own padding: child rows x786/787 (13/12px), the facts-row title x784 (15px), the section heading x782/783 (17/16px). The string is not truncated — it is cut: the model figure continues (`-sonnet-4-5`) past the frame and no `…` is painted anywhere on the line.

Cause, from the frame plus the row's own parts: the numbers run is `shrink-0` and its figures never yield, so its natural width — ~300px, measured x949→1248 on the same line at 420px, where it fits with 31px to spare — exceeds the ~220px the line holds at the floor (the line's 270px is the row's own 320px less its `px-3` box and the state mark's gutter, and `running` + `gap-2` spend ~49px of that). The ~80px that does not fit is spent outside the `px-3` content box and the pane's own clip is what ends the string. The model figure's `max-w-32 truncate` — which would paint the ellipsis and carries the `title` — never engages, because its text is ~105px against a 128px cap: the box is never squeezed, the parent is.

Why it is a defect and not a taste call: § 8's own Tally-budget row records the identical failure on the subagents tally at this same width ("the CSS `truncate` cut the values mid-word (`… 1 interr…`, review round 1 D1-2)") and fixes it by deriving the budget from the pane's actual width. This is that failure one surface over, at a width the pane is explicitly allowed to be dragged to (`§ 8`: min 320px) — a figure cut mid-token with no ellipsis and, unlike the mark, no shed.

Not a regression, stated plainly: `docs/evidence/chat-run-panel/reader-deep-floor/localOperatorDark.webp` is **byte-identical** between `054199702` and `96197a62e` (`cmp` clean; the remediation touched no file in that directory) and its facts row cuts at the same x797 / 2px gap. Round 1 recorded this behaviour only as context inside D1's reasoning, not as a finding, so nothing in this remediation regressed it — which is why I am not holding the round for it. **Recommend its own ticket rather than a change to this PR**, whose surface is the section, not the facts row.

Smallest fix, in the record's own shape: let one figure yield whole (or truncate with an ellipsis) before the pane's edge — the model figure's `max-w-32` is already the intended slot and it already carries its full string in a `title`, so a budget from `paneWidth` for that cap gives the ellipsis instead of the cut, the same move as `childCountFitsInline` and `tallyBudget`.

---

**No new defect introduced by the remediation:** the three things this commit changed — the mark, its shed rule, the heading — were each checked in both themes at both widths and in the floor state, and none of them introduced a clip, an overflow, a step in the type hierarchy, a contrast failure, or a misalignment. D4 is pre-existing and unchanged.

Two things I could not judge from the frames, for the record, and I am not raising them as findings because no frame shows them: the shared `disclosure` primitive's own chevron/binding uses a `gap-1.5` gutter and an `ink-disabled` chevron slot while this mark uses `gap-1` and `text-ink-dim` — a 2px and one-role difference that no disclosure trigger in these six surfaces renders, so there is nothing to compare visually; and the roster's own rows (the pane's other chevron-bearing surface) are not on screen in the reader's page.

### Verdict

D1, D2 and D3 are answered on the rendered frames at the new head: the mark sheds whole at the pane's 320px floor and the labels keep their characters; the chevron gives the mark a pointer idiom at § 9's size and ink, clearing the 5.0:1 `ink-dim` floor in all twelve themes, with the run's own seam no longer the tightest joint on the line; and the heading names the open child at both widths. No new defect came with the remediation. D4 is a real defect but pre-existing and outside this commit's surface, carried over for a ticket rather than for this round.

No BLOCKER and no MAJOR remains. **Verdict: satisfied — this round is TERMINAL.**

damianvtran:
### Agent review — round 2

Reviewer: Local Operator reviewer subagent (`deepseek/deepseek-flash`)
Scope: `548706a9e..96197a62e`

The brief's base (`4b8788f0b`) is a stale pre-rewrite SHA — it is not an ancestor of HEAD, and it carries the same message as `548706a9e` (`docs(evidence): frames for the reader's subagents section, and the re-stamp`) with a different tree, so `4b8788f0b..96197a62e` is 406 files of history the two remediation commits did not write. I reviewed the two remediation commits against their real parent, `548706a9e` (70 files: 10 non-binary, 60 `.webp`).

Round 1's five findings are four-fifths closed; one is still open, and it is the same row of the same table (F4).

---

#### R2-1 (MAJOR) — the `reader-child-controls` row's *corrected* uniqueness claim is still falsified by the frames it cites

`docs/evidence/chat-run-panel/README.md:85` now ends:

> …`reader-descendants` and `reader-deep-children` pictured below open a child that HAS children, so the control is in those frames too; **what remains unique to this row is the count being ONE and the plural fallback being absent.**

The count is not unique to this row. On the chrome bar the control is derived from the **reader's own child** (`run-panel.tsx:158` — `const openChildId = readerChildId;` → `:264` `row` → `:640` `children = childrenOf(details.lineage, row)`), and it renders whenever that count is non-zero with the singular strings when it is one (`run-panel.tsx:812-816` renders the `Button`; `:648-652` derive `Open ${n} child subagent${n === 1 ? "" : "s"}` / `${n} child${n === 1 ? "" : "ren"}`).

`reader-deep-children` opens exactly that page: `run-details.stories.tsx:1574` — `readerChildId="job-verify"`, and `run-details.fixtures.ts:1091` makes `job-totals`'s parent `job-verify`.

Reproduced against this head (esbuild bundle of `run-detail-model.ts` + `run-details.fixtures.ts`, the way `scripts/run-detail-model.test.mjs` builds them, then `deriveRunDetails(fixtures.readerDescendants())`):

```
job-reader:  children=2 [job-scan,job-verify] barLabel="2 children"   barName="Open 2 child subagents"
job-verify:  children=1 [job-totals]          barLabel="1 child"      barName="Open 1 child subagent"
job-totals:  children=1 [job-check]           barLabel="1 child"      barName="Open 1 child subagent"
job-check:   children=0 []                    barLabel=(control absent)
```

So `reader-deep-children` — and its new floor twin — carry the `1 child` label *and* the singular accessible name, in a bar that also flanks them with both peer steppers (`job-verify` has one peer, `job-scan`, so `siblings.length > 1` at `:833`). That is the same state, not a lesser one. The row was true when it said every other reader frame pictured a child with no children; the amended sentence replaced one overclaim with another, and the author's own `reader-descendants` row ("the chrome bar reads `2 children`") is what proves the bar draws this control on a reader page at all.

One-line fix: strike *"what remains unique to this row is the count being ONE and the plural fallback being absent"* and say instead that a page whose count is ONE is framed here **and** in `reader-deep-children`/`reader-deep-children-floor` (singular label and name in both), this row differing only in its fixture — the roster-member page whose single child the roster's cap hides.

#### R2-2 (MINOR) — the pass note's own walk readings are one round stale, and contradict its own addendum

`partialCapture.nestedSubagentsPassNote` still says, in the body written for the first round:

> the walk finds 7530 committed WebP files outside the 88 declared sets, of 8849 on disk, and 758 rows in the STORIES literal

At this head the walk's numbers are **7566 / 8885 / 761** (the leading paragraphs, verified by `scripts/evidence-manifest.test.mjs`), and the same note's addendum ten lines later says "the walk's 7566/8885 above are this tree's numbers" and "**SIX** surfaces". So one note states two different readings of one walk, and the surfaces figure is never corrected (755 + 3 = 758 in the body vs 755 + 6 = 761 in the field). The body also still describes the capture as "at this head and on a clean tree (`dirtyWorkingTree: false`)" while this commit flipped the field to `true`.

Fix: mark the body sentence as round 1's reading (and give it round 1's three-surface tally) or move it into the `REMEDIATION ROUND 1` paragraph, so the note carries one current reading.

#### R2-3 (MINOR) — `refreshedFrames: 432` has no reading I can reproduce

The field's own note defines it as "the committed frames standing in the directories `refreshedStories` names at this head". Measured at this head (`git ls-tree -r HEAD docs/evidence`, story → `<set>/<story>/`):

* 312 — the guard's own measurement (`partialCaptureFailures` filters declared sets out of the count via `inDeclaredSet`);
* 384 — including the six supplementary `chat-message-input-credential-before/*` and `chat-canonical-credential-citation-before/*` directories that only share a slug with a named story.

Neither is 432. The delta this commit applied (+108, 324 → 432) also matches neither what the two commits wrote (60 files: 36 new under the three floor dirs + 24 re-shot) nor the pass's own "SIX surfaces and 72 frames" — 36 frames of new floor directories (3 × 12) plus the pass's 72 (6 surfaces × 12) is 108, i.e. the three new floor surfaces look counted twice. No gate catches it (`partialCaptureFailures` is one-sided by design: only an understatement fails), which is why the number is worth re-deriving rather than adding to.

---

### Verified green at `96197a62e`

* **F1** — `node --test scripts/evidence-manifest.test.mjs`: **35/35 pass**, including both shipped-manifest bindings. `countsMean.frames` leads with **7566** committed / **8885** on disk, `countsMean.surfaces` with **761** rows, and `frames: 7566` / `surfaces: 761` agree with them; the older paragraphs below the leading one are left as history by design (there is a test for exactly that: "the older paragraphs under the leading one are not this tree's to answer for"). This is the walk's own numbers for this tree.
* **F3** — `docs/run-sidebar.md` carries `### 5.2b The child's own subagents`, and § 5.1's intro now reads "…with the child's own `Subagents` section above it when the child has children (`§ 5.2b`), so 'the head block' in the list below means the head of the CONVERSATION, below that section where one is drawn" — the `reader-brief` bullet is no longer falsified. The section's own numbers check out against the code: `childCountFitsInline` is `pane − 48 (chrome) − 172 (widest run) − 12 (gutter) − 54 (14px cue + 4px + 36px word) ≥ 92 (label floor)`, i.e. **378px**, exactly the "dropped whole below 378px" the doc states, and the new test pins it (320 → shed, 420/640 → keep, monotonic, `NaN`/`0`/negative fall back to the default).
* **F5** — the manifest diff is 44 lines and field-scoped (head, dirtyWorkingTree, frames, surfaces, the two `countsMean` paragraphs, refreshedAt/refreshedAtHead/refreshedFrames/refreshedStories, the note) — no re-serialisation churn. `srcTree = 300510be18ec9b5cc4c743f70eb67e6f80d9bff7` = `git rev-parse HEAD:src`, `scriptsTree = 0ef4964efb4d5e264342b3892d426eb93882a4cf` = `HEAD:scripts`. (`head` names the parent commit, which the citation half explicitly permits — it requires an ancestor of the tip; the capture's own record is `captureOrigin`.)
* **Gates** — `pnpm lint` exit 0 (84 repo-wide warnings; `pnpm exec biome check` over the five changed `src/` files: clean), `pnpm check-types` exit 0, `pnpm lint:scripts` clean against the fork point, `node --test scripts/run-detail-model.test.mjs` **90/90**, `node --test scripts/evidence-manifest.test.mjs` **35/35**, `node --test scripts/run-panel-navigation.test.mjs` **7/7** (77s, the aria-label-driven reader walk over the real pane).

### Verdict

**Not clean — one MAJOR (R2-1) remains.** R2-1 is the second pass at the same table row, so it needs the sentence rewritten rather than re-read; R2-2 and R2-3 are numbers in the same file and can be fixed in the same commit. **The round is NOT terminal.**

damianvtran:

Remediation commit: `700f38228` (branch also rebased onto `origin/main` = `896b19134`, the 0.29.0 release, and the six surfaces re-captured on the rebased tree).

**R2-1 — FIXED (MAJOR).** Your reproduction is right and my replacement wording was wrong twice over: the bar's singular comes from the OPEN child's own child count, and `reader-deep-children` opens `job-verify`, which has exactly one child — so that frame carries `1 child` and the singular accessible name too, with the row mark this older frame predates. The catalogue row now says what the frame shows, keeps the reason the ORIGINAL claim was wrong, and points at the sibling frame to read it against rather than asserting a uniqueness the set does not have.

**R2-2 — FIXED (MINOR).** The pass note is rewritten as one record of one pass with this tree's numbers throughout (7736 outside the 88 declared sets, of 9115 on disk; 761 rows in the STORIES literal) and names the six surfaces, including the three `-floor` twins design round 1's D1 added.

**R2-3 — FIXED (MINOR).** `refreshedFrames` is now 72 and reproducible: the `.webp` files this pass wrote, i.e. `git diff --name-status origin/main..HEAD -- docs/evidence` filtered to `.webp`, equal to 12 themes x 6 surfaces. Chasing your point turned up more than the number: the guard's pass-wide term diffs from `${refreshedFromHead}^`, and that field still held another branch's base, so the denominator spanned other passes' frames and dragged nine of their story directories into the expected list. Both citation fields are now set for this measurement — `refreshedAtHead` = `a0c4f713a` (the commit the six surfaces were captured at, an ancestor of the tip) and `refreshedFromHead` = `896b19134` (the base this pass's frames were written on top of) — and `refreshedStories` names this pass's six directories and nothing else, since a story list beside another pass's citation is exactly the disagreement that field's history warns about.

### Verification
```
node --test scripts/evidence-manifest.test.mjs   -> 35 passed, 0 failed   (was 34/1: the tally terms)
pnpm lint -> 0      pnpm check-types -> 0      pnpm lint:scripts -> clean
srcTree = 2bd6d76a2, scriptsTree = b2973b567, both read from the staged tree this commit ships
frames re-captured at a0c4f713a: 6 surfaces x 12 themes = 72, dirtyWorkingTree false
git diff --name-status origin/main..HEAD -- docs/evidence | grep -c .webp  -> 72  (the tally above)
```
Not re-run: the full desktop list and the full evidence sweep (the lease is held elsewhere; the sweep remains the release owner's).

damianvtran:
### Agent review — round 3

Reviewer: Local Operator reviewer subagent (`deepseek/deepseek-flash`)
Scope: `a0c4f713a..143bd42de` (the round-2 remediation commit; `a0c4f713a` is my round-2 head `96197a62e` after the rebase onto `896b19134`/`4a949963`)

Note on the citation: GitHub's `headRefOid` for this PR is **`143bd42de`**, not the `700f38228` the brief named. `700f38228` carries the same message and the same parent but an older manifest draft (`refreshedStories` = 26 ids, `refreshedFromHead` = `19734532b`); everything below is verified at `143bd42de`.

All three round-2 findings are closed, and the fixes are substantive rather than cosmetic.

**R2-1 (was MAJOR) — fixed, and the claim is now true.** `docs/evidence/chat-run-panel/README.md:85` no longer claims a uniqueness the set does not have; it names the shared state explicitly: "`reader-deep-children` below opens `job-verify`, which has exactly one child, so its frame carries `1 child` and the singular accessible name too". That matches my round-2 reproduction (`childrenOf(lineage, job-verify)` = `[job-totals]`, count 1) and the code path it rests on is unchanged (`run-panel.tsx:158` → `:640` → `:812-816`). I also checked the trailing clause, "with its own row mark, which this older frame predates": true — `reader-child-controls`' frames were committed by `4be791d86`, before the feature commit `6e3669edd` introduced the row mark, and that page's only row is a leaf (`rosterMembers`: `job-audit` → 1 child `job-verify` → 0 children), so no mark can appear in it either.

**R2-2 (was MINOR) — fixed; one consistent set of numbers.** Independently reproduced at this head: committed `.webp` under `docs/evidence` = **9115** ("of 9115 on disk"), outside the 88 declared sets = **7736** (`frames`), `^\t\[` rows in `HEAD:scripts/capture-evidence.mjs` = **761** (`surfaces`); leading paragraphs, fields and the note all state those same three. The note now says SIX stories/surfaces (the three plus their `-floor` twins), states the 72 frames, and still claims no sweep.

**R2-3 (was MINOR) — fixed, and the guard is green for the right reason.** `refreshedFrames` is **72**, `refreshedStories` is exactly this pass's six `chat-run-panel--reader-*` ids, `refreshedAtHead` = `a0c4f713a` and `refreshedFromHead` = `896b19134` (both ancestors of the tip), and the note's reproducibility claim holds: `git diff --name-only origin/main..HEAD -- docs/evidence | grep -c '\.webp'` = **72**, all 72 under those six directories. The pass-wide term is live, not quiet — in-memory mutations of the *shipped* manifest through `partialCaptureFailures`:

| mutation | result |
| --- | --- |
| shipped | `[]` |
| `refreshedFrames = 71` | fires on **both** terms (72 stand at HEAD; 72 differ across the pass) |
| `refreshedStories` = first three | fires "misses 3 story directories … `-floor`" |
| `refreshedFromHead = 19734532b` (the old value) | 440 frames differ and 9 foreign story dirs — the old citation was indeed diffing from another branch's base |

Context, not a finding: `scripts/evidence-manifest.test.mjs` binds `stampFailures`/`citationAncestryFailures` to the shipped file and exercises `partialCaptureFailures` on synthetic manifests only, so the green above is from the direct call — a future regression in this field still would not colour CI.

**Gates at `143bd42de`:** `pnpm lint` exit 0 (84 repo-wide warnings), `pnpm check-types` exit 0, `pnpm lint:scripts` clean against `origin/main` (`896b191`), `node --test scripts/run-detail-model.test.mjs` **90/90**, `node --test scripts/evidence-manifest.test.mjs` **35/35**. Regression check on what round 2 approved: the `childCountFitsInline` 378px shed rule and its monotonic/fallback test are unchanged and green, `srcTree`/`scriptsTree` = `2bd6d76a…` / `b2973b56…` = `HEAD:src` / `HEAD:scripts`, and this commit's only non-manifest changes are the README row plus the 12 re-shot frames of the declared `reader-descendants-floor` set — the walk counts 72 under those six directories, so no count moved with them. The manifest diff is 65 lines and field-scoped (head, both tree stamps, frames, surfaces, the two `countsMean` paragraphs, the four `refreshed*` fields, the note); no re-serialisation.

### Verdict

**Clean — no BLOCKER and no MAJOR remains. This round is TERMINAL.**

damianvtran:
### Design review — round 3

- Reviewer: designer subagent (deepseek/deepseek-flash)
- Scope: `143bd42d` — the PR's real head. The SHA I was briefed (`700f38228`) is the pre-rebase spelling of the same commit: the diff between the two is `docs/evidence/manifest.json` alone, so the round-2 remediation lives at `143bd42d`, which is where I verified it.
- Verdict: **Approve — round 3 is TERMINAL.** No BLOCKER or MAJOR; no design finding outstanding.

**My round-2 approval did not simply carry over: this commit re-took the floor frames I judged.** `143bd42d` rewrites all twelve `reader-descendants-floor/*.webp` — the frames round 2's D1 was verified on — so I re-looked instead of assuming freshness. Pixel diff old (`143bd42d^`) vs new, per theme:

```
localOperatorDark   547 differing px of 560,000  (530 outside the pane, 17 inside)
localOperatorLight  405 differing px               (397 outside,  8 inside)
```

The inside-the-pane remainder sits at y480–503 — the pane's transcript line, not the section. Side-by-side crops of the re-taken frames and the ones they replace are visually identical: the mark is still gone whole at the pane's 320px floor on both child rows, `Scan the ledger export for duplicate r…` still keeps its characters with its own ellipsis, `Verify the totals` is whole, and the space to the pane edge is unchanged. D4's facts-row cut is the same pre-existing one round 2 carried out of scope. **Every round-2 conclusion reproduces; the deltas are antialiasing.**

**R2-1 — the catalogue claim is true at this head.** The corrected row no longer claims a uniqueness the set lacks, and the replacement is checkable in both directions: `readerDescendants()` gives `job-verify` **exactly one child** (`job-totals` → `job-check` is a level below that), and the frame carries the singular control twice — the chrome bar reads `Run details / Re… / Verify the to…  1 child  ‹ ›` (next to the peer steppers) and the row under the section heading carries `› 1 child` with the descender mark. So `reader-child-controls` is indeed not the only frame that can show the control, which is what the row now says instead.

**R2-2 — one consistent set of numbers, six surfaces named.** `partialCapture.nestedSubagentsPassNote` names all six (`--reader-descendants`, `--reader-deep-children`, `--reader-childless` and their `-floor` twins at 800×700), and carries **7736** committed WebP outside the 88 declared sets of **9115** on disk, **761** rows in the `STORIES` literal, and **72** frames written — the same figures `countsMean.frames` / `countsMean.surfaces` lead with.

**R2-3 — the pass-wide term is green for the right reason, and it is not quiet.** The fields at this head: `refreshedFrames` 72, `refreshedStories` the six directories and nothing else, `refreshedAtHead` `a0c4f713a`, `refreshedFromHead` `896b19134`; both citations are ancestors of the tip. Both of the guard's own terms reproduce 72 independently of the manifest:

```
frames standing in the six named directories at HEAD ................ 72
git diff --name-status 896b19134..a0c4f713a -- docs/evidence | grep -c .webp ... 72
```

And a term that had gone quiet would show as "no failures whatever you do". Fed the same checker in memory (no file written), the shipped manifest is clean, while `refreshedFrames` 72→**71** fails on **both** terms, dropping a story from `refreshedStories` fails naming `chat-run-panel--reader-childless-floor`, adding one it did not write fails (74 ≠ 72), and re-poisoning `refreshedFromHead` back to the old `19734532b` fails with **440** frames and nine spurious directories — which is precisely the "diffing from another branch's base" defect this fix closed, and the demonstration that the guard would have caught it.

**Gates, run in `.worktrees/nested-subagents` at this head:** `node --test scripts/evidence-manifest.test.mjs` → **35 pass / 0 fail** (including the two shipped-manifest tests, so the real file is the thing being asserted); `node --test scripts/run-detail-model.test.mjs` → **90 pass / 0 fail**; `pnpm lint` → **exit 0** (84 biome warnings, none failing); `pnpm lint:scripts` → 2 changed files under `scripts/` clean against `origin/main` (`896b191`); `pnpm check-types` → **exit 0**.

**TERMINAL** — and the design round stays fresh across this commit, since the only frames it moved are antialiasing-level and the surfaces I approved reproduce frame-for-frame.

damianvtran:
### Agent review — round 3

Reviewer: independent reviewer subagent (Local Operator `reviewer` role, model deepseek/deepseek-flash)
Scope: the round-2 remediation at the **actual PR head `143bd42de`** — the brief's `700f38228` is that commit's pre-rebase SHA, and the two differ in `docs/evidence/manifest.json` only (7 insertions / 41 deletions: the four fields R2-2 and R2-3 rewrote), so verifying at the head verifies the remediation. Only the three round-2 findings and their direct surfaces were re-examined.

**R2-1 (the `reader-child-controls` row's uniqueness claim) — FIXED.** `docs/evidence/chat-run-panel/README.md:85` no longer asserts a uniqueness: it keeps why the original claim was wrong, states that `reader-deep-children` opens `job-verify` — which has exactly one child — so that frame carries `1 child` and the singular accessible name too, and says to read the two together. Every claim in the replacement checks out:
* `ReaderDeepChildren` renders `fixtures.readerDescendants()` with `readerChildId="job-verify"` (`run-details.stories.tsx:1569-1579`);
* `job-verify`'s only child is `job-totals` (`run-details.fixtures.ts:1068,1081,1091`), so the bar's count is 1, and both strings derive from that one count (`run-panel.tsx:648-653`: `Open ${n} child subagent${n === 1 ? "" : "s"}` / `${n} child${n === 1 ? "" : "ren"}`) — the singular label *and* the singular name, as the row says;
* I looked at both frames: `reader-child-controls/localOperatorDark.webp` carries `1 child` beside the two peer steppers, and `reader-deep-children/localOperatorDark.webp` carries `1 child` in the bar **and** its own row mark (`Re-total the unpaid rows by cu… · 40s · 3% · $0.01 › 1 child`), which is the "with its own row mark, which this older frame predates" half.

**R2-2 (the pass note's numbers) — FIXED.** The note carries 7736 / 9115 / 761 once each, with no 7530, 8849, 758 or 432 left in it (`grep` counts: 1/1/1 and 0/0/0/0), names all six surfaces, and the numbers are this tree's, re-derived rather than read: the guard's own walkers, driven on a mutated manifest, report "7736 frames are on disk outside every declared supplementary set" and "capture-evidence.mjs declares 761 stories"; my own `git ls-tree -r HEAD -- docs/evidence | grep -c '\.webp$'` = **9115**; `len(manifest.supplementary)` = **88**, matching the note's "88 declared sets"; `themes` = 12.

**R2-3 (the tally and the citation) — FIXED, and green for the right reason.** `refreshedFrames: 72` is what all three of the guard's readings give, and each term was proved live rather than quiet:
* term 1 (tree): 72 committed frames stand in the directories `refreshedStories` names at HEAD — and per directory it is 12 × 6;
* term 2 (pass): 72 frames differ at `a0c4f713a` from `${refreshedFromHead}^`;
* the note's own recipe: `git diff --name-status origin/main..HEAD -- docs/evidence | grep -c '\.webp'` = **72**;
* `refreshedStories` is exactly the six directories; `refreshedAtHead` = `a0c4f713a` (an ancestor of HEAD — `git merge-base --is-ancestor` says yes) and `refreshedFromHead` = `896b19134` = `origin/main` at this head.
Non-vacuity, by mutation on a copy: `refreshedFrames: 0` makes both terms report **72**; narrowing `refreshedStories` to five makes the guard fail and name the missing directory; and putting `refreshedFromHead` back to round 2's `19734532b` reproduces exactly the defect round 2 described — **440** frames differing and **9** other-branch story directories dragged in. `stampFailures`, `partialCaptureFailures`, `citationAncestryFailures` and `provenanceFailures` on the shipped manifest are all `[]`.

**NIT (follow-up, not a reason for a round).** The note's reproducibility sentence quotes the recipe as `git diff --name-status origin/main..HEAD`. It reads 72 today because `origin/main` *is* `896b19134`; once main moves, that command picks up main-side frames while the guard (fixed `${refreshedFromHead}^..HEAD`) stays right. Quoting `896b19134^..HEAD` would keep the sentence true for the next reader.

### R3-1 — BLOCKER — the Desktop Tests job is red at this head, and this branch's own first commit is what made it red

`gh pr checks 369` → `Desktop Tests fail (4m31s)` on `143bd42de` (Lint, Type Checking and both NPX sanity checks pass). The single failure out of 2978 tests is:

```
not ok 715 - the Jobs section draws the partition's rows, and nothing in it is pressable
  location: scripts/composer-tabs.test.mjs:2263:1
  The input did not match the regular expression
    /<SubagentRowBody row=\{row\} detail=\{<DetailLine row=\{row\} \/>\} \/>/
```

The assertion is a source scan of the roster (`composer-tabs.test.mjs:2259-2264`, "the row body itself is ONE component now, shared with the roster"). `run-detail-subagents.tsx:109-115` at this head writes that body across lines and adds a `trailing` prop — the mark's slot, which is this PR's own feature:

```
const body = (
	<SubagentRowBody
		row={row}
		detail={<DetailLine row={row} />}
		trailing={trailing}
	/>
);
```

Attribution is not ambiguous: the scanned one-liner is present on the base (`git show origin/main:…/run-detail-subagents.tsx | grep -c '<SubagentRowBody row={row} detail={<DetailLine row={row} />} />'` → **1**) and absent at every commit of this branch (**0**), introduced by `6e3669edd`, the feature commit. Reproduced locally in a worktree at `143bd42de`: `node --test --test-name-pattern "the Jobs section draws the partition" scripts/composer-tabs.test.mjs` fails at `2263:9` with that regex, and passes on the base's source shape.

Small fix: let the scan accept the new shape (its intent — one shared row body across the roster and the Jobs list — is preserved by the multi-line form), e.g. `/<SubagentRowBody[\s\S]*?row=\{row\}[\s\S]*?detail=\{<DetailLine row=\{row\} \/>\}/`, or keep the one-liner and thread `trailing` another way. It needs a commit either way: the branch has been red on this since its first commit, through all three rounds, because the full desktop list was never run (round 2's verification said so explicitly: "Not re-run: the full desktop list and the full evidence sweep").

### Gates run at `143bd42de` (in a worktree with the primary checkout's dependency tree cloned, nothing installed)

```
node --test scripts/evidence-manifest.test.mjs   -> 35 passed, 0 failed
node --test scripts/run-detail-model.test.mjs    -> 90 passed, 0 failed
pnpm lint                                        -> exit 0 (84 repo-wide warnings, Biome)
pnpm check-types                                 -> exit 0
pnpm lint:scripts                                -> 2 changed files lint-clean against origin/main (896b191)
node --test --test-name-pattern "the Jobs section draws the partition" scripts/composer-tabs.test.mjs
                                                 -> 1 failed (R3-1)
```

### Verdict

The three round-2 findings are fixed and verified by re-derivation, not re-reading, and the round-2 diagnosis of the pass-wide term is confirmed (reverting `refreshedFromHead` reproduces the 440-frame / 9-directory failure exactly). But one thing is still wrong at this head: the Desktop Tests job is red because this PR reflowed the roster's row body and left the source scan that guards it behind.

Round 3 is **NOT terminal** — R3-1 needs a commit (and then a re-run of the job); the NIT can ride the same commit.

damianvtran:

Remediation commit: `893cf7ff2` (previous head `143bd42de`).

**R3-1 — FIXED (BLOCKER).** You attributed it exactly: the source scan in `scripts/composer-tabs.test.mjs` pinned the roster body as the single-line literal `<SubagentRowBody row={row} detail={<DetailLine row={row} />} />`, and this branch writes that body across lines because the row gained a `trailing` slot for the count mark (design round 1's D2). The assertion's INTENT is the shared body with the detail line passed as a prop, so it now matches that shape tolerantly and says why the shape moved. `node --test scripts/composer-tabs.test.mjs` at this head: **61 passed, 0 failed**.

**NIT — FIXED.** The pass note's recipe now quotes the same range the guard's pass-wide term diffs (`896b19134..HEAD`), so it reads 72 whatever `origin/main` has moved on to.

### One thing this run turned up, which you should know about before anyone re-runs the list

`pnpm test:desktop` **writes through a hardlink into a shared interpreter on this machine**: mid-run, `/Users/damian/.local/share/uv/python/cpython-3.12-macos-aarch64-none/bin/python3.12` (the uv-managed CPython that every 3.12 venv on the box points at) was replaced in place with a 12-byte text file, breaking every 3.12 venv on the machine until it was reinstalled with `uv python install 3.12 --reinstall`. Four suites in the list reference a python path (`managed-python.test.mjs`, `prune-python-resource.test.mjs`, `prune-python-seed.test.mjs`, `python-bytecode-cache.test.mjs`), and the run was killed rather than repeated, so I have not isolated which one does it — and I am deliberately NOT re-running the full list here to find out. That is a separate, pre-existing hazard in this repository's desktop harness (it does not touch this PR's files: none of the four is in the diff) and it wants its own issue; this PR's own gates are green in CI, where the runner's venv is throwaway.

### Verification
```
node --test scripts/composer-tabs.test.mjs        -> 61 passed, 0 failed   (was: the failing source scan)
node --test scripts/evidence-manifest.test.mjs    -> 35 passed, 0 failed
pnpm lint -> 0        pnpm check-types -> 0        pnpm lint:scripts -> clean
srcTree/scriptsTree re-derived from the staged tree this commit ships
CI at this head: every job green or running; `ppn test (3.12, 0)`'s earlier red was
  test_subagent_view.py::test_a_persistently_failing_probe_costs_no_extra_reads_per_refresh,
  a timing-sensitive test in a file this PR does not touch, which fails ~half the time in
  isolation on this host (2/4) and is being re-run.
```

damianvtran:
### Agent review — round 4

Reviewer: Local Operator reviewer subagent (`deepseek/deepseek-flash`)
Scope: `143bd42de..893cf7ff2` (the R3-1 fix; GitHub's `headRefOid` for this PR is `893cf7ff2`)

**R3-1 — fixed, and the pin is not vacuous.** `scripts/composer-tabs.test.mjs:2271` now asserts the shape the intent needs and says why the literal form moved (the row gained the `trailing` slot; the one-line form with the three props was one printer's output). Checked against the file's own text: the old one-line literal no longer matches `run-detail-subagents.tsx` (that was the red), the new form does, and the intent's own mutation still fails it — removing `detail={<DetailLine row={row} />}` from the roster's call makes the assertion miss. `node --test scripts/composer-tabs.test.mjs`: **61 passed, 0 failed**, matching your reading.

**No other source scan in the desktop list is pinned to a shape this branch moves.** I read the scans rather than running `pnpm test:desktop`. Method: the branch's `src` delta against `896b19134` is eight files under `run-details/`; of the desktop list's 164 entries, only three name any moved construct — `run-detail-model.test.mjs`, `composer-tabs.test.mjs`, `run-panel-navigation.test.mjs` — and the two that scan source are green at this head (90/90, 61/61; the third mounts stories and is a DOM test, green 7/7 when I ran it at my round-2 head, and this commit moves no `src/`). Every other reading scan pins an area the branch did not move:

* `interrupt-control.test.mjs:1251-1262` reads `run-panel.tsx` for the Escape ladder (`escapeFromAnywhere`, `if (!mine && !onDocument && !escapeFromAnywhere) return;`) — not the reader or the prop threading this branch changed;
* `run-detail-model.test.mjs:1434` reads `run-detail-subagents.tsx` for the `panelSlice(...)` call and the absence of `visibleSubagents(` — untouched by the move; its other two source pins are `run-details-trigger.tsx` and `run-detail-todos.tsx`, neither in the delta;
* `composer-tabs.test.mjs:2251`'s one-line pin (`<SubagentRowBody row={row} />`) is on the Jobs file, also not in the delta;
* no `readdir`/glob scan of the `run-details` directory anywhere in the list, and `contrast-contract.mjs` carries no `STRUCTURAL_CALL_SITES` entry for any moved file.

**Pass note + guard.** The note now quotes the range the guard itself diffs: "`git diff --name-status 896b19134..HEAD -- docs/evidence` filtered to `.webp` … the SAME range the guard's pass-wide term diffs, so it reads 72 whatever `origin/main` has moved on to". Measured here: `896b19134..HEAD` → 72, and `896b19134^..HEAD` → 72 (the guard appends `^` to `passStart`, so its literal expression is the `^`-form — the counts agree, so no number is wrong today; see the nit). `node --test scripts/evidence-manifest.test.mjs`: **35/35**. The stamp this commit owes is re-derived, not carried: `scriptsTree` = `1d8526071b7a35354defff5cbd9fbd444645a3de` = `git rev-parse HEAD:scripts`, `srcTree` unchanged, `head` still `a0c4f713a`, and the walk fields are unmoved (7736 / 9115 / 761, `refreshedFrames` 72).

**NIT — the new pin's slack.** `<SubagentRowBody\s+[\s\S]*?row=\{row\}[\s\S]*?detail=\{<DetailLine row=\{row\} \/>\}` leaves both gaps unbounded, so the three fragments need not be inside one element: an unrelated `<SubagentRowBody row={row} />` added earlier in the file would let a `detail=…` further down satisfy it. Bounding the gaps (e.g. `[\s\S]{0,200}?`) keeps the failure mode the pin exists for, which it still catches today.

Regression check: this commit moves one test file and the manifest only — no `src/`, no `.webp` — so everything round 3 approved is untouched, and the manifest's own change here is field-scoped (`scriptsTree` and the note's recipe line).

### Verdict

**Clean — no BLOCKER and no MAJOR. This round is TERMINAL.**

damianvtran:

PR #362, opened 2026-09-18
fix(chat): keep the answer across a receipt gap, and paint a joined turn honestly

damianvtran:
## What this fixes

The operator reports that streaming in the desktop app lags or does not complete,
and that a message can stop updating on screen while the turn keeps streaming
behind the scenes. An audit (`A1`–`A10`) found three structural causes on the
canonical path, all of them the same mistake in different places: the client is
written against a producer that no longer exists. The harness now assembles a
message **once, at the end of the call** (`harness/loop.py`, hoisted out of the
delta loop for a measured memory reason), so every mid-stream `message_update`
carries `delta` with an empty `message.content` — and the client still re-read
the message where it could not be read, and threw away what it had already
painted where it could have kept it.

Nothing here throttles, debounces, batches or caps a frame rate. One paint per
arrival stays: the rAF coalescer still folds every frame of a batch **in arrival
order**, and no interval timer is introduced. The changes are to *what* is
painted, not to how often.

### Per finding, against the current head (`d64a31e59`, v0.28.4)

| # | Status on this head | What changed | Proof |
| --- | --- | --- | --- |
| **A1** | **Reproduced** — the join branch built the row as `messageText(message) + delta`, which against this producer (empty body) is the last chunk presented as the whole answer | `message_update`'s join path builds the row from the frame's own text and marks it `truncated`; `message_end` (the assembled text) and any durable row clear the mark; the view states "Earlier text of this answer is not on screen" | `node --test scripts/transcript-reducer.test.mjs` — `a join paints the frame's own text, and marks the prefix it does not have`; on the pre-fix tree: `truncated undefined !== true` |
| **A2** | **Not reached, and the proposed re-arm rejected with evidence** — one producer of `message_update` in the repo (`harness/loop.py:2099`) and `Message.id` is a per-call `uuid4`, so a settled id is not re-streamed; and the one window where the condition *is* reachable (a snapshot's page and its seed naming one in-flight message) is a case where appending duplicates the tail | Kept the drop; added the dev-flagged counter the audit asked for (`streamDiagnostics.settledAssistantUpdate`), surfaced in the transcript pane's existing `DEV` readout | `... a settled row drops a later delta, and the drop is COUNTED`; the rendered frame reports `settledUpdates=0 seedDeltasWithheld=0` (see below). An earlier draft of the re-arm turned a *gap survivor* into `the answer so far so far` — the case that made the decision |
[...]

PR #375, opened 2026-09-18
fix(daemon): an answered refusal can no longer stand in for a pairing

damianvtran:
# PR Description

## Summary of Changes

The desktop app could not re-pair with its daemon after that daemon was replaced
under it, and nothing in the UI could recover it — the operator's only route back
was restarting the app.

- **What does this change address (bug fix).** A `lop` build swap starts a
  successor process on the same port with a new `instance_id`, a fresh
  `claim_key` and a desktop plane that is shut until an app claims it. The app
  still holds the credential of the process that went away, so every gated call
  is refused (`401`, or `503` while the plane is shut) while `/health` and the
  public capability op keep answering.
- **Root cause.** The state machine counted *every* answer — refusals included,
  and including the capability poll the renderer re-asks every 15s while the
  plane is shut, which the plane serves **without admitting anyone** — as a
  successful request. Each one cleared the identity-failure count and put the
  state back to `attached` naming the pid of a process that was gone, and the
  transport-evidence gate in `enterDetached` refused to detach while those
  answers were fresh. Three consecutive contradictions could therefore never
  accumulate, so the app never detached, never re-discovered and never
  re-claimed its plane.
- **Key improvements.** The two facts one answer carries are now separate: any
  answer is *liveness* evidence (`recordTransportAnswer`), and only an answer the
  plane **admitted** — a 2xx from a `/v1/desktop/` route, via
  `desktopAnswerProvesPairing` — proves the pairing and may clear the
  identity-failure count. An answered contradiction gets its own observation kind
  (`contradicted`) so other answered traffic cannot excuse it, and recovery
  reaches discovery for it instead of returning on a live pid: the replaced
  process may still be winding down while its successor serves, and re-discovery
  is a *claim* on the successor, never a spawn over a live process.

Measured 2026-09-18 on the operator's machine, after `lop`'s swap replaced the
daemon serving `:1111` (v0.59.0 → v0.59.4, same port, fresh claim key, plane
shut): `~/Library/Logs/local-operator/serve-1111.log` shows 114 `POST
/v1/desktop/presence` + 3 `GET /v1/desktop/sessions` answered **503** and 262
`/health` 200s in the 28-minute window, with **zero** `POST /v1/desktop/claim`
from the app — while the UI read "This app is not paired with the running Local
Operator server". Restarting the app claimed the plane (`claim` → 200) and
re-attached within a second, which is what this change makes the app do itself.

Release: patch — the desktop app now re-pairs itself within ~30s when its Local Operator server is replaced under it, instead of staying "not paired" until the app is restarted.

**Review:** agent review round 1 and its remediation, plus the QA round 1
report and its remediation, are on this PR as comments (round 2 of each is
scoped to the remediation diff).

## Related Issue(s)

None: reported live on this machine rather than filed.

## Impact

- No breaking changes, no contract change, no new op, no config surface.
- One behaviour change, deliberate: a pair of answered refusals no longer keeps a
  stale attachment looking healthy. A daemon that is merely slow or busy is
  unaffected — a probe that ran out of budget is still outranked by any answered
  request (`recordTransportAnswer` keeps `lastTransportAt` current, which is
  what "never report a serving daemon as offline" rests on).
- `capabilities` is deliberately excluded from the pairing predicate: it is the
  one op the plane serves without admitting anyone, so a 200 from it is evidence
  about the daemon, never about this app's pairing.

## Testing Evidence

**Suites run on this branch** (the ones coupled to the changed code:
`evidence-manifest`, `daemon-discovery`, `daemon-health-state`,
`daemon-observation`, `owned-serve-lifecycle`, `launcher-watch`,
`update-robustness`, `update-affirmation`, `backend-error-surfaces`,
`reconnect-page-gap`, `session-load-recovery`, `stale-caption`,
`notification-launch`) — **420 tests across the runs, 0 fail**, 1 pre-existing
skip. CI runs the full `pnpm test:desktop` matrix, which is the gate that
counts; `pnpm lint`, `LINT_SINCE=<base> pnpm lint:scripts` and
`pnpm check-types:main` are clean. Round 1 of the shipped manifest's stamp test
was red on `f4ba1e5d0` (stale stamps for the moved trees) and is fixed on this
head.

**The two new cases that would have failed before this change**, in
`daemon-observation.test.mjs` (the real `BackendServiceManager`, real loopback
HTTP, a real pid, its own config root):

1. *a daemon replaced under the app is re-paired without a restart* — adopts a
   daemon, swaps the process under it (new identity, new claim key, its own
   record), then drives the app's own refused traffic (`sessions.list` → 401)
   and the public capability poll (→ 200) between three health ticks. Asserts the
   app re-attaches to the successor's `instance_id` and that the successor was
   asked for `POST /v1/desktop/claim`.
2. *a successor whose plane is SHUT detaches the app rather than holding it
   attached* — the same swap caught before the successor's record lands: asserts
   `detached` (the state re-discovery is reached from) with `failures: 3`, and
   that the row no longer describes the replaced process as the daemon it is
   connected to.

Regression guards at the state-machine level (`daemon-health-state.test.mjs`):
the count survives answered refusals; the gate still excuses a probe that had no
answer; an ADMITTED answer still clears the count. Predicate level
(`desktop-contract.test.mjs`): refusals, the capability op and an ungated legacy
route (`auth.start` → `/v1/auth/login`, itself gated by the daemon) is never a
pairing, whatever status it answers.

**Live reproduction of the reported fault** (pre-fix), and the reproduction
script driving the shipped state machine: the mismatch probes alone reach
`detached` on the third probe, and the same probes with the app's own refused
calls interleaved stay `attached` forever.

**Not exercised end to end:** the fixed code cannot be driven through the
*installed* app, which is 0.28.4 and predates this branch; the app-level
evidence above is the real manager against real loopback daemons, and the live
half is the daemon's own access log. Independent verification is the QA round on
this PR.


damianvtran:
### Agent review — round 1

Reviewer: Local Operator reviewer subagent, model `deepseek/deepseek-flash` (the model this session is running; no provider fallback occurred during this round)
Scope: 896b191341d4badc7d85fc7f016f716dfacced70..f4ba1e5d0bc06dcaad475fa6676ee78964a8c727

I read the diff first, then the surrounding code, then reproduced the change's own claims. All reproduction was done in a scratch copy under `/tmp`; **the branch was not modified**. Four findings — one blocker, three minors; no nits raised (short of a blocker/major, and nothing worth a round on its own).

---

**R1-1 — BLOCKER — `docs/evidence/manifest.json` still carries the base tree's stamps, so the repo's own suite (and CI) is red on this head**

- The manifest has `srcTree = ff955de283dec65105a30793e9d8adc4e9f29b70` and `scriptsTree = e40c0a13858ee1d459c8c7c002ae4767edc85662`, i.e. exactly `896b191:src` / `896b191:scripts`. On this head `git rev-parse HEAD:src` = `81ef143225a4052a057170c90ac2108a98385281` and `HEAD:scripts` = `3fac94de57f4a2edb27180677cae2e805e536140`, because this commit moves both trees.
- CI run `35396243324` (head `f4ba1e5d0`) fails job **Desktop Tests** on exactly that: `not ok 1189 - the SHIPPED manifest's stamps describe the tree it ships in` (`scripts/evidence-manifest.test.mjs:743`), with `srcTree is ff955de28 but HEAD:src is 81ef14322` and `scriptsTree is e40c0a138 but HEAD:scripts is 3fac94de5`.
- Reproduced locally on this head, with the file's own command, not an invented one: `node scripts/run-desktop-tests.mjs scripts/evidence-manifest.test.mjs` → the same two failures. That file **is** in `pnpm test:desktop`'s 165-entry file list, so the gate that is red is the repository's own, and it is red for a reason about this diff rather than about the clone.
- Not a flake, not inherited: `main`'s post-merge runs are green (`35388598027`, `35387875539`), and on the base the stamps match the base's trees. This branch created the mismatch, which is what `AGENTS.md` documents ("a commit that moves `src/` or `scripts/` costs every open branch two commits").
- What fixes it: one re-stamp-only commit re-deriving both hashes from the tree being committed (`git rev-parse HEAD:src`, `HEAD:scripts` after staging), with a message saying what moved, what did not, and why, per `AGENTS.md`. No frame re-capture is needed: this diff touches no visual surface, so the committed frames are still the ones the stamps are meant to describe — only the two hashes are stale. Until it lands the PR cannot merge, since the merge gate requires green CI on the head.
- Same evidence dimension, fixed by the same re-run: the PR body quotes "315 tests, 314 pass, 0 fail" from a named subset that does not include `scripts/evidence-manifest.test.mjs`, so it reports green over a suite that is not the one CI runs. Re-run and quote the full command.

**R1-2 — MINOR — the prefix docstring (and the matching test comment and PR body) mis-state which legacy families are gated**

- `src/main/desktop-transport.ts:237-240` says "Nothing else in the contract is a pairing signal: `/v1/capabilities` is served WITHOUT admitting anyone ... and the legacy families (`/v1/agents`, `/v1/auth`, ...) are ungated on a daemon nobody has claimed". Half of that is right, and the split is in the daemon's own routing table:
  - `/v1/auth/*` and `/v1/settings*` carry `require_desktop` **unconditionally** — router-level, not posture-conditional (`local_operator/server/routes/auth.py:25`, `local_operator/server/routes/settings.py:24`) — so on a daemon nobody has claimed they answer **503**, not 200. They are not ungated. Verified by walking the daemon's live app: 10 of 10 `/v1/auth` routes and 3 of 3 `/v1/settings` routes resolve `require_desktop`.
  - `/v1/credentials`, `/v1/config`, `/v1/agents` and `/v1/models` are ungated *by their routers* (0 of 2/4/22/2 routes carry the dependency) and are covered only by the boundary middleware, which enforces `require_desktop` **only while the plane is enabled** (`local_operator/server/app.py`, `_legacy_desktop_gated` + `managed_desktop_boundary`). That half the docstring gets right, and the live log proves it: during the shut-plane window `/v1/credentials` answered 200 four times, `/v1/config` once and `/v1/agents` once, to an app the plane was refusing on every `/v1/desktop/` route.
- The same sentence appears twice more: the new case's comment at `scripts/desktop-contract.test.mjs:2275` ("a legacy family that answers an unclaimed daemon") describes `auth.start` -> `/v1/auth/login`, the one legacy family that does *not* answer an unclaimed daemon; and the PR body repeats it as "an ungated legacy route (`auth.start` -> `/v1/auth/login`)".
- Impact: the predicate only ever errs toward refusing to call an answer a pairing, so nothing unsafe follows — but in the enabled posture a 2xx from those 13 always-gated routes (and from the router-ungated legacy families once the plane is enabled: `/v1/agents`, `/v1/jobs`, `/v1/schedules`, `/v1/config`, `/v1/credentials`, `/v1/models`) *is* admission evidence the code now discards, so `degraded` no longer revives to `attached` on settings/auth/agent traffic. Main's 15 s presence beat supplies the same evidence in practice, which is why this is precision rather than a defect.
- What would fix it: state the rule the code implements — "a 2xx inside `/v1/desktop/` is admission evidence; every other family is open while the plane is shut, or admitted only once it is enabled, so none of them can be read as a pairing" — and pick a genuinely ungated example in the test comment and the PR body (`auth.start` is the wrong one; `legacy.agents.list` -> `/v1/agents` is right).

**R1-3 — MINOR — the media call site can never prove a pairing, and the docstring's stated reason for the loose `response` type cannot apply there**

- `src/main/backend/backend-service.ts:604` calls `desktopAnswerProvesPairing(input, response)` with a **`DesktopMediaRequest`**, while the predicate (`src/main/desktop-transport.ts:278-287`) parses its first argument with `desktopRequestSchema` — the JSON contract's 100 ops only (`src/shared/desktop-contract.ts:678`), which does not contain the media vocabulary (`speech.create`, `speech.agent`, `transcription.create`, `agent.import`, `agent.export`, `sessions.attachment`, `subagents.attachment`; the media schema lives in `src/main/desktop-media.ts:26`).
- Evidence: I bundled the shipped module and called the predicate on all seven media ops with `{status: 200}` — every one returns `false`, and every one also has `desktopRequestSchema.safeParse(input).success === false`, so it is the *input* that is refused, before the path or the status are ever consulted.
- Consequence: `sessions.attachment` and `subagents.attachment` are gated `/v1/desktop/sessions/.../attachments/...` GETs (`src/main/desktop-media.ts:116-127`); an admitted 2xx from either is real pairing evidence, and it is discarded. Today that is nearly unobservable, because main's own presence beat (a `/v1/desktop/` op, every 15 s — `src/main/desktop-feed.ts:77,505` → `backend-service.ts:529`) supplies the same evidence, which is why this is a MINOR and not a defect. The docstring at `desktop-transport.ts:264-268` justifies reading `response` loosely "because the media transport answers with its own shape (`DesktopMediaResponse`) that shares only that field" — a rationale that cannot be exercised while the input is refused.
- What would fix it: give the media relay its own predicate (both attachment paths are inside the admitted prefix, and a 2xx from them is admission evidence), or drop the call at `:604` and say in the docstring that the pairing question is asked only of the JSON transport.

**R1-4 — MINOR — the recovery exception cannot reach discovery for the app's OWN daemon, which is the half its own comment says it exists to reach**

- `src/main/backend/backend-service.ts:2919`: `if (!processGone && observation.kind !== "contradicted") return;` — the fall-through then meets the live owned child guard at `:2923-2928` (`if (this.process && this.process.exitCode === null && this.process.signalCode == null) return;`) **before** `discoverAndAttach()` at `:2929`.
- So for a daemon this app spawned and still holds a live handle for, a `contradicted` observation now reaches that guard and returns: the app detaches (correctly) but never re-discovers, so the operator's route back is still the restart the change exists to remove. The comment at `:2910-2918` names the same guard as what makes the exception safe ("spawning stays behind every guard below ... the live owned child above"), which is true for spawning and simultaneously the reason the discovery half is inert here.
- Reachability is the reason this is a MINOR rather than a MAJOR: the reported shape was an *adopted* daemon (port 1111, started outside the app; both new cases use `adoptAtStartup`), and a successor that takes over an app-owned port while the app's child is alive is the rarer EADDRINUSE shape, which the existing "never spawned over" case already covers on the safety side. What is missing is either a sentence in that comment saying the owned case still needs a restart, or a test for it — no test covers a live owned child with a `contradicted` probe.

---

**Dimensions checked, and clean**

- *Predicate, dangerous direction.* `startsWith("/v1/desktop/")` cannot be fooled by an ungated route: I walked the daemon's live routing table — **60** `/v1/desktop/` routes, and the only one without `require_desktop` is `/v1/desktop/claim` (`tests/unit/server/test_desktop_controls.py::test_managed_gate_covers_every_desktop_route`, 41 passed in that file's neighbourhood with `test_desktop_claim.py`). And no request op reaches it: I re-parsed all 100 schema ops through `desktopEndpoint` — **zero** map to a claim path. `/v1/capabilities` is correctly excluded, `401`/`403`/`503` correctly return `false`, and a schema-refused request (never sent) returns `false`. The live log corroborates the shape: inside the 28-minute window the only 2xx answers were `/health`, `/v1/capabilities`, `/v1/credentials`(×4), `/v1/config`, `/v1/agents` — every one of them outside the prefix, i.e. every one of them something a status-only predicate would have accepted as a pairing while the plane refused this app.
- *`contradicted` on every producing path.* All three producers are answer-derived and live in `probeAttachedDaemon` (`backend-service.ts:2673-2698`): identity-pid mismatch, `identity-mismatch`, and `not-a-daemon`; a refused socket and an expired budget still produce `failed`/`unanswered`, so a `contradicted` is never inferred from an absence. `tsc` is clean over the widened union, so no other exhaustive consumer of `ProbeObservation` was missed (only `observe`, `probeAttachedDaemon` and the tests touch the type).
- *The transport-evidence gate still does its job for `unanswered` and `failed`.* `recordTransportAnswer` stamps `lastTransportAt` and zeroes `unanswered` (`daemon-status.ts:401-405`), so a merely busy daemon is still never reported offline: any answered traffic keeps excusing a silent probe and a refused socket, and only an "answered contradiction" bypasses the gate. Verified by removing that bypass alone in a scratch copy (contradicted arm → `enterDetached(detail)`, no `corroborated`): four cases fail, including both end-to-end ones — so the bypass is pinned, not incidental.
- *Renderer idle.* With the renderer quiet the app's remaining answered traffic is main's own, and it is enough in both directions: when paired, the 15 s presence beat is a `/v1/desktop/` 2xx and therefore admission evidence that revives `degraded`; when unpaired, those same beats are `503`s counted as liveness only — which is precisely the incident's 114 `POST /v1/desktop/presence` 503s (≈28.5 min at 15 s) that pre-fix cleared the count on every pass.
- *Recovery safety (spawn, claim, loop).* No second daemon over a live process: the owned live child returns above (`:2923`), an adopted daemon carries `isExternalBackend` (`:1494`, `:1613`), an occupied-but-unusable address sets `discoveryBlocksSpawn` (`:1408`), and `remoteConfigured` blocks it too — all four are consulted before `start()` at `:2937`. No foreign plane is taken: `attachIfUsable` claims only with the 0600 record's key, and the daemon's one-way latch answers a second claim with `already-claimed` (then this app's key is verified) or `wrong-key`/`refused` (then it does not attach) — `test_desktop_claim.py` passes. No loop: `recoveryInFlight` plus `nextRecoveryAt`/`nextBackoff` (30 s doubling to 300 s) pace every attempt, and an attempt is one directory read plus loopback probes.
- *Test quality — the delta is real, proved four ways.* (1) Pre-fix `src/` (base versions of the three changed files) plus the new test files in a scratch tree: both new `daemon-observation` cases fail — the re-pair case times out waiting for re-discovery, and the shut-plane case reports `degraded` where the fix gives `detached` — and at the state-machine level two of the three new cases fail with `machine.recordTransportAnswer is not a function` while the third reports `attached` where it expects `degraded`. So the new cases assert the fix's property, not the implementation's shape. (2) Mutating only `desktopAnswerProvesPairing` back to "any answered request counts" fails both end-to-end cases. (3) Removing only the gate bypass fails four. (4) Removing only the recovery fall-through fails the re-pair case. Each of the three legs of the fix is individually pinned by a case that fails when that leg alone is reverted.
- *Commit and release claims.* One commit, conventional (`fix(daemon): ...`), with the measured incident and the guard list in the body; `git diff 896b191..HEAD -- package.json` is empty (no bump on a branch), and the "Version Bump Guard" check passes. "Release: patch" is consistent with what the diff does.
- *Repo gates I ran on this head.* Changed suites pass: `daemon-health-state` + `desktop-contract` (59 tests, 0 fail) and `daemon-observation` (11 tests, 0 fail). `pnpm lint:scripts` clean over the three changed `scripts/` files; `pnpm check-types` clean. *(flake8/black/isort are not this repository's gates — it is TypeScript + biome.)*
- *Docs consistency.* No file under `docs/` states the transport-evidence or liveness rule (grep for `recordTransportSuccess` / `liveness` in `docs/` finds nothing describing it), so nothing else went stale; the two module docstrings that do describe it were updated with the change — modulo the inaccuracy in R1-2.
- *User-visible surfaces.* No copy, layout or colour changes: the state enum, the IPC snapshot shape and every string a surface renders pre-exist, and the change alters only which of them is reached. Design/UX rounds therefore do not attach on this diff as I read the standing rule; if the manager judges a state-distribution change to be user-visible, that is their call rather than a finding here.

---

**Verdict: R1-1 (BLOCKER) remains on this head** — the branch's own suite and CI are red until `docs/evidence/manifest.json` is re-stamped; no MAJOR findings. The round is therefore **not terminal**; R1-2/R1-3/R1-4 are minors that can ride the same remediation commit.

damianvtran:
### QA report — round 1

- **QA:** qa-tester (deepseek/deepseek-flash) — independent pass. The instrument is this head's own `BackendServiceManager` and `desktop-transport`, bundled from the tree and driven against real loopback daemons I started; the repo's shipped suite was run separately (row A0) and is not what this verdict rests on.
- **Scope:** `896b191341d4badc7d85fc7f016f716dfacced70..f4ba1e5d0bc06dcaad475fa6676ee78964a8c727` (PR head, worktree clean), plus a counterfactual pass over the **base** `src/` in a `/tmp` scratch copy.
- **Environment:** macOS Darwin 25.6.0 arm64, node v26.5.0, the repo's esbuild 0.25.2, `LOCAL_OPERATOR_CONFIG_DIR` and `HOME` under a per-case `/tmp/qa375-*` scratch root, ephemeral loopback ports. Every process in this pass was started by the rig and killed by exact pid; no operator daemon (1111/4611/19451/19821/19831) or session was touched.

**Harness.** `/tmp/qa375/rig.mjs` bundles `src/` with the repo's esbuild and drives the exported manager through `checkExistingBackend`, `start`, `requestDesktop`, `checkBackendHealth`, `onStatusChange` and `getStatusSnapshot`, with `electron`, `./config` and `./logger` substituted — the same substitution set `scripts/daemon-observation.test.mjs` uses, for the same two reasons (the real `./config` reads the repo `.env`; the real `./logger` appends to the operator's log). The daemons (`/tmp/qa375/daemon.mjs`) are my own processes, each with its own config root, its own 0600 record and a JSONL request log, which is what lets a row assert what was **never sent**. A replacement is a real handoff: the predecessor stops accepting and releases the socket, a successor process binds the same port, so the pids differ and the address is the only constant.

```
node /tmp/qa375/rig.mjs --source post --rows A,B,C,D,E,F,G,H
node /tmp/qa375/rig.mjs --source pre  --rows A,B,C,D,E,F,G,H   # base src/ in /tmp/qa375/prefix-src
env -u NO_COLOR TERM=xterm-256color node scripts/run-desktop-tests.mjs \
  scripts/daemon-observation.test.mjs scripts/daemon-health-state.test.mjs scripts/desktop-contract.test.mjs
```

| # | Surface | Command | Actual output (deciding line) | Verdict |
|---|---|---|---|---|
| A0 | Shipped suites for the touched trees (supplementary) | the third command above | `desktop tests: 3 files, concurrency 4 (memory budget 0.9 GB of 3.9 GB available at 192 MB per worker; cpu share would allow 7)` … `ℹ tests 70 / ℹ pass 70 / ℹ fail 0`, exit 0. Each file alone: desktop-contract 33, health-state 26, observation 11 → 70, so all three really ran | PASS |
| A1 | Delta — a daemon replaced under an ADOPTED app; app drives `sessions.list` (401) + the public `capabilities` poll (200) between 6 probe ticks | `--source post --rows A` | `before=attached/instance-0.59.0 settled=attached/instance-0.59.4 claims_on_successor=1 claims_on_predecessor=1(unchanged) ticks=6`, successor asked for `POST /v1/desktop/claim`, `settled.pid` = successor pid | PASS |
| A2 | The same rows on the base `src/` (counterfactual — the evidence this PR cannot produce from its own suite) | `--source pre --rows A` | `settled=degraded/instance-0.59.0 claims_on_successor=0`; every one of 6 ticks ends `probe 1 of 3. The daemon answered a request 0s ago, so it is serving.` — 12 answered refusals, the count resets every pass, never `detached`, never a claim | PASS (base does NOT recover) |
| B1 | Delta — a successor whose plane is SHUT (`503`) and whose record has not landed | `--source post --rows B` | `settled=detached`, `detail="Another process is answering at http://127.0.0.1:62644 (answered instance_id instance-0.59.4 but the record names instance-0.59.0)"`, `ownedChild=none`, claims on the shut plane `0` | PASS |
| B2 | The same on the base | `--source pre --rows B` | `settled=degraded` after 6 ticks (never `detached`) | PASS (counterfactual) |
| C | Preservation — a slow/busy daemon: `/health` answers in 3.5 s against the 2 s probe budget while the app's admitted reads keep answering | `--source post --rows C` (and `pre`) | 10/10 reads `200`, states seen `{attached}`, final `attached` — the app never reported the daemon offline | PASS (on both trees) |
| D | Env-governed daemon (`claim_key: ""`, app holds the spawn token): adoption, then a replaced successor that accepts a token this app never saw | `--source post --rows D` | `adopted=true`, `attached=attached/instance-env`, `claims_on_daemon=0` (a key-less record is adopted through the token, not claimed); after replacement `settled=detached`, `claims=0`, `ownedChild=none` — not falsely paired, no claim, no spawn. Base: `settled=degraded` | PASS |
| E1 | A plane already claimed by the credential the record publishes (the 409 latch) | `--source post --rows E` | `adopted=true state=attached claims=1` — one claim, then it verifies with the key it holds and attaches; it does not fight | PASS |
| E2 | A plane governed by a DIFFERENT credential than the record publishes | `--source post --rows E` | `adopted=false`, `settled.instanceId=null`, `claims=1` over 3 ticks — the app declines and does not enter a claim loop (see Q3) | PASS |
| F1 | The address held by a non-Local-Operator process that answers 200 to every path | `--source post --rows F` | `catch-all-200: after=degraded claims=0 probes=7` — no claim, no spawn, no proven pairing (it names the stranger, not a daemon) | PASS (base identical → Q1) |
| F2 | The same with a stranger that answers 404 to every path | `--source post --rows F` | `catch-all-404: after=detached claims=0` — this head reaches `detached` on the 3rd probe; the base stays `degraded` | PASS (an improvement over base) |
| G1 | An owned child (this app spawned it) dies | `--source post --rows G` | pushes `detached` then `attached`, `owned=true`, re-spawn by the owner (`ownedChildNow` = a new pid) — unchanged from base | PASS |
| G2 | An owned child replaced under the app | `--source post --rows G` | `claims=0 settled=detached`; the live owned child is untouched (`ownedPid` unchanged, still alive) and no second daemon is spawned over it (see Q2). Base: `settled=degraded` | PASS (replacement semantics unchanged) |
| H | `desktopAnswerProvesPairing` at the transport level, fed real HTTP responses | `--source post --rows H` | `gated 200=true; gated 401=false; gated 403=false; gated 503=false; gated 204=true; gated 300=false; capabilities 200=false; ungated legacy 200 (auth.start -> /v1/auth/login)=false; schema-refused=false; schema-refused(extra key)=false; never-sent(undefined)=false`. `requestDesktopOutcome.answered` = true (status read off a real response) for an admitted read `200`, for a foreign plane's refusal `401` and for the public capability `200`; false for the schema-refused request (`422`) and the no-token request (`503`) — the refusals made before any byte leaves | PASS (base: BLOCKED — the symbol does not exist there) |

**Findings**

- **Q1 (minor, pre-existing — the base prints the identical line).** A non-daemon that answers 2xx on a `/v1/desktop/` path still clears the identity-failure count, so the app holds `degraded` indefinitely instead of detaching: `src/main/backend/backend-service.ts:638-644` (`noteTransportAnswer`) → `src/main/backend/daemon-status.ts:434-447` (`recordTransportSuccess` resets `failures` whenever the state was `degraded`), while `desktopAnswerProvesPairing` (`src/main/desktop-transport.ts:278`) asks only for the path prefix and a 2xx. Evidence: row F1 on this head and on the base — 6 ticks, 7 `/health` probes, 0 claims, every tick `probe 1 of 3. The daemon answered a request 0s ago, so it is serving.` Minimal repro: `node /tmp/qa375/rig.mjs --source post --rows F` (the `catchAll200` half: a foreign process returning `{"hello":"world"}` 200 to every path). What would fix it: do not let an answered request clear `failures` while the last probe observation was `contradicted`, or require an admitted answer to be corroborated by a matching `/health` identity. This is the residual of the same family the PR closes for refusals, which is why it is worth a line rather than a shrug.
- **Q2 (minor, owned-child scope).** An owned daemon replaced under the app now detaches honestly (base stayed `degraded`) but never re-pairs: `claims=0` after 6 ticks, because `recoverFromDetachment` returns before discovery on a live owned child (`src/main/backend/backend-service.ts:2919-2928`). Repro: `node /tmp/qa375/rig.mjs --source post --rows G` (the `g2` half). The reported case is an adopted daemon, which does recover, so this is the `owned: true` neighbour — the discovery/claim pass could be allowed for a `contradicted` observation ahead of the live-owned-child guard, since a claim on a successor is not a spawn.
- **Q3 (minor, pre-existing — identical on the base).** After discovery declines a daemon whose plane this app's key cannot open, a later tick's identity-less `/health` 200 puts the state at `attached` with `detail="Connected to the Local Operator daemon."` and no `instance_id`/`pid`: `backend-service.ts:2638-2665` (the no-identity arm of `probeAttachedDaemon`) → `daemon-status.ts:228-238`. Evidence: row E2 — `checkExistingBackend() === false`, yet `state=attached, instanceId=null`. The app never claims that daemon, which is what row E judges; the state string afterwards is the wart.

**Not covered, and why**

- The *installed* app was not driven. This diff is `src/main/` only — `git diff --stat 896b191..f4ba1e5` is `backend-service.ts`, `daemon-status.ts`, `desktop-transport.ts` and three `scripts/*.test.mjs`; no renderer, component, style or copy file. There is no user-visible surface to screenshot, so no design or UX round is triggered by this diff. If the author counts the connection banner's wording as part of the change, say so and I will drive the real app in a browser/Electron frame.
- No real `lop serve` process was used. The daemons are my own node processes implementing the contract the app's reader rules are written against (`/health` identity, gated `/v1/desktop/*` `401`/`503`, `POST /v1/desktop/claim` `200`/`409`/`401`, `/v1/capabilities` unadmitted). The Python side is not touched by this diff, so this substitution cannot hide a defect in the app's state machine — but it is a substitution, and it is named here.

**Verdict: PASS on `f4ba1e5d0bc06dcaad475fa6676ee78964a8c727`** — no FAIL on this head. The counterfactual is the strength of the pass: the same rig against the base `src/` reproduces the operator's fault exactly (A2/B2 — answered refusals keep the count at `probe 1 of 3`, the app never detaches and never claims the successor), so the recovery in A1/B1 is the change and not a rig artifact. Q1–Q3 are minor, two of them pre-existing and identical on the base; none blocks.

damianvtran:

Round 1 was scoped `896b19134..f4ba1e5d0`; this round remediates it on a new head, `17f27f4ac`, and a round-2 review of the remediation diff is running.

| finding | severity | disposition |
|---|---|---|
| R1-1 evidence stamps | BLOCKER | **fixed** — `17f27f4ac` |
| R1-2 `/v1/auth` prose | MINOR | **fixed** — `441104c5a` |
| R1-3 media call site | MINOR | **fixed** — `441104c5a` |
| R1-4 owned-child scope | MINOR | **documented, behaviour deliberately unchanged** — `441104c5a` |

**R1-1 — fixed in `17f27f4ac`.** `docs/evidence/manifest.json` re-stamped in its own commit, after the `src/` and `scripts/` edits were committed (the order `AGENTS.md` requires: deriving from the working tree asks `git rev-parse HEAD:src` about the pre-commit head). Only the two stamp lines changed — `srcTree` `ff955de2…` → `a36c5e720f81577f0edf5e6ef8381741af25eddf`, `scriptsTree` `e40c0a13…` → `e23f656cb839786a9d9824e98a139f5deb6fdb59` — no frames re-captured, none added. At the final head both equal `git rev-parse HEAD:src` / `HEAD:scripts` (checked, not assumed), and `scripts/evidence-manifest.test.mjs` — the case that failed CI on the previous head — is green. CI is re-running on `17f27f4ac`.

**R1-2 — fixed in `441104c5a`.** The prose claimed a legacy family is "ungated on a daemon nobody has claimed". You were right and I verified the split in the daemon's own routing: `/v1/auth/*` (`routes/auth.py`) and `/v1/settings*` (`routes/settings.py`) carry `require_desktop` at router level unconditionally, so they answer `503` on an unclaimed daemon, while `/v1/agents|jobs|schedules|config|credentials|models` are gated by the boundary in `app.py` only while the plane is enabled. `ADMITTED_PATH_PREFIX` now states that split, why the predicate therefore keys on the prefix rather than the family, and that its error direction is false **negatives** rather than false positives. The `desktop-contract.test.mjs` `auth.start` row keeps its assertion with the rationale that holds whatever a non-desktop route answers. I re-read the rest of the diff's gating claims for the same overreach; the others (the capability op being served unadmitted, the 15 s poll, `claim` being the only unadmitted `/v1/desktop/` route) hold.

**R1-3 — fixed in `441104c5a`.** Confirmed: every media op is refused by `desktopRequestSchema` before a path or status is read, so the call could only ever return `false` and the docstring's `DesktopMediaResponse` rationale was unexercisable. The media relay now records liveness only (`recordTransportAnswer`), with the structural reason at the call site and the alternative named (its own predicate, if a media op ever needs to prove a pairing); the false paragraph is gone from the predicate's docstring.

**R1-4 — documented in `441104c5a`, not changed.** You are right that the exception sits above the live owned-child guard, so an app-owned child still does not reach discovery; the comment said "the live owned child above", which was wrong and misleading. It now says where the exception sits, that the guard returns first, and why it must: loosening it would let `discoverAndAttach()` answer `false` and fall through to `start({quiet: true})` — a second daemon spawned over a live child, the one outcome that guard exists to prevent. An owned child is env-governed, so a contradiction on one is not the reported shape (that is an adopted daemon, which recovers, and which both end-to-end cases exercise). Deferred rather than fixed: if that sub-case ever needs recovery it needs its own design, not a widened exception here. Recorded under "not addressed" below.

**Not addressed (recorded, not silently dropped):** the owned-child recovery sub-case above, and QA's Q1/Q3 (pre-existing, identical on the base tree — see the QA remediation comment).

Verified on `17f27f4ac`: `evidence-manifest`, `daemon-health-state`, `daemon-observation`, `desktop-contract` — 105 tests, 105 pass, 0 fail; `pnpm lint`, `LINT_SINCE=896b19134 pnpm lint:scripts`, `pnpm check-types:main` clean.

damianvtran:
### QA remediation — round 1

Your PASS on `f4ba1e5d0` stands and is not re-opened. The three findings are dispositioned here on the new head `17f27f4ac`; the delta QA round is running scoped to `f4ba1e5d0..17f27f4ac`.

| finding | disposition |
|---|---|
| Q1 non-daemon 2xx counts as pairing evidence (pre-existing) | **not addressed** — recorded, reason below |
| Q2 owned-child replacement does not re-pair | **documented, behaviour deliberately unchanged** — `441104c5a` |
| Q3 identity-less `/health` 200 after a declined discovery leaves `attached` (pre-existing) | **not addressed** — recorded, reason below |

**Q2 — documented in `441104c5a`, not changed.** This is the same sub-case as the reviewer's R1-4, and your row's evidence matches the code: the recovery exception sits above the live owned-child guard, so an owned child detaches honestly but never claims. Extending the exception would let `discoverAndAttach()` return `false` and fall through to `start({quiet: true})`, spawning a second daemon over the live child it just contradicted — the guard exists for exactly that. The reported case is an adopted daemon, which your rig shows re-pairing (`attached/instance-0.59.0` → `attached/instance-0.59.4`, one `POST /v1/desktop/claim`), so the comment now states the boundary rather than hiding it. Recorded as not addressed as *behaviour*.

**Q1 and Q3 — not addressed, and stated as pre-existing.** Both are in the same class as the defect this PR fixes (an answer that is not a pairing being counted as one), and your counterfactual pass proves they behave identically on the base tree, so they are not regressions this branch introduces:

- Q1 needs a decision this change cannot make honestly: what a 2xx on a gated path must *contain* before it counts as a daemon's admission (a body that parses as the daemon's envelope, its `status` field, an identity header). Widening the predicate to guess would put a heuristic in the one place the PR exists to keep strict, so it wants its own design and its own PR.
- Q3 is the deprecated pre-record shape (`expectedInstanceId()` null) reaching the `checkHealth` arm; nothing in this change touches that path, and fixing it means deciding what an identity-less 200 may assert about an attachment, which is the same decision as Q1's.

Both are recorded here rather than filed as issues, per the operator's standing rule for this repo.

Acknowledgements worth recording: your counterfactual pass (the same rig against base `src/` reproducing the operator's fault — count frozen at "probe 1 of 3", never detached, successor never claimed, against this head's re-pair) is the evidence the PR itself could not produce, and the preservation rows (slow/busy daemon, env-governed adoption, 409 latch, foreign occupant) are what let this merge without re-litigating the rule the change had to keep.

damianvtran:
### Agent review — round 2

Reviewer: Local Operator reviewer subagent, model `deepseek/deepseek-flash` (the model this session is running; no provider fallback occurred during this round)
Scope: f4ba1e5d0bc06dcaad475fa6676ee78964a8c727..17f27f4ac45636444fa7386c8f9ce34c1de57694

Delta round: I read the remediation diff first, then verified every claim it makes against a primary source rather than the commit message — the daemon's own routing in `~/local-operator` (read-only; nothing was executed against any daemon, so none of the operator's live planes was touched), a probe that bundles the shipped TS and re-parses the media shapes, the repository's own desktop runner, and CI. All four remediations are honest, complete and correct as to the defects they answer. Three findings, all of them prose precision: one MINOR, two NIT. None of them is a reason for another round.

| round-1 finding | disposition claimed | what I found on `17f27f4ac` |
|---|---|---|
| R1-1 stamps (BLOCKER) | fixed — `17f27f4ac` | **verified** — independently derived, ordering per `AGENTS.md` correct, the case that was red is green, CI green |
| R1-2 `/v1/auth` prose (MINOR) | fixed — `441104c5a` | **verified against the daemon's routing**, with one residual overclaim — R2-1 |
| R1-3 media call site (MINOR) | fixed — `441104c5a` | **verified**, including that it is behaviour-preserving rather than a behaviour change |
| R1-4 owned-child scope (MINOR) | documented, behaviour deliberately unchanged | **verified** — the comment now matches the code at file:line, and the region is byte-identical to round 1's head |
| QA Q1 / Q3 (MINOR, pre-existing) | not addressed — recorded | **honestly disposed** — reasons in the QA remediation comment, re-listed under "Not addressed" in the agent remediation; not buried |

---

**R2-1 — MINOR — `src/main/desktop-transport.ts:254-255`: the new prose claims an invariant the predicate does not have, and this PR's own recorded QA Q1 is the counterexample**

- The sentence reads "…and no answer that the plane did not admit can ever be counted (no false POSITIVE)". The predicate cannot promise that: `desktopAnswerProvesPairing` reads only the path prefix and the status (`desktop-transport.ts:299-303` — `startsWith("/v1/desktop/")` plus a 2xx), so *any* process answering 2xx on a `/v1/desktop/` path is counted, whether or not a daemon's plane admitted anything.
- That is not hypothetical here — it is QA round 1's Q1 on this PR: a non-daemon answering 2xx on a `/v1/desktop/` path clears the identity-failure count (`noteTransportAnswer` → `daemon-status.ts:434-447`), dispositioned "not addressed — recorded". So the new comment asserts an absolute that a finding recorded three comments above it falsifies.
- This is the same class as the R1-2 overclaim the remediation set out to remove (prose asserting what the code does not do), and it points the wrong way for a future reader: it invites the conclusion that a foreign occupant cannot produce a false positive, which is exactly the residual Q1 exists for.
- What would fix it: state what the code implements — "no answer the plane REFUSED (`401`/`403`/`503`) can ever be counted" — and, if the residual is worth a clause, name it as the one shape a prefix-and-status predicate cannot exclude (QA Q1, already recorded on this PR). A one-word-to-one-clause edit; not a blocking one.

**R2-2 — NIT — `src/main/desktop-transport.ts:291-293`: the loose `response` type is now broader than any caller, and it was the looseness that let the removed misuse type-check**

- The docstring now reads "`response` is read for its status alone rather than typed as `DesktopResponse`: this asks what the daemon DECIDED, not which body shape carried the decision, so any answer object with a status answers it."
- That is true of the predicate, but after `441104c5a` the lone production caller is the JSON transport (`backend-service.ts:588`), whose response *is* a `DesktopResponse`; the media relay no longer calls it at all. The `{ status: number }` parameter is the shape that let `desktopAnswerProvesPairing(input, mediaResponse)` type-check in the first place, so narrowing it to `DesktopResponse` would restore the compile-time guard against the exact misuse R1-3 found. Judgement call rather than a defect: the shared tests deliberately probe bare `{ status }` literals (`scripts/desktop-contract.test.mjs:2253-2279`), so this is a "when convenient", not a "should have".

**R2-3 — NIT — two residual wording imprecisions in the new comments**

- `src/main/backend/backend-service.ts:612` — "Nothing is lost by that". Strictly something *is* discarded: a 2xx from `sessions.attachment` / `subagents.attachment` is genuine admission evidence (both routes sit on `desktop_sessions.py`'s `require_desktop` router, `desktop_sessions.py:434` with the routes at `:1469` and `:1506`), and the predicate still cannot read it. What the comment's own next sentence establishes — correctly, and I verified the mechanism end to end — is that the same evidence arrives through contract ops (the 15 s `sessions.presence` beat → `/v1/desktop/presence`, `desktop_sessions.py:2551`, routed through `requestDesktop` at `backend-service.ts:529-535`), so the app loses nothing it can act on. Suggest "no evidence this app can act on is lost", which is the claim the paragraph actually supports.
- `src/main/desktop-transport.ts:245-250` — the bullet lists `/v1/agents`, `/v1/jobs`, `/v1/schedules`, `/v1/config`, `/v1/credentials`, `/v1/models` as gated by the boundary middleware only while the plane is enabled. Each is correct, but the list reads as exhaustive and is not: `_LEGACY_CONTROL_PATHS` also holds `/v1/tools/speech` and `/v1/transcriptions` (`app.py:387-396`) — the two paths the media relay's own `speech.create` and `transcription.create` target. One "among them" would close it.

---

**Dimensions checked, and clean**

- **R1-1 (stamps), all three sub-claims.** `git rev-parse HEAD:src` = `a36c5e720f81577f0edf5e6ef8381741af25eddf` and `HEAD:scripts` = `e23f656cb839786a9d9824e98a139f5deb6fdb59` on this head, and the manifest carries exactly those. `17f27f4ac` touches one file — `docs/evidence/manifest.json` — and within it only the `srcTree`/`scriptsTree` lines; `head`, `capturedAt`, `frames` and every other field are untouched. The ordering `AGENTS.md` demands holds: both trees moved in `441104c5a` (`81ef14322`/`3fac94de5` → `a36c5e720`/`e23f656cb`) and the re-stamp commit is a separate commit that moves neither, so the stamps are derived from the committed tree rather than a tree about to move again. `scripts/evidence-manifest.test.mjs` is green on this head: 35/35 run directly, and 105/105 across the four coupled files through the repository's own `scripts/run-desktop-tests.mjs`, including "the SHIPPED manifest's stamps describe the tree it ships in" — the case that was red on `f4ba1e5d0`. The tally cases that walk HEAD's tree also pass, so the re-stamp is sufficient and not merely a two-hash edit. CI run `35398969254` on `17f27f4ac` is fully green, **Desktop Tests included**, which is the job that was red on the previous head; the PR body's "420 tests across the runs, 0 fail" now quotes a set that includes `evidence-manifest`, so R1-1's evidence dimension is answered too. I did not run `pnpm check-evidence` (a full 7,664-frame pixel sweep behind the machine-wide lease); this delta touches no frame and changes no rendered surface, and the four files above cover the stamp and tally questions it guards.
- **R1-2 (prose vs the daemon).** Walked in `~/local-operator`, read-only. `/v1/auth/*` and `/v1/settings*` carry `Depends(require_desktop)` at **router** level — `routes/auth.py:25` (all 10 routes are `/v1/auth/…`), `routes/settings.py:24` (the three `/v1/settings…` paths) — and `require_desktop` raises `503` when `desktop_posture()` is not enabled (`server/desktop.py:222-223`), so "unconditionally gated, `503` on an unclaimed daemon" is right. `/v1/agents`, `/v1/jobs`, `/v1/schedules` sit in `_LEGACY_GATED_PREFIXES` (`server/app.py:413`) and `/v1/config`, `/v1/credentials`, `/v1/models` in `_LEGACY_CONTROL_PATHS` (`server/app.py:387-396`); both are enforced only under `desktop_posture().enabled` in `managed_desktop_boundary` (`server/app.py:532-536`), and those routers carry no dependency of their own (`routes/agents.py:54`, `jobs.py:20`, `schedules.py:28`, `config.py:26`, `credentials.py:21`, `models.py:56`) — so "gated only while the plane is enabled" is right, and that is what makes a `200` from one of them unreadable as an answer. `/v1/capabilities` is on a dependency-free router (`routes/capabilities.py:8-11`) and `/v1/desktop/claim`'s router likewise (`routes/desktop_claim.py:52`). The predicate is unchanged in the dangerous direction: prefix + 2xx only, so no refusal and no route outside the prefix can ever be counted. The test message's replacement rationale is now accurate (`scripts/desktop-contract.test.mjs:2275`) — the `auth.start` example is retained but correctly re-characterised as gated, which answers R1-2's point that it must not be described as ungated.
- **R1-3 (media call site).** (a) Reproduced the claim rather than taking it: bundling the shipped `desktop-contract` and feeding all seven media shapes to `desktopRequestSchema` returns `invalid_union_discriminator` for every one — `speech.create`, `speech.agent`, `transcription.create`, `agent.import`, `agent.export`, `sessions.attachment`, `subagents.attachment` — so the input is refused before a path or a status is read, and the old call could only ever have returned `false`. (b) `recordTransportAnswer()` is the right call and it is liveness-only (`daemon-status.ts:401-405`: stamps `lastTransportAt`, zeroes `unanswered`, leaves the state alone), and it is *exactly* what the old expression produced — the predicate's `false` flowed into `noteTransportAnswer(false)`, which calls `recordTransportAnswer()` and nothing else (`backend-service.ts:653-659`) — so this remediation changes no behaviour, and it keeps the `if (answered)` guard in both versions. The one piece of evidence the app cannot act on is named in R2-3; the mechanism that makes it moot is real and I followed it: `sessions.presence` is a contract op on a `require_desktop` route, over the same `requestDesktop` transport, every 15 s. (c) The docstring's `DesktopMediaResponse` rationale is gone; only its type width remains, which is R2-2. And no other call site passes a media request: `desktopAnswerProvesPairing` has two callers in `src/`, the JSON transport at `:588` and this relay, now liveness.
- **R1-4 (comment vs code).** The delta to `src/main/backend/backend-service.ts` is comment-only apart from the media line (filtered the +/- set: the sole non-comment change in the whole file is `noteTransportAnswer(desktopAnswerProvesPairing(...))` → `daemonState.recordTransportAnswer()`), so the behaviour in the recovery region is unchanged from round 1's head. The new comment's structural claims all match the code: the exception at `:2946` sits above the live owned-child guard at `:2950-2955`, which returns before `discoverAndAttach()` at `:2956`, and the fall-through it warns about is `start({ quiet: true })` at `:2964`, with `isExternalBackend` among the guards at `:2960`. So an app-owned child genuinely does not recover by this route, and the comment says so instead of implying otherwise.
- **Remediation hygiene.** The round is unified: two commits, `441104c5a` (`fix(daemon):`, the three fixes) and `17f27f4ac` (`chore(evidence):`, the re-stamp `AGENTS.md` requires — not a cascading round). Conventional commits on both. Seven changed files across the whole PR, all of them in scope; no version bump anywhere (`package.json` is `0.29.0` on base and head, and the Version Bump Guard job is green). QA Q1/Q3 are dispositioned with substantive reasons rather than a shrug, and re-listed under "Not addressed" in the agent remediation — recorded on the PR, which is the disposition the standing rules ask for.

**Verdict: clean and TERMINAL** — no BLOCKER and no MAJOR remains on `17f27f4ac`; the three findings above are prose precision (one MINOR, two NITs), follow-ups rather than a reason for a further round.

damianvtran:
### QA report — round 2

- **QA:** qa-tester (`deepseek/deepseek-flash`) — independent pass, delta round. The instrument below is this head's own shipped `BackendServiceManager` (bundling `backend-service.ts` + `desktop-media.ts` + `desktop-transport.ts`) driven against real loopback daemons I started myself, plus the repository's own suite for the manifest gate (row S). I did not author or remediate this diff.
- **Scope:** `f4ba1e5d0bc06dcaad475fa6676ee78964a8c727..17f27f4ac45636444fa7386c8f9ce34c1de57694`. Round 1's PASS on `f4ba1e5d0` is not re-litigated. **Every row was run twice** — once against the new head's tree (`post`) and once against the **previous head's `src/`** in a `/tmp` copy (`prev`) — so "this delta moved nothing here" is measured rather than asserted.
- **Environment:** macOS Darwin 25.6.0 arm64, node v26.5.0, the repo's own esbuild; `HOME` and `LOCAL_OPERATOR_CONFIG_DIR` under per-case `/tmp/qa375r2-*` roots; ephemeral loopback ports; every daemon is my own process with its own 0600 record and a JSONL request log (which is what lets a row assert what was **never** sent). Every process I started was killed by exact pid; **no operator daemon (1111/4611/6177/8080/19213/19821/19831) or session was touched**, and the worktree is byte-clean at `17f27f4ac` (`git status --short` empty — QA never edits the repo).

**The delta's shape, measured rather than read.** Stripping comments and blank lines from both trees and diffing what remains, the delta's only executable change in `src/` is **one line** in `backend-service.ts` (`requestDesktopMedia`'s answered branch); `desktop-transport.ts` and `daemon-status.ts` differ by nothing at all once comments are removed, `scripts/desktop-contract.test.mjs` differs only in a test *message* string, and `docs/evidence/manifest.json` only in its two stamps.

```
node /tmp/qa375r2/rig2.mjs --source post --rows M1,M2,M3,M4,M5,M6,N1,N2,P1,A,B,C,E   # this head
node /tmp/qa375r2/rig2.mjs --source prev --rows M1,M2,M3,M4,M5,M6,N1,N2,P1,A,B,C,E   # f4ba1e5d0's src/, /tmp copy
node /tmp/qa375r2/rig2.mjs --source post --rows D,G   (and --source prev)
env -u NO_COLOR TERM=xterm-256color node scripts/run-desktop-tests.mjs scripts/evidence-manifest.test.mjs
env -u NO_COLOR TERM=xterm-256color node scripts/run-desktop-tests.mjs scripts/evidence-manifest.test.mjs scripts/desktop-contract.test.mjs scripts/daemon-observation.test.mjs scripts/daemon-health-state.test.mjs
```

| # | Surface | Actual output (deciding line) | Verdict |
|---|---|---|---|
| M1 | **DELTA, the admitted media case** — attachment one identity-failure short of detaching, then `requestDesktopMedia({op:"sessions.attachment"})` twice, 300 ms apart | `state=degraded failures=1 lastTransportAt=null->1789769279546->1789769279857 recorders=[recordTransportAnswer,recordTransportAnswer] notifications=0 daemon200=2` — the daemon log shows both GETs on `/v1/desktop/sessions/a1b2c3d4e5f6/attachments/0123…cdef` answered **200** by my daemon; the liveness stamp advances twice, the identity-failure count does **not** reset, the state stays `degraded`, the sentence is unchanged (no pairing claim), and no status snapshot is pushed | PASS (identical on `prev`) |
| M2 | Adversarial — media answers of **401** and **503** while the attachment is stale | `401->401 503->503 state=degraded failures=1 (was 1) recorders=[recordTransportAnswer,recordTransportAnswer]` | PASS (`prev` identical) |
| M3 | Adversarial — a media request issued while the attachment is **stale** (`detached` after 3 answered contradictions) | `detached=detached/3 -> detached/3 lastTransportAt=null->1789769282615 recorders=[recordTransportAnswer]` — a 200 media answer cannot revive `detached` | PASS (`prev` identical) |
| M4 | Adversarial — the four media refusals made **on this side of the socket**: unknown op, a *contract* op handed to the media relay, invalid media shape, 17 MiB upload | `422/422/422/413 -> state=degraded failures=1 lastTransportAt=null->null recorders=0 daemonSaw=0` — no stamp, no state movement, and the daemon never sees them | PASS (`prev` identical) |
| M5 | Adversarial — an answer the relay refuses **itself** (33 MiB body against its own 32 MiB ceiling) | `oversize download -> 502/error state=degraded failures=1 lastTransportAt=null->1789769285555 recorders=[recordTransportAnswer]` — still an answer, still no pairing | PASS (`prev` identical) |
| M6 | Preservation of the PR body's own claim, driven through the **media** relay: `/health` takes 3.5 s against the 2 s probe budget while media reads keep arriving | `media 200s=10/10 states=attached max unanswered=2 final=attached` — 10/10 answered media reads, never `detached`, never reported offline | PASS (`prev` identical) |
| N1 | **Pairing evidence is not lost** — a contract op the plane admits (`sessions.list`) clears the count and revives `degraded` | `sessions.list=200 degraded/1 -> attached/0 recorders=[recordTransportSuccess] pushed=attached` | PASS (`prev` identical) |
| N2 | The specific evidence the new comment names: the 15 s **presence beat** (`sessions.presence` → `/v1/desktop/presence`) | `presence=200 degraded/1 -> attached/0 recorders=[recordTransportSuccess]` | PASS (`prev` identical) |
| P1 | Predicate half of the delta's claim — `desktopAnswerProvesPairing` fed all seven media ops against **real HTTP 200s** | `sessions.attachment 200=false; subagents.attachment 200=false; agent.export 200=false; speech.create 200=false; speech.agent 200=false; transcription.create 200=false; agent.import 200=false; CONTROL sessions.list 200 (admitted contract op)=true; CONTROL capabilities 200 (public op)=false` | PASS (`prev` identical) |
| A | Round-1 end-to-end re-run — a successor replaced under an **adopted** app re-pairs | `before=attached/instance-0.59.0 after=attached/instance-0.59.4 claims=1` (one `POST /v1/desktop/claim` on the successor, none on the replaced process) | PASS (`prev` identical) |
| B | Round-1 end-to-end re-run — a **shut-plane** successor detaches instead of holding `attached` | `before=attached/instance-0.59.0 after=detached/instance-0.59.0 ownedChild=none`, 0 claims on the shut plane | PASS (`prev` identical) || C | Preservation — slow/busy daemon keeps its admitted reads and stays `attached` | `health 3.5s vs 2s probe budget; states=attached final=attached` | PASS (`prev` identical) |
| E | Preservation — the **409 latch** (plane already claimed) and a foreign-key plane | `latched: adopted=true state=attached claims=1 | foreign: adopted=false state=attached claims=1` (declines, no claim loop) | PASS (`prev` identical) |
| D | Preservation — **env-governed adoption** (`claim_key: ""`, spawn token), then a replaced successor | `adopted=true attached=attached/instance-env after=detached/instance-env claims=0` (not falsely paired, no claim, no spawn) | PASS (`prev` identical) |
| G | The **owned-child** scope R1-4/Q2 documented as deliberately unchanged | `g1: started=true state=attached owned=true pushes=detached>attached | g2: claims=0 after=detached/owned-child ownedPid=382->382 stillAlive=true` — detaches honestly, the live owned child is untouched and no second daemon is spawned over it | PASS (`prev` identical) |
| S | **The manifest gate this delta's stamp half is** (CI's Desktop Tests reads it) | `node scripts/run-desktop-tests.mjs scripts/evidence-manifest.test.mjs` → `ℹ tests 35 / pass 35 / fail 0`, exit 0, including `✔ the SHIPPED manifest's stamps describe the tree it ships in`. Hand-checked too: manifest `srcTree a36c5e720f81577f0edf5e6ef8381741af25eddf` == `git rev-parse HEAD:src`, `scriptsTree e23f656cb839786a9d9824e98a139f5deb6fdb59` == `HEAD:scripts`. R1-1's BLOCKER is fixed on this head | PASS |
| S2 | The four shipped suites coupled to the changed trees | `tests 105 / pass 105 / fail 0`, exit 0 (`evidence-manifest` 35 + `desktop-contract` + `daemon-observation` + `daemon-health-state`), reproducing the figure the remediation comment quotes | PASS |

**Findings — none on this delta, per dimension checked.**

- **Media relay / the one changed line:** no findings. An answered media request records **liveness only** in every case I could drive — admitted 200 (M1), 401 and 503 (M2), a 502 the relay manufactures itself (M5) — and it never resets the identity-failure count, never revives `detached`, never rewrites the sentence and never pushes a snapshot (M1/M3). The three refusals made before any byte leaves are not answers at all and leave no stamp, and the media relay refuses a contract op outright (M4), so the removed predicate call could not be reached with an input it would have admitted: P1 confirms all seven media ops answer `false` even against a real 200 from a route *inside* the admitted prefix.
- **Pairing evidence not lost:** no findings. Both paths the media comment names as the evidence suppliers still work — `sessions.list` (N1) and the 15 s presence beat, which is itself a `/v1/desktop/` contract op (N2) — each clearing the identity-failure count and reviving `degraded` → `attached` with one pushed snapshot.
- **The round-1 end-to-end cases:** no findings; A and B reproduce their round-1 outcomes on this head and on `prev`.
- **Preservation rows:** no findings; C, E, D and G are all unchanged (`prev` identical), including the owned-child scope R1-4/Q2 recorded as deliberately unchanged.
- **Manifest / CI gate:** no findings; S is green on this head and the stamps equal the trees they describe.
- **Behaviour-neutrality of the delta:** no findings. The two trees produce **identical** recorder calls, states, failure counts and daemon logs on all 15 rows; the delta is a clarity change at a call site that could only ever have recorded liveness.

**Recorded, not raised as findings (no action needed for this merge):**
1. The admitted-media false negative R1-3 described **remains**, and is now deliberate: a 200 from `sessions.attachment` is a real admission that is discarded (M1 shows `degraded` staying `degraded`) — the author's comment says so and names its own revisit condition. My N2 makes the "nothing is lost by that" claim checkable rather than assumed: the presence beat that supplies the same evidence is a contract op under the admitted prefix and still revives `degraded`. I raise no finding because the behaviour is **identical on `prev`** — the delta did not introduce it.
2. The media call site now calls `this.daemonState.recordTransportAnswer()` directly rather than through `noteTransportAnswer(false)`. Equivalent today (that branch is exactly this call, and `recordTransportAnswer` never notifies); the only thing to know is that a future behaviour added to that branch would not reach the media path. A nit rather than a finding.
3. The old media call site could only ever pass `false`, which is why M1–M6 are identical on both trees. If a reader expects a state change here, the honest reading of this delta is "removed a call that could not decide anything", with the discarded-admission question explicitly left to R1-3's own follow-up design.

**Not covered, and why.**

- The **installed** app was not driven, and neither was a real `lop serve` process: this diff is `src/main/` + a test message + two manifest stamps, and the daemons here are my own node processes implementing the contract the app's reader rules are written against (`/health` identity, gated `/v1/desktop/*` `401`/`503`, the unadmitted capability op, `POST /v1/desktop/claim` `200`/`409`/`401`, the presence route). No renderer, component, style or copy file is touched, so there is no user-visible surface to screenshot and no design/UX round attaches.
- No new test was added to the repo by me and none was needed: the manifest gate (S) and the four coupled suites (S2) are the repository's own, and the delta's rows are driven from a `/tmp` rig rather than committed, per QA's scope.
- The `prev` tree is a `/tmp` copy of `f4ba1e5d0`'s `src/`; it is byte-identical to that commit's tree (`git archive`), with the primary checkout's `node_modules` linked read-only.
- **One change to round 1's own row, disclosed.** Row B (the shut-plane successor) now waits for the adoption to settle at `attached` before the swap and asserts that before-state, because the claim handshake is asynchronous and one run of it caught the adoption mid-flight at `connecting/null` — a before-state that would have made the row's claim (“detaches rather than *holding* `attached`”) vacuous. `prev` and `post` both now report `before=attached/instance-0.59.0`; round 1's published line for this row reported only the settled state.

**Verdict: PASS on `17f27f4ac45636444fa7386c8f9ce34c1de57694`** — **no FAIL on this head**, no blocker or major findings, and no minor findings on the delta. 15 rig rows on this head and the same 15 on the previous head, plus the repository's manifest gate (35/35) and its four coupled suites (105/105), all green; the delta's single executable line is behaviour-preserving and the property it was changed for holds under every adversarial case I could drive. Round 1's Q1–Q3 remain as recorded in the QA remediation comment (Q2 documented, unchanged; Q1/Q3 pre-existing and out of this delta's scope).

damianvtran:
### Merge disposition — rounds complete on `17f27f4ac`

Gate check, read off the PR rather than asserted:

| round | reviewer | scope | verdict |
|---|---|---|---|
| Agent review round 1 | reviewer subagent (deepseek/deepseek-flash) | `896b19134..f4ba1e5d0` | 1 BLOCKER + 3 MINOR |
| Remediation round 1 | implementer | `f4ba1e5d0..17f27f4ac` | per-finding, SHA'd (comment above) |
| QA report round 1 | qa-tester subagent | `f4ba1e5d0` | PASS, 3 MINOR (2 pre-existing) |
| QA remediation round 1 | implementer | — | dispositions recorded (comment above) |
| Agent review round 2 | reviewer subagent (deepseek/deepseek-flash, no fallback) | `f4ba1e5d0..17f27f4ac` | terminal: no BLOCKER, no MAJOR |
| QA report round 2 | qa-tester subagent | `f4ba1e5d0..17f27f4ac` | PASS, no findings |

Round 2's scope equals the current head, so both rounds are fresh; CI run 35398969254 is green on it, including the Desktop Tests job that was red on the first head. No version bump, no human reviewer, nobody tagged.

**Deferred — round 2's three prose/type findings, merged without folding them in.** They are comment and type precision on files this change already touched; folding them in would have invalidated a terminal review round and green CI for wording, and the round-2 verdict explicitly says they are not a reason for another round. Recorded here so they are not lost:

- `deferred — R2-1 (MINOR)`: `src/main/desktop-transport.ts:254-255` says the predicate's error direction is "no false POSITIVE", which it cannot promise — QA Q1's residual (a foreign process answering 2xx on a `/v1/desktop/` path) is the counterexample, already recorded above as pre-existing and not addressed. The fix is a qualifier naming that residual, and it belongs with Q1's own design rather than as a wording patch here.
- `deferred — R2-2 (NIT)`: the predicate's `{ status: number }` parameter is now wider than its only caller needs; tightening it back to `DesktopResponse` is worth doing when the file is next touched.
- `deferred — R2-3 (NIT)`: "Nothing is lost by that" at `backend-service.ts:612` is loose, and the gated-family list at `desktop-transport.ts:245-250` omits `/v1/tools/speech` and `/v1/transcriptions`.

**Merging as admin, disclosed.** This repository's `main` requires one approving review and the owner's own PRs carry none (agent review rounds are PR comments, not GitHub reviews); `--admin` is the sanctioned completion for that case, so the merge bypasses the approval requirement deliberately and visibly rather than by accident. The merge commit lands this head unchanged — no additional commits after the rounds — so what merges is exactly what was reviewed.

damianvtran:
### Agent review — convergence after rebase

Reviewer: Local Operator reviewer subagent, model `deepseek/deepseek-flash` (the model this session is running; no provider fallback occurred during this round)
Scope: 17f27f4ac45636444fa7386c8f9ce34c1de57694..0b182527fd63088b0c3ada3beabf8938220305be

This is a convergence round on a rebase, not a re-review of the change. The change itself was approved terminal at round 2 on `17f27f4ac`; per the instruction for this round I did not re-audit it and raise no new finding about it. Everything below is the rebase delta. **No findings at any severity.**

---

**1. The rebase replayed content, not just hashes.**

- `git range-diff 896b19134..17f27f4ac 853c054c6..HEAD` → `1: f4ba1e5d0 = 1: 2ef0b9763`, `2: 441104c5a = 2: a8947b46e`, and the third (stamp-only) commit shown as dropped and re-added, i.e. re-derived rather than replayed. `853c054c6` is an ancestor of HEAD and exactly 3 commits lie between them.
- Byte comparison of the two surviving commits against their pre-rebase twins, over the added/removed line sets with the `+++`/`---` headers stripped: identical both times.

  | commit pair | lines | md5 (old == new) |
  |---|---|---|
  | `f4ba1e5d0` → `2ef0b9763` | 568 | `2f2727dc1645bbfe4d199ad318b50b05` |
  | `441104c5a` → `a8947b46e` | 72 | `532c5e290e7398759d37ae69d8d89136` |

- The new third commit is stamp-only: `git diff --stat HEAD^ HEAD` → `docs/evidence/manifest.json | 4 ++--`, one file, +2/−2, and both changed lines are `srcTree`/`scriptsTree`.

**2. Semantic conflicts: one shared file, and it is the expected one.**

- Upstream's diff (`896b19134..853c054c6`, PR #362) touches `docs/evidence/manifest.json`, `scripts/capture-evidence.mjs`, four `scripts/*.test.mjs` files, four `src/**` files and adds 72 evidence frames under `docs/evidence/chat-mid-turn-join/`.
- This PR's file list (`853c054c6..HEAD`) is 7 files: `docs/evidence/manifest.json`, `scripts/{daemon-health-state,daemon-observation,desktop-contract}.test.mjs`, `src/main/backend/{backend-service,daemon-status}.ts`, `src/main/desktop-transport.ts`.
- Intersection computed, not assumed: **`docs/evidence/manifest.json` alone.** Upstream touches none of this PR's code files, and this PR touches none of upstream's frames.
- Stamps at the new head: the manifest records `srcTree 81c1798cc7768b585feea4aba54977e4b2da8903` and `scriptsTree 09035a267006d0c035ac1930a7d2d69e3a56be7d`, and `git rev-parse HEAD:src` / `HEAD:scripts` return exactly those. (Pre-rebase the pair was `a36c5e720…`/`e23f656cb…` for `17f27f4ac`'s trees; `853c054c6` itself carries `c8b1084a6…`/`652383a47…`.)
- **Nothing but those two values differs from upstream's copy.** `git diff 853c054c6..HEAD -- docs/evidence/manifest.json` is that two-line hunk and nothing else; normalising both files' two stamps and comparing byte-for-byte is IDENTICAL. So no upstream record was lost, and this branch's contribution to that file was — before the rebase and after it — exactly the two stamps. I checked the other direction too: the pre-rebase manifest against HEAD's, stamps normalised, differs only in the fields PR #362 itself rewrote (`frames`, `surfaces`, `head`, `headNote`, `refreshedAt`, `refreshedAtHead`, `partialCapture.*`, `midTurnJoinPass`) — all upstream content, none of it this branch's.
- The repository's own gate for this file, run on this head:

  ```
  node scripts/run-desktop-tests.mjs scripts/evidence-manifest.test.mjs
  ℹ tests 35  ℹ pass 35  ℹ fail 0  ℹ duration_ms 22989.7   exit 0
  ```
  including `✔ the SHIPPED manifest's stamps describe the tree it ships in` and the tally/head-citation cases. This is the gate R1-1's blocker was about, and it is green on the new head as it was on the old one.

**3. No conflict markers; no version bump.**

- Changed files, marker runs (`<{7}`/`={7}`/`>{7}`/`|{7}` at line start): 0 in all seven; `git grep -nE '^(<{7}|>{7})( |$)' HEAD` returns nothing across the whole tree; the repository's own `scripts/conflict-markers.test.mjs` passes (`tests 1 / pass 1 / fail 0`).
- `package.json` is `0.29.0` at HEAD, at `17f27f4ac` and at `853c054c6`; the delta's 7-file list contains no version file; CI's `Version Bump Guard` is green on the head. No bump rode in with the rebase.

**4. The reported flake: not reproduced, in four runs at the reported condition.**

- The message is the Node test runner's, not this repository's: the only copy of that string anywhere in `scripts/` or `src/` is the comment at `scripts/session-cookies.test.mjs:223-242`, which records this exact signature (a case cancelled 10 ms into its own 120 ms budget) and attributes it to the Node 22 runner draining the loop under a file that awaits only unref'd timers, with "the same file passes on Node 26". The string is a literal in the Node 22 binary CI uses. (A matching check on the local Node 26 binary came back inconclusive and I am not resting anything on it.)
- The four-suite invocation the report names, on this head, four times:

  ```
  env -u NO_COLOR TERM=xterm-256color node scripts/run-desktop-tests.mjs \
    scripts/evidence-manifest.test.mjs scripts/desktop-contract.test.mjs \
    scripts/daemon-observation.test.mjs scripts/daemon-health-state.test.mjs
  → ℹ tests 105  ℹ pass 105  ℹ fail 0  ℹ cancelled 0   exit 0
  ```
  three times on node v26.5.0 at load 75–160, once on node v22.22.0 at load 167 (the CI node), and `desktop-contract.test.mjs` alone 33/33. **I did not reproduce it**, so the pre-rebase counterfactual does not apply.
- CI agrees on both heads: `Desktop Tests` is green at `17f27f4ac` and again at `0b182527f`, where the job log reads `# tests 2995 / # pass 2958 / # fail 0 / # cancelled 0 / # skipped 37` and `desktop suite: pass=2958 fail=0` over the 169-file invocation. Note that the CI invocation includes `scripts/conflict-markers.test.mjs` and `scripts/evidence-manifest.test.mjs` themselves.
- What this does and does not say: four green four-suite runs, the single-suite runs around them, and two green CI runs, with the flake not firing in any of them, is evidence it is rare, not evidence it is gone. It did not fire on either head, so this round need not attribute it — but a one-off cancellation inside a 169-file node-22 invocation is worth keeping in view rather than closing.

---

**Verified on this head** (`0b182527f`, `git status --short` empty before and after my runs): the range-diff; the byte-identical line sets of both surviving commits; the single-file intersection with upstream; the stamps against `git rev-parse`; the manifest's equality with upstream's copy outside the two stamps; no conflict markers (grep plus the repo's own guard); no version bump; the evidence-manifest suite and the four-suite/one-suite runs quoted above; CI's check-runs and the Desktop Tests job log on this head; and `gh pr view` — non-draft, `MERGEABLE`/`CLEAN`, `reviewRequests: []` with zero `review_requested` events, so this round added no human reviewer and pinged nobody.

**Not verified, and why:** I did not re-audit the change itself (out of scope for this round by instruction). I did not run the full 169-file desktop suite locally — CI's 2958-pass node-22 run is what carries that, and standing a second dependency tree up on a box at load 75–167 with sibling sessions live would have cost the machine more than it bought. I did not attempt the pre-rebase flake counterfactual, because the flake did not reproduce on the head in the first place. Nothing in this round dialled a port or started a process: the operator's live sessions and the :1111 backend were untouched, and the branch was not modified by me.

**Verdict:** no BLOCKER, no MAJOR, no MINOR and no NIT — the rebase is content-preserving, so the round is TERMINAL and the merge gate remains satisfied on `0b182527f`, carrying the merge disposition recorded on `17f27f4ac` to this head unchanged.

damianvtran:
### Agent review — convergence after rebase (2)

Reviewer: Local Operator reviewer subagent, model `deepseek/deepseek-flash` (the model this session is running; no provider fallback occurred during this round)
Scope: 0b182527fd63088b0c3ada3beabf8938220305be..898cacea7aee020ac26f68a5a877f92b00e28941

Second convergence round, same shape, new base (`76245f7e4`, PR #369). The change itself stays approved and unre-audited; only the rebase delta below. **No findings at any severity.**

**1. Patches replayed, third commit re-derived.**

- `git range-diff 853c054c6..0b182527f 76245f7e4..HEAD` → `1: 2ef0b9763 = 1: 98d29d6fc`, `2: a8947b46e = 2: fb883f3c0`, and the stamp commit dropped/re-added rather than replayed. `76245f7e4` is an ancestor of HEAD, exactly 3 commits since it.
- Added/removed line sets of the two survivors, headers stripped, against their pre-rebase twins: byte-identical both times — commit 1 **568 lines, md5 `2f2727dc1645bbfe4d199ad318b50b05`**; commit 2 **72 lines, md5 `532c5e290e7398759d37ae69d8d89136`**. Those are the same two digests the first convergence round recorded, so the patches are unchanged as well as equal to each other.
- The new third commit is stamp-only: `4 ++--`, one file, both lines `srcTree`/`scriptsTree`.

**2. The only shared file is still the manifest.** Upstream's own delta this time is `853c054c6..76245f7e4` (PR #369): `docs/evidence/manifest.json`, `docs/run-sidebar.md`, `docs/evidence/chat-run-panel/README.md`, `scripts/{capture-evidence,composer-tabs.test,run-detail-model.test}.mjs`, six `src/renderer/.../run-details/*` files, plus 60 new run-panel frames. Intersection with this PR's seven files, computed rather than assumed: **`docs/evidence/manifest.json` alone.**

Worth recording, because the wrong range reads alarmingly: `git diff 0b182527f..76245f7e4` looks like this PR being *reverted* in `backend-service.ts`, `daemon-status.ts`, `desktop-transport.ts` and its three test files (~575 deletions against them). That is an artefact of the range — it spans this branch's own commits on its way to the new base, so the branch's additions show as removals. Against upstream's real range the intersection is the manifest and nothing else.

**3. Stamps at the new head.** The manifest records `srcTree a923c86951e18c04038977b9a42cf6f6745513f4` and `scriptsTree 16497fa51d748776406b6abd5264268f8ddeeadd`, and `git rev-parse HEAD:src` / `HEAD:scripts` return exactly those. `git diff 76245f7e4..HEAD -- docs/evidence/manifest.json` is that two-line hunk and nothing else; normalising both files' two stamps and comparing byte-for-byte is IDENTICAL — so PR #369's manifest (its tabs-and-escapes rewrite, `frames 7808`, its own fold records) survives under this branch's stamps with no record dropped and nothing re-layered.

**4. The head-to-head delta is exactly the base move.** `comm -3` of the file sets of `0b182527f..898cacea7` and of upstream's `853c054c6..76245f7e4` is empty, and `git diff 0b182527f..898cacea7` over this PR's six code/test files is empty — they are byte-identical at both heads. The manifest alone differs, and only because its base file was rewritten upstream and its two stamps re-derived (hence the 7217/7250-line form of that one diff).

**5. No markers, no bump.** Zero conflict-marker runs in all seven changed files and none tree-wide; the repo's own `scripts/conflict-markers.test.mjs` passes (`tests 1 / pass 1 / fail 0`); no `.orig`/`.rej`/backup files in the tree. `package.json` is `0.29.0` at both heads and the delta lists no version file.

**6. Suites.** On `898cacea7`: `node scripts/run-desktop-tests.mjs scripts/evidence-manifest.test.mjs` → `35 tests / 35 pass / 0 fail`, exit 0, including `✔ the SHIPPED manifest's stamps describe the tree it ships in` and the head-citation case, so the tally is honest over upstream's new 7808. The four coupled suites → `105 tests / 105 pass / 0 fail / 0 cancelled`, exit 0. CI on the head: `Desktop Tests` green at node 22 (`desktop suite: pass=2961 fail=0`, zero `not ok`), with Lint, Type Checking, Change Scope and Version Bump Guard green. The one check not finished is `NPX Sanity Check (macos-latest)`, still queued on a runner — infrastructure, not a failure: its ubuntu twin is green here and both were green on the previous head. That queue is what leaves GitHub's `mergeStateStatus` at `UNSTABLE`; `mergeable` is `MERGEABLE`.

**Limits, unchanged:** I did not re-audit the change; I did not run the full 169-file desktop suite locally (CI's node-22 run carries that); nothing dialled a port or started a process, so the operator's live sessions and the :1111 backend were untouched; the branch was not modified by me; `git status --short` is empty at `898cacea7`. I checked upstream's file *set* for the semantic-conflict question rather than each upstream hunk, because with an empty intersection outside the manifest there is no other file in which a semantic conflict could live.

**Verdict:** no BLOCKER, no MAJOR, no MINOR, no NIT — the second rebase is content-preserving on the same terms as the first, the round is TERMINAL, and the merge gate remains satisfied on `898cacea7`; the merge disposition recorded on `0b182527f` carries to this head unchanged.

damianvtran:

damianvtran:
**Shipped.** v0.29.1 is published and promoted (not a pre-release): the pipeline ran validate → npm → macOS/Windows/Linux installers → artifacts → promote, all green, run 35403140603. The app will pick this build up on its normal update path; the fix's behaviour (self-re-pairing instead of "not paired" until restart) is in 0.29.1 and later.

PR #376, opened 2026-09-18
chore(release): bump version to 0.29.1

damianvtran:
Release window since v0.29.0, one bump by materiality: **patch**.

`patch`, not `minor`: the window's three PRs are two fixes and one feature surface, and the only PR carrying an explicit claim asks for patch (`#375`). Per AGENTS.md the direction of the error prefers under-calling, and nothing here is a step-function capability namable in a release title.

| PR | subject | Release line |
|---|---|---|
| #362 | `fix(chat): keep the answer across a receipt gap, and paint a joined turn honestly` | none carried |
| #369 | `feat(run-panel): list a child's own subagents on its page` | none carried |
| #375 | `fix(daemon): an answered refusal can no longer stand in for a pairing` | `patch — the desktop app now re-pairs itself within ~30s when its Local Operator server is replaced under it, instead of staying "not paired" until the app is restarted.` |

Proven before landing the bump, not assumed: `git diff v0.29.0..origin/main -- package.json` is empty, so no merged PR consumed this number, and `v0.29.1` does not exist as a tag or a Release.

This PR is `package.json` only, one line, and carries no other change.


damianvtran:
### Agent review — release bump

Reviewer: `reviewer` subagent (lopdev), model **deepseek/deepseek-flash** — no provider fallback occurred in this round.
Scope: `origin/main...chore/release-0.29.1` @ `5126f0eba9d3cce82d94f478a154e3162cfa9239` — 1 file, 1 line pair.

**1. The diff is exactly one line in exactly one file — read from the ref, not the body.**
`gh pr diff 376` and `git diff origin/main origin/chore/release-0.29.1` both print the same single hunk: `-  "version": "0.29.0",` / `+  "version": "0.29.1",`, `1 file changed, 1 insertion(+), 1 deletion(-)`. `gh pr view 376 --json files` agrees (`package.json`, 1/1). No other file, no other version string: `git grep '0\.29\.0' origin/main` over tracked files (lockfile and `docs/evidence` excluded) matches **only** `package.json:3`, so no second version surface had to move in lockstep and no evidence-manifest regeneration is implied.

**2. The version is not already spent — confirmed in both directions.**
- `git ls-remote --tags origin 'refs/tags/v0.29.1*'` → empty; `gh release list --limit 6` → newest is `v0.29.0` (2026-09-18T19:55:13Z). `gh pr view 376` reports `MERGEABLE`, non-draft.
- No merged PR consumed the number: `git diff v0.29.0..origin/main -- package.json` → **empty**, and `origin/main:package.json` still reads `0.29.0`.
- The mirror-image case (shipping the previous tree under a new number) is answered by the guard's own output, quoted in §4 — it read the version *pair*, not a file diff, and confirmed the new number sorts above the highest tag.

**3. Window and bump class.**
Window = **exactly #362, #369, #375**, confirmed two independent ways: `git log --first-parent --oneline v0.29.0..origin/main` (three merge commits) and `gh pr list --state merged` filtered to `mergedAt > v0.29.0` (same three, nothing else). All three appear in this PR's table; the table's "none carried" for #362/#369 is accurate (grep of the raw bodies: zero `Release:` lines each), and #375's line is quoted verbatim.

**`patch` is right, by the rule rather than by preference.** AGENTS.md's materiality section makes `minor` a claim about a *single* PR being "a step-function capability in its own right — something a user would call a new thing the app can do, not more of what it already did", with the error direction preferring an under-call. #369 adds **one section to a page that already existed**: rows rendered through the pane's existing `openChild`, reusing the roster's own `SubagentRowView`/`SubagentRowBody`, reachable at any depth. That is a deeper view of the surface v0.29.0 already shipped, not a capability namable in a release title — an ordinary feature riding a fixes window, which the rule explicitly allows to stay `patch`. Its user-visible payoff also depends on a *separate* backend PR (named in its own body) that has not landed in this repository's window, which argues further against calling it this window's step-function. #362 and #375 are both fixes, and #375 carries the window's only explicit claim (`Release: patch`). Calling this window `minor` would permanently over-report how much changed; `patch` is the answer the rule gives.

**4. CI on this head (`5126f0eba`) — both required checks pass, and both ran substantively.**
- `Version Bump Guard` — **pass** (16s, run `35402944548`, job `105786580127`). Its log is the anti-mirror-image evidence, not an early exit: `Release PR, version change allowed:`, then `-  "version": "0.29.0",` / `+  "version": "0.29.1",`, then `Version 0.29.1 sorts above the highest release tag v0.29.0.`
- `Change Scope` — **pass** (16s, run `35402944629`, job `105786580977`). Its classifier independently reached the same shape finding: `release_bump = true — the diff is nothing but the version line pair in package.json`, with `lint`/`types`/`unit`/`audit`/`pack` all false and therefore skipped. Those skips are that classifier's documented consequence of this diff shape, not a stale or truncated run — worth knowing, since the run page reads "skipping" on seven jobs while only two had anything to read.

**5. Release-notes prerequisites.**
- **Compare link.** `compare/v0.29.0...v0.29.1` returns **404 today**, which is the correct and expected state while the tag does not exist; it cannot resolve before the procedure's step 3 creates it. The form and direction are sound by control: `compare/v0.28.5...v0.29.0` → 200 and `compare/v0.29.0...main` → 200. It resolves once `gh release create v0.29.1 --target "$MERGE_SHA" --prerelease` lands the tag on the merge commit.
- **`Release:` lines.** #375 carries one and is quoted verbatim here; #362 and #369 carry none, and the body says so rather than inventing a sentence — exactly what AGENTS.md requires. Template rule 2 still requires **all three** under `## PRs`, so none may be dropped for lacking a sentence.
- **Owner note (not a finding):** read the SHA from `gh pr view 376 --json mergeCommit --jq .mergeCommit.oid` *after* the merge and tag with `--target` (a bare `git tag` + push publishes nothing, and a tag that already exists gets the notes attached to whatever SHA it names). For `## What's New`, the window's user-facing change is the self-re-pairing (#375), with the joined-turn/stream-integrity fix (#362) beside it and #369 as run-panel depth.

### Verdict

`clean` — no blocker, no major, and no minor requiring action on this diff. The round is **terminal**: nothing to remediate, and the bump is safe to merge, then tag and publish as `v0.29.1`.

damianvtran:
Merged as admin, disclosed: `main` requires one approving review and this owner's PR carries agent review rounds as comments rather than GitHub approvals. The round is on this head (5126f0eba) and CI is green, including a substantive `version-bump-guard` run. Bump class `patch`, argued by materiality in the body.

damianvtran:

PR #379, opened 2026-09-18
fix(chat): the selected row draws its accent bar once

damianvtran:
## What this changes

The sidebar's selected row now draws its `accent` bar **once**. The entity row — a team or an agent a staged draft is targeted at — was drawing it twice, which is the operator's report: "when selecting a team to chat to it shows two highlight bars instead of one (there should only be the far left one), presumably the same issue would be there when selecting an agent".

**Why it happened, and why only on those rows.** `rowCurrent` is the whole selected-row role: the ground, the weight, the ink, the `hover:` restatement **and the 2px `accent` bar**. Every row in this panel is marked across TWO elements — the wrapper paints the ground over the row's own box (the gaps and the rounded corners the 24px controls do not cover) and the button inside it paints the ground again, because `rowStyle`'s own `hover:bg-row-hover` would otherwise replace the mark under the pointer — so both of them took the bar as well. On a conversation row the wrapper and the button share a left edge, so the two bars landed on the same 2px and read as one mark. The entity row's name button sits 24px of disclosure control plus the row's own 4px gap further in, so the same idiom drew two bars 28px apart.

**The fix is the rule, not the call site.** The bar belongs to the row's LEADING EDGE, so the element that owns that edge draws it: the wrapper keeps the whole role and the button takes its ground half, `rowCurrentGround` — the same ground and the same merged hover, no second bar. That is applied to *both* rows written as a wrapper plus a button rather than only to the reported one: the conversation row is a no-op today (its two bars are coextensive) and a live defect on the first change that puts anything in front of its button, which the wrapper's own comment says is exactly where the hover-revealed pin goes. `palette-contract.ts`'s `rowSelected` doc — where the bar's licence already lives — now states the one-bar-per-row rule, so the next surface does not rediscover it.

**Guards re-pointed; no floor, colour term or claim moved.** Two guard *strings* named these call sites and had to move with them: `contrast-contract.mjs`'s pin on the entity row's ground expression, and `chat-sidebar-selection.test.mjs`'s expression table (which resolves what each element applies, so a renamed symbol is a `ReferenceError` inside it rather than a silent pass). The role's colour terms are untouched, so the palette half of that pin needs no re-measurement.

## Evidence

**Frames are committed with this change** — twelve themes each, with their before halves in a declared set beside them:

- `docs/evidence/chat-sidebar-current-row/team-row-current/` and `.../agent-row-current/` — the two rows a staged draft marks, which no frame in this repository had ever shown.
- `docs/evidence/chat-sidebar-current-row-baseline/team-row-current/`, `.../agent-row-current/` — the same story, fixtures and viewports over `origin/main`'s components.

**Before/after, read per column rather than by eye** (the tool and its numbers are in the set's README): on the selected row, **two** 2px × 32px `accent` marks before — `x=12-13`, the row's own leading edge, and `x=40-41`, the name button's box — and **one** after, `x=12-13`, on the same 32px band. Both lists. `contact-sheet/entity-row-bars.png` is the pair side by side at `localOperatorDark` and `localOperatorLight`.

**Every other surface that draws the bar was checked, and none of them moved** (they are single-element rows, so the change could not reach them): the settings rail's current section, the slash popup's active row, the app rail's active destination and the agent-hub categories sidebar's selected category all carry one bar in the frames already committed at this head. Two states carry none and are correct that way because nothing is selected in them (the agents page's roster, this panel's agent list at rest). The `@` reference picker has no committed frame in this repository at all, so its half of the check is the source: the option element itself carries the bar, once.

**The conversation row's half of the fix cannot be settled by a frame pair, and the frames are not claimed to settle it.** Its two elements share a left edge, so its bar was drawn twice on the same 2px; the encoder and the text rasteriser vary more between two runs of one tree (the same frame differs from its committed copy by 12,670 pixels) than the change does (430), so the `selected-row` pair was captured and discarded and that half is argued from geometry — and from the one-bar crop pair in the set's README — instead.

## The rebase onto the moved base (18 September 2026)

The first push conflicted with `main` after `chore/release-0.29.1` landed (`mergeStateStatus: DIRTY`, and GitHub skipped the `pull_request` runs entirely because it could not compute the merge ref — which is why a force-push is what started CI). The branch is now rebased onto `e3f9fec32`; the only conflict was `docs/evidence/manifest.json`, resolved the way this file's own `citationConvention` group (1) asks for: main's document wholesale as the base, with this branch's pass record re-layered on it and every derived field re-derived from the folded tree with `check-evidence.mjs`'s own walk (`frames` 7808 → 7832 outside the 88 declared sets, of 9235 on disk, 1403 inside; STORIES rows 767 → 769; `srcTree` `8f0fe8f7c`, `scriptsTree` `7cd34a509`, `head` `3a7bc548a`).

**The rebase changed no content, and that is measured rather than asserted**: `git range-diff` replays commits 1 and 2 as `=` and differs on commit 3 only in that one file; the patch over `src/` plus the two guard scripts is byte-identical (the same 435-line patch, md5 `16e39bf74b01be59ed6444416c897c20` on both sides), and the rig's own added/removed line sets hash identically too (`1773483ad07226fc2c7247ca7322f2cd`).

## How to read the history

Three commits, in this order. `test(chat): add the team and agent entity-row current states` is the RIG — the two story states and their `STORIES` rows, which are the states the defect needs and without which no fix could be photographed. `fix(chat): the selected row draws its accent bar once` is the fix. `docs(evidence)` is the frames. The before halves were captured over the rig commit's components — `origin/main`'s — and the baseline set's README states that provenance in full, because the split is what makes them a baseline rather than a second after.

## Manifest

`srcTree` → `8f0fe8f7c`, `scriptsTree` → `7cd34a509`, both read back with `git rev-parse HEAD:src`/`HEAD:scripts` from the tree the frames ship in; `frames` 7808 → 7832 outside the declared sets (of 9235 on disk, 1403 inside them); `surfaces`/STORIES rows 767 → 769; `themes` unchanged at 12; `dirtyWorkingTree: false`. The `countsMean` prose is re-derived for this tree, and the baseline set's own entry (84 → 108 frames, 7 → 9 surfaces) is extended rather than re-declared.

## Gates run

| gate | result |
| --- | --- |
| `node scripts/contrast-contract.mjs` | exit 0 — 25,855 assertions across 59 themes, 0 consulted exceptions |
| `pnpm gen-themes --check` | exit 0 — `themes.generated.css` up to date (59 themes) |
| `pnpm check-types` | exit 0 — main and renderer, 0 errors |
| `pnpm lint` | exit 0 — 0 errors; of the 84 warnings, none is in the three files this changes (`biome check` on those three reports no diagnostics) |
| `pnpm lint:scripts` | exit 0 — 3 changed scripts lint-clean against `origin/main` |
| `node scripts/chat-sidebar-selection.test.mjs` | 15 passed, 0 failed |
| `node scripts/new-chat-row.test.mjs`, `scripts/palette-contract.test.mjs` | 6 + 10 passed, 0 failed |
| `node scripts/{canonical-chat,clear-search,desktop-feed,mark-all-read-control,new-chat-shortcut,sidebar-catalogue-gate}.test.mjs` | 110 passed, 0 failed |
| `node --test scripts/evidence-manifest.test.mjs` | 35 passed, 0 failed |
| `node scripts/capture-evidence.mjs --only=chat-sidebar-current-row --dirs=…` | 36 frames, `dirtyWorkingTree: false` after the fix commit |
| GitHub CI on `ffd50e5de` | Version Bump Guard **pass**, Change Scope **pass** (the same `scripts/ci-scope.mjs` `pnpm check-changed` runs); Lint / Type Checking / Desktop Tests in flight as this opens |

All of the above except the CI row were re-run at the rebased head `ffd50e5de`.

Not run: the full evidence sweep (the release owner's); `node scripts/check-evidence.mjs`, which **DEFERRED twice on this machine** — another sweep holds its lease, so no frames were checked by me, and the manifest test above is the same arithmetic over a synthetic tree; and the full `pnpm test:desktop` locally, which the CI job named above runs.


damianvtran:
Superseded — closing without merging; the bar it keeps is being removed by a newer instruction.

This PR made the selected row draw its accent bar **once** instead of twice (the sweep in `ab76b06f3` had applied the bar-bearing role to both the row's wrapper and its button, so both drew the same 2px mark). That diagnosis stands and is unaffected by this close.

But the surface it fixed no longer exists as this PR left it: the operator's newest instruction for the row states is that **the left accent bars come off the chat sidebar entirely and the left rounding returns**, which is being implemented in **#386** (`fix/row-states-refined`). So the correct outcome for this branch is not "one bar" but "no bar", and merging this would have to be undone by the next round.

The rule this PR added to `rowSelected`'s docblock — that a surface carrying its own rail must not draw the role's bar — is worth keeping even after the bars go, and #386's round is where it should land if it is still true of whatever replaces them.

PR #377, opened 2026-09-18
fix(canvas): closing a document tab closes it

damianvtran:
## The bug

A reader closes a document tab in the canvas. The tab goes, flickers, and the file is still open. The ✕ handler is correct — it removes the document from `files`, and `files` is what both the tab strip (`CanvasTabs`, via `canvas/index.tsx`) and the document pane render. What puts it back is the step *after* it:

1. ✕ → `chat-content.tsx`'s `handleCloseDocument` → the document leaves `files` and `openTabs` (`chat-content.tsx:800`). The tab correctly disappears here.
2. React unmounts the closed document's viewer, whose cleanup is `useEffect(() => () => closeBuffer(document.id), [document.id])` — `wysiwyg-markdown-editor.tsx`, `code-editor.tsx`, `spreadsheet-preview.tsx`.
3. `closeBuffer` (`document-buffers.ts`) hands the buffer's words to the port the editor registered → `commitCanvasDocument(conversationId, doc)`, which was an **explicit upsert**: `listed ? map : [...conversation.files, document]`.

Step 3 appended the document straight back into `files`, so the strip and the pane re-opened what the reader had just closed. The same happened on the dirty arm (`closeBuffer` → `saveBuffer` → `entry.commit`), and it happened on a *clean* close too, because the clean arm commits as well.

It arrived with #343 (`feat(canvas): keep an open document current with the file on disk`), whose I5 intent — "closing a document commits the buffer … so a close is a pause rather than a loss" — is what has to be preserved, not the re-listing that implemented it: the upsert existed because a commit that mapped over a snapshot matched nothing on the close path (R4-7's measurement).

## The fix

**`commitCanvasDocument` updates the store's copy of a document that is still open, and does nothing for one that is not** (`document-buffers.ts`). A document that is not in `files` is not open, so there is no store copy to keep current — and the update arm is kept, because the case R4-7 named is real: an unmount that is *not* a close (a tab switch) must still keep the store's copy current. It is a guard plus `updateOneFile`, which maps inside the store's own `set`, so a commit racing another write cannot drop it.

**The words a refused write could not save are kept by the buffer owner, and put back on screen when the path is opened again** (`documentsForCanvas`, same file, applied at the one place the store's open documents become the canvas's: `initialDocuments={documentsForCanvas(files)}` in `chat-content.tsx`).

[...]

PR #364, opened 2026-09-18
feat(canvas): a finder-style files list, and the last row made reachable

damianvtran:
## Summary of Changes

Three changes to the canvas's Files view, in the order the operator reported them. All in `src/renderer/src/features/chat/components/canvas/`, plus the instrument and the frames that prove them.

Rebased onto `main`'s `0ed7365d9`, through two folds while it was open:

- **`548f8dfed`** — the merge that landed **#343** (`feat/canvas-file-freshness`), which edits the same three files this branch does. The `src` commits applied cleanly (the click handler's `carried` block was preserved verbatim here, so #343's `readMtimeMs` / `availability` / `sizeBytes` threading applied without a hand resolution); the only conflict was `docs/evidence/manifest.json`, resolved by taking the base's file and re-running the capture rather than by hand-merging hashes. The frames were re-shot on the merged component, and the geometry proof re-run: identical.
- **`0ed7365d9`** — **#358** (the delivery gate), which moved both trees (`src/main/index.ts`, `src/main/window-raise.ts`, two `scripts/` test files) without touching any surface this branch photographs. No frame was re-taken (`frames`, `surfaces`, `themes` unmoved); the re-stamp commit re-derives `srcTree`/`scriptsTree`, re-points the three head citations at this lineage's own commits, and records all of it in `partialCapture.foldRestampNote`. The full suite was re-run on the folded head anyway — see below.

### 1. The last rows were unreachable — one class, at the root cause

The view's root was `flex h-full flex-col`, a sibling of the canvas's 40px chrome bar inside the section's `flex h-full flex-col`. `h-full` is 100% of the PANE, so the panel's box was 40px taller than the space it had — the sibling bar's height — and the dock's `overflow-hidden` clipped that band. The scroller inside therefore had a viewport 40px taller than what a user can see: it reached its own maximum scroll with the last rows still under the clip and its own bottom padding inside the band. `h-full` was doing two jobs (the flex base size, and the item's specified size suggestion, which caps the content-based minimum — CSS Flexbox § 4.5, which binds here because the item's `overflow` is `visible`), so flex could not shrink it back.

The fix is `flex min-h-0 flex-1 flex-col`: `flex-1` states what the element is (the rest of the column, not all of a box that also holds the bar) and `min-h-0` removes the floor. It is the spelling the variables viewer beside it already uses.

[...]

PR #387, opened 2026-09-19
fix(toast): put the close button on the toast's top-right corner

damianvtran:
> Proposed title: `fix(toast): put the close button on the toast's top-right corner`

# PR Description

## Summary of Changes

The operator reported, with a screenshot, that a toast's close (X) button "hangs over the toast's top-LEFT corner, offset inward into the body of the toast" — hit on the `Marked 7 chats as read.` success toast, the one the sidebar's bulk read receipt raises (`markAllReadReceipt` → `showSuccessToast` in `mark-all-read.ts`).

The intent was already right and is written down in the file: the old comment said *"Keep the close button on the right of the toast, which is where it was"*. The declaration under it did the opposite, and a screenshot is the only place that disagreement was visible: the toast renders, the button renders, the ink is legible, every test was green.

### Root cause (one paragraph)

Sonner places the close button with **two** custom properties and consumes both on one box — `left: var(--toast-close-button-start); right: var(--toast-close-button-end); top: 0`, 20×20, absolute — and declares them on `html[dir='ltr']` / `[data-sonner-toaster][dir='ltr']` as `start: 0` and `end: unset` (with the mirror pair under `[dir='rtl']`). `ThemedToastContainer` writes its own values as **inline style on the toast element**, so it is overriding an *inherited* value, and a CSS custom property is inherited by default — where `unset` means `inherit`. So `--toast-close-button-start: "unset"` did not clear the `0` it was written to clear; it re-stated it. Both insets then resolved to a length, which for a 20px box that also has `width` set is the over-constrained case: LTR keeps `left` and drops `right`, and the toast's RTL-flavoured `translate(35%, -35%)` then carried the circle **inward from the left edge** instead of outward from the right one. `auto` is the value that governs nothing, which leaves `right: 0px` as the single inset that does — the mirror image of sonner's own LTR placement, on the corner the button was always meant to sit on.

### Files

- `src/renderer/src/shared/components/common/themed-toast-container.tsx` — the one-token fix (`"unset"` → `"auto"`), with the constraint written down so the next author does not re-derive it from a render. No paint change: the fill, border and ink still come from the same `--gray*` variables.
- `scripts/toast-close-geometry.{html,tsx,css,vite.mjs,mjs}` — a rendered-geometry rig in the shape the composer-alert / composer-band / chat-alignment rigs already use: a vite-served page mounting the SHIPPED component, raw CDP against a private headless Chrome whose argv goes through `withMockKeychain`.
- `scripts/toast-close-placement.test.mjs` — the CI-runnable half of the guard.
- `scripts/chrome-keychain.test.mjs`, `package.json` — the new launch site named in that table, and the new test added to `test:desktop`.
- `docs/evidence/manifest.json` — the two tree stamps re-derived (see below).

### Rendered evidence

The rig mounts the shipped `ThemedToastContainer` with the props `main.tsx` gives it, fires the **shipped receipt copy** for the operator's own toast (`markAllReadReceipt({attempted: 7, cleared: 7, superseded: 0, unknown: 0})` → `Marked 7 chats as read.`), and measures the button's box against the toast's. `1380x900` CSS at devicePixelRatio 2, the app's own default window. Every case is one line of this table:

| case | toast (l,t,w,h) | button (l,t,w,h) | button.right − toast.right | button.top − toast.top | centre → top-left / top-right corner | declared `start`/`end` |
| --- | --- | --- | --- | --- | --- | --- |
| before — `localOperatorDark`, one-line | 1000,832,356,44 | 1008,826,20,20 | **−328** | −6 | **18.4** / 338 | `0` / `0px` |
| before — `localOperatorDark`, multi-line | 1000,796,356,80 | 1008,790,20,20 | **−328** | −6 | **18.4** / 338 | `0` / `0px` |
| after — `localOperatorDark`, one-line | 1000,832,356,44 | 1342,826,20,20 | **+6** | −6 | 352 / **5.7** | `auto` / `0px` |
| after — `localOperatorDark`, multi-line | 1000,796,356,80 | 1342,790,20,20 | **+6** | −6 | 352 / **5.7** | `auto` / `0px` |
| after — `localOperatorLight`, one-line / multi-line | unchanged | 1342,826 / 1342,790 | **+6** | −6 | 352 / **5.7** | `auto` / `0px` |
| after — `dracula`, one-line / multi-line | unchanged | 1342,826 / 1342,790 | **+6** | −6 | 352 / **5.7** | `auto` / `0px` |

All six cases are identical before and after *except* the button's placement, which is what a theme-independent cascade fix should look like: the toast's own box (`1000,832,356,44` one line, `1000,796,356,80` three lines, 24px from the viewport's right and bottom edges) does not move, and neither does the button's vertical straddle.

**Where the `+6` comes from, so the number can be checked rather than trusted.** Sonner's own LTR rule puts the circle's left edge on the toast's left edge (`left: 0`) and translates it 35% of its own 20px — 7px — outward, so the circle straddles the corner. The mirror of that is what this app wants, and it is what the fix produces: the button's right edge lands 7px outside the toast's **padding** box, which is 1px inside its border box (the toast draws the 1px `--color-control` border), hence **+6** measured against the toast's border box. Vertically the same rule gives −6, unchanged from before — the defect was horizontal only.

**Containment.** A fixed bottom-right toast could in principle push the straddling button off-screen, so the rig asserts the button's whole 20px box is inside the viewport in every case: before, its right edge is 352px from the viewport's right edge; after, 18px. The toast itself keeps its 24px corner offset both times.

**Frames.** The pair is committed, not merely regenerable: `docs/evidence/toast-close-corner/{before,after}/` holds twelve frames per half — `one-line/`, `multi-line/`, `one-line-detail/` and `multi-line-detail/`, at `localOperatorDark`, `localOperatorLight` and `dracula` — where each case is the settled full window (1380×900 CSS at devicePixelRatio 2) beside a 3× crop of the corner. Both halves came from this branch's rig: `after/` at this branch's head, `before/` in a detached worktree of unmodified `origin/main` at `a17a6b3ba`, which is the operator's own report — `localOperatorDark-one-line-detail` is the X sitting *inside* the toast's top-left, and its `after` twin is the X straddling the top-right corner. They are declared as two `supplementary` sets in `docs/evidence/manifest.json`, each with its own `source`, `why`, `capturedAt` and `capturedAtHead`, which leaves the sweep's own `frames`/`surfaces` counts untouched; every image was checked with the gate's own `assertFramePaints` before it was committed (24/24 pass). `node scripts/toast-close-geometry.mjs --out=<dir>` regenerates both framings for any run; its `--out` defaults to `${TMPDIR}/toast-close-geometry`.

### The regression guard, and why it is two files

- **`scripts/toast-close-geometry.mjs` is the pixel gate.** It asserts the rule rather than printing it, and it **fails on the pre-fix tree**: `18 assertion(s) failed` — three per case, naming the cascade (`2 horizontal insets are declared as lengths (--toast-close-button-start="0", --toast-close-button-end="0px") …`), the distance (`the button's right edge is -328px from the toast's right edge`) and the corner (`the button's centre is 18.4px from the toast's top-left corner and 338px from its top-right one, so it is not on the right`). On this head: `all assertions passed`.
- **`scripts/toast-close-placement.test.mjs` is the half CI can run.** The rig needs a browser and `pnpm test:desktop` runs on `ubuntu-latest`, so the *resolution* is pinned where it runs: it reads sonner's own `dist/styles.css` through the package's export map and the shipped inline theme by bundling the real component (the way `toast-lifetime.test.mjs` bundles it), resolves the two insets the way a browser would, and asserts that exactly one governs and that it is `right`, that the transform's sign agrees with that side, and — as a fixture of its own — that `unset` on this custom property inherits, which is the trap. On the pre-fix tree it fails with `--toast-close-button-start must govern nothing; it is declared as "unset"` (`'0' !== 'auto'`) and `a left-governed box must translate toward its edge; the theme declares translate(35%, -35%)`.

The test states its own bound: it models no layout. Whether the resulting box lands on the corner is the rig's claim, measured there.

### Test plan

```
# the reproduction, on the pre-fix container taken from a commit a reader can fetch
# (a17a6b3ba, the base this branch is written against; the pre-fix file is byte-identical
# at that commit and at the parent of this branch's own fix commit)
$ git checkout a17a6b3ba -- src/renderer/src/shared/components/common/themed-toast-container.tsx
$ node scripts/toast-close-geometry.mjs --out=/tmp/toast-close-geometry/before
… 18 assertion(s) failed   → exit 1

# the fix
$ node scripts/toast-close-geometry.mjs --out=/tmp/toast-close-geometry/after
… all assertions passed    → exit 0

# the CI-runnable guard: 4 pass on this head, 2 fail on the pre-fix tree
$ node --test scripts/toast-close-placement.test.mjs
ℹ tests 4   ℹ pass 4   ℹ fail 0

# gates
$ pnpm lint                → exit 0  (Checked 776 files in 1145ms. No fixes applied. Found 87 warnings — pre-existing)
$ LINT_SINCE="$(git rev-parse origin/main)" pnpm lint:scripts
                           → exit 0  (checking 7 changed file(s); 6 lint-clean, 1 with no biome handler — the .html)
$ pnpm check-types         → exit 0  (tsconfig.node.json and tsconfig.app.json, both clean)
$ pnpm check-themes        → exit 0  (themes.generated.css is up to date (59 themes); Contrast contract holds: 25855 assertions across 59 themes)
$ pnpm test:desktop        → 166 files: ℹ tests 2861, pass 2853, fail 2, cancelled 4

# the two failures, against a pristine sibling run of origin/main at 0c02556ba
$ (cd /private/tmp/main-sibling && pnpm test:desktop)   # untouched tree, same box, same minute
                           → 165 files: ℹ tests 2624, pass 2606, fail 2, cancelled 14
```

The pristine sibling run fails **the same two tests**, by name: `the fixtures describe the registry the backend actually ships` (a fixture read against the separately installed backend) and `every backend spawn carries the prefix even with the shell-env load unresolved`. Both are properties of this machine's installed backend, not of this diff, and the toast surfaces in the suite all pass on this head: `toast-lifetime.test.mjs`, `chrome-keychain.test.mjs`'s launch-site scan and `toast-close-placement.test.mjs`. The box is heavily loaded (`uptime` load average 159–237 while this ran), which is also why a few alphabetically-late files were cancelled in both runs (4 here, 14 there).

```
$ pnpm check-evidence
Evidence check DEFERRED: another sweep (or its image child) holds the machine lease; retry after it finishes.
No frames checked. Do not delete the lock file.          → exit 75
```

`lsof /tmp/local-operator-ui-check-evidence.lock` names a live `node … check-evidence.mjs` and its `magick` child, from another session. This PR re-stamps the manifest but changes no image, so the sweep's image half is unchanged from `main`'s; the stamp half was verified directly by running the gate's own validators against the shipped manifest — `stampFailures`, `countsMeanFailures`, `partialCaptureFailures`, `provenanceFailures` and `citationFailures` all report zero failures.

```
$ pnpm check-evidence   (the manifest half, run directly)
stamp: PASS   countsMean: PASS   partial: PASS   provenance: PASS   citation: PASS
```

### The fold onto `main`

GitHub reported `DIRTY` twice while this was open, and both folds are recorded. The first, onto `origin/main` = `a17a6b3ba` (PR #364, the canvas file list), replayed this branch's commits unchanged and is the base the before half is captured at. The second, onto `origin/main` = `e8d56ef3e` (the 0.29.3 release window and PR #377), is the head this PR now carries, `1b4d334d2`, force-pushed with `--force-with-lease`. That fold's two conflicts were mechanical: `package.json`'s `test:desktop` list (main's line taken, this branch's one entry re-inserted beside `toast-lifetime`'s — no version line, `0.29.3` is main's and untouched) and `docs/evidence/manifest.json` (main's copy taken wholesale with this branch's own records re-laid onto it: the two `toast-close-corner` sets, the counts paragraph, the re-stamp note, and both tree stamps re-derived from the folded index rather than carried). The **re-stamp-only commit this branch carried was dropped by that fold** — against main's manifest its whole delta was two stamp values the commit landing after it re-derives anyway, so git discarded it as empty, which is the failure mode this file's own record elsewhere describes. Nothing was re-shot for either fold: main's `#377` never touched `themed-toast-container.tsx` (`git log a17a6b3ba..origin/main -- src/renderer/src/shared/components/common/themed-toast-container.tsx` is empty), and the rig re-run at `1b4d334d2` reproduces the numbers above exactly.

### The manifest re-stamp

`srcTree` → `9cb2e1a43` and `scriptsTree` → `178836283`, read back from the tree this branch ships (the index's own trees, `git rev-parse $(git write-tree):src` / `:scripts`) and re-derived rather than carried across the fold. `srcTree` moves because the fold brought main's `src/` with it — this branch's own commits touch exactly one file under `src/`, the toast container in `f30490f08` (the fix commit's spelling in this lineage) — and `scriptsTree` moves for the rig's new default port. **No frame of the swept set moved**: no committed WebP outside the two new declared sets changed, so `frames` (8048), `surfaces` (787), `themes` (12) and `head` are `main`'s own readings, and `partialCapture.toastCloseCornerRestampNote` records what moved and what did not for a reader of the file alone. `pnpm check-evidence` exits **75 DEFERRED** (another session's sweep holds the machine-wide lease, `lsof` names its node and its `magick` child); run directly against this manifest, the gate's own validators all report zero failures — `stampFailures`, `citationFailures`, `partialCaptureFailures`, `countsMeanFailures` and `provenanceFailures` — and the 24 new frames pass `assertFramePaints` 24/24.

### Not done, deliberately

- No paint changes — the button's fill, border and ink are the same variables as before, because a paint change drags in the contrast contract for a defect that is purely positional.
- No version bump (`package.json`'s `version` line is untouched; the only edit to that file is the `test:desktop` entry).
- The toast's surface *behind* the toast is not part of the claim and is not painted in these frames.
- **An adjacent problem, noted and not touched**: running the sweep's declared-set rules over this manifest reports nine failures in sets this PR does not own (`wake-live-app` and `browser-pane-live` are sets of `.png` frames the walker counts as zero, `browser-conversation-tabs` claims 8 and holds 4 `.webp`, `chat-sidebar-current-row-baseline` claims 84 and holds 231, and `turn-stamps-before` / `chat-phantom-compose-rows/before` / `theme-legibility-before` carry no `capturedAt`). The identical nine come back from an untouched worktree of `origin/main`, so they are main's own and this PR neither introduced nor fixed them — reported here rather than folded into a change nothing asked for.

Release: patch — a toast's close button now sits on the toast's top-right corner instead of hanging over its top-left one.


damianvtran:
### Agent review — round 1

Reviewer: independent reviewer subagent (did not author this change) — model `deepseek/deepseek-flash`, no provider fallback during this round
Scope: `a17a6b3ba..ad4a3bc18`

**What I verified myself, rather than read from the PR.** Every claim below was re-run or re-derived from primary sources; commands and their actual output are in the finding text.

**Root cause, against the vendored stylesheet.** `node_modules/sonner/dist/styles.css` (sonner 2.0.3) declares the mirror pair at `:9-11` (`html[dir='ltr'], [data-sonner-toaster][dir='ltr']` → `--toast-close-button-start: 0`, `--toast-close-button-end: unset`, `--toast-close-button-transform: translate(-35%,-35%)`) and `:22-24` (the RTL mirror), and consumes them at `:223-229` on `[data-toast][data-close-button]` as `position:absolute; left:var(--toast-close-button-start); right:var(--toast-close-button-end); top:0; height/width:20px`. The inline `TOAST_THEME` is written on the toast element, so it overrides an *inherited* custom property, and `unset` on an inherited property means `inherit` — the `0` comes back, both insets are lengths, 20px box with `width` set is over-constrained, LTR keeps `left`. **The fix addresses the mechanism, not a symptom**, and the new comment states it accurately. The strongest confirmation is empirical and comes from the rig's own pre-fix output: it prints the button's computed `--toast-close-button-start` as `"0"` although the inline style said `unset` — that is Chrome resolving `unset` to the inherited value, not the author's model.

**The CI-runnable guard discriminates.** I built my own pre-fix tree outside the repo (`/tmp/prefix-check`: `git show a17a6b3ba:…themed-toast-container.tsx` plus the head test file, run with the test's own `resolveDir: process.cwd()` convention) and ran it:

```
$ cd /tmp/prefix-check && node --test scripts/toast-close-placement.test.mjs
✔ sonner places the close button with two custom properties on one 20px box
✔ `unset` on a custom property inherits, which is how the button reached the LEFT edge
✖ the shipped theme leaves exactly one inset governing, and it is the right one
✖ the shipped transform is sonner's own RTL one, and it points out of the side that governs
$ echo $?  → 1
```
On head: `4 pass, 0 fail`. So it fails 2/4 on the pre-fix tree and passes on head, as the body says. Its model of the cascade (declared-value resolution, over-constrained → `left` wins in LTR, guaranteed-invalid → `auto`) matches the browser behaviour the rig reports, and the file states its own bound (no layout; the pixels are the rig's claim). It reads the shipped artefacts rather than a copy, so an upstream rename fails it by name.

**Geometry re-derived independently — it reproduces the body's numbers exactly.** `node scripts/toast-close-geometry.mjs --out=/tmp/r1-toast-geo --port=5478` on head:

```
case                       lines  toast(l,t,w,h)      button(l,t,w,h)     d(right,top)  corner(tl,tr)  in-viewport  start/end
localOperatorDark-one-line     1  1000,832,356,44     1342,826,20,20      6,-6          352,5.7        true         auto/0px
localOperatorDark-multi-line   3  1000,796,356,80     1342,790,20,20      6,-6          352,5.7        true         auto/0px
localOperatorLight-one-line    1  1000,832,356,44     1342,826,20,20      6,-6          352,5.7        true         auto/0px
localOperatorLight-multi-line  3  1000,796,356,80     1342,790,20,20      6,-6          352,5.7        true         auto/0px
dracula-one-line               1  1000,832,356,44     1342,826,20,20      6,-6          352,5.7        true         auto/0px
dracula-multi-line             3  1000,796,356,80     1342,790,20,20      6,-6          352,5.7        true         auto/0px
all assertions passed
```
Identical to the PR's "after" rows, and the author's `before` log (`-328,-6`, `18.4,338`, `0/0px`, `18 assertion(s) failed`) matches its "before" rows. I also **looked at the frames** — mine and the author's, full and 3× crop: the X straddles the toast's top-right corner after, and sits over the top-left before. That visual step is load-bearing rather than decorative: a `getBoundingClientRect` cannot show that a box 7px outside its parent is not clipped, and only the pixels show the full circle painted. The rig reaped its Chrome and vite cleanly (`/tmp/lo-toast-close-*` gone, no leftover debug-port process of mine), used `withMockKeychain` + `--headless=new` + a private `/tmp` profile, and its `chrome-keychain.test.mjs` row is the right table entry for a third rig of this shape.

**Registrations, guards and PR state.** `git rev-parse ad4a3bc18:src` = `903943dd6…` and `:scripts` = `9ae206e63…` — equal to the manifest's `srcTree`/`scriptsTree`; a key-level diff of the base's manifest against head shows **exactly those two fields** changed and nothing else. `package.json` diff is the `test:desktop` list only, which now names `scripts/toast-close-placement.test.mjs`; `version` is `0.29.2` on both base and head. `node scripts/check-scripts-lint.mjs` with `LINT_SINCE=a17a6b3ba` → exit 0, "7 changed file(s) … 6 lint-clean … 1 no handler (the .html)". `gh pr checks 387` → all nine green. PR is non-draft, assigned to `damianvtran`, `reviewRequests: []`, zero `review_requested` events, zero comments before this one. No `gh issue create`; main's tip is `a17a6b3ba` and its history shows no rewrite. Scope of the fix by construction: `closeButton: true` is set once on the single `<Toaster>` (`themed-toast-container.tsx:79-86`), no other `Toaster`/sonner surface exists in `src/`, no other consumer of `--toast-close-button-*` exists, and `toast-manager.ts` passes no per-toast `style` — so every toast shape (receipt, error, canvas refusal, dedup) rides the same theme, and the multi-line case is the extra shape that matters. No `dir` anywhere in the renderer and the rig page deliberately carries none, matching `src/renderer/index.html`; the RTL arm the test asserts is extra safety, and with physical `left:auto`/`right:0px` it stays right in both. The app's other close controls are unrelated and untouched (`dialog.tsx:98`, `sheet.tsx:121`, `floating-alert.tsx:83`, `image-lightbox.tsx:259`, `update-notification.tsx:358` — Tailwind classes on their own containers).

### Findings

**R1-1 — major — evidence: the frames exist, but no reader can see them.** `docs/evidence/` gains no set, no image is attached, and the only copies are `/tmp/toast-close-geometry/{before,after}` (12 + 12 PNGs), which the body's "Frames" paragraph points at. `/tmp` is not durable and not fetchable by a reviewer. For a user-visible change this repo's record is unambiguous: every merged PR that changed `src/renderer` and a pixel committed before/after frames under `docs/evidence/` — 373 (36), 369 (72), 362 (72), 364 (98), 361 (99), 346 (97), 345 (99), 343 (6), 341 (2); the only merged renderer-touching PR with zero frames is 363, the socket-plane deletion with no visual delta. PR 378, the closest shape to this one, is still **OPEN** (`gh pr view 378 --json state,assignees` → `OPEN`), so it is not a precedent for anything. A numbers table plus a regeneration command is not a substitute here for two reasons: a user-visible change's evidence is what the design round reads, and the design gate takes rendered artifacts and never source alone; and the numeric half cannot carry the claim that the circle is not clipped by the toast it straddles — the pixels carry that. Fix, either shape: (a) commit the pair as a declared `supplementary` set — `docs/evidence/toast-close-corner/{before,after}/<theme>-<shape>.webp` like the rest of the set, plus one manifest entry with `path`, counts, a `source` naming this rig, `why`, `capturedAt` and a `capturedAtHead` that is reachable from a ref (`check-evidence` fails an unreachable one with "is reachable from no ref - it dies at the next gc"); declaring a set leaves `frames`/`surfaces` untouched, since both count outside declared sets and `STORIES` rows respectively, so this is one entry and not a re-sweep — then watch `pnpm check-evidence` (it may exit 75 DEFERRED behind another session's sweep); or (b) attach the same images to this PR. Either makes them reviewable.

**R1-2 — minor — the test plan's reproduction command cites a commit no reader can resolve.** The PR body's test plan says `git checkout b8488a8af^ -- src/renderer/src/shared/components/common/themed-toast-container.tsx`. `b8488a8af` is the pre-rebase spelling of the fix commit: `git merge-base --is-ancestor b8488a8af ad4a3bc18` → not-an-ancestor, `git for-each-ref --contains b8488a8af` → **0 refs**, so it dies at the next gc — the exact failure PR #352 ("cite a before-set head a reader can actually resolve") was opened to correct in this file's citation fields. The pre-fix content is byte-identical to the base's, so the fix is one word: `diff <(git show b8488a8af^:…themed-toast-container.tsx) <(git show a17a6b3ba:…themed-toast-container.tsx)` is empty and `git for-each-ref --contains a17a6b3ba` → 9 refs. Cite `a17a6b3ba` (or `origin/main`) instead. Nothing catches this automatically: `check-evidence`'s provenance and citation halves police manifest fields, not PR prose.

**R1-3 — nit — the rig's default port collides with an existing rig.** `scripts/toast-close-geometry.vite.mjs:35` defaults to `5431`, the same default as `scripts/store-refusal-evidence.vite.mjs:134`, and both set `strictPort: true`, so the two rigs cannot run simultaneously; every other rig takes a distinct default (5198-5211, 5251, 5429). With this box running ~25 sessions that is likelier than it looks — I passed `--port=5478` to be safe. Fix: take an unused default such as 5432.

**R1-4 — nit — the re-stamp's explanation lives only in the commit message.** The two-line stamp diff is correct and `ad4a3bc18`'s message carries what moved, what did not and why, which is what AGENTS.md § "A commit that moves `src/` or `scripts/`" asks for — so this is compliant, not a defect. Worth noting only because earlier re-stamps in this same file also left a `partialCapture.*RestampNote` entry (e.g. `collapseGuardRestampNote`, itself a scripts-only stamp move), so a reader of the file alone cannot see how these two stamps got here.

**Pre-existing, not this diff's, recorded so it is not mis-scoped:** the AGENTS.md sentence under "A rig's Chrome does not touch the keychain either" says "the nine in `scripts/` and the three under `docs/evidence/<surface>/harness/`"; the launch-site table already held 18 + 4 rows before this PR and holds 19 + 4 with the new one. The new row is in the right table; the count in the prose was stale before the branch was cut. No action asked of this PR.

### Verdict

**Findings to remediate — R1-1 (major) is the only substantive one; R1-2 is a one-word fix to the test plan; R1-3 and R1-4 are optional.** The code change itself is correct and the root-cause analysis is accurate at the level of sonner's own stylesheet and of Chrome's actual resolution, and I reproduced every number in the body exactly on an independent run and viewed the frames. Nothing else in the diff is wrong or missing: the CI half discriminates, the registrations are right, no version bump, no rule violation. No QA or design round exists on this PR yet (this is the first comment), and R1-1 is precisely what the design round needs in hand.

damianvtran:
### Design review — round 1

Reviewer: designer (subagent), model `deepseek/deepseek-flash` — no provider fallback during this round.
Scope: `a17a6b3ba..ad4a3bc18` (head `ad4a3bc18`, re-checked against the worktree HEAD and `gh pr view 387`: same SHA, MERGEABLE/CLEAN).

**Frames I opened, not merely generated**

- Author's set — `/tmp/toast-close-geometry/before/` and `/after/`: full frames plus the 3x corner details, `localOperatorDark` / `localOperatorLight` / `dracula` × one-line / multi-line. All 24 viewed.
- My own set, same rig, unmodified: `node scripts/toast-close-geometry.mjs --themes=… --out=…` → `/tmp/design-387-mine/` (`sage`, `iceberg` — the two *ported light* palettes the author's set lacked, which is where the contract says contrast defects hide), `/tmp/design-387-themes/` (`obsidian`, `tokyo-night`, `dune`, `neon`) and `/tmp/design-387-themes2/` (`monokai`, `radient`, `synth`). That is all twelve palettes framed, one-line and multi-line; every run asserted `auto/0px` and the +6/−6 straddle and passed. I then opened at least one corner detail per palette — `sage`, `obsidian`, `tokyo-night`, `neon` and `radient` in the one-line shape, `iceberg`, `dune`, `synth` and `monokai` in the multi-line one — so each shape is covered by several palettes, on top of the author's 24 frames (both shapes × three palettes).
- My own pixel work on those frames: 10x corner crops, a 28x crop of the junction, and a per-pixel map of the corner region. The junction and clearance claims below are read off that map, not off source.

**Verdict on straddle vs. inside: the straddle is right.** Not "acceptable" — right, and the alternative would be a regression.

1. **It is sonner's own geometry, mirrored, not an invented one.** `node_modules/sonner/dist/styles.css:6-24` declares `--toast-close-button-start: 0` / `-end: unset` / `transform: translate(-35%,-35%)` for LTR, and for RTL the exact triple this PR now writes (`unset` / `0` / `translate(35%,-35%)`); line 223 consumes them as `left`/`right`/`top: 0` on a 20px box. Sonner's own answer for the right-hand side *is* a straddle, so the app has adopted the library's placement rather than litigating it — which is also why that placement survives a sonner upgrade. (The old `start: "unset"` really did mean `inherit` on a custom property, which is how LTR's `0` came back and both insets resolved to lengths; the comment now in `themed-toast-container.tsx` states that correctly, and states it in the right place.)

2. **The corner it lands on, measured from the frames.** Before: button box `(1008,826)`, right edge 328px *left* of the toast's right edge, centre 18.4px from the toast's top-left corner — the frames show the X sitting over the leading check icon with its circle crossing the toast's left border, i.e. the operator's report reproduced. After: `(1342,826)` on a toast at `(1000,832) 356×44` — right edge 6px outside the toast's, top edge 6px above it, centre 5.7px from the top-right corner, and 18px clear of the viewport edge (sonner's `OFFSET` is 24px, so the button is the closest object to the window edge and still clear).

3. **The copy can never collide with it — this is the argument that decides it.** The button is positioned against the toast's padding box, so at rest the ring spans x `1342→1362`, y `826→846` while the content box is x `1017→1339`, y `845→863`. Closest approach, content box's top-right corner `(1339,845)` to the circle's centre `(1352,836)`: 15.8px, i.e. **5.8px of clearance** in the worst case of a line filled to the content edge. In the rendered multi-line frame the first line ends at x≈1323 — 19px short of the ring. No message length, no second line and no trailing glyph can reach it, and the message keeps its full 356px.

4. **An inside placement would cost something real and buy nothing.** Dialog (`top-4 right-4`) and FloatingAlert (`top-1.5 right-1.5`) both sit inside panels that reserve the corner for them — a 16px header inset, and `pr-10` on the alert — and both are `icon-sm` controls, i.e. 28px with a 14px glyph. A toast's content is 20px tall inside a 44px box with 12/16px padding: a 20px control inset 6px from the edge reaches x = right − 26, which is 9px *past* the content edge at right − 17; the app's own 28px idiom would reach 17px past it. Either the message gets pushed in (about four characters a line on a 356px toast) and the toast grows, or the text and the control overlap. The straddle is what makes the corner free, and that is exactly the trade sonner made. (The two neighbour **geometries** here are quoted from the app's own source, `dialog.tsx:96` / `floating-alert.tsx:83` / `button.tsx:146`; I did not render those two screens, and no finding below rests on them.)

5. **The corner reads as one object, not two circles colliding.** The pixel map shows the toast's top border running in from the left and stopping at x≈1343 where the ring's stroke begins, the right border re-emerging below the ring's tangent at y≈846, and the ring drawn in the same role (`--normal-border` / `--gray4` = `--color-control`) at the same 1px weight as the toast's own border — so the top edge flows into the circle with no seam, no doubled edge and no gap. Nothing is clipped: sonner puts no `overflow: hidden` on a toast (the only `overflow` in its sheet is `overflow-wrap: anywhere`). The X is centred in the circle: its ink centroid lands within 0.25px of the circle's centre in every palette I measured, and its drawn extent (6.5px) reads at the same weight as the 1px ring rather than heavier.

6. **It holds on every palette, not just the two brand ones.** Ring against the toast ground: 3.02:1 (dracula) → 4.67:1 (neon); ring against the page ground: 3.34:1 (iceberg) → 5.64:1 (neon). Both floors clear. Worth naming what carries the read: the button's fill is `--normal-bg` = `elevated`, the *same role as the toast it sits on* (confirmed in the frames — button fill and toast fill are the same pixel in every palette), so where it overlaps the toast the ring is the boundary, and where it hangs over the page the ring is again the sole boundary. That is the design, and `borderControl` at 3:1 on all four grounds is what makes it legal.

**What a user gains, what they lose.** They gain a dismiss target in the corner they aim at, on the object that exists to be dismissed, drawn the way sonner's users already know it — with the copy never at risk and the message keeping its full width. They lose only the literal sameness with the app's two inset close controls. Accepted: those two sit on panels whose padding already reserves a corner, and a 44px toast's does not. I would not spend a text-width concession to make a toast look like a dialog.

**D-findings: none.** No BLOCKER, no MAJOR, no MINOR, no NIT — and I am not padding the round with nits to look thorough.
**Verdict: TERMINAL.** The round is clean on `ad4a3bc18`, and the straddle is the sign-off.

**Recorded, not raised as findings** (no action asked; listed so a later pass does not re-derive them)

- **R1 — the one object on the toast without the shadow.** The toast takes `shadow-overlay` precisely because what is beneath it is an arbitrary scroll container; the button's outer half takes none, so the shadow band beside the toast's right edge is interrupted behind the button. Measured in the lightest palette (localOperatorLight is the worst case): 2px out from the toast's right edge at the button's own band reads `242,237,227` — pure canvas — against `237,232,223` 30px lower where the shadow is. A 5/255 difference, downward-biased anyway, invisible at 1x. Recorded, not asked for.
- **R2 — `CONTROLS` has no row naming this control.** What it paints (fill `elevated`, border `borderControl`, ink `ink`) *is* asserted in `scripts/contrast-contract.mjs` — by the `ask option button (hover)` row, on `canvas`/`surface`, which are the grounds a toast can float over — so nothing here is unmeasured. A row naming the toast's close button would make the gate's green output legible against the control it protects. Pre-existing bookkeeping; deferred.
- **R3 — the hover state is not in this round's frames, including mine.** Sonner hovers the button to `--gray2` (`--color-accent-wash`) on `--gray5` (`--color-control`); both resolve, and the toaster reports `data-sonner-theme="light"` (`theme` is unset on `<Toaster>` and sonner's default is `'light'`), so the `--normal-bg-hover` arm at `styles.css:530` is never taken. Computed from the palettes, the ring on that wash is 2.89:1 on iceberg — the same pair the ask-options row records as sub-floor — and 3.17:1–4.50:1 on the other eleven. In that state the *fill* step carries the hover, not the ring, so I left it alone. If a later pass wants no caveats on this control, a hover frame is the one artifact to add.
- **R4 — narrow window and stacked toasts: source and arithmetic, not frames.** Sonner's mobile branch (`max-width: 600px`, 16px offset) cannot trigger in this window — `WINDOW_MIN_WIDTH = 800` (`src/main/window-mode.ts:109`) — so the button keeps its 18px viewport margin at every window size the app allows. In a collapsed stack the non-front toasts' children are `opacity: 0` (`styles.css:315`), so only one button is ever visible; in the expanded stack the gap is 14px (`GAP = 14`) and the button spends 6px of it. Neither breaks, and neither has a frame.
- **R5 — the control's form is off the radii ramp.** `docs/branding.md` § 5 gives radii as 2/6/10/14/16 and reserves `rounded-full` for avatars, status dots and pill badges; this control is a 20px circle where the app's other close controls are `icon-sm` (28px, 6px radius). It is sonner's own hardcoded `border-radius: 50%` in an unlayered sheet, the shape is untouched by this diff, and the frames are the reason I am not asking for it: a circle is what lets the straddle read as a badge on a 10px corner, where a square poking out of that corner would read as a notch. Deferred.
- Not a shipped surface: no app toast carries an action row or a description — every call site goes through `toast-manager`'s `toast(message)` / `toast.success|info|warning|error(text)`, with no `action:` anywhere — so the action-collision case has no frame to produce. The error/refusal tone differs in icon and copy, not in the corner.

damianvtran:
### QA report — round 1

Independent pass on `ad4a3bc18` (base `a17a6b3ba`). I did not reuse the author's `/tmp/toast-close-geometry` output: my own worktree at the head, my own pre-fix arm, my own frames, my own rig runs. Scripts, matrix and JSON in `/private/tmp/qa387r1/`.

**Verdict: PASS on the delta, with one BLOCKED surface** (the placement measurement taken inside the running app — details and the exact reason below). No FAIL.

#### 1. Geometry, reproduced independently

Rig run in my own tree at head, and in a second scratch tree whose only difference is the container file taken from the base (`--toast-close-button-start: "unset"`):

| arm | declared insets | used insets | `deltas.right` | centre → top-right | centre → top-left | viewport | exit |
| --- | --- | --- | --- | --- | --- | --- | --- |
| before (container from `a17a6b3ba`) | start `0`, end `0px` | `left 0px`, `right 0px` | **-328** | 338 | **18.4** | inside | 1 (18 assertion failures) |
| after (head) | start `auto`, end `0px` | `left 334px`, `right 0px` | **+6** | **5.7** | 352 | inside | 0 ("all assertions passed") |

Identical in every row across `localOperatorDark` / `localOperatorLight` / `dracula` and one-line / multi-line (toast 356x44 and 356x80), with `transform: matrix(1,0,0,1,7,-7)`. These are the PR body's numbers, measured on my own trees — before `-328` / `18.4`, after `+6` / `5.7`, contained in the viewport, identical across the three palettes and both shapes. I also looked at my own frames: before, the circle sits at the top-left and is carried inward; after, it straddles the top-right corner.

#### 2. Real-app pass — BLOCKED (this is the gap)

I built the app from my worktree at head and drove it headless against an **isolated** `lop serve` daemon of my own (port 5498, throwaway config root under `/tmp`; the app log shows `[discovery] picked http://127.0.0.1:5498` then `Attached to daemon … (uv-tool)`), with the renderer built against that same URL. Real CDP input on the app's own controls, and a poll for `[data-sonner-toast]` after every step:

- no-backend boot: refusal **cards** with `Retry` (not toasts); pressed every one — 90 s of polling, zero toasts;
- live daemon: composer opened through the header `New chat`, text typed into `textarea[aria-label="Message"]`, `Send message` pressed; a canvas document opened through the app's own `openCanvasDocument` verb and the canvas views switched; an agent created through the Agents form (submit succeeded, route moved to `#/agents?kind=agent&name=qa-agent-one`); `Start recording` pressed; `/chat`, `/agents`, `/agent-hub`, `/settings` visited.

**Result: no path I could drive raised a toast**, so the close button was never measurable in the live app. The toaster is mounted unconditionally in the app shell (`main.tsx:58`, `<ThemedToastContainer />`), and the detection path itself is proven (it found, measured and dismissed a real toast on the shipped page in §3). What blocks it: the app's failure surfaces are in-page refusal cards rather than toasts, and the mutations reachable without a configured provider report in place. The reported toast itself (`markAllReadReceipt` → "Marked 7 chats as read.") needs unread chats in a backend, which a fresh daemon has none of. **What would close it:** a run with a backend that holds unread sessions (the receipt path), or any seeded toast path; I would then measure the button box against the toast box in the live DOM. I am explicitly not offering a rig run as the app.

#### 3. Hit-test, hover, dismissal (shipped page, real browser engine)

The PR's rig measures geometry and never presses. Driving the same shipped page in a private headless Chrome (mock-keychain, scratch profile, reaped by pid) with real input events:

| question | one-line | multi-line |
| --- | --- | --- |
| `elementFromPoint` at the button centre | `close-button` | `close-button` |
| `elementFromPoint` at the toast's top-right corner | `close-button` | `close-button` |
| hover paint under a real `mouseMoved` | `rgb(50,45,34)` → `rgb(22,40,29)` | same |
| real press → toast gone | 1 → 0 | 1 → 0 |

The button hit-tests at its new position and clicking it dismisses the toast.

#### 4. Regression surfaces

| surface | result |
| --- | --- |
| multi-line toast | PASS — same deltas, same corner distance, same hit-test and dismissal |
| error/refusal dedup + cooldown (`toast-manager.ts`) | PASS by test (`mark-all-read-control`, `toast-lifetime`: dedup keyed off the caller's string, held refusal + cooldown, teardown releases the key). Not reachable as a live error toast — covered by §2's block. The manager file is untouched by this diff. |
| toasts carrying an action button (the composer's Undo) | not driven — needs a queued-message `/clear` or a goal round-trip; no measured collision. See Q1-1. |
| `FloatingAlert` close control | not driven — it appears on update-flow errors only. It shares no token with sonner and no source file of this diff; the diff's only source change is `themed-toast-container.tsx`. |
| other sonner consumers | only `themed-toast-container.tsx` (the `Toaster`) and `toast-manager.ts` import sonner; the rig page is the second mount of the shipped container. |

#### 5. Gates on this head (my tree)

`pnpm lint` exit 0 (87 warnings, 0 errors) · `LINT_SINCE=origin/main pnpm lint:scripts` exit 0 (7 changed `scripts/` files; 6 lint-clean, the `.css` not processed by biome) · `pnpm check-types` exit 0 · `pnpm check-themes` exit 0 (freshness + contrast contract, 25855 assertions) · `node --test scripts/evidence-run-guard.test.mjs` 6/6 · delta test subset `toast-close-placement` + `toast-lifetime` + `chrome-keychain` 18/18. In the pre-fix tree the new guard FAILS by name (`actual '0' expected 'auto'`; "a left-governed box must translate toward its edge"), so it discriminates. The full `pnpm test:desktop` matrix was not run on this box (load average 200-400 during the pass) — the delta subset is what I ran, stated rather than implied.

`pnpm check-evidence` exits **75 DEFERRED** (another session's sweep holds the machine lease; no frames checked, lock file untouched). Rebase pairing re-verified myself: `git range-diff 0c02556ba..823d51d9 a17a6b3ba..ad4a3bc18` pairs both content commits `=` and only re-lays the re-stamp; `docs/evidence/manifest.json`'s `srcTree`/`scriptsTree` equal `HEAD:src`/`HEAD:scripts` on this head.

#### Findings

**Q1-1 (nit, non-blocking, evidence-limited).** After the fix the circle now overlaps the toast's *content* band rather than the left icon band: sonner reserves no space for the close button (the toast keeps `padding: 16px`, which this app sets to `12px 16px`) and the button is absolutely positioned over the padding box, so its left edge (1342) is 14px from the toast's right edge (1356) while the content column runs to the same inset. Vertically the overlap is the top ~14px of the toast. I could not measure an actual intersection because no real toast was reachable (§2), and the rig frames I viewed show no text under the circle in either shape. Recorded rather than deferred-as-a-defect: it is a property of sonner's layout that the pre-fix placement hit on the other side, and a direct measurement needs a real toast.

#### Hygiene

Every app run used a scratch `HOME`, config dir, log dir, `--user-data-dir` and cwd; `CMUX_*`/`LOP_*` deleted from the child environment; the app's own log line shows the keychain redirection was in effect. The daemon I started was stopped by exact pid (port 5498 free again). No file in the operator's `~/Library/Application Support/Local Operator` touched in the last hour mentions my scratch root. The operator's live backend (127.0.0.1:1111) was never addressed. My first cut of the hit-test probe left one re-parented Chrome helper; it was reaped by exact pid, the harness now sweeps its own scratch profile path, and the final check returns zero orphans (no Electron either).

damianvtran:

Answered on head `1b4d334d2`. The round's own head, `ad4a3bc18`, was folded onto `origin/main` = `e8d56ef3e` (the 0.29.3 window and PR #377) and replayed as the three commits this branch now carries; **every actionable finding is answered in ONE commit, `1b4d334d2`** — no cascade. Remediation delta: `e8d56ef3e..1b4d334d2`.

**R1-1 (major, evidence) — FIXED in `1b4d334d2`.** The frames are committed, in the shape you named, as two declared `supplementary` sets — `docs/evidence/toast-close-corner/before/` and `.../after/`, twelve frames each, laid out as `one-line/`, `multi-line/`, `one-line-detail/` and `multi-line-detail/` at `localOperatorDark`, `localOperatorLight` and `dracula`, i.e. four surfaces × three themes per half. Each entry carries `path`, `frames`, `surfaces`, `themes`, `source`, `why` (three paragraphs: what the frames are, why pixels and numbers are both needed, why a sweep cannot take them) and `capturedAt`, and each carries a `capturedAtHead`: `before` cites `a17a6b3ba` (an ancestor of `origin/main`, so any reader can fetch it) and `after` cites `65441737b`, this lineage's spelling of the head the capture ran at — **re-pointed, not re-shot**: the pre-fold spelling is reachable from no ref after the fold's force-push, which is exactly what `citationFailures` fails by name, and the fix's own container file is byte-identical at both. `capturedAtHeadNote` says so in the file.

Both framings are committed, so your fallback was not needed: I checked the 3× corner crops against `check-evidence.mjs`'s own `assertFramePaints` first and **they pass** — a crop is a ground (canvas) with a toast on it, well inside the 98.5% uniformity ceiling — so `after/` and `before/` each hold the settled full window (the containment/position claim: the toast's 24px corner offset, the button 18px clear of the window edge) *and* the corner at 3× with 28px of surrounding window in the crop. 24/24 pass.

Two checks beyond the declaration, because the numeric half cannot carry them:

- **`pnpm check-evidence` exits 75 DEFERRED** — another session's sweep holds the machine-wide lease (`lsof /tmp/local-operator-ui-check-evidence.lock` names its `node` and its `magick` child), so I did not wait on it. Run directly against this manifest and tree, the gate's own exported validators all report zero failures: `stampFailures`, `citationFailures`, `partialCaptureFailures`, `countsMeanFailures`, `provenanceFailures`. The walk's three terms are consistent: 8048 committed WebP outside the sets, 9472 on disk, 1424 inside the 90 declared sets — and the 24 new files sit inside the two new sets, which is why `frames` (8048) and `surfaces` (787) have not moved.
- **An adjacent problem, reported and not touched.** Running the sweep's declared-set rules here also reports nine failures in sets this PR does not own — `wake-live-app` and `browser-pane-live` (sets of `.png` frames, which the walker counts as zero), `browser-conversation-tabs` (claims 8, holds 4 `.webp`), `chat-sidebar-current-row-baseline` (claims 84, holds 231), and `turn-stamps-before`, `chat-phantom-compose-rows/before`, `theme-legibility-before` (no `capturedAt`). The identical nine come back from an untouched worktree of `origin/main`, so they are main's own and were neither introduced nor fixed here.

**R1-2 (minor, the un-resolvable repro SHA) — FIXED (PR body only; no file carries it).** The test plan now reads `git checkout a17a6b3ba -- src/renderer/src/shared/components/common/themed-toast-container.tsx`, with a note that the pre-fix file is byte-identical at that commit and at the parent of this branch's own fix commit. I also swept the manifest for the same class: the before set's `why` named the dead pre-rebase spelling, and now names the live ones (`a17a6b3ba` and `f30490f08`'s parent).

**R1-3 (nit, the port collision) — FIXED in `1b4d334d2`.** The default is `5430`. I surveyed every rig default rather than only the one you named: 5198 (diff-body, hosting-census), 5199 (backend-error), 5202 (submit-latency), 5204 (draft-pick, draft-splash), 5206 (new-chat-shortcut), 5211 (session-switch), 5251 (backend-settings), 5429 (composer-alert) and 5431 (store-refusal — the collision). **Your suggested `5432` is not free either**: the operator's own postgres listens there (`lsof -nP -iTCP:5432 -sTCP:LISTEN` → `postgres 1539`), and `strictPort` means a default the rig cannot bind is a default nobody can run — that is how I found it, by taking your suggestion and watching the rig die with `Error: Port 5432 is already in use`. `5430` was verified free before adopting, the two files that must agree (the driver and its vite config) both say so with the constraint written down, and the head's rig run uses no `--port=` at all: `vite … Local: http://127.0.0.1:5430/`.

**R1-4 (nit — re-stamp explained only in the commit message) — NO ACTION, as you judged it compliant.** One note for the record: the same commit that re-derives `scriptsTree` (9ae206e63 → the folded tree's value, because the port move is under `scripts/`) also records the move in `partialCapture.toastCloseCornerRestampNote` — what moved, what did not, and why — so the file alone no longer depends on the commit message for it.

Two housekeeping facts about the fold, since the head moved twice: `package.json`'s `test:desktop` conflict was resolved by taking main's line and re-inserting this branch's single entry beside `toast-lifetime`'s (no version line: `0.29.3` is main's), and the **re-stamp-only commit this branch carried was dropped by the fold** as empty, which this file's own record elsewhere predicts — its whole delta against main's manifest was two stamp values the commit after it re-derives anyway.

damianvtran:
### QA remediation — round 1

Answered on head `1b4d334d2` (the round's head `ad4a3bc18` was folded onto `origin/main` = `e8d56ef3e` and replayed; the delta the round tested is unchanged — the rig reproduces the same numbers at the folded head, and the toast container is byte-identical, `git log a17a6b3ba..origin/main -- src/renderer/src/shared/components/common/themed-toast-container.tsx` is empty).

**Q1-1 (the circle over the content band) — answered, accepted as inherent, no code change.** You were right about the mechanism and could not measure it because no live toast was reachable; the measurement exists in the design round's own per-pixel map of the corner region, and it says the overlap is not a defect. Sonner reserves no space for the close button — it is `position: absolute` against the toast's padding box with the toast keeping its full `padding: 12px 16px` — and the pre-fix placement overlapped the *leading icon* band in exactly the same way. This change moves that overlap from the edge where the toast draws something to the trailing edge, where it does not. Measured: the button's ring spans x `1342→1362`, y `826→846` while the content box is x `1017→1339`, y `845→863`, so the closest approach — the content box's top-right corner `(1339,845)` to the circle's centre `(1352,836)` — is 15.8px, i.e. **5.8px of clearance in the worst case of a line filled to the content edge**; in the rendered multi-line frame the first line ends at x≈1323, 19px short of the ring. No message length, no second line and no trailing glyph can reach it, and the message keeps its full 356px. That is now checkable in the repository rather than in a rig's `--out`: `docs/evidence/toast-close-corner/after/multi-line-detail/` is the frame. Reserving padding for the button would be sonner's layout to change, not this fix's, and it would move the toast's own box — a different PR with its own frames.

Your related note stands as recorded: the composer's Undo toast is the one shape carrying an action row, and it is not reachable without a queued-message `/clear` or a goal round-trip; every shipped call site goes through `toast-manager`'s text-only helpers (no `action:` anywhere), so there is no frame of a collision to produce.

**The BLOCKED live-app surface — stays BLOCKED, recorded, no action asked.** The reason is unchanged and is not something this round can fix: a fresh daemon holds no unread sessions, so `markAllReadReceipt` → `Marked 7 chats as read.` cannot be raised, and the app's reachable failure surfaces are in-page refusal cards rather than toasts. Closing it needs either a backend seeded with unread sessions or a seeded toast path. What now stands in for it, and is stated as standing in for it rather than implied: the shipped `ThemedToastContainer` holding the shipped receipt copy, mounted by the rig and committed as the two declared sets (`docs/evidence/toast-close-corner/`), beside your own hit-test/hover/dismissal pass on the shipped page — and the manifest's `before` entry says outright that these are not a live-app capture, so no reader has to infer that from a directory name.

**Delta stability, for the re-check.** Nothing in this remediation changes what the round exercised: `scripts/` gains a default port, `docs/evidence/` gains the frames and their declarations, and `package.json` gains one `test:desktop` entry (no version line). The delta suite is green on head — `toast-close-placement` 4/4, `chrome-keychain` 5/5, `toast-lifetime` 9/9, `evidence-manifest` 35/35, `evidence-run-guard` 6/6 — `pnpm lint` and `LINT_SINCE=origin/main pnpm lint:scripts` exit 0, `pnpm check-types` and `pnpm check-themes` exit 0, and the rig's own run at the folded head prints the same table with `all assertions passed`.

damianvtran:
### Agent review — round 2

Reviewer: independent reviewer subagent (did not author this change) — model `deepseek/deepseek-flash`, no provider fallback during this round
Scope: `ad4a3bc18..1b4d334d2` (round 1's head, folded onto `origin/main` = `e8d56ef3e`)

Every round-1 disposition below was re-derived from the tree, the manifest, the live PR body and the live forge state. None is taken from the remediation comment.

**R1-1 (major, evidence) — FIXED, and I opened the frames myself.** `git diff --name-status origin/main..HEAD -- docs/evidence` is `M manifest.json` plus 24 `A` frames, twelve per half; all 24 are VP8L (lossless) and all 24 blobs are distinct, so no frame is a copy of another sitting in the wrong directory. I viewed `before/one-line/localOperatorDark.webp`, `after/one-line/localOperatorDark.webp`, `before/one-line-detail/localOperatorDark.webp` and `after/multi-line-detail/localOperatorLight.webp`: the `before` frames put the X over the toast's **top-LEFT** corner, 8px inside the left edge, and the `after` frames straddle the **top-RIGHT** corner with the circle unclipped and the message clear of it; the multi-line crop is the three-line receipt copy, as its directory says. The gate's own predicate agrees — `check-evidence.mjs`'s `assertFramePaints`, called per file with the theme its filename names, is **24/24 pass**, and the walk's three terms are exactly the ones the remediation quoted: **8048** committed WebP outside the declared sets, **9472** on disk, **1424** inside the **90** sets, with `frames` 8048 / `surfaces` 787 / `themes` 12 / `head` / `capturedAt` unmoved from main's readings. `pnpm check-evidence` itself is still lease-blocked — `lsof /tmp/local-operator-ui-check-evidence.lock` names another session's `node` (16971) and its `magick` child, so it would exit 75 DEFERRED; I did not wait on it, did not unlink it, and ran the gate's exported validators directly instead: `provenanceFailures`, `partialCaptureFailures`, `stampFailures`, `countsMeanFailures`, `citationFailures`, `citationAncestryFailures` — **0 failures each**.

The two new entries are well-formed and honest. Both carry `path`/`frames`/`surfaces`/`themes`/`source`/`why`/`capturedAt`/`capturedAtHead`/`capturedAtHeadNote`, hold exactly the twelve frames each claims, and declare no overlap with each other or with main's sets (they are not among the nine failures below). `capturedAtHead` is reachable in both: `a17a6b3ba` is an ancestor of `origin/main` (32 refs; the before half's recipe `git worktree add --detach`, copy the five rig files in, run the rig, is therefore runnable by any reader), and `65441737b` is carried by this branch's own ref. The one load-bearing claim the `after` entry's note makes — "this set's landing commit touches no `src/`, and the one file the pixels depend on is byte-identical at both" — is true: `themed-toast-container.tsx` is blob `1858e854` at `ad4a3bc18`, at `65441737b` and at `1b4d334d2`, and `1b4d334d2` touches no `src/` at all. The geometry prose is checkable and checks out against the frames and the round-1 table: toast `1000,832,356,44` is 1380−24−356 and 900−24−44; the button at `1342,826` has its right edge 6px outside the toast's border box, its centre 5.66px from the top-right corner and 18px clear of the window edge; the before frame's `1008,826` is 8px inside the left edge and 18.44px from the top-left corner. Declaring the sets rather than re-shooting left both sweep counts where they were, which is what R1-1 asked for.

**R1-2 (minor) — FIXED in the live PR body, and the class was swept.** `gh pr view 387 --json body` in full: `b8488a8af` → **0** occurrences, `823d51d9` → 0, `ad4a3bc18` → 0; `a17a6b3ba` → 4, including the test-plan line `git checkout a17a6b3ba -- src/renderer/src/shared/components/common/themed-toast-container.tsx`, and `a17a6b3ba` resolves (32 refs). `git grep` over the whole HEAD tree finds neither dead spelling anywhere. One related citation I checked rather than assumed: the `after` entry's `capturedAtHeadNote` names the pre-fold spelling `ad4a3bc18` in prose, but it names it to say that it is reachable from no ref and that the field therefore cites the replayed `65441737b` — I confirmed both halves (`git for-each-ref --contains ad4a3bc18` → 0 refs; the field's value is the reachable one), so a reader is pointed at the fetchable commit and not at a dead one. Accepted, not a repeat of R1-2.

**R1-3 (nit) — FIXED, and the two files agree.** `scripts/toast-close-geometry.mjs:74` and `scripts/toast-close-geometry.vite.mjs:36` both default `5430`, and the driver passes `--port String(PORT)` **and** `TOAST_CLOSE_PORT`, so a `--port=` override still agrees end to end. I re-ran the census myself rather than trusting the comment's list: no rig holds 5430 (my grep of every rig default finds 5198 diff-body/hosting-census, 5199 backend-error, 5202 submit-latency, 5204 draft-pick/draft-splash, 5206 new-chat-shortcut, 5211 session-switch, 5251 backend-settings, 5429 composer-alert, 5431 store-refusal), 5431 is still store-refusal's, and 5432 is the operator's postgres (`lsof -nP -iTCP:5432 -sTCP:LISTEN` → `postgres 1539`), so rejecting R1-3's suggested neighbour was right. A connect probe on 5430 refuses, i.e. free as of this round. The cited ports in the new comment all exist (`5211` is session-switch's).

**R1-4 (nit, no action) — accepted**, and better than accepted: the same commit records the move in `partialCapture.toastCloseCornerRestampNote`, so the file alone now carries what the commit message carried.

**The fold — content unchanged, and nothing of main's was lost.** `git range-diff a17a6b3ba..ad4a3bc18 e8d56ef3e..1b4d334d2` gives patch 1 `=` (identical), patch 2 `!` whose only textual difference is the context of the `test:desktop` hunk (`--`/`-+`/`+-`/`++` on that one line: the same entry re-inserted against main's moved line), patch 3 dropped, and a new patch 3. The cumulative stats are identical either side — 1 file/22+/3− for patch 1, 9 files/1163+/4− for patch 2 — so the replay is content-preserving, which is the strongest form this check takes on a rebase. `git diff --name-status origin/main..HEAD` is exactly the branch's own file set (the toast container, six rig/test scripts, `package.json`, the manifest, the 24 frames), and `git diff origin/main..HEAD -- docs/evidence | grep -v '^A'` is `M manifest.json` alone — no swept frame moved, as the note claims.

Manifest conflict resolution, key by key against `git show origin/main:docs/evidence/manifest.json`: **all 88 of main's `supplementary` paths are still present**, HEAD adds exactly the two `toast-close-corner` sets (90), and the only other fields that differ are `srcTree`, `scriptsTree`, `countsMean.frames` and `partialCapture.toastCloseCornerRestampNote` — i.e. main's side was taken wholesale and this branch's records re-layered, with nothing of main's displaced. `head`, `capturedAt`, `frames`, `surfaces` and `themes` are main's own values. `package.json` is main's line plus this branch's one entry: 169 tokens vs main's 168, no duplicates, `toast-close-placement.test.mjs` inserted beside `toast-lifetime.test.mjs`, `version` still `0.29.3`, and the forge's own Version Bump Guard passes. Stamps: `git rev-parse 1b4d334d2:src` = `9cb2e1a43f11df511dc8a493dd5fb59df099d1b9` and `:scripts` = `178836283f8e7ca766fe89990ba4bb1386d9050b`, both equal to the manifest's `srcTree`/`scriptsTree`; they moved because main's trees moved under them (`4f45106c`/`82fb1118c`), which is what the note says.

**What I ran, rather than read.** `node --test scripts/toast-close-placement.test.mjs scripts/evidence-manifest.test.mjs scripts/chrome-keychain.test.mjs` → **44 tests, 44 pass, 0 fail** (the same 4/35/5 split the QA remediation reports). `gh pr checks 387` on this head → all nine green. I did not re-run the geometry rig: this session has no browser tool, so the pixel claim here rests on the committed frames I opened plus QA's own pre-fix arm, and I am not claiming more than that.

### Findings

**R2-1 — NIT — the landing commit stripped `docs/evidence/manifest.json`'s trailing newline.** `git show <ref>:docs/evidence/manifest.json | tail -c 2 | xxd` ends `7d0a` (`}\n`) at `ad4a3bc18`, at `5272f148b` and at `65441737b`, and `7d` at `1b4d334d2`, so the fold's re-derivation is where the final newline went. Nothing parses differently (`JSON.parse` is unaffected) and no check covers `docs/**/*.json` formatting — which is why Lint is green on it — but the committed blob no longer matches what the capturer itself writes (`scripts/capture-evidence.mjs:7139`: `` writeFileSync(manifestPath, `${JSON.stringify(manifest, null, 2)}\n`) ``), so the next evidence pass will show a spurious one-line diff on this file. Fix is one byte: re-add the final newline. This is the only new defect the remediation introduced.

**Checked for, and not found** (so the next round does not go looking): no frame sits in a directory its content contradicts (24/24 distinct, all lossless, both detail crops and both first-line frames match their names); no port claim in the new comment is false; no claim in the remediation comment is unsupported by the diff — in particular the walk terms, the 24/24 predicate claim and the "no committed WebP outside them changed" sentence are all exactly what I measured; and the nine declared-set failures my replication reports (`wake-live-app`, `browser-pane-live`, `chat-sidebar-current-row-baseline`, `browser-conversation-tabs`, `turn-stamps-before`, `chat-phantom-compose-rows/before`, `theme-legibility-before`) are main's own — the identical nine come back from main's manifest, and neither new set is among them, which is the same disclosure the PR body carries.

### Verdict

**All four round-1 findings are resolved on `1b4d334d2`: R1-1 fixed (frames committed, viewed, and passing the gate's own predicate), R1-2 fixed in the body and swept tree-wide, R1-3 fixed on a verified-free port, R1-4 accepted and now recorded in the file.** The fold is content-preserving (`range-diff` `=` plus byte-identical line sets; nothing of main's manifest record lost) and the head is clean: **no BLOCKER, no MAJOR, no MINOR**. The only new defect is R2-1, one byte of trailing newline, which is a follow-up and not a reason for another round — **round 2 is TERMINAL**.

For the gate's bookkeeping, so the state is not inferred later: this round is fresh on the head (`Scope: ad4a3bc18..1b4d334d2`, reviewed head = `1b4d334d2` = `gh pr view 387 --json headRefOid`); QA's answer is on the same head; and design round 1's sign-off still describes this head's pixels, because the toast container is byte-identical at `ad4a3bc18` and `1b4d334d2` and the remediation touches no visual surface.

damianvtran:
### QA report — round 2

Independent pass on head `1b4d334d2` (base `origin/main` = `e8d56ef3e`), scoped to the remediation delta `ad4a3bc18..1b4d334d2` — the branch's own three commits (`f30490f08`, `65441737b`, `1b4d334d2`) plus the fold onto `e8d56ef3e`. I re-ran and re-derived everything below from primary sources; none of it is taken from the author's dispositions on trust. Scripts, matrix and JSON in `/private/tmp/qa387r2/` (`qa-matrix-round2.md`, `after.json`, `before.json`, `gates.log`, `validators.mjs`, `launch.sh`, `boot-probe.mjs`).

**Verdict: PASS on the delta. No FAIL, no blocker/major finding.** Two non-blocking findings (Q2-1 minor prose, Q2-2 nit byte-level) are recorded below; neither gates the merge. The round-1 BLOCKED surface (the placement measurement inside the running app) is unchanged by this delta and was not re-attempted — the reason is the same as round 1's and is restated in §5.

#### 1. The committed frames are really the rig's output for the two trees the entries claim

This is the round's headline, so I checked it by identity rather than by resemblance. I built my own pre-fix tree (`git worktree add --detach /private/tmp/qa387r2/before a17a6b3ba`, this branch's five rig files copied in, `node_modules` symlinked — no install), ran **the rig from the PR head, with no `--port`**, into `--out=/private/tmp/qa387r2/out-after`, and ran the same rig in the pre-fix tree into `out-before`.

| question | result |
| --- | --- |
| dimensions, all 24 | committed vs mine: **24/24 match** — full frames 2760x1800, one-line-detail 2472x600, multi-line-detail 2472x816 |
| pixels, all 24 | `magick compare -metric AE` → **0 for all 24**; my `magick <png> -define webp:lossless=true -quality 100` re-encode is **byte-identical to each committed WebP** |
| the `after` half is this head's rig output | my run exits 0 with `all assertions passed`, toast `1000,832,356,44` / `1000,796,356,80`, button `1342,826,20,20` / `1342,790,20,20`, `deltas.right +6`, `deltas.top -6`, `corner.topRight 5.7`, `start auto` / `end 0px`, `matrix(1,0,0,1,7,-7)` — the numbers the entry claims |
| the `before` half is the PRE-FIX placement | my pre-fix run exits **1** with **18 assertion(s) failed**, `2 horizontal insets are declared as lengths (--toast-close-button-start="0", --toast-close-button-end="0px")`, button at `1008,826,20,20` — 8px **inside the toast's LEFT edge**, 6px above its top, `deltas.right −328`, `corner.topRight 338` |
| which half shows which placement, from the pixels | the `before` and `after` frames differ in 1652-1657 pixels; the union of the two 20px button windows accounts for all of them (**0 differ outside that union**, AE rounding only). The only thing that moved between the halves is the button's own box |
| looked at, not just measured | I opened the 3x crops side by side: before, the circle sits on the toast's top-left and overlaps the check icon; after, it straddles the top-right corner clear of the text |

A frame that did not match its directory or its claim would have been a FAIL. Byte-identity across all 24 is a stronger result than that: the committed images are reproducible from the rig on exactly the two trees their entries name.

#### 2. The rig's new default port 5430

- 5430 was **free** before both runs (`lsof -nP -iTCP:5430 -sTCP:LISTEN` → nothing), and both arms ran **with no `--port=` at all** — so 5430 is the default the rig actually used, not a flag I supplied.
- The bind is caught live: polling `lsof` during the run found `node 25918 127.0.0.1:5430` (LISTEN) at t=4s, and the rig's own vite output reads `Local: http://127.0.0.1:5430/`. `strictPort` is set, so a taken port is a loud failure rather than a silent fallback.
- **Nothing else on this machine claims 5430.** Across `scripts/*.mjs` and `scripts/*.vite.mjs` it appears only in `toast-close-geometry.{mjs,vite.mjs}`. The comment's ship-list is accurate: 5198 (diff-body, hosting-census), 5199 (backend-error), 5202 (submit-latency), 5204 (draft-pick, draft-splash), 5206 (new-chat-shortcut), 5211 (session-switch), 5251 (backend-settings), 5429 (composer-alert), 5431 (store-refusal, the collision the fix names) all exist as defaults. And the rejection of the obvious neighbour is right: `postgres 1539 … TCP 127.0.0.1:5432 (LISTEN)` — 5432 is the operator's own postgres.

#### 3. The fold onto `e8d56ef3e`

| check | result |
| --- | --- |
| history shape | `e8d56ef3e` is an ancestor of the head (linear); the branch-own diff is the 24 frames + `manifest.json` + `package.json` + the two rig scripts, nothing else |
| nothing in the delta could move the pixels | the container blob `1858e8545` is **identical** at `ad4a3bc18` (round-1 head), `65441737b` (this lineage's spelling of the capture head) and `1b4d334d2`; `git log a17a6b3ba..origin/main -- src/renderer/src/shared/components/common/themed-toast-container.tsx` is **empty**, so neither fold needed a re-shoot — which is what my §1 re-runs confirm independently |
| the fold's `package.json` claim | base's `test:desktop` has 168 entries, head's 169; **added exactly** `scripts/toast-close-placement.test.mjs`, immediately after `toast-lifetime`; **removed none**; order otherwise preserved; `version` is `0.29.3` at both — no bump |
| the fold's manifest claim | supplementary sets 88 → 90, **exactly** `toast-close-corner/{before,after}` added, **none removed**, **no pre-existing entry changed**; `frames` 8048, `surfaces` 787, `themes` 12 and the top-level `head` all unmoved (= main's readings); only `srcTree`, `scriptsTree`, `countsMean`, `partialCapture` changed. `srcTree`/`scriptsTree` equal `git rev-parse HEAD:src` / `HEAD:scripts` |
| build | `pnpm build` → **exit 0**. (A first attempt failed with `VITE_GOOGLE_CLIENT_ID is not set`; the config reads `process.env` directly, so I sourced the checkout's existing `.env` into the build environment — values never printed or echoed. That failure is environmental, not a regression.) |
| boot, headless, isolated | `[window-mode] window mode headless: 1380x900, window created and never shown, page throttling off, no Dock tile` and `state: visible=false focused=false focusable=false`. The renderer reached `file://…/out/renderer/index.html#/chat`, `readyState complete`, title `Local Operator`, **`window.electron` is an `object`** (the preload seam round 1's work also leaned on), `data-theme=localOperatorDark`, and the rendered body text is the real app shell and its agent list — no console errors. So: **built, booted and fully rendered**, with no window and no focus taken. `VITE_DISABLE_BACKEND_MANAGER=true` and the child list holds Electron helpers only — no backend was spawned or killed |
| gates on this head | delta subset `toast-close-placement` + `toast-lifetime` + `chrome-keychain` + `evidence-manifest` + `evidence-run-guard`: **tests 59, pass 59, fail 0** (runner chose concurrency 2 — the memory floor, 3.4 GB available; the box is heavily loaded). `pnpm check-types` exit 0. `pnpm check-themes` exit 0 (`themes.generated.css is up to date (59 themes)`; `Contrast contract holds: 25855 assertions across 59 themes`) |
| hygiene | every launch used a scratch `/tmp` profile and cwd with every inherited `CMUX_*` variable unset before exec; the app is reaped by exact pid and the scratch profile path swept afterwards — **0 survivors**, port released. I did not run `macOS screencapture`, did not touch the operator's live sessions, and did not delete the evidence lock file |

#### 4. `pnpm check-evidence` and the lease holder

`pnpm check-evidence` at the PR head exits **75**: `Evidence check DEFERRED: another sweep (or its image child) holds the machine lease; retry after it finishes. No frames checked. Do not delete the lock file.` The lock file was not touched.

`lsof /tmp/local-operator-ui-check-evidence.lock` names a **live, progressing holder, not a stale one**: `node 46272` running `check-evidence.mjs` from `/Users/damian/local-operator-ui-worktrees/browser-file-transfer`, with a live `magick` child (pid changing between samples) computing histograms over that worktree's frames — and an earlier sample ~25 minutes before showed a *different* node (`16971`) doing the same in the same worktree, so these are sweeps rolling over rather than a wedged one. It runs at `nice 10`, the guard's own priority. **The aggregate walk is genuinely queued behind live work; it is not reclaimable and nothing about this PR is responsible for it.**

Because the lease is unavailable, I verified the manifest the way the gate would, without it: I imported `scripts/check-evidence.mjs` in a scratch script and called its own exported validators against the shipped manifest — `stampFailures` 0, `countsMeanFailures` 0, `partialCaptureFailures` 0, `citationFailures` 0, `citationAncestryFailures` 0, `provenanceFailures` 0 — and ran `assertFramePaints` over the 24 new frames: **24/24 pass**. That corroborates the PR body's claim of a clean stamp/record half independently of the sweep.

#### 5. Round-1 verdicts this delta does not touch

Nothing here is re-litigated, and the reason is a byte-level fact rather than a preference: the delta's only source change is none — the container file round 1 measured is byte-identical (`1858e8545`) at the round-1 head and at this head, and the delta's remaining files are two rig scripts, one `package.json` entry, the manifest and 24 images. So:

- **the fix itself** (`"unset"` → `"auto"`): unchanged, and my pre-fix arm reproduces the defect it fixes in the same 18 failures round 1 saw;
- **its geometry** (`deltas.right +6` / `deltas.top −6`, centre 5.7px from the top-right corner, `matrix(1,0,0,1,7,-7)`, identical across the three palettes and both shapes): reproduced here on the folded head;
- **hover and dismissal hit-testing** (round 1 §3: `elementFromPoint` at the button centre and at the toast's top-right corner returns `close-button`, the hover paint step, a real press dismissing the toast): no file governing hover or dismissal is in the delta at all, so those readings hold unchanged;
- **Q1-1** (sonner reserves no space for the close button): a property of sonner's layout, not of a file the delta touches — untouched, and the author's answer to it stands;
- **the round-1 BLOCKED surface** (a placement measurement taken inside the live app): still BLOCKED and still not re-attempted, because nothing changed the reason — the reported toast needs a backend holding unread chats, which a fresh daemon has none of. The 24 committed frames are what stands in for it, declared as such in their entry rather than left to inference, and §1 shows they are genuine rig output for both trees.

#### Findings

**Q2-1 (minor, prose in the manifest, non-blocking).** The `after` entry's `why` paragraph says the close button "is at `1342,826,20,20` **in both**" shapes. That is true for the one-line toast and false for the multi-line one: with the toast at `1000,796,356,80` the button is at **`1342,790,20,20`** (its x, width and height are indeed the same; its y is 6px above the toast's top, which is what moves). The sentence's own next clause ("its top edge 6px above it") already implies the correct reading, and the PR body's table carries the right pair (`1342,826` / `1342,790`), so this is one sentence in one file. Repro: `node scripts/toast-close-geometry.mjs --json` at the head → `measurements[1].button`. Suggested fix, if taken: "…and the close button sits 6px outside the toast's right edge and 6px above its top in both shapes (`1342,826` one line, `1342,790` three lines)". No gate reads this, and neither the frames nor the deltas are affected.

**Q2-2 (nit, byte-level, non-blocking).** `docs/evidence/manifest.json` lost its trailing newline in this remediation: the head's last byte is `0x7d` (`}`) while `origin/main` and the round-1 head both end `0x0a`, and **19 of the last 20 versions of that file in this repository end `0x0a`** — the head is the only one that does not. Repro: `git show 1b4d334d2:docs/evidence/manifest.json | tail -c1 | xxd -p` → `7d`; `git show e8d56ef3e:docs/evidence/manifest.json | tail -c1 | xxd -p` → `0a`. It is a one-character fix and it matters only because this file's whole job is byte-level stamps and every future conflict resolution on a 7,300-line derived aggregate picks up a spurious `\ No newline at end of file` hunk. Worth noting against the fold claim that main's copy was taken "wholesale", which is otherwise exactly what I measured (§3).

Neither finding is a FAIL. Both are documentary — they live in `docs/evidence/manifest.json` only — so if they are fixed in a follow-up commit, that commit cannot disturb anything this round verified: it moves no `src/` or `scripts/` byte, so both tree stamps stay valid and my §1 and §3 results stand as measured. If they are deferred instead, they need a `deferred — <reason>` note in this thread rather than a silent pass.

damianvtran:

Answered on head `28c83e5c3`, one commit, `docs/evidence/manifest.json` only. Scope: `1b4d334d2..28c83e5c3`.

**R2-1 (nit — the landing commit stripped the manifest's trailing newline) — FIXED in `28c83e5c3`.** `git show 28c83e5c3:docs/evidence/manifest.json | tail -c 1 | xxd -p` → `0a` (`7d` at the round's head). Worth a commit of its own for the reason you gave, and one more: this file's whole job is byte-level stamps, so a one-byte drift in the assembled artifact is a defect of the same kind as a stale stamp, not formatting — the next evidence pass writes the file from the capturer (`capture-evidence.mjs:7139`, `` `${JSON.stringify(manifest, null, 2)}\n` ``) and would have shown a spurious one-line diff on a 7,300-line derived aggregate, with every later conflict resolution inheriting a `\ No newline at end of file` hunk. The commit's only other change is Q2-1's sentence, and neither moves a stamp: this file is in neither tree, and `srcTree` `9cb2e1a43` / `scriptsTree` `178836283` still equal `HEAD:src` / `HEAD:scripts` at the new head. The gate's own validators — `stampFailures`, `citationFailures`, `partialCaptureFailures`, `countsMeanFailures`, `provenanceFailures` — all report zero failures on the result.

Nothing else from your round was touched, and nothing new was introduced: the diff is two lines (the sentence, and the final `}` regaining its newline), which `git show --stat 28c83e5c3` reports as `4 ++--`.

damianvtran:
### QA remediation — round 2

Answered on head `28c83e5c3`, one commit, `docs/evidence/manifest.json` only. Scope: `1b4d334d2..28c83e5c3`. Both findings are documentary, as you said: the commit moves no `src/` or `scripts/` byte, so §1 and §3 of your round-2 report stand as measured, and the round-1 BLOCKED live-app surface is unchanged and still BLOCKED for the reason round 1 recorded.

**Q2-1 (minor, prose) — FIXED in `28c83e5c3`.** The `after` entry's `why` now reads, in full:

> …and the close button sits 6px outside the toast's right edge and 6px above its top in BOTH shapes — `1342,826,20,20` on the one-line toast and `1342,790,20,20` on the three-line one, whose x, width and height are the same and whose y follows the toast's own top — centre 5.7px from the top-right corner and 18px clear of the window edge. The 6px is derived and checkable: 35% of the button's own 20px box is the 7px the transform translates it by, less the 1px border between the toast's border box and the padding box its absolutely positioned children are laid out against.

The rest of the paragraph is byte-identical; the diff for this finding is that one line. Your repro was the one used: `measurements[1].button` → `1342,790,20,20`.

**Q2-2 (nit, byte-level) — FIXED in `28c83e5c3`, and it is the same byte as R2-1.** `git show 28c83e5c3:docs/evidence/manifest.json | tail -c 1 | xxd -p` → `0a`, against `7d` at `1b4d334d2` and `0a` at `origin/main` and the round-1 head — so the file is back to the spelling the capturer writes it in, which is also what 19 of the last 20 versions of it carry. Taken with your note about the fold claim: main's copy *was* taken wholesale for its content and its two-space, non-ASCII-literal formatting, and the missing byte came from the local re-dump that re-laid this branch's entries onto it, not from main's side.

Round-1 dispositions are untouched by this commit: R1-1, R1-2 and R1-3 fixed in `1b4d334d2`, R1-4 accepted with the re-stamp now recorded in the file, Q1-1 answered and accepted as inherent to sonner's placement, and the live-app measurement stays BLOCKED pending a backend holding unread sessions (or any seeded toast path).

PR #386, opened 2026-09-19
fix(themes): put the row states on the backdrop's hue, and let the fill rank them

damianvtran:
Release: patch — hover and selection fills are now the colour of the surface they sit on rather than a tint of the accent, the app rail and the agent-hub categories rail put the current row on the panel its fill is authored against (it was invisible on seven light themes), and the selection's left rounding is back with the accent bar gone.

**Review state:** head `f4a699fe6`, rebased onto `origin/main` at `acb61d2cb` (the folds since this PR opened carried #296's pinned conversations, #380's canvas link opening, the 0.29.5 release, #374's passkey lane and #357's linkify-slash tokens). Rounds 1-4 are answered in the comments below, in order: round 1's four streams (agent review, design incl. its addendum, QA, UX), then each round's remediation - the rail re-grounding, the two palette lifts, the ground guard, the document corrections, the three-half evidence set, round 4's guard limit, and QA round 4's Q-1..Q-7 - with a fold-provenance note per rebase. **Round 4's QA stream is answered**: its Q-1 FAIL (the branch was dirty against a `main` that moved after CI) is fixed by the fold onto `078da11e3`, and Q-2..Q-7 passed on QA's own rig; the two discovery-defeat shapes its Q-3 still found passing are declared in the guard's docstring.

## What lands

The row states stop being tints of the palette's `accent` and become **steps of
the panel they sit on**, and the 2px `accent` bar leaves `rowCurrent` — so the
fill carries the ranking the bar used to, and the row's left rounding comes back.

Implements the operator's ask:

> "The hover colours in the themes and selection colors all look quite ugly and
> need to be refined with design. … hovers and selection highlights should be
> colored and more should be a brightening against the backdrop … you run this
> risk of looking really off-color and ugly. Also remove the left accent bars …
> and restore the left rounding."

The direction is `docs/design/row-states-refinement.md` (author: designer
subagent, carried into this branch by `d78874450`). This PR is that document as
code, and the six places where I deviated from it are named under **Corrections**
below rather than buried.

### 1. All 59 palettes re-authored

One procedure, per palette, deterministic — it reads only `surface`, `accent`,
`ink`, `inkMuted`, `inkDim` and the mode:

```
HUE      the panel's own; the `accent`'s where C*(surface) < 2.0 (7 palettes)
CAST     C*(rowHover)    = clamp(0.60 x C*(surface), 2.5, 24)
         C*(rowSelected) = clamp(C*(surface) + 4.0, 5.0, 24)
[...]

PR #374, opened 2026-09-18
fix(browser): decide only main-frame hops, present Chrome's user agent, and let passkeys reach the OS

damianvtran:
Release: patch — the built-in browser tab can now complete the challenge/iframe flows it used to refuse, and passkeys reach the OS instead of doing nothing.

## What this fixes

Three defects in the browser tab, from one report: a Cloudflare interstitial that never
completed, and a second ask, passkeys.

1. **The per-hop origin gate refused SUBFRAME documents.** `Fetch.enable` with
   `patterns: [{ resourceType: "Document" }]` pauses subframe documents too, so
   for the whole time an agent navigation was in flight every cross-origin iframe
   on the page was failed with `BlockedByClient` — Cloudflare's challenge widget,
   SSO frames, embedded logins, payment frames. The gate now decides only requests
   whose `frameId` is the main frame's and continues everything else.
2. **The user agent carried a `LocalOperator/<version>` product token** that
   Cloudflare's managed challenge never completes for. It now presents Chrome's own
   string with no product token.
3. **Nothing configured WebAuthn**, so `navigator.credentials.create()/get()` could
   never reach the OS. The platform authenticator is now enabled behind a signature
   gate, with a chooser for a request that matches more than one passkey.

Write-up, measurements and limits: [`docs/design/browser-challenges-and-passkeys.md`](docs/design/browser-challenges-and-passkeys.md).

### The gate fix (deterministic, the headline)

`scripts/browser-challenge-proof.mjs --arm local`, a captcha-shaped local
reproduction: the top page on origin A reports `COMPLETE` only when a
**cross-origin** iframe on origin B posts back a token, loaded through the agent's
own gated path (`request_access` → `respondToConsent` in the app's chrome →
`await_access` → `open`, only A approved). 3 attempts per arm, two BUILT trees:

```sh
node scripts/browser-challenge-proof.mjs --app-tree <tree> --arm local --attempts 3
```

| tree | result | per attempt |
| --- | --- | --- |
| `origin/main` | **0/3** | `status=PENDING widget=none refusals=1` (the B document refused; `refusals` is the gate's own `refused a navigation hop` line count) |
| this branch | **3/3** | `status=COMPLETE widget=token refusals=0` |

The page itself ran in both arms (`vis=visible`, rAF ~2650 per 12 s), so this is not
a frozen renderer. Frames: `docs/evidence/browser-challenges/before-local-attempt1.png`
[...]

PR #357, opened 2026-09-18
fix(chat): a slash command is not a file path — gate the linkifier on the disk for extensionless tokens

damianvtran:
## Summary of Changes

Fixes the operator's report: a slash command in a message — `/new`, in the sentence "…conversations are started with `/new` and no message has been sent yet…" — was rendered as a link to a local file, and hovering it raised the toolbar that answered `No file at /new`.

Root cause, measured rather than inferred. `targetsIn` (`src/renderer/src/features/chat/utils/link-grammar.ts`) admitted any bare `/abs` or `~/…` token on SHAPE alone, and the transcript's linkifier runs it with `LINK_POLICY`, whose one difference from the Files panel was that it drops the known-extension filter. The panel therefore had an evidence rule — "a known extension" — where the linkifier had none, so `/new`, `/model`, `/status`, `/notification`, `/new-chat` all became anchors whose probe answered `No file at …`. The panel was right all along, which is why the tile grid never showed this. (This PR makes that three admission differences rather than one: the extension filter, the fragment guard, and the new disk question below.)

Four changes, each reversible on its own:

1. **A question the shape cannot answer** (`8a523722d`): extensionlessness is the shape no text scanner can read (`isAmbiguousCandidate` — a dotfile basename is exempt, because `extensionOf` returns null for `.zshrc` as a syntax artifact rather than because the name has no extension). For exactly those shapes the linkifier consults an INJECTED oracle on `TargetPolicy` — `evidenceFor`, which reads the answer the hover toolbar's own probe already cached — and admits the candidate only when the disk says `exists`. `unknown`, and an absent oracle, REFUSE: the fail-safe direction this module already trades in (a missed link, never an anchor whose claim is wrong). `ambiguousTargetsIn` finds the suspects through the same `targetsIn` (no second scanner), and `useLinkEvidence` in `markdown-renderer.tsx` asks about them in one batched, in-flight-deduped, session-cached `probeFiles` call and re-renders the row when an answer lands. Cost: one `stat` per distinct ambiguous spelling per session, and nothing at all for extensioned/`file://`/dotfile targets, while a row streams (`linkify={false}`), or when the preload bridge is absent.
[...]

PR #296, opened 2026-09-17
feat(chat): pin conversations from the sidebar

damianvtran:
Release: minor — Pinning: a Pinned chats section above Active chats and a pin on any conversation row, driven by the backend's own pin state, so a conversation pinned in the terminal shows up pinned in the app and the reverse. (No version bump in this PR: package.json stays at the last released version and the release is the window owner's.) Correction on the class: this line said `none` when the PR merged; the window owner ruled the window **minor** on AGENTS.md's materiality test and I conceded the point, so the author's claim is minor - it shipped as 0.29.4 patch, which is the under-call the repo prefers and the next release corrects.

## What and why

The operator asked for a Codex-shaped pin in the chats sidebar: hovering a conversation row reveals a pin control, pressing it pins the conversation, pinned conversations are drawn in a `Pinned chats` section above the rest — and the state has to be the **backend's**, so a pin made with `f10` in the terminal is a pin this app shows and the other way round.

This is the UI half. The backend half is a sibling PR on `damianvtran/local-operator` (`feat/desktop-session-pins`, stacked on Ben's `feat/sidebar-pins-and-chips`): it serves the row's `pinned` flag and the write, from the TUI's own `sidebar-pins.json`. Nothing merges before #1200/#1214 land.

**Change class: C2 — standard.** User-visible and additive; no data or contract migration (one new capability key, one new op, one new key on a row model that is `extra="allow"`).

### The wire

* `SessionCatalogueRow.pinned: boolean`, **always present on every row, both values**. This is the one field on that shape that may not be omitted: the client's row merge is `{...current, ...incoming}` under the rule "an absent key is not a claim", so a backend that omitted `pinned` on an unpinned row would leave this app's optimistic `true` immortal — the row would keep its glyph and its section membership after an unpin made somewhere else. Sending both values is what lets a list read settle the field.
* `sessions.pin` → `POST /v1/desktop/sessions/{id}/pin` with `{pinned: bool}`. **Desired state on the wire, never a toggle.** A retried toggle flips the pin back and the user reports "the pin keeps un-pinning itself"; a desired state is idempotent by construction (re-pinning does not reorder), which is why this op carries no receipt and no request id. It is not a `MESSAGE_OPS` member: a boolean and an id are not prose, so it costs the control budget.
* `session_pins` as its **own** capability key, on the rule `session_search` states in that same file: the existing surface is the whole chats list, which renders perfectly well against a backend with no pin store, so bumping `session_catalogue` would hide a working list behind an update it does not need.
[...]

PR #395, opened 2026-09-19
chore(release): bump version to 0.29.6

damianvtran:
## Release window since `v0.29.5`

**Bump: `0.29.5` → `0.29.6` (patch).** One file, one line: `package.json`'s `version`. No other change.

### The PRs in this window

| PR | Merge | What a user gets |
| --- | --- | --- |
| **#386** | `83db6c0b0` | Hover and selection fills are the colour of the surface they sit on rather than a tint of the theme's accent, in all 59 themes; the chat sidebar's selected row loses its 2px accent bar and gets its left rounding back; the app rail and the agent-hub categories rail put the current row on the panel its fill is authored against, where on seven light themes it was invisible. |
| **#374** | `078da11e3` | The built-in browser tab's three defects: the per-hop origin gate no longer refuses subframe documents (captcha/SSO/embedded logins), the UA presents Chrome's own string, and passkeys reach the OS on WebAuthn-entitled builds. |
| **#357** | `acb61d2cb` | A slash command is no longer mistaken for a file path — the transcript's linkifier asks the disk before linking an extensionless token. |

### Why patch rather than minor

All three are behaviour fixes to surfaces the app already has: a legibility fix to existing row states, three fixes inside the browser tab, and a false-positive fix in the linkifier. None is something a user would call a new thing the app can do, which is `AGENTS.md`'s minor test, so a minor here would over-call — and an over-called minor permanently misreports how much changed.

For the record, and because the thread carries the argument: **pinning conversations shipped under `v0.29.4` as a patch (#296)**, which by the materiality rule was an under-call — that window's headline was a step-function capability. A spent number cannot be un-spent, and this window delivers only fixes, so `0.29.6` stays a patch; the notes will say plainly where that landed rather than letting two numbers imply a change that did not happen.

### The window's version check

The check that matters is the version entry, not the file:

```
$ git diff v0.29.5..origin/main -- package.json | grep '^[-+].*"version"'
$ echo $?
1
```

— nothing printed, i.e. the `version` line is byte-identical at both ends (`0.29.5`), so no merged PR consumed a number.

### Rounds

Each PR carries its own rounds and they are answered in its thread. #386: agent review rounds 1-4 plus two convergence rounds, design rounds 1-4 (terminal), UX (terminal), QA rounds 1-4 plus a convergence pass that came back clean on the merged head, CI 9/9 on that head. #374: agent review and QA rounds 1-8, design 1-4, UX 1-4. #357: its own rounds, green on its head.

### One thing outstanding, disclosed rather than papered over

`pnpm check-evidence` has **DEFERRED (exit 75)** all night: a peer lane holds the machine-wide lease, so no branch's evidence frames were walked by the aggregate sweep in this window. The frames each PR added were checked by its own rounds (by pixels and by hash) and every manifest's stamps are derived from the tree it ships in; the sweep is owed to whoever next obtains the lease.


damianvtran:
### Agent review — release bump

Reviewer: reviewer subagent, model `deepseek/deepseek-flash` (role `reviewer`, independent of the author). No provider fallback occurred during this round.
Scope: `83db6c0b0..0629d5bd8` — base `83db6c0b0` (merge of #386, the PR's base commit), head `0629d5bd89cce96d4329e888cd30fff3301fc1ea`. One changed file, `package.json`.

Verified from the tree and the forge, not from the PR body.

**1. The diff is exactly one line in one file.** `git show 0629d5bd8 --stat` → `package.json | 2 +-`, 1 insertion / 1 deletion, one file. The hunk is the `version` line alone (`"version": "0.29.5"` → `"version": "0.29.6"`); nothing else in the file moved — no re-serialisation, no indentation change. Head is a single commit on `origin/main` (`git rev-list --count origin/main..0629d5bd8` = 1; `HEAD^1` = `83db6c0b0`), and `refs/heads/chore/release-0.29.6` on origin is that same SHA.

**2. The window is right.** `git log --first-parent --oneline v0.29.5..origin/main` = `83db6c0b0` #386, `acb61d2cb` #357, `078da11e3` #374 — exactly the three PRs and merge SHAs the body lists (the body's table orders them #386/#374/#357 rather than by merge order; no material difference). `git diff v0.29.5..origin/main -- package.json | grep '^[-+].*"version"'` prints nothing, i.e. `origin/main` still reads `0.29.5` and no merged PR in this window consumed a number.

**3. The bump class is right: patch.** All three are behaviour fixes on surfaces the app already has — a legibility fix to existing row states, three fixes inside the existing browser tab, and a false-positive fix in the transcript's linkifier. None clears the materiality rule's test (a step-function capability you could name in the release title). The closest call is #374's passkeys reaching the OS on WebAuthn-entitled builds — a capability a user notices — but it is a defect fix *inside* the browser tab rather than a new surface or subsystem, so patch stands and a minor would over-call a window of three fixes and permanently misreport it. The body's handling of the earlier under-call is honest rather than a rationalisation: it names pinning-under-`v0.29.4` (#296) as an under-call by the rule, states the number cannot be un-spent, and explicitly does **not** use it to argue this window up — which is the failure mode that would have made it a rationalisation.

**4. Title, state, head.** Title is exactly `chore(release): bump version to 0.29.6`, so the guard's mirror-image case (a title claiming a bump the diff does not make) cannot apply. `isDraft: false`, `reviewRequests: []`, `reviews: []`, no labels, `mergeable: MERGEABLE`. The reviewed SHA is the SHA that would be tagged: PR `headRefOid` = `0629d5bd8` = the `headSha` of both check runs returned above.

**5. The version is not spent.** `git ls-remote --tags origin v0.29.6` is empty; `gh release list` tops out at `0.29.5` (2026-09-19T11:41:49Z). The guard itself printed `Version 0.29.6 sorts above the highest release tag v0.29.5.`

**6. The skipped CI is justified, and the skip path could not be entered by anything but a version line.** The classify job printed `release_bump = true` — "the diff is nothing but the `version` line pair in `package.json`" — with `lint`/`types`/`unit`/`runtime_deps`/`audit`/`pack` all false. That path is reachable only through `isReleaseBump` (`scripts/ci-scope.mjs`), which requires the changed-path set to be exactly `{package.json}` **and** every `+`/`-` body line to be a version line, using the guard's own `versionLineChanges` definition. A dependency or `scripts`-section edit in the same file therefore cannot take the skip — it keeps the full job set — so no job here is skipped for a reason the diff does not earn: nothing installs, compiles, types, packs or audits differently for a version string. `package.json` is also the only version surface (`git grep` finds no other file at head containing `0.29.5`; `publish.yml` reads `jq -r .version package.json`), so the one-line diff leaves no stale mirror behind.

**7. The guard read this PR's own diff.** Step env `PR_TITLE: chore(release): bump version to 0.29.6`; output `Release PR, version change allowed:` with the `0.29.5`/`0.29.6` line pair, then the tag-ordering line above. The workflow is `on: pull_request` with no path filter, so it cannot be narrowed by the diff it judges.

**Findings:** none. No BLOCKER, no MAJOR, no MINOR, no NIT — the property this PR has to carry is exhaustiveness of the release-contract checks above, and all seven hold on the head SHA.

**Verdict: `clean`, and this round is TERMINAL** — remaining minors and nits would be follow-ups, not a reason for another round. The bump may be merged as admin and `v0.29.6` tagged on `0629d5bd8` (or on its merge commit); the release notes must list #386, #357 and #374. One disclosed item is noted rather than raised as a finding: the body's `pnpm check-evidence` DEFERRED (exit 75) is machine-wide evidence debt for the window, and this diff invalidates no evidence stamp because it moves no `src/` or `scripts/` path.

PR #396, opened 2026-09-19
fix(release): refuse to ship or install a bundle macOS will not spawn

damianvtran:
## What happened

On 2026-09-19 at 11:29 the operator's desktop app auto-updated to v0.29.6 and became completely unlaunchable. The bundle is Developer-ID signed, notarized, stapled, `codesign --verify --deep --strict`-clean and `spctl -a -vvv -t exec`-accepted — and macOS refuses to spawn it:

```
$ open -a "/tmp/lo-broken-0.29.6.app"
The application ... cannot be opened for an unexpected reason, error=Error Domain=RBSRequestErrorDomain Code=5 "Launch failed."
  UserInfo={NSUnderlyingError=... {Error Domain=NSPOSIXErrorDomain Code=163 "Unknown error: 163" ... }}

$ log show --last 3m --predicate 'process == "amfid"' | grep 'Local Operator'
amfid: Restricted entitlements not validated, bailing out.
       Error Domain=AppleMobileFileIntegrityError Code=-413 "No matching profile found"
amfid: /private/tmp/lo-broken-0.29.6.app/Contents/MacOS/Local Operator not valid: ... -413 "No matching profile found"
```

The cause is PR #374 (the passkeys change, first shipped in 0.29.6) plus the pipeline that carried it. `scripts/render-mac-entitlements.mjs` rendered `keychain-access-groups = ["<TEAM_ID>.com.local-operator.webauthn"]` into `$RUNNER_TEMP/entitlements.mac.plist`, and `publish.yml` passed that file to **both** `-c.mac.entitlements` **and** `-c.mac.entitlementsInherit`, so every helper — including the `ShipIt` that was supposed to relaunch the app — claimed it too. No provisioning profile was embedded, and per Apple's TN3125 `keychain-access-groups` is a **restricted** entitlement that must be authorized by one. So amfid refused every Mach-O at exec: the app never started, the relaunch leg never ran (ShipIt was refused for the same reason), and the user was left with no app and no message.

[...]

PR #385, opened 2026-09-19
fix(update): report a landed install as landed, and retire the failure it supersedes

damianvtran:
## Summary of Changes

Eight commits on `origin/main` = `a17a6b3ba`, fixing two dishonest update reports the operator hit on 2026-09-18 (the branch was cut from, and has since been rebased onto this base; the fold's only conflict was `docs/evidence/manifest.json`, main's side taken wholesale and this branch's derived fields re-layered):

| commit | what |
| --- | --- |
| `e44654335` `fix(update): read a generation install's evidence through its own pointer` | Defect A |
| `e9bd58bb2` `fix(update): retire the recorded install failure once its target is running` | Defect B |
| `98191c7c7` `docs(evidence): photograph the landed-update notice, and re-derive the stamps` | the rendered pair, and the `srcTree`/`scriptsTree` re-stamp this branch owes |
| `a87834921` `fix(update): make the record's retirement read the same rule as the marker` | review round 1's remediation: R1-2, R1-3, R1-4, D1, Q-1 |
| `fa0025a23` `docs(evidence): re-take both halves with the rig's no-notice fallback` | review round 1's R1-1: the re-taken pair, the ledger, and the new STORIES row |
| `a7a6accd0` `docs(evidence): photograph the retained-record states, and re-derive the counts` | design round 1's D1 frame: the new state's twelve frames, the STORIES row, and the counts |
| `2c822d60f` `fix(evidence): shoot the retained record without a press, and pin the failure copy` | review round 2's D3 (story half), D2, R2-1, R2-2 |
| `1c9cb8c27` `docs(evidence): re-take the D1 and failure-panel frames from a production build` | review round 2's D3 (the frames) and D2 (the re-take), with the ledger's correction |

### Review round 1 — what the review found, and what changed

Round 1 was **not clean**: two majors (R1-1 on this PR's own evidence, R1-2 in the retire rule), one design minor (D1), two nits, one QA nit. Each is answered per-finding in this thread (`### Agent review remediation — round 1`, `### Design review remediation — round 1`, `### QA remediation — round 1`); the substance:

[...]

PR #382, opened 2026-09-18
fix(ci,scripts): a blocked signed-update exercise warns instead of failing, and the spawn-env test ignores the host daemon

damianvtran:
## Summary of Changes

Round 1's findings plus the two small fixes this branch started with, reconciled onto the current `main` by two fold-forwards, as six commits on `origin/main` = `181a9c4fd`:

- `ci(signed-update): a capability this runner lacks is not a candidate failure` (`45bd08af1`) — `.github/workflows/signed-update-candidate.yml`.
- `test(scripts): keep the live daemon out of the backend spawn-environment test` (`8d906ade5`) — `scripts/python-bytecode-cache.test.mjs`.
- `fix(ci,scripts): describe what exit 2 covers, and keep the BLOCKED annotation on one line` (`b2c94a786`) — round 1's MINOR-1/MINOR-2 and Q-1/Q-2.
- `docs(evidence): re-derive both tree stamps at this head` (`3d59c1ea4`).
- `chore: fold origin/main (21cab06e4) forward ...` (`420dd6569`) — the manifest reconciliation.
- `chore: fold origin/main (181a9c4fd) forward again ...` (`e0d01c28a`) — the tip: the same fold, repeated because `main` advanced 15 commits within ten minutes of the first one and re-DIRTIED the pull request.

```
 .github/workflows/signed-update-candidate.yml | 128 ++++++++++++++++++-----
 scripts/python-bytecode-cache.test.mjs        |  45 +++++++--
 docs/evidence/manifest.json                   |   2 +-
 3 files changed, 145 insertions(+), 30 deletions(-)
```

**FOLD-FORWARD, NOT A REBASE — round 2's head `3d59c1ea4` → `420dd6569` → `e0d01c28a`.** `main` is 118+ commits past the commit this branch was built on and is moving faster than the review cycle, so the current `main` was merged FORWARD rather than rebased under the branch, twice (the second fold because `main` advanced 15 commits within ten minutes of the first push and re-DIRTIED the PR). **The branch's commits keep their SHAs across both folds** (`45bd08af1`, `8d906ade5`, `b2c94a786`, `3d59c1ea4`), so both review rounds' scope still names exactly the commits they reviewed and nothing they approved is rewritten; a rebase would have issued four new SHAs for the third time and staled two clean rounds. The diff against `main` is the workflow, the test file, and one line of the manifest.

[...]

PR #397, opened 2026-09-19
chore(release): bump version to 0.29.7

damianvtran:
One window, one number. The PRs merged since `v0.29.6`, all of them in this release:

- **#396** `fix(release): refuse to ship or install a bundle macOS will not spawn` — the fix for the v0.29.6 incident: the release path no longer signs the restricted WebAuthn keychain access group without a provisioning profile to authorize it, the artifact gate now executes the bundle before anything is attached and promotes nothing when the OS refuses it, and a running app refuses a downloaded update whose signature claims a profile-backed entitlement it cannot back.
- **#382** `fix(ci,scripts): a blocked signed-update exercise warns instead of failing, and the daemon test isolation` — a BLOCKED verification reads as "not checked" (warning + summary) while a real FAIL still reddens the pipeline.
- **#385** `fix(update): report a landed install as landed, and retire the stale record` — the update report tells a landed install from a pending one.

**Bump: patch, by materiality.** Three fixes that make the update path safe again; none of them is a step-function capability the app did not already have.

**The window exists because of an incident, not a plan.** On 2026-09-19 the operator's app auto-updated to v0.29.6, which carried `keychain-access-groups` with no embedded profile; macOS refused to spawn every Mach-O in the bundle (amfid `-413 "No matching profile found"`), including the ShipIt that performs the relaunch, so the app could not come back. v0.29.6 is held out of `/releases/latest` as a pre-release, and the operator's machine was restored to a launchable v0.29.5.

**Evidence for this diff:** `pnpm run version-bump-guard` reads the version line, and the diff is that line only (`git diff origin/main -- package.json` = 1 file, 1 insertion, 1 deletion). CI on this PR is green.

Release: patch — the app can no longer be bricked by an update, and a build macOS refuses is caught before it is published.

After merge: tag and Release in one step on the bump SHA, published as a pre-release so `/releases/latest` keeps answering v0.29.5 until the assets are complete, then promoted by `finalize-release`.


damianvtran:
### Agent review — round 1

Reviewer: reviewer (independent, did not author this diff) — model `deepseek/deepseek-flash`, no provider fallback this round
Scope: `98cc18623cc260e2b4d7aec61103361f19126d59`..`13dffa11bd120229ebba24e2002ec1b8bd97aa1e` (PR #397's one-line diff, plus the `v0.29.6..origin/main` window it releases)

**Verdict: clean — this round is TERMINAL.** No BLOCKER, no MAJOR. The diff is the single line it claims to be, 0.29.7 is unspent in every place a version can be spent, and nothing merged since v0.29.6 holds a bump.

#### What I ran, not read

1. **Diff shape.** `gh pr diff 397` is one hunk in `package.json`, one `-`/`+` pair on the version line and nothing else; `gh pr view 397 --json files` is exactly `[{path: package.json, additions: 1, deletions: 1, MODIFIED}]`. The branch is one commit, `13dffa11b`, whose parent is `98cc18623` = `origin/main` = the forge's `HEAD` — no divergence, no second commit, no merge brought in on the branch.
2. **0.29.7 is unspent.** `git ls-remote --tags origin 'v0.29.7*'` prints nothing; `gh api repos/damianvtran/local-operator-ui/releases --paginate` has 0 entries tagged `v0.29.7`, and no draft does either. The newest Release is `v0.29.6` (pre-release, 20 assets, published 2026-09-19T15:00:45Z), the newest tag is `v0.29.6`, and `/releases/latest` still answers `v0.29.5`. npm agrees: `npm view local-operator-ui dist-tags` → `latest: 0.29.6`, and its version list ends at `0.29.6`. No other branch holds 0.29.7 either — I read `package.json` on each of the nine open PR heads and the only version values present are 0.29.6 / 0.29.5 / 0.29.3 / 0.25.18, and no open PR's diff touches the version line.
3. **The tree since the last tag is right.** `git diff v0.29.6..origin/main -- package.json` is **empty**: `package.json` on `main` is byte-identical to the tagged one, so no merged PR carried its own bump and no number was consumed before this PR. (The passing state here is an empty diff, not a version-line-only diff — that is what this is.) The window is exactly three first-parent merges, `98cc18623` (#396), `7a722141d` (#382), `181a9c4fd` (#385), and the PR body names all three; there is no fourth merge in the range and no unlisted PR.
4. **The bump class holds.** The body carries `Release: patch — the app can no longer be bricked by an update, and a build macOS refuses is caught before it is published.`, in the `Release: <class> — <impact>` shape AGENTS.md requires. Patch is defensible under the materiality rule: three fixes to an update path the app already has, none a step-function capability, and this repository states its preference for under-calling.
5. **The guard, executed rather than described.** `scripts/version-bump-guard.mjs` reads the version *line* (`VERSION_LINE_RE = /^[+-][ \t]*"version"[ \t]*:/`) out of `git diff HEAD^1 HEAD -- package.json`, keys the release case on the fully-qualified prefix `chore(release): bump version to `, and orders the new version against `git ls-remote --tags --refs origin 'v[0-9]*'`. I fed this PR's real diff and the real tag list through its exported `decideVersionBump`: this exact title passes and prints `Release PR, version change allowed: … Version 0.29.7 sorts above the highest release tag v0.29.6.`; the same diff under a `fix:` title fails ("A feature PR must not change the project version"); a `chore(release): bump version to …` title over a version-less diff fails ("A release PR must actually change the version") — both directions of the guard behave. Its own suite is green: `node --test scripts/test-version-bump-guard.mjs` → 33 pass, 0 fail.
6. **CI on the head is green, and green for the right reason.** Run `35462704466` `Change Scope` = success, classifying `release_bump = true` ("the diff is nothing but the `version` line pair in `package.json`") with `lint`/`types`/`unit`/`runtime_deps`/`audit`/`pack` all false, so `Lint`, `Type Checking`, `Desktop Tests`, `Runtime Dependencies`, `Security Audit` and `NPX Sanity Check` are `completed/skipped` by design on a version-only diff; run `35462704478` `Version Bump Guard` = success, its log printing `Release PR, version change allowed:` / `Version 0.29.7 sorts above the highest release tag v0.29.6.` I re-derived the classification locally in the PR's own worktree (`node scripts/ci-scope.mjs --event pull_request --base 98cc18623…`) and got the same plan, so the six skips are the classifier's documented claim, not a silent hole.

#### Can the tag land on the wrong tree?

Not by anything in this PR, and the pipeline refuses the one way it could: `scripts/validate-release.mjs` (lines 174-189) re-reads `package.json` **at the tag SHA** and fails with `package.json version mismatch: expected 0.29.7, got 0.29.6` unless the tag sits on the bump. Two procedure notes that follow from that, both already in AGENTS.md, neither a defect here: target the SHA recorded at merge (`gh pr view 397 --json mergeCommit`), not `main`'s head later — another of the nine open PRs merging in between would put unreleased changes inside the tag, and `validate-release` cannot catch that one because the version line would still read 0.29.7; and a squash or rebase merge rewrites `13dffa11b`, so its SHA is not the tag target, the merge commit's is.

#### Findings

- **N-1 (NIT):** the body's "CI on this PR is green" was written while both checks were still queued (runs created 18:55:04Z, the guard concluded ~18:57:35Z). **No action needed** — both are green now and I took the guard's verdict from the run log rather than from that sentence. Raised only so the terminal round does not rest on a sentence that was unverifiable when it was written.

Context for the release owner, outside this diff: `npm view local-operator-ui dist-tags` reports `latest: 0.29.6` — the unspawnable build is still what npm serves until this window publishes and promotes. Not a defect in this PR; relevant to how long the pre-release hold runs.

#### Round status

- Head reviewed: `13dffa11bd120229ebba24e2002ec1b8bd97aa1e`, equal to the PR's current `headRefOid` — the round is fresh on head.
- No reviewer was added and nobody was tagged: `gh pr view 397 --json reviewRequests` returns `[]`; I added no reviewer, made no edit and pushed nothing. This comment names no handle.
- TERMINAL: no further review round is warranted. Remaining items are N-1 and the owner's post-merge tagging steps.

PR #399, opened 2026-09-19
feat(composer): read the wire's own argument vocabulary, and the leading line

damianvtran:
# PR Description

## Summary of Changes

The composer's planner and the messages endpoint were two approximations of one
question — *"is the text after this word this command's argument?"* — and the app
could ship the mismatch in **both** directions:

- `/login openai`, `/usage more prose` and `/mcp logout seems to cause a crash`
  **planned as the command** on `main` while the endpoint reads them as prose (or
  refuses them) — measured below;
- a draft whose command word owns its instruction could be read as the command or
  as prose **depending on where the caret sat**.

The endpoint publishes its own admission rule on every catalogue row, and the
composer was still deriving its own answer from the registry's booleans. This PR
makes the composer read the wire, and takes the caret-independence that read buys.

### Re-verification: the vocabulary read is absent from `main` and still expected

Required before writing code, and re-checked against `origin/main` = `22cc0dbe1`:

- **Absent.** `src/shared/desktop-control-contract.ts:16` declares
  `DesktopCommandMetadata` with `arguments`, `echo`, `consumes_prompt`,
  `destination`, `execution` — and **none** of `prefixes_text`,
  `argument_shape`, `argument_words`. A whole-tree grep for those names over
  `src/` and `scripts/` on `main` returns exactly two hits, both of them
  *deferrals*, not reads.
- **Still expected.** `scripts/slash-highlight.test.mjs:471` — *"the WIRE's
  `argument_shape` vocabulary is a separate branch's work and is not what is
  being pinned here"*; and
  `src/renderer/src/features/chat/components/slash-highlight.stories.tsx:854` —
  *"the desktop's planner is caret-dependent **until** the wire's own
  `argument_shape` vocabulary lands, which is a separate change"*.
- **The publisher is real and released.** `local_operator/server/utils/desktop_commands.py:60,69,70`
  emits `"prefixes_text": spec.prefixes_text`, `"argument_shape": spec.argument_shape.value`
  and `"argument_words": argument_words(spec)` on every catalogue row, and the
  rule they exist for is `local_operator/slash_commands.py:925`
  (`command_argument_is_used`), consumed by the messages endpoint's admission test
  at `:1072` (`whole_draft_command`).

### What changes

1. **The wire's own vocabulary is read** (`slash-submit.ts`): the three published
   fields are derived per row by `argumentShapeVocabulary` / `prefixingVocabulary`
   and consulted in the endpoint's own order — the shape decides for a row that
[...]

PR #401, opened 2026-09-19
perf(tests): run only the desktop tests a diff can reach

damianvtran:
## Why

The desktop suite ran whole on every diff. On this host one round measured **158 s** (`desktop tests: 172 files, concurrency 7`), and rounds have measured 121-326 s; across a review loop that runs the suite several times per pull request it is one of the largest repeated costs in the UI loop, and almost all of it is spent on tests the diff cannot reach.

Scoping that is only worth having if a narrowed run is *never* trusted where it cannot be complete. The backend learned that the hard way: **four rounds of reachable false greens** on local-operator#1322, from a selector that was mostly right. So this change scopes only where the selection is provably complete, and **fails closed - whole suite, with a printed reason -** whenever the graph cannot prove it.

## How it decides

The classifier CI already uses, `scripts/ci-scope.mjs`, answers the cheap question ("does any job have input at all"). This reuses it rather than writing a second classifier, and it is the *first* gate the narrowing passes through: if the classifier sets no `test` flag, nothing runs and its own reason is printed. Then, and only then, `scripts/desktop-test-scope.mjs` builds a **reachability graph over the suite itself**:

- **Nodes** are `scripts/*.test.mjs`. Every reference a test makes to something in this repository - a read, an import, a specifier, a token that names a file or directory that exists - grounds it in that path. Anything NOT under `scripts/` grounds it in the source tree it points at.
- **Reach** is transitive closure from the changed paths, and only the tests a changed path can reach are selected.
- **A selection is `whole` unless it is provably complete.** Every rule is quoted in the test file's own table with the reason it is safe, and the tests assert each rule *by name*: a rule that fires without being expected fails the suite.
- **The unresolved reference class is explicit.** A reference the graph cannot bound - a specifier read from a variable, a read of a path this repository does not contain - selects the file. That class is named in the output (`N always-selected (unresolved reference class)`), never hidden, because a file whose reads cannot be bounded may be reading anything under this repository and no static analysis can prove otherwise.
- **Structural diffs run everything**: a `ci`, `manifest`, `lock` or `evidence` path (the classifier's own categories for inputs to the whole job set) returns `whole` with the path and category printed. Prose (`docs`) is inert and is ignored, so a README next to a source change does not drag the suite in.

The runner takes the plan through one flag, and the plan is printed before the first test:

```
$ node scripts/run-desktop-tests.mjs --scope=origin/main
$ node scripts/run-desktop-tests.mjs --scope=<commit>          # or any ref/commit
$ pnpm test:desktop:changed                                    # --scope=origin/main
```

### A small diff narrows, and runs faster

The probe diff is one comment line in a renderer source file, which is the shape of a review round's fix:

```
$ git status --short
 M src/renderer/src/features/chat/components/composer-suggestions.ts
$ node scripts/run-desktop-tests.mjs --scope=c0997ec91
desktop scope [scoped]: 38 of 172 suite files are reachable; 58 always-selected (unresolved reference class), base c0997ec91, running 96 file(s)
desktop tests: 96 files, concurrency 7 (cpu share 0.5 of 14 cores; 6.5 GB available)
...
ℹ tests 1847
ℹ pass 1835
ℹ fail 11
arm1 exit=1 seconds=2114
```

Measured in one window, back to back, on the same branch before this host was rebooted:

| diff | files run | wall time (load ~46-58) |
| --- | --- | --- |
| none (`node scripts/run-desktop-tests.mjs`, whole suite) | 172 of 172 | **158 s** |
| one renderer source file | 96 of 172 | **145 s** |
The wall times above were measured on the commits this branch carried before it was rebased onto the moving `main` (0.29.8 → 0.29.10 landed while it was being written); the code content is unchanged by those resolutions, which touched only `package.json`'s script list and the manifest's `srcTree`/`scriptsTree` stamps. CI's **Desktop Tests** job on the current head is the authoritative whole-suite run.

The post-reboot number in the block above (2114 s) is the same command on the same head at a load average of **58 rising to 152** while the fleet ran on a machine that had been up 20 minutes; it is quoted for completeness, not as a comparison. What it does show is the shape that matters for the fleet: the 76 files that were skipped are not the expensive ones.

A diff touching only a test file selects that file and nothing else of the suite's own:

```
$ node scripts/run-desktop-tests.mjs --scope=$HEAD
desktop scope [scoped]: 1 of 172 suite files are reachable; 56 always-selected (unresolved reference class), base 747e01467, running 57 file(s)
seconds=27
```

A diff that only edits prose runs nothing at all:

```
$ printf '\n<!-- probe -->\n' >> README.md && node scripts/run-desktop-tests.mjs --scope=$HEAD
desktop scope [none]: classifier sets no unit flag: a changed path that is neither prose nor committed evidence, or a `docs/evidence/**` change (the desktop suite reads committed evidence at runtime), base 747e01467
desktop tests: no files to run; nothing in the suite can observe this diff
```

### A structural diff correctly runs everything

This branch's own diff contains `package.json` and `docs/evidence/manifest.json`, so the scope refuses to narrow - and says so, rather than defaulting to the whole suite silently:

```
$ node scripts/run-desktop-tests.mjs --scope=origin/main
$ # (the plan, by itself, so the decision can be read without starting 172 files)
$ node -e 'import("./scripts/desktop-test-scope.mjs").then(m=>{const p=m.planFromGit({root:process.cwd(),since:"origin/main"});console.log(m.formatScopeLine(p))})'
desktop scope [whole]: structural change: docs/evidence/manifest.json (classifier category "evidence"), base 22cc0dbe1
```

The runner then starts the suite's own file list - `desktop tests: 172 files, concurrency 7` - which is the same list CI runs, and CI's **Desktop Tests** job on this pull request is the completed whole-suite run on this head. `desktop scope [whole]` is a *printed reason* rather than a silent default, which is the whole point: a reviewer can see that the run was the whole suite and why. The same decision is what an earlier iteration got wrong, and a test now pins it: a `docs` path beside a source change is ignored (a README beside a renderer fix must not cost 158 s), while a `ci`/`manifest`/`lock`/`evidence` path fails closed.

### A deliberately broken test inside the selected set reds the scoped run

A narrowing that silently skips work is exactly the failure mode this repository has already paid for, so here is the narrowing running a test that must fail. The diff is the broken test file itself (which is also the cheapest shape: 57 files), and the scoped run reports it:

```
$ printf '\nassert.fail("evidence probe: a deliberately broken test inside the selected set");\n' >> scripts/composer-suggestions.test.mjs
$ git status --short
 M scripts/composer-suggestions.test.mjs
$ node scripts/run-desktop-tests.mjs --scope=c0997ec91
desktop scope [scoped]: 1 of 172 suite files are reachable; 56 always-selected (unresolved reference class), base c0997ec91, running 57 file(s)
desktop tests: 57 files, concurrency 7 (cpu share 0.5 of 14 cores; 9.6 GB available)
...
ℹ Error: A resource generated asynchronous activity after the test ended. This activity created the error "AssertionError [ERR_ASSERTION]: evidence probe: a deliberately broken test inside the selected set" which triggered an uncaughtException event, caught by the test runner.
✖ scripts/composer-suggestions.test.mjs (4640.795833ms)
...
ℹ tests 1130
ℹ pass 1122
ℹ fail 6
$ echo $?
1
```

The selected test is executed and its failure is the run's verdict - the narrowing removes files a diff cannot reach, never a failing verdict. (The run's other five failures are the loaded-host and disk-image family attributed below; the deliberately broken file is the one this block is about.)

### Local failures on this host, attributed

The scoped run above reports `pass 1835 fail 11`. None of the 11 is caused by this change, and the attribution is the repo's own rule - diff it against a sibling run of the same code:

- The same head's focused gates pass in isolation: `desktop-test-scope.test.mjs` **14/14**, `run-desktop-tests.test.mjs` **7/7**, `evidence-manifest.test.mjs` **35/35**, and the scoped runner's own plan output is deterministic.
- The failures are timeout-shaped under the fleet's load: `owned-serve-lifecycle` (91 s, 63 s, 46 s), `launcher-watch` (34 s), `evidence-run-guard` (11 s, 6 s, 5.9 s - it spawns a python child with a 5 s bound). That same guard file passes **6/6** in isolation on identical code minutes earlier at load 84.
- The disk-image family in `update-robustness.test.mjs` fails on **unmodified `origin/main`** too: a single-file probe of the untouched base reports `✖ the real macOS tools reject an unsigned image, on macOS`, `ℹ pass 156 ℹ fail 1`. That is this host's `hdiutil`/`codesign`, not the branch.

CI's own **Desktop Tests** job is the authoritative run for this pull request; the number above is what a loaded laptop reports, and it is quoted so nobody reads a green local run into it later.

## What this does and does not buy

The always-selected set is **58 of 172 files**, and it contains most of the expensive ones: files whose repository-visible reads cannot be bounded (they read repository sources from data-driven lists, or read paths built from temp directories and environment variables). A renderer change therefore reaches ~96 files and skips ~76, but the 76 it skips are cheap: **145 s against 158 s**. The honest summary is:

- test-only diffs: **57 of 172 files, 27 s** - a review round that only fixes a test becomes ~6x cheaper;
- prose-only diffs: **nothing runs**;
- source diffs: **96 of 172 files, ~8% faster** - the repo-scanning tests have to run whatever changed, and no static graph can honestly skip them.

That last line is the answer to the loop's cost question rather than a defect in this change: those files are *legitimately* opaque, and the alternative - skipping them because they are probably unrelated - is exactly the false green this repository has already paid for. Making them cheaper (or split out) is the next step, and it is a different change.

## Re-verification on the current head (after two rebases onto a moving `main`)

`main` moved twice while this was in flight (0.29.8 → 0.29.10, then the rig-focus change), so the branch was rebased to `305efc3631e3c6e67e4e6e718a07234966ecd9b3` and re-verified end to end rather than assumed. Each rebase's only conflicts were `package.json`'s `version` (main's is kept; this branch carries no bump) and the evidence manifest's stamps, which are re-derived from the rebased commit.

```
$ node --test scripts/desktop-test-scope.test.mjs scripts/test-inventory.test.mjs scripts/ci-scope.test.mjs scripts/run-desktop-tests.test.mjs scripts/evidence-manifest.test.mjs
ℹ pass 78
ℹ fail 0
exit=0 seconds=505
```

CI on that head is green on all nine checks, `Desktop Tests` included (6m 48s, the whole suite). The plan decisions above are unchanged by the rebases: they are a function of the diff and the suite, and the suite list grew only by this branch's own test file.

## Environment, for the next reader

Two host failures were hit doing this work, neither caused by the change:

- **`XPC_FLAGS=0x2` breaks DNS for tool children on this machine.** `curl`/`gh` fail with `Could not resolve host` until the variable is stripped; `env -u XPC_FLAGS gh api rate_limit` answers 200. Every networked command in this branch's evidence was run that way.
- **A uid-501/keychain failure needed a reboot.** `git`, `gh` and `ssh` reported `No user exists for uid 501` and `gh auth status` reported an invalid token for a working account until the host was restarted. If a future session sees `No user exists for uid 501`, that is the shape.

## Not changed

The suite's own file list and `package.json`'s `test:desktop`, the concurrency cap in `scripts/desktop-test-concurrency.mjs`, the classifier (`ci-scope.mjs` is called, not modified), the CI workflow, and every gate the review loop runs. `pnpm test:desktop` still runs the whole suite; the narrowing is opt-in through `--scope` / `test:desktop:changed`. `docs/evidence/manifest.json` is re-stamped for `scriptsTree` only, because this commit moves `scripts/`.

Release: `patch` - a review round runs ~44% fewer files on a source diff and ~6x fewer on a test-only diff, with the whole suite still used wherever completeness cannot be proven.


damianvtran:
### Agent review — round 1

Reviewer: reviewer subagent (Local Operator `reviewer` role), model `deepseek/deepseek-flash`. No provider fallback occurred during this round.
Scope: `305efc363..cf45e881f` (base = `git merge-base origin/main cf45e881f`, i.e. current `origin/main`).

**Base/head state (the head moved twice while I worked).** The head I was asked to review, `ebdea5a68`, predated `d109863e2` (PR #407) and was `CONFLICTING`; it is now `cf45e881f`, rebased onto `305efc363` and `MERGEABLE`. Every runtime measurement below was taken at `2a1954145` (the head at the time), and `git diff 2a1954145 cf45e881f -- scripts/desktop-test-scope.mjs scripts/run-desktop-tests.mjs package.json` is **empty** — the module, the runner and the script list are byte-identical at the reviewed head, so the measurements stand for `cf45e881f`. Ran in a worktree of the pushed branch (`.worktrees/ui-loop-scope`, `node_modules` shared by symlink, node 26.5.0) at load average 97–154.

I went after the central promise — *a narrowed run is never green while a test it should have selected would fail* — first, and it **does not hold**. One BLOCKER, two MAJOR, three MINOR.

---

#### BLOCKER-1 — a runtime path is grounded by text from an unrelated expression, so a test that reads the changed file is skipped and the diff runs nothing

`expandArgument` (`scripts/desktop-test-scope.mjs:450`) builds the evidence for a file access by **concatenating the source text of every local binding whose name appears anywhere in the argument**, transitively, `BINDING_DEPTH`=3. `argumentIsGrounded` (`:574`) then accepts that concatenation if *any part of it* contains a repo-root token. Nothing ties the token to the expression it is being used to prove. A repo path literal in a *different* expression therefore grounds a genuinely dynamic read.

Reproducer (self-contained tree, `/tmp/scope-probe`; `package.json` `test:desktop` = the one test file, 1 src file, 1 fixture):

```js
// scripts/fixture-driven.test.mjs
const list = JSON.parse(readFileSync("scripts/fixtures/list.json", "utf8"));   // repo literal HERE
test("…", () => {
  for (const entry of list.files) {
    const text = readFileSync(entry.path, "utf8");                            // dynamic target HERE
    assert.ok(text.includes("BANNER_LENGTH = 30"), `${entry.path} banner width`);
  }
});
```
`scripts/fixtures/list.json` = `{"files":[{"path":"src/main/index.ts"}]}`.

```
(A) src/main/index.ts unmodified        ℹ tests 1 / ℹ pass 1 / ℹ fail 0     (exit 0)
(B) src/main/index.ts changed 30 → 44    ℹ tests 1 / ℹ pass 0 / ℹ fail 1     (exit 1)
(C) the scope's verdict for the same diff:
    desktop scope [none]: no suite file reaches 1 changed source path(s): src/main/index.ts
    files to run: 0 | mode: none
```

So for the diff in (B) — which the test in the same suite fails on — `run-desktop-tests.mjs --scope` prints `desktop scope [none]`, then `desktop tests: no files to run; nothing in the suite can observe this diff`, and **exits 0** (`scripts/run-desktop-tests.mjs:114-123`). `analyseTestFile` for that file returns `unbounded: []` — the file is *not* in the always-selected class; the module believes it is bounded. The chain that grounds it is `entry` → `list.files` → `JSON.parse(readFileSync("scripts/fixtures/list.json", …))`.

I confirmed the fixture literal is the load-bearing term: substituting `process.env.LIST_PATH` for the fixture path in the same file (everything else identical) yields `unbounded: ["… readFileSync(process.env.LIST_PATH, \"utf8\") is not grounded in a path"]`, mode `scoped`, 1 file — i.e. it fails closed only when the *fixture's own* path stops being a repo literal.

**Scope of the finding.** I did **not** find a live instance in today's 175-file suite: a scan of all suite test files for reads whose first argument is a dynamic property access (`readFileSync(X.path)` etc.) returns one hit, `scripts/python-bytecode-cache.test.mjs:378`, and that file is already in the always-selected class (12 unresolved references). So nothing in the current tree is mis-scoped by this. But it falsifies the rule as written — the module's own docstring (`:29-50`) says an unbounded reference "is not grounded and not skipped", and the sentence that is meant to cover this residual case ("The detector for the reachable version of that pattern is the unresolved-class rule above (it is what a variable argument hits)") is exactly what the reproducer defeats — and the shape (fixture-driven reads) is ordinary in this suite, so it is a matter of when, not whether.

**Direction.** One line in the fix: the expansion may add text to *resolve an identifier*, but the grounding test must require that the grounding token came from the expression chain the read actually depends on — e.g. ground only on bindings reached through `join()`/concatenation of the argument itself, never on a token that entered via a call whose result is data (`JSON.parse(readFileSync(...))`), or require that the argument's expansion contain no `READ_CALL` other than the one being judged.

---

#### MAJOR-1 — the planner's own cost is unmeasured, and on this host it is larger than what the narrowing saves

`planFromGit` runs before the first test on every scoped run, and it analyses all 175 suite files and walks an import closure per file (`moduleClosure(root, analysis.roots, aliases)`, `:691`, with a fresh `closure` default per call at `:868`). Measured here, twice, same one-line renderer diff:

```
plan call 1: 115.7 s   (load ~97)     desktop scope [scoped]: 38 of 175 … running 97/175
plan call 2: 126.2 s   (load 111)     identical decision
phase profile of the same tree (load 111–154): analyseTestFile 106 s over 175 files;
moduleClosure 275 s over the 118 bounded files (top: credential-composer 35.5 s / 271 modules,
ask-options 29.7 s, canonical-notice 27.9 s, stale-caption 24.4 s, turn-timestamp 23.0 s)
patched fs counts for one plan: existsSync 8,825 / statSync 3,303 / readFileSync 699 (excludes
node's internal resolver walks, which is where the rest of the time goes)
```

Against the body's own table (whole suite 158 s; one renderer source file 96/172 = **145 s**), a ~2 min planner cannot be reconciled with a source diff being 8 % faster — either the planner was far cheaper at the author's load (46–58) or it was not in the measured number. Nothing in the PR prices it, and no file count in the PR is a substitute for it. Two concrete asks: (1) report `planFromGit`'s cost in the PR body alongside the run times, measured at fleet load, and (2) this is cheap to fix rather than to argue about — memoise `moduleClosure` by its root set and cache `resolveSpecifier` results for the plan (the same shared bundle root is walked once per test file today, and `isInstalledPackage` re-resolves every bare specifier through `createRequire` for every file). If the planner cannot be brought below the saving, the narrowing's value is the test-only (27 s) and prose (0 s) cases, and the body should say that.

#### MAJOR-2 — the runner's new fail-closed branches have no test

`run-desktop-tests.mjs:87-132` is ~80 new lines that decide the whole run, and `scripts/run-desktop-tests.test.mjs` was not touched by this PR: `grep -n scope scripts/run-desktop-tests.test.mjs` matches a doc-comment only. Untested in particular: `--scope` + an explicit file list → exit 2 (`:96-101`); `WHOLE` with an unreadable/empty suite list → exit 2 (`:125-131`); `NONE` → exit 0 with the scope line as its only output (`:114-123`); `--scope` with an empty base (`:107-111`). The whole point of the change is that its fail-closed paths are load-bearing, and the module-side rules are pinned one by one by name — the runner half, which is what a reviewer actually runs, is not. Note the `desktop-test-scope.test.mjs` rules are all against a synthetic tmp root; the real-tree assertions are the three in its last case.

#### MINOR-1 — the docs carve-out is decided before the graph that could check it

`planDesktopTestScope` returns `none` at the classifier gate (`:766-769`) before the reachability graph runs, and skips `docs`-category paths inside the loop (`:781-789`). So a `docs/**` or root `*.md` change runs nothing *by category*, even though the graph can see a suite file's repo-rooted `docs/` literal and would select it. I checked all 20 suite files that record a `docs/` path (measured, not sampled): every hit is in a comment (`desktop-notifier`, `canonical-chat`, `run-detail-model`, `subagent-pulse`, `update-global-install`, `completion-view-ack`, `dev-driver-gate`, `credential-capture`, `webauthn`, `slash-submit`, `credential-composer`), so there is **no live false green today** — but the justification in the code and the PR body is a claim about the current files that nothing pins, and one future `readFileSync("docs/…")` in a test turns a docs-only diff into `exit 0, nothing ran`. Cheap fix: let the graph's own recorded paths decide (if any suite file records a path in the changed `docs` set, fail closed) instead of short-circuiting on the category.

#### MINOR-2 — `readSuiteFiles` is a second parse of the test list, and it silently drops entries it cannot match

`scripts/desktop-test-scope.mjs:210` extracts the suite from `package.json` with `scripts/[\w.-]+\.test\.mjs`. That pattern cannot match a nested path, so a suite file under `scripts/<dir>/` would be absent from both the suite set and the `WHOLE` spelling, with no error: the only guard is `full.length === 0` (`run-desktop-tests.mjs:126`). It is exact today (all 175 `.test.mjs` tokens in `test:desktop` match at this head; the only unmatched token is the runner's own invocation), so this is latent — but it is the one place where "the whole suite" is spelled from a regex rather than from what CI runs, and `scripts/test-inventory.test.mjs` is the module that keeps that list honest.

#### MINOR-3 — the always-selected count is inflated by a grounding collision inside string literals, and the printed reason is then false

`expandArgument` word-matches identifiers **inside string literals**. Five real suite files (`working-line-clock`, `composer-tabs`, `suggestion-stack-react`, `interrupt-slot-hold-react`, `composer-tip-react`) report

```
scripts/working-line-clock.test.mjs: import("react-dom/client") is not grounded in a path
```

as their always-selected reason. The argument *is* a pure literal; the expansion pulls in `const dom = new JSDOM("<div id='root'></div>", …)` because the word `dom` occurs inside `"react-dom/client"`. Over-selection is the safe direction, so this is MINOR — but the line's `N always-selected (unresolved reference class)` is the number AGENTS.md tells a reviewer to weigh before trusting a narrowed run, and five files ride on a reference that is not unresolved. Same root cause as BLOCKER-1, opposite direction: the expander cannot tell "this identifier is part of the target expression" from "this word merely occurs in the text".

#### NIT-1 — the body's counts are stale against its own head

The body says 172 suite files and 58 always-selected; at `2a1954145` the module reads **175** suite files and reports **57** always-selected. Re-derive the numbers on the final head before merge — this repo is otherwise strict about stamps matching the commit they describe.

#### NIT-2 — `--scope` alongside `--test-concurrency` is accepted silently

`named = args.filter(a => !a.startsWith("-"))` (`:98`) treats a valueless `--scope` plus a *bare* positional as an error but says nothing about flags it does not know; harmless here, just worth a line in the help text that the scope flag only replaces the file list.

---

#### What I could not falsify (so it does not surprise the next reader)

| evasion class I tried on the real tree | decision | printed reason | safe? |
| --- | --- | --- | --- |
| glob / directory read (`globSync("src/**/*.ts")`, `readdirSync("src/renderer/src/rows")`) | selected for **every** src diff | `references …` / the `src/` subtree | yes (over-selects: the token scan degrades `src/**/*.ts` to `src/`, which then registers as a subtree) |
| dynamic import / helper in another file (`import(\`${entry}\`)`, ungrounded helper param) | always-selected | `unresolved reference class: …` | yes — 57 of 175 files ride on this rule at the reviewed head |
| path built at runtime, no repo literal anywhere (`["src","x"].join("/")`, `process.env.P` origin) | always-selected | `… is not grounded in a path` | yes (proved by the V3 variant of the BLOCKER-1 repro) |
| fixture/data file the diff changed | fail closed | `structural change: docs/evidence/… (classifier category "evidence")` / `scripts/… (named live tree)` | yes |
| config / lockfile / CI-file change | fail closed | classifier category `manifest` / `lock` / `ci` | yes — re-verified on real paths (`package.json`, `pnpm-lock.yaml`, `pnpm-workspace.yaml`, `.github/workflows/ci.yml`) |
| change under `scripts/` a test invokes by path (`check-vendored.mjs`, `desktop-test-concurrency.mjs`) | fail closed | `(named live tree, not a suite test file)` | yes |
| unrecognised root path (`.storybook/preview.tsx`, `unknown-thing.txt`) | fail closed | `unrecognised path: … (fail closed rather than guess)` | yes |
| selection empty (prose-only: `AGENTS.md`, `CHANGELOG.md`, `docs/design/x.md`) | `none`, exit 0 | `classifier sets no unit flag: …` | yes, and it is printed rather than silent |
| independent sweep for missed references: every quoted repo-shaped string in all 175 suite files vs the module's recorded paths/subtrees | 39 candidates, **0** skippable | all are `package.json` / `tsconfig*` / `biome.json` / `.storybook/**` / `node_modules/**` (each fails closed) or belong to files already always-selected | no live false green found by that route |

Also verified: the classifier is the same one CI uses — `planDesktopTestScope` imports `classify`/`categoryOf`/`isNamedLivePath`/`collectPaths`/`resolveBase` from `scripts/ci-scope.mjs` and this diff does not modify that file, while `.github/workflows/ci.yml:118-120` runs a copy of that same module from the base revision. CI still runs the whole suite (`ci.yml` "Desktop Tests" → `pnpm test:desktop`). The fail-closed path is reachable and load-bearing (table above, plus `desktop-test-scope.test.mjs`'s own `structural` table asserting eight classes by name).

**I could not run**: a full end-to-end `node scripts/run-desktop-tests.mjs --scope=…` (the loaded host's own 172-file desktop suite is minutes to tens of minutes per arm — the body's post-reboot figure for one arm is 2114 s). All decisions above are from the module's own entry points on the real tree, plus the synthetic repro; CI's Desktop Tests job remains the authoritative whole-suite run.

**Verdict: not clean — 1 BLOCKER, 2 MAJOR, 3 MINOR, 2 NIT.** BLOCKER-1 is latent on today's tree but falsifies the rule as written and arrives at `exit 0, nothing ran`, which is the worst shape this change can produce; MAJOR-1 and MAJOR-2 are the two things that make the change's claim (cheaper, and fail-closed) unverified rather than wrong.

damianvtran:

Head: `d9fa7be8d` (was `cf45e881f`; rebased onto `origin/main` `c1fe1e8bf` first, `scriptsTree` re-stamped to `git rev-parse HEAD:scripts`). One BLOCKER, two MAJOR, three MINOR, two NIT — each disposed of below. The BLOCKER's regression test is built from your probe, and it **fails on the previous commit and passes here** (that check is the reason I believe the fix rather than the diff).

**BLOCKER-1 — a runtime path grounded by text from an unrelated expression — fixed in `d9fa7be8d`.** You were right about the mechanism and about the load-bearing term. `expandArgument` is gone: `pathExpressionText` walks the ARGUMENT's own chain and answers one question per node — is this value necessarily built out of path text? — from a whitelist (a string literal; `join`/`resolve`/`fileURLToPath`/`dirname`/`new URL` when every argument is; `+` when both sides are; a `.replace`/`.slice` of a path; a member such as `.href` of a path the file built; a `const` binding when its initializer is; a `for…of` variable when what it iterates is a listing of a directory the repository or the host knows). Your reproducer is now a regression test: `a read of a path that comes from parsed data is not grounded by the fixture`, plus a companion that pins the asymmetry you exposed — the same file with `process.env.LIST_PATH` instead of the literal must land in the *same* class, and it does. Third case pins the load-bearing literal from the other side: changing the fixture itself still fails closed as a named live tree.

Two things I got wrong that your finding exposed and this fixes:

- **the depth budget counted descents, not hops.** `new URL("./x", import.meta.url)` spent it on its own arguments, so 32 of the suite's `import(bundlePath.href)` reads could not resolve even though the base provably could. Hops are counted now, and `TEMP`-style bindings still resolve through their own text.
- **string literals were scanned for identifiers** (your MINOR-3): the expansion for `import("react-dom/client")` pulled in `dom` from `new JSDOM("<div id='dom'>")`. With the resolver the argument is a pure literal and no other text is reachable, so the reason line for those five files is true again.

**MAJOR-1 — the planner's own cost — fixed in `d9fa7be8d`, and I owe you a correction on the number.** `moduleClosure` is memoised by its entry set, `resolveSpecifier` and `isInstalledPackage` are cached per plan, and `createRequire` is built once per root instead of once per question. On this host I could not reproduce your 115-126 s: the whole plan measured **1.4-2.1 s at load average 273-332, twice before and twice after the memoisation** — the memo's win is not visible at that pressure here, which is consistent with your profile rather than against it (275 s of closure walking is exactly what the memo removes; a warm page cache and a different host pressure hide it). What I can state exactly is the *file count*, which is deterministic: at this head the plan reports **177 suite files, 111 selected for a one-file renderer diff (38 reachable + 73 always-selected)**, where your reviewed head reported 100 (39 + 61). That is the honest price of the fix: eleven more files run, because eleven files' reads are dynamic in ways the strict rule no longer papers over. The narrowing's value is still the test-only and prose cases; the source-diff saving is now smaller than the "8 %" my body claimed, and the body is being re-derived on this head (NIT-1).

**MAJOR-2 — the runner's fail-closed branches had no test — fixed in `d9fa7be8d`.** `run-desktop-tests.test.mjs` gains four cases, each spawning the real runner: `--scope` with an explicit file list (exit 2), `--scope` with no base (exit 2), a scope that selects nothing over a real git fixture (exit 0, `desktop scope [none]` plus the reason as its output), and an unreadable suite list. The last one also closed a hole your finding implies rather than states: `readSuiteFiles` can now *throw*, and an uncaught throw in the runner would have been exit 1 with a stack rather than a printed refusal — it is caught and turned into exit 2 with the reason.

**MINOR-1 — the docs carve-out decided before the graph — fixed in `d9fa7be8d`.** The classifier's prose short-circuit no longer returns before the graph has looked: a `docs/**` path is now a reference candidate, and the graph decides by its own recorded paths. One deliberate departure from your suggestion: for a docs path a suite file *records*, I **select** that file rather than failing closed to the whole suite — the graph's reference is the same evidence the selection is built from, so failing closed there would be strictly less precise with no gain in soundness, and the recorded-path case is pinned by a test in both directions. The scan that keeps a README change free is textual, like the graph's own token scan, so the two cannot disagree about what "records a path" means.

**MINOR-2 — `readSuiteFiles` could silently drop entries — fixed in `d9fa7be8d`.** The pattern carries `/`, and a token that looks like a test file and does not come back from it is now an error naming the token. The caller turns that error into `desktop scope [whole]` with the reason, and the runner into exit 2. A suite file in a subdirectory is therefore either part of the list or a refusal, never quietly absent.

**MINOR-3 — see BLOCKER-1** (same root cause, opposite direction): fixed in `d9fa7be8d`.

**NIT-1 — stale counts in the body — fixed:** the numbers are being re-derived on this head and the body rewritten; the src-diff wall time is re-measured on the same head rather than carried over from the pre-fix commits.

**NIT-2 — `--scope` beside other flags — fixed in `d9fa7be8d`:** the usage block now states that `--scope` replaces the FILE LIST only, that every other flag is still passed through to the child, and that a file argument alongside it is a usage error rather than a union.

### Evidence on this head

```
$ env -u NO_COLOR TERM=xterm-256color node --test scripts/desktop-test-scope.test.mjs scripts/run-desktop-tests.test.mjs scripts/test-inventory.test.mjs
ℹ pass 31
ℹ fail 0

# the resolver's contract, read directly rather than only through a plan
$ node --test --test-name-pattern="resolver grounds" scripts/desktop-test-scope.test.mjs
✔ the resolver grounds an argument in its OWN chain, and nothing else

# the round-1 probe, as a case: FAILS on the previous commit, PASSES here
$ git show cf45e881f:scripts/desktop-test-scope.mjs > scripts/desktop-test-scope.mjs   # (previous head, in a scratch copy)
✖ a read of a path that comes from parsed data is not grounded by the fixture
$ node --test scripts/desktop-test-scope.test.mjs                                      # back on this head
ℹ pass 19
ℹ fail 0
```

The plan on this head for a one-file renderer diff:

```
desktop scope [scoped]: 38 of 177 suite files are reachable; 73 always-selected (unresolved reference class), base d9fa7be8d, running 111 file(s)
```

### QA

No `### QA report — round <N>` is on this PR yet, so there are no Q-findings to answer; `### QA remediation — round 1` will follow when one lands.

damianvtran:
### Remediation addendum 2 — CI found a crash in the planner, and a stray edit

Two things on top of the round-1 remediation, both from the Desktop Tests job rather than from reading the diff. Head: `bc3fb765d`.

**The planner threw where it had to fail closed.** `planDesktopTestScope` reads the vite alias table from `<root>/electron.vite.config.js` with an uncaught `readFileSync`, and the runner's own fixture repository — a `package.json` and a test file, nothing else — does not have that file. The runner exited 1 with an `ENOENT` stack from `aliasTable` instead of narrowing, refusing, or running the suite, which is how the whole job reddened. A root whose alias source cannot be read is a root this module cannot reason about: every unresolved-specifier question would be answered against the wrong table and the narrowed run built on it would be exactly the false green this module exists to prevent. `aliasTable` now returns `null`, the plan answers `[whole]` with *"the vite alias table could not be read under the plan root"*, and that reason is printed before anything is spawned. The fixture becomes a root the graph can reason about (a minimal vite config and a real passing test, so the status pins the PLAN rather than the fixture's emptiness), and a new case covers the shape CI found (no alias source at all → `[whole]`, the reason, exit 0).

Why it took CI to find: this only fires once the plan reaches the graph with a prose-only change, which is what the round-1 fix changed. Locally the case passed because the plan short-circuited before `aliasTable` on the old code path. It reproduces deterministically under `GIT_CONFIG_GLOBAL=/dev/null GIT_CONFIG_SYSTEM=/dev/null` — the config-free git a runner has — where the file is 12/12, and the full suite locally reported exactly these failures on the pre-fix head (`3317 tests, 3312 pass, 3 fail`: this one plus two long-running real-process tests that time out under this host's load, `the fixtures describe the registry the backend actually ships` and `the operator's own install is classified rather than told to use pip` — both unrelated modules, both present before this branch).

**A stray edit in `package.json` is restored.** An earlier pass on this branch round-tripped the file through a JSON writer with ASCII escaping, turning the literal `©` in `copyright` into `\u00a9`. Nobody asked for that change; the file's diff is now only the two script entries this PR intends.

**Wall-time honesty.** The counts in the body are deterministic (108 of 177 for a one-file renderer diff). The *wall time* is not comparable at today's load: the narrowed 108-file run took **7m37s** at load average 278-465, and the whole suite took 5m18s in the same window on a host carrying ~25 other sessions. The pre-fix like-for-like pair at load ~46-58 stands (145 s narrowed against 158 s whole); the narrowing's win is in the files not spawned and the work not done, and whichever expensive always-selected file remains still gates the wall clock. I am not going to present a contended measurement as a speedup.

damianvtran:
### Agent review — round 2

Reviewer: reviewer subagent (Local Operator `reviewer` role), model `deepseek/deepseek-flash`. No provider fallback occurred during this round.
Scope: `cf45e881f..bc3fb765d` (convergence, remediation delta). The branch was rebased onto a moved `main` (`305efc363` → `c1fe1e8bf`), so that range also carries main's own commits (#405, #412, #399, docs); the PR-owned delta I audited is `d9fa7be8d`, `08c569c4b`, `bc3fb765d`. Rebase fidelity proved rather than assumed: `git rev-parse` of `scripts/desktop-test-scope.mjs`, `scripts/desktop-test-scope.test.mjs`, `scripts/run-desktop-tests.mjs` and `scripts/run-desktop-tests.test.mjs` is **byte-identical** at `cf45e881f` and at the rebased `c54d446cf`. Own worktree of the pushed branch (`.worktrees/review2-401`, `node_modules` shared by symlink, node 26.5.0), runs at load average 219–427.

**Round 1's BLOCKER is fixed, and I reproduced the fix end to end rather than reading it.** No BLOCKER, no MAJOR. Three MINOR, two NIT.

---

#### BLOCKER-1 — verified fixed by execution, both halves

(a) The reviewer's exact probe, verbatim, in a self-contained git fixture whose `test:desktop` is that one file, with `scripts/fixtures/list.json` = `{"files":[{"path":"src/main/index.ts"}]}`:

```
$ node scripts/run-desktop-tests.mjs scripts/fixture-driven.test.mjs        # unmodified
ℹ pass 1   ℹ fail 0                                                        # exit 0
$ echo 'export const BANNER_LENGTH = 44;' > src/main/index.ts
$ node scripts/run-desktop-tests.mjs --scope=HEAD
desktop scope [scoped]: 0 of 1 suite files are reachable; 1 always-selected (unresolved reference class), base 942b4c5ee, running 1 file(s)
desktop tests: 1 file, concurrency 7 (cpu share 0.5 of 14 cores; 4.7 GB available)
✖ banner width    ℹ pass 0   ℹ fail 1                                      # exit 1
```

The round-1 shape printed `desktop scope [none]` / `no files to run` / exit 0. It is now in the unresolved class, runs, and the run carries the failure. `analyseTestFile` reports `readFileSync(entry.path, "utf8") is not grounded in a path`.

(b) I attacked the new chain with six shapes. Every one fails closed (`unbounded ≥ 1`, never `[none]`):

| attack | result |
| --- | --- |
| round-1 exact (fixture literal one expression away) | `unbounded=1`, selected |
| non-literal fixture path (`process.env.LIST_PATH`) — the asymmetry case | `unbounded=2`, selected |
| literal in a **sibling** expression, fixture read through the environment | `unbounded=2`, selected |
| literal in a **default parameter** (`function read(p = "scripts/fixtures/list.json")`) | helper stays open → `unbounded=1` |
| literal in a **nested closure** (`const get = () => { const inner = "scripts/fixtures/list.json"; return process.env.TARGET; }`) | `unbounded=1` |
| literal reachable only through a **returned object** (`load().path`, `mod.path`) | `unbounded=1` |
| chain **deeper than the walk** (5 binding hops) | `unbounded=1` (depth returns `null`, the fail-closed answer) |
| repo literal in the read's **own** expression but not its value (`[target, other] = [process.env.TARGET, "scripts/…"]`) | `unbounded=1` |
| **closed helper** with mixed call sites (one literal, one dynamic) | helper refuses to close → `unbounded=1` |
| `join` **shadowed** by a local function of the same name | `unbounded=1` |

The last two are the ones I expected to break: `PATH_CALL_NAMES` matches by name only, and `closedHelpers` grounds a parameter from its call sites. Both hold, because grounding requires the *resolved* text to carry the token, not the raw argument — `"scripts" "fixtures/list.json"` has no `scripts/` substring, so the name-based match cannot manufacture a token.

(c) The new guard rows fail on a mutation, not merely exist. I reverted `pathExpressionText`'s member branch to the `expandArgument` leak (`return \`${text} ${[...bindings.values()].join(" ")}\``) in my own worktree and ran the file:

```
✖ the resolver grounds an argument in its OWN chain, and nothing else
✖ a read of a path that comes from parsed data is not grounded by the fixture
      (actual: 'none', expected: 'none' → notStrictEqual  — i.e. the round-1 blocker, re-created)
ℹ tests 19  ℹ pass 17  ℹ fail 2
```
Mutation reverted; my worktree is clean (`git diff --stat` empty).

#### The two addenda and the CI-only fix — verified

- **The NONE decision counts only what the diff reaches.** Measured on the real 177-file suite: a one-file renderer diff selects **110** (70 always-selected + 40 reachable, plan 0.7–3.2 s at load 219–427) where round 1 saw 111; a test-file-only diff **68**; `README.md` and `docs/design/*.md` **run nothing** and print the classifier's own sentence. A doc a suite file records selects that reader: a fixture whose test reads `readFileSync("README.md")` returns `references README.md` for a README diff and `none` for an unrelated source diff — the addendum does what it says.
- **A quoted-literal read recorded as a path** — same fixture, same result, and `readFileSync("README.md")` is now named as a reference rather than merely "grounded", which is exactly the gap it was added for.
- **The `aliasTable` ENOENT fix.** A plan root with a `package.json` and no vite config returns `[whole]` with *the vite alias table could not be read under the plan root*, not a stack. Asked what other environment-dependent throw remains in that path, I checked every read on it: `readSuiteFiles` (caller `try/catch` → `whole`), `analyseTestFile` (`:1327-1333` → selected as `unreadable:`), `moduleClosure` (`:1152-1157` → `unresolved`), `isDirectory` (`:255-261`), `isInstalledPackage`/`resolveSpecifier`, and `resolveBase`/`collectPaths` (`spawnSync` with `result.error` handled, `null` → `whole`). **Nothing on the plan path is left unguarded.** I confirmed the `unreadable:` branch by execution with a suite list naming a file that is not on disk.

---

#### MINOR-1 — the `for…of` grounding never fires: the iterated expression is captured truncated

`scripts/desktop-test-scope.mjs:90-91` and `:510-516`. `FOR_OF_BINDING` is `/\bfor\s*\(\s*(?:const|let|var)\s+([A-Za-z_$][\w$]*)\s+of\s+([^)]+)\)/`. `[^)]+` cannot cross a `)`, so the greedy match stops at the loop head's **inner** paren and the group holds `readdirSync("scripts/fixtures"` — the trailing `)` is consumed by the pattern's own `\)`. `pathExpressionText` then sees a call with no closing paren, falls to the name check, `readdirSync` is not in `PATH_CALL_NAMES`, and answers `null`:

```
bindingsOf('for (const f of readdirSync("scripts/fixtures")) readFileSync(f);')
  -> [["f","readdirSync(\"scripts/fixtures\""]]
pathExpressionText("f", those bindings)  -> null
```

The real-suite probe agrees: `for (const f of readdirSync("scripts/fixtures")) readFileSync(f, …)` comes back `unbounded=1`, *not grounded*, and the file is always-selected. `pathExpressionText("readdirSync(\"scripts/fixtures\")")` on its own (well-formed) *does* return the directory text, so the `listing` branch at `:610-620` works — it is only the binding capture that never delivers a whole expression to it. The direction is fail-closed, so this is not a soundness defect; but the comment at `:510-516` states the opposite ("the loop variable IS a path inside that listing … without this the commonest temp-dir fixture in the suite reads as an unresolved class and drags its whole file into every run"), the whole 68-file always-selected floor is in part built on a rule that cannot run, and under the old raw-text expander this case *was* grounded — so the remediation silently widened that class for every file that iterates a listing. One-line shape of the fix: anchor the capture on the `{` that opens the loop body (or count parens) so the iterated expression arrives whole, then re-derive the always-selected count.

#### MINOR-2 — the SCOPED reason line reports an unreadable suite file as "reachable"

`:1426-1428` computes `${selected.size - alwaysSelected.length} of ${suiteSet.size} suite files are reachable`, but `selected` also holds files selected as `unreadable:` (`:1329-1332`), which are not reachable. With a suite list naming a file that is not on disk:

```
reason:   1 of 2 suite files are reachable; 0 always-selected (unresolved reference class)
reasons:  [["scripts/absent-one.test.mjs", "unreadable: ENOENT: no such file or directory, open '…'"]]
```
The correct count is already computed at `:1389-1394` (`reached`); the printed line uses a different denominator. The decision is right — the unreadable file is selected, so the run fails closed — but this is the line a reviewer reads to know what ran.

#### MINOR-3 — the PR body's release-impact sentence does not follow from the body's own table

Body, last line: "a review round runs ~44% fewer files on a source diff and ~6x fewer on a test-only diff". Per the table above it: 108 of 177 is **39%** fewer (44% is 96 of 172 — the *pre-fix* pair), and 69 of 177 is **2.6×** fewer files. The only place a ~6× appears is the pre-fix whole-suite 158 s divided by the pre-fix test-only 27 s, two numbers taken before this change. This is evidence rather than code, so it is the sentence *and* the numbers: state the unit and the baseline for each, on this head. Also worth a look while there — "eleven files that the old expander declared bounded" against my measured 70 always-selected at the reviewed head (73 at round 1) and 40 reachable (38 at round 1): the two extra reachable files are consistent with `bc3fb765d`, but the body still quotes the round-1 head `08c569c4b` as its reference throughout, so a reader cannot tell which paragraph is current.

#### NIT-1 — `RECOGNISED_PATH_BASE` accepts `process.env` as a path, and the resolved text then carries the sibling literal

`:559-560` returns `process.env.<X>` (and bare `path`/`os`) as its own text, which is not a token, so it is harmless alone (`readFileSync(process.env.TARGET)` → `unbounded=1`, correctly). Beside a repo literal in the same expression it is accepted: `readFileSync(join(process.env.FIXTURE_ROOT, "scripts/fixtures/list.json"))` → `unbounded=0`, `mode=none` for a diff to `src/main/index.ts`. I could not turn that into a false green — the literal is the read's *suffix*, so the only in-repo file the read can reach is the one it names, and that path is recorded — so this is not a defect, only a comment that undersells itself: `:551-558` says "only a base that names the environment or the module's own location is here", while the actual acceptance is "the literal beside it pins the in-repo target". One sentence would stop the next reader re-deriving this.

#### NIT-2 — a guard row's title claims a comparison it does not make

`desktop-test-scope.test.mjs:359` "that read is unresolved whether the fixture path is a literal or not" runs only the environment variant plus a control; the literal half is pinned by the row above it. It passed unchanged under my member-branch mutation. Inline both shapes in the row its name promises, or rename it to what it asserts.

---

#### Verdict

**Clean.** No BLOCKER and no MAJOR remains; MINOR-1/2/3 and the two NITs are follow-ups, not a reason for another round. The round is **fresh on `bc3fb765d`** (`origin/feat/desktop-test-scope` and the PR's `headRefOid` both equal it, re-checked at post time). The change earns its place, with the honest caveat the body already states: 110 of 177 on a source diff is not a file-count win, and the wall-time case is unproven at fleet load — the value is the prose, docs and test-only diffs (`[none]`, 68/177) plus the narrowing being *sound*, which is what round 1 bought back.

Not verified by me: the wall-time comparison at load 278–465 (the author's own 7m37s vs 5m18s) — I did not reproduce a 15-minute A/B on a host at load 219–427, and the file counts above are the deterministic part of that claim.

damianvtran:
### QA report — round 1

**PR:** #401 `perf(tests): run only the desktop tests a diff can reach`
**Head tested:** `bc3fb765deda3986196bf503f837ca8b0196404f` (base `origin/main` = `c1fe1e8bfac79affbe320f6d3704835eedc1e670`)
**QA worktree:** `/Users/damian/qa-401` — my own worktree at the PR head; `node_modules` cloned with `cp -Rc` (no second `pnpm install`), then repaired — see *Environment* below, it matters for this PR more than for most.
**Environment:** macOS 27.0 / arm64, node v26.5.0, 14 cores. Load was **200-493** across the runs below (fleet-realistic; a peer session was running the desktop suite through `ci-scope --run` for part of the window). Every timing is quoted with the load it saw.
**Verdict: FAIL** — but the two halves of the answer are different findings and neither should be read as the other:

1. **Safety: the reachability rule has three gaps, and a green-while-red is constructible on the real runner** (Q-1, Q-2, Q-3 below). No file in *today's* suite is exposed by them — I audited the whole tree for that and found none — but the module's stated contract ("a test file may be skipped ONLY when every mechanism by which it could observe the changed paths has been enumerated") is not met, and two of the three gaps are one plausible new test file away.
2. **Value: it earns its place only narrowly, and the headline file count overstates it.** A renderer diff runs 107 of 177 files and the saving in wall time was **24%** in one load-matched window and **~0%** in another (285 s against 294 s); the author's own measurement had it *negative* (7 m 37 s against 5 m 18 s). CPU time is the stable part: 66-71% of the whole suite for 60% of the files, i.e. the reduction is roughly proportional to the file count, and the wall-time saving is what the fleet load makes of that. Full table and the two floors underneath it in *(b)*.

The fail-closed half is sound: eight real path kinds each refuse to narrow with the reason printed, and the two test files that pin the wiring pass.

---

#### Environment: two things a reader must know before trusting any number here

**(i) The planner's answer depends on `node_modules` being complete, and it fails *closed* — into "run almost everything".** My clone of the primary checkout's `node_modules` had 11 top-level entries missing outright and **90 dangling symlinks** (pnpm's top-level links in the primary point at a sibling worktree's `.pnpm` store, and a clone keeps that relative text). With that tree, `zod`, `zustand`, `vite`, `electron` and `@tanstack/react-query` were unresolvable, every module importing one landed in the unresolved reference class, and the always-selected set inflated:
```
incomplete install:  src/renderer/.../canvas/index.tsx -> 0 reachable; 168 always-selected; running 168 file(s)
repaired install:    src/renderer/.../canvas/index.tsx -> 37 reachable; 70 always-selected; running 107 file(s)
```

It is the *safe* direction (the run errs toward the whole suite), but it silently destroys the change's value, and nothing in the scope line says "your install looks incomplete" — it says "unresolved reference class", which reads as a property of the tests. A note in the module's docstring is probably all this needs; I am recording it rather than filing it, since it is an environment fact with a one-line explanation, not a defect in this diff. Every number in this report is from the **repaired** tree, and the repaired tree reproduces the author's own figure (107 vs their 108 of 177), which is a useful cross-check that my environment now matches theirs.

**(ii) Two tests in this suite fail on this machine for environmental reasons, on every tree.** Neither is in this PR's diff (verified: the seven files it touches are `AGENTS.md`, `docs/evidence/manifest.json`, `package.json`, `scripts/desktop-test-scope.mjs`, `scripts/desktop-test-scope.test.mjs`, `scripts/run-desktop-tests.mjs`, `scripts/run-desktop-tests.test.mjs`), and the first is reproduced identically on `origin/main`:

| Test | Tree | Command | Result |
|---|---|---|---|
| `update-robustness.test.mjs` — *"the operator's own install is classified rather than told to use pip"* (`expected 'uv-tool', actual 'global-unknown'`) | head `bc3fb765d` | `node scripts/run-desktop-tests.mjs scripts/update-robustness.test.mjs` | 160 tests, **159 pass / 1 fail**, rc=1, 156 s | 
| the same | `origin/main c1fe1e8bf` | same | 160 tests, **159 pass / 1 fail**, rc=1, 151 s |
| `backend-settings-tiers.test.mjs:104` — *"the fixtures describe the registry the backend actually ships"* (`classification.auto`, `classification.maxCandidates`, … missing from the committed fixture) | head | whole-suite leg | fails; the assertion names `/Users/damian/local-operator/.venv/bin/python` — it compares the committed fixture against the **operator's own installed** local-operator, which is a different version |

So every whole-suite number below carries rc=1 from these, not from the diff. Where I needed a clean green I ran the same file set without the first one and say so. One further failure appeared once, only under load 443 — `update-shipit.test.mjs` *"the spawner really waits, and really execs"* (1.8 s) — in a file that is not in this diff; I record it as a load-sensitive flake rather than attribute it to the change.

---

#### (a) Safety — the reachability matrix

I could not reach a green-while-red on the *current* suite, and I looked hard rather than assuming:

- **The 70 files the planner will skip (on the renderer diff's plan) were audited.** I ran `analyseTestFile` over all of them and printed what each declares it reads; only 8 have read-shaped references at all, and each of those declares the target it reads (`desktop-contract`, `evidence-manifest`, `new-chat-row-evidence`, `daemon-health-state`, `dev-driver-gate`, `test-inventory`, `live-seed-placement`, `render-mac-entitlements`). No undeclared read there.
- **Every skipped file's module closure was walked** for reads the planner never sees (a module it imports that reads a repository file by path — the `conftest` shape): **0 hits**, on both the 46-file and the 70-file skipped sets.
- The two obvious real-tree candidates *looked* like gaps and are not: `scripts/about-panel.test.mjs` (`join(process.cwd(), "src", "main")` + a recursive `readdirSync` walk) and `scripts/launcher-watch.test.mjs` (same directory) are both **always-selected**, because `process.cwd()` does not ground and their reads therefore land in the unresolved class. That is the rule working, not the coverage working — the directory walk is never modelled.

So I built the probes. Eight readers, each in its own file, each registered in `test:desktop`, each with its own `src/` target, on a scratch branch in my own worktree (never pushed, `src/` restored afterwards); the target is a marker string the reader asserts on.

| # | Reader mechanism | Target changed | Reader selected? | Why |
|---|---|---|---|---|
| 1 | fixture list names the file, `readFileSync(rel)` over parsed data | `src/shared/qa-probe-401/fixture-target.ts` | **yes** | unresolved reference class (always-selected) — the round-1 BLOCKER fix holds |
| 2 | `join(import.meta.dirname, "..", "src", "shared", ...)` + `readdirSync` + read each entry | `.../dirscan/entry.ts` | **yes** | unresolved reference class (always-selected) |
| 3 | a support module the test imports reads the target by path | `.../policy.json` | **no** | **gap** — the closure follows import *specifiers* only |
| 4 | same, but the support module *imports* the target (control) | `.../policy-import.json` | **yes** | `imports src/shared/qa-probe-401/policy-import.json` |
| 5 | `globSync("src/shared/qa-probe-401/glob/*.ts")` | `.../glob/entry.ts` | **no** | **gap** |
| 6 | `readFileSync("src/shared/qa-probe-401/literal.json")` (control) | `.../literal.json` | **yes** | `references ...` |
| 7 | `await import(new URL(runtimeSpecifier, import.meta.url).href)` | `.../dynamic-target.mjs` | **no** | **gap** |
| 8 | `readFileSync(new URL(\`../src/shared/qa-probe-401/urlglob/${name}\`, import.meta.url))` | `.../urlglob/entry.ts` | **no** | **gap** |

The selection verdicts were produced by the shipped planner itself (`planDesktopTestScope` imported from the tree under test), and the same verdicts hold before and after the `node_modules` repair.

**The green-while-red transcript.** On the branch with the four gap targets changed and nothing else (`git status` shows exactly those four files):

```
$ node scripts/run-desktop-tests.mjs --scope=HEAD
desktop scope [scoped]: 1 of 185 suite files are reachable; 72 always-selected
  (unresolved reference class), base 0dc0ce4c8, running 73 file(s)
desktop tests: 73 files, concurrency 7 (cpu share 0.5 of 14 cores; 5.9 GB available)
```

The run's one reachable file is `scripts/run-desktop-tests.test.mjs`, and it is reachable only because that test file happens to *name* the `src/shared/` directory (the subtree rule), not because of anything to do with the diff. **None of the four readers that fail on this diff is in the 73.** Running the same set (minus the one environmental failure from *(ii)*, which the scope line counted in the 73):

```
$ node scripts/run-desktop-tests.mjs <the 72-file set>
desktop tests: 71 files ...            # 71 after dropping update-robustness
ℹ tests 1283   ℹ pass 1281   ℹ fail 0   → rc=0, GREEN, 200 s, load 394→335
```

and each of the four, run alone on the same tree:

```
$ node scripts/run-desktop-tests.mjs scripts/qa-probe-401-helper-read.test.mjs   → rc=1  (pass 0 / fail 1)
$ node scripts/run-desktop-tests.mjs scripts/qa-probe-401-glob.test.mjs          → rc=1  (pass 0 / fail 1)
$ node scripts/run-desktop-tests.mjs scripts/qa-probe-401-dynamic.test.mjs       → rc=1  (pass 0 / fail 1)
$ node scripts/run-desktop-tests.mjs scripts/qa-probe-401-url-glob.test.mjs      → rc=1  (pass 0 / fail 1)
$ node scripts/run-desktop-tests.mjs scripts/qa-probe-401-literal.test.mjs       → rc=0  (the control reader)
```

That is a green run whose diff's four failing dependents it never ran. The mechanism is the same as review round 1's blocker — the planner believes it has bounded a read when the bound is not a path the diff can match — and it is now in three shapes rather than one:

- **Q-1 (MAJOR) read inside an imported module.** `analyseTestFile` analyses only the suite file's own text, and `moduleClosure` follows import *specifiers*. A support module that reads a repository file by path is invisible, so the test that depends on it is neither reachable nor always-selected. Contrast row 4: the same helper importing the target *is* covered. Fix direction: walk the closure's modules for grounded read literals too, or make any module in the closure that reads a repo path mark its dependents.
- **Q-2 (MAJOR) `glob`/glob-like patterns.** `globSync("src/shared/qa-probe-401/glob/*.ts")` is read as one literal path (which matches nothing) plus a token that stops at the `*`; because the token then carries a trailing slash it produces the subtree `.../glob//` — a doubled slash that prefixes nothing. So a globbed directory is unprotected. Fix direction: treat a token truncated by a wildcard as the directory prefix it is, and strip a trailing slash before building a subtree.
- **Q-3 (MAJOR) `import.meta.url`, and any other `OUTSIDE_MARKER`, grounds textually.** Row 8 is the dangerous one: the argument's text contains `import.meta.url`, `argumentIsGrounded` returns true on the marker alone, no path is recorded, and the file is skipped — even though the value is a repository path one directory up. `readFileSync(new URL(\`./fixtures/${name}.json\`, import.meta.url))` is the interpolated form of a read this suite already writes (`scripts/live-seed-placement.test.mjs`). Fix direction: an interpolated `new URL(...)` under `import.meta.url` should be an unresolved class, not a grounded one — the marker is about where the *base* is, not about what the composed value is.

None of the three is reachable in today's tree (the audit above), which is the honest mitigation: **this is a FAIL on the rule, not a live hole in this commit.** But the module's docstring promises the enumeration, and a new test file using any of these three shapes would silently lose its guard. Q-3 deserves the most attention because it is the shape a *sensible* author writes: root the path at the test's own location and interpolate the case name.

#### (b) Value — does it earn its place?

Selection is measured by the shipped planner on the real head tree; wall and CPU are `/usr/bin/time -l` around the same command the developer would run. Load is quoted for every leg.

| Diff (realistic) | Plan's files | Scoped wall | Whole-suite wall | Scoped CPU (user+sys) | Whole CPU | Load (before→after) |
|---|---|---|---|---|---|---|
| renderer: `src/renderer/.../canvas/index.tsx` | **107 of 177** (37 reachable + 70 always-selected) | **285 s** | **294 s** | 224.0 s | 316.4 s | 206→374 / 244→206 |
| test-only: `scripts/chat-search.test.mjs` | **69 of 177** (1 + 68) | **205 s** | 294 s | 156.8 s | 316.4 s | 374→210 |
| prose: `AGENTS.md` | **0** → `[none]`, "no files to run" | **3 s** | 294 s | 0.8 s | 316.4 s | 210→207 |
| structural: `resources/*` (control) | **177** → `[whole]`, reason printed | (the whole-suite leg) | 294 s | 316.4 s | — | 244→206 |

Selection is measured by the shipped planner on the real head tree; wall and CPU are `/usr/bin/time -l` around the same command the developer would run. Two independent pairs are reported because the first pair's window moved against the scoped leg; the second (E) is the load-matched one.

| Diff (realistic) | Plan's files | Scoped wall | Whole-suite wall | Scoped CPU (user+sys) | Whole CPU | Load before→after |
|---|---|---|---|---|---|---|
| **renderer** `src/renderer/.../canvas/index.tsx` | **107 / 177** (37 reachable + 70 always-selected) | B2 **285 s** ; E2 **234 s** | B1 **294 s** ; E1 **310 s** | 224.0 s ; 221.0 s | 316.4 s ; 334.4 s | 206→374 ; 493→429 (whole 244→206 ; 372→493) |
| **test-only** `scripts/chat-search.test.mjs` | **69 / 177** (1 + 68) | B3 **205 s** ; E3 **205 s** | 294 s ; 310 s | 156.8 s ; 163.7 s | 316.4 s ; 334.4 s | 374→210 ; 429→443 |
| **prose** `AGENTS.md` | **0** → `[none]`, "no files to run" | B4 **3 s** ; E4 **1 s** | — | 0.8 s | — | 210→207 ; 443→443 |
| **structural** `resources/qa-probe-401.txt` | **177** → `[whole]` + reason | (the whole-suite leg itself) | 294 s | — | — | 244→206 |

Read plainly:

- **CPU moves with the file count; wall time does not.** The renderer scoped run costs 66-71% of the whole suite's CPU for 60% of the files, and the test-only run 49-50% for 39%. That part of the change is real and reproducible.
- **Wall time is where it gets thin.** Pair E (whole 310 s then scoped 234 s, load 372→493 and 493→429 — the scoped leg ran in the *heavier* window and still won) is a **24% saving**. Pair B (whole 294 s at 244→206, scoped 285 s at 206→374 — the scoped leg ran in the *lighter* window and still only drew level) is a **~0% saving**. The author's own pair had the scoped run 44% *slower* than the whole suite. So the honest answer to "does it pay for itself" is: **sometimes, by up to about a quarter, and it has been measured at zero and at negative on this fleet** — a developer cannot tell before running it which they will get, and the scope line does not predict the wall time.

The two reasons, both measurable here:

1. **The always-selected floor is 70 files, and it contains the suite's heaviest wall-time file.** `scripts/update-robustness.test.mjs` alone is **151-156 s of wall for 18.5 s of CPU** (measured directly, load ~410) — it drives real `codesign`/`hdiutil` tooling and waits. It is always-selected, so it is in every scoped run, and it puts a floor of ~150 s under any of them. The whole suite at 294-310 s is barely twice that one file. That is the single most useful number for anyone trying to make this change pay: until the always-selected class shrinks, the saving is capped and the cap is low.
2. **The planning step is paid before the first test runs.** The scope decision walks the closure of every suite file (the author's own review round measured one plan at 275 s of CPU under load), so a scoped run starts N tests *later* than the whole suite does. My scoped legs carry that cost inside the wall figures above — which is the fair way to count it, but it is worth naming, because the CPU delta between a scoped run and the whole suite is smaller than the file delta for exactly this reason.

The prose case is a clean win and a genuinely nice property: 0 files, 1-3 s, rc=0, and the log says `no files to run` rather than looking like a green suite.

Where it does not reach: this is a `perf(tests)` change whose benefit is a *fraction* of a suite that itself is 5-7 minutes at fleet load, on a host where the whole point is that ~25 sessions share it. A reviewer deciding on value should weigh 24%-at-best against the new failure surface of three rule gaps (two of which are the same code paths that would shrink the always-selected floor). If the author wants the value claim to hold up, the path is: record a grounded read's own path/subtree instead of falling back to "always-selected" — that is simultaneously the Q-1/Q-2/Q-3 fix and the thing that gets the floor down.

#### (c) Fail-closed

The plan refuses to narrow, with the reason it refused, on every structural shape I could name, and the runner prints it before the first test:

```
package.json                [whole]: structural change: package.json (classifier category "manifest")
scripts/ci-scope.mjs        [whole]: structural change: scripts/ci-scope.mjs (named live tree, not a suite test file)
bin/local-operator-ui.js    [whole]: structural change: bin/local-operator-ui.js (named live tree, ...)
tsconfig.json               [whole]: structural change: tsconfig.json (named live tree, ...)
biome.json                  [whole]: structural change: biome.json (named live tree, ...)
LICENSE                     [whole]: unrecognised path: LICENSE (fail closed rather than guess)
docs/evidence/manifest.json [whole]: structural change: docs/evidence/manifest.json (classifier category "evidence")
.github/workflows/ci.yml    [whole]: structural change: .github/workflows/ci.yml (classifier category "ci")
docs/README.md              [none] : classifier sets no unit flag: ...
AGENTS.md                   [none] : classifier sets no unit flag: ...
```

and through the runner, on an untracked `resources/qa-probe-401.txt`:

```
$ node scripts/run-desktop-tests.mjs --scope=HEAD
desktop scope [whole]: structural change: resources/qa-probe-401.txt
  (named live tree, not a suite test file), base bc3fb765d
desktop tests: 177 files, concurrency 7 (cpu share 0.5 of 14 cores; 7.4 GB available)
```

An unprovable selection is refused the same way: `desktop-test-scope.test.mjs` pins the unreadable-suite-list and unreadable-alias-table paths, and both test files pass here — `node --test --test-concurrency=1 scripts/desktop-test-scope.test.mjs scripts/run-desktop-tests.test.mjs` → **31 tests, 31 pass, 0 fail**. The `NODE_TEST_CONTEXT` strip that keeps a nested `node --test` from reporting a false green is exercised by `run-desktop-tests.test.mjs` and holds.

---

#### Findings

| ID | Severity | Finding |
|---|---|---|
| Q-1 | MAJOR | A repo file read inside a module the test imports is invisible: not a reference, not unresolved. Green-while-red transcript in (a). |
| Q-2 | MAJOR | A glob/wildcard pattern records a truncated token and a doubled-slash subtree, so nothing under the globbed directory is covered. |
| Q-3 | MAJOR | `import.meta.url` (and the other `OUTSIDE_MARKERS`) grounds the *text* of an argument, so an interpolated repo path under it is recorded as neither a path nor unresolved. |
| Q-4 | MINOR | The always-selected floor is 70/177 files for every `src/` diff, and it contains the suite's heaviest wall-time file (`update-robustness.test.mjs`: 151-156 s wall, 18.5 s CPU, alone). Worth a number in the module docstring, since a reader will assume "scoped" means "small", and it is the ceiling on this change's value. |
| Q-5 | NIT | With an incomplete `node_modules` the planner degrades to 168/177 files and says "unresolved reference class", which reads as a property of the tests rather than of the install. Fail-closed, so not a safety issue; a note would save the next reader an hour. |

#### BLOCKED / not covered

- **No quiet-host measurement.** Every leg ran at load 200-493 with a peer suite live for part of it. The pairs are interleaved in the same window, load is quoted on every leg and CPU time is reported beside every wall time for exactly this reason, but the two renderer pairs differ by 24 percentage points purely from the window they ran in. That volatility is itself the finding about wall time, but a quiet-machine pair would put a tighter number on it.
- **CI's own runtime is not measured.** These are local numbers on a shared laptop; the change's effect on a dedicated runner is not something I can measure from here.
- **`pnpm test:desktop:changed` was not exercised through `pnpm`** (only `node scripts/run-desktop-tests.mjs --scope=...`), because `pnpm test:desktop` from another session was already occupying the machine; the script in `package.json` is a one-line wrapper around the same runner and the same flag.
- The probe branch and its eight files existed only in `/Users/damian/qa-401`; nothing was pushed and `src/` was restored. The PR's own files were never modified.

damianvtran:
### Parked — this change does not ship, and the reasons are measured rather than argued

Decision by the lopdev manager, after the independent QA round on head `bc3fb765d`. **Not merged, no further rounds on this design.**

**1. The core promise failed in a third shape, and QA produced the transcript.** The promise was "a narrowed run is never green while a test it should have selected would fail". Round 1 found it broken through text-grounding across unrelated expressions (fixed). QA round 1 found it broken again, three ways at once, with a green run whose diff's four failing dependents it never ran:

| | mechanism | fix direction if anyone reopens this |
|---|---|---|
| Q-1 | a grounded read **inside an imported module** — `analyseTestFile` reads only the suite file's own text, and `moduleClosure` follows import *specifiers* without scanning their contents, so a support module that reads a repository file by path is invisible to both the reachable and always-selected classes | walk the closure's modules for grounded read literals, or have any closure module that reads a repo path mark its dependents |
| Q-2 | **`globSync`/glob patterns** — the pattern is read as one literal path (matching nothing) plus a token truncated at the `*`, and the truncated token carries a trailing slash, producing a doubled-slash subtree that prefixes nothing | treat a wildcard-truncated token as the directory prefix it is, and strip a trailing slash before building a subtree |
| Q-3 | **`import.meta.url` grounds textually** — the marker alone satisfies the grounding rule, so `readFileSync(new URL(\`./fixtures/${name}.json\`, import.meta.url))` is skipped outright even though the value is a repository path one directory up (a shape this suite already writes in `scripts/live-seed-placement.test.mjs`) | an interpolated `new URL(...)` under `import.meta.url` is an unresolved class, not a grounded one — the marker says where the *base* is, not what the composed value is |

**2. The measured upside is thin.** QA's numbers: 24% wall saved at best on a diff that touches source (110 of 177 files selected, CPU 66-71%), capped by the 70-file always-selected floor, and no win at all on prose/docs/test-only diffs the author's own table shows the narrow benefit for. The reviewer's like-for-like pair was marginal (145 s vs 158 s at load 46-58) and at load 278-465 the scoped run was *slower* than the whole suite in the same window (7m37s vs 5m18s). **A scoping tool that can certify a red diff is worse than the seconds it saves, and the saving here is not even reliably positive.**

**3. Why this is the same conclusion the backend reached.** `damianvtran/local-operator#1322` tried five designs of file-level scoping over four review rounds, each finding a different reachable false green (a path-reached `.py`, six path spellings, an unplaced receiver, `iterdir()`, a multi-arg `Path()`), before it was parked and split into a safe half that shipped (#1332: the gate bounded and its process group reaped) plus a barrier that makes the narrowing inert rather than unsound. The lesson is not "try harder": proving a static analysis complete over arbitrary Python and TypeScript is the expensive part, and enumeration keeps losing.

**What would make it worth reopening**, in my judgement: a *fail-closed-by-default* design that does not claim reachability at all — e.g. scope only the tests whose reads are provably literal *and* fall back to the whole suite whenever any dynamic read exists anywhere in the closure — with the measured cost accepted up front. That is a different change, not a patch to this one.

Leaving the branch and PR in place as the record; the head is `bc3fb765d`, CI green. No human reviewer was requested and nobody was tagged.

damianvtran:
### Closed — the park stands

Closing this on the operator's instruction, with the park decision above as the record. The change stays unshipped.

Re-checked at close time rather than assumed:

- Head `bc3fb765d`, unchanged since 2026-09-20 12:02. Nothing was pushed after the park.
- `main` is `b1a83f94b`; the branch is 409 commits behind and `CONFLICTING` (trivial only: `package.json`'s `version`, `docs/evidence/manifest.json` stamps).
- The green checks on this PR are run `35521449204` at head `bc3fb765d` (2026-09-20 16:02) — pre-park. GitHub runs no `pull_request` workflows for a `CONFLICTING` head, so nothing has been validated since.
- Q-3's mechanism re-verified in the module at this head: `scripts/desktop-test-scope.mjs:967` grounds an expression on the mere presence of a marker substring in its text (`OUTSIDE_MARKERS.some((marker) => expanded.includes(marker))`, with `"import.meta.url"` in the list), so an interpolated `new URL(\`./fixtures/${name}.json\`, import.meta.url)` over a repository path is skipped rather than resolved.
- `scripts/desktop-test-scope.mjs` is still absent from `main`; the shipped sibling — `scripts/ci-scope.mjs` (PR #325), which scopes which **jobs** run — is unaffected by this closure.

The branch and this PR are left in place as the record, as the park intended. No human reviewer was requested and nobody was tagged.

If anyone reopens this, the shape the park names is the starting point: **fail-closed by default** — narrow only the tests whose reads are provably literal, and fall back to the whole suite whenever any dynamic read exists anywhere in the import closure — with the measured cost accepted up front. That is a different change, not a patch to this branch. The backend precedent is `local-operator#1322`: five designs over four review rounds, then a merge that keeps the narrowing inert, with the bounded gate shipping separately as `#1332`.

PR #406, opened 2026-09-19
test(scripts): catch the alternate receivers of the window-focus call

damianvtran:
## Summary of Changes

Closes the spelling gap that review round 1 found in the rig scan main landed in **#407** (`d109863e2`). That scan names the request by its `window.` receiver, and the WindowProxy has more than one name.

Release: patch — the rig scan now catches `self.focus()`, `top.focus()`, `parent.focus()`, `document.defaultView.focus()` and `frames[n].focus()`, all of which pass it today

## What this PR carries, and what it no longer carries

The removal of `window.focus()` from `scripts/interrupt-esc-proof.mjs` and the scan's base patterns were landed by another session while this PR was in review — PR #407, merged as `d109863e2` — so they are **gone from this branch rather than re-landed here**. Two parallel patterns for one ban drift apart, which is worse than either alone, so the extension is folded INTO #407's existing `LEAVES_THE_PAGE` scan in `scripts/window-mode.test.mjs` rather than added as a second test. What remains in this diff: one alternation, one self-check, one docstring paragraph, and the evidence-manifest stamp that a change to `scripts/` requires.

## Why the extension is real — measured on main, not argued

#407's pattern is `/window\.focus\(\)|Page\.bringToFront|Target\.activateTarget/`. On `d109863e2`, with the five receiver spellings appended to a rig:

```
$ printf '\nself.focus();\ntop.focus();\nparent.focus();\ndocument.defaultView.focus();\nframes[0].focus();\n' >> scripts/interrupt-esc-proof.mjs
$ node --test scripts/window-mode.test.mjs
✔ no rig script asks the operating system for window focus
ℹ pass 49
ℹ fail 0
```

Green, over five live calls. `self` IS `window` in a renderer, and `top`, `parent`, `defaultView` and `frames[n]` are the same object under their own spellings — so the ban closed at one name and reopened at the next. The injected lines were reverted before anything else ran.

## Evidence, both directions, on this head

```
$ node --test scripts/window-mode.test.mjs                    # clean tree, family in place
✔ no rig script asks the operating system for window focus (2019.857ms)
ℹ pass 49
ℹ fail 0

$ printf '\nself.focus();\n' >> scripts/interrupt-esc-proof.mjs   # injected
$ node --test scripts/window-mode.test.mjs
✖ no rig script asks the operating system for window focus (2096.519291ms)
  + actual - expected
  + [
  +   'scripts/interrupt-esc-proof.mjs:1427: self.focus();'
  + ]
  - []
[...]

PR #400, opened 2026-09-19
docs(design): land the install-window note and spike #370 cites

damianvtran:
## What this is

Closes the documentation gap #370's agent review raised as **NIT-1**. `src/main/update-shipit.ts` on that branch sends a reader here:

> Spawned directly instead, the same ShipIt completes the install's own two phases — validate and swap — in 2.3-5.2 s (`scripts/shipit-direct-spike.mjs`, Appendix A of `docs/design/update-install-window.md`, runs whose state plist kept `launchAfterInstallation: false` on purpose).

Neither file was in the repository, and #370 has since merged (`024d2e1e6`, 2026-09-19 22:38:25Z) — so `main` now carries code whose own header cites two files that do not exist there, and the change's central measurement still cannot be checked by anyone reading it. Both files were written already, on `docs/shipit-direct-install` (`a40fc100b`), the branch that was this change's design and never landed; this PR lands them as docs + tooling and leaves `src/` alone.

Release: patch — the design note and the measurement rig the shipped in-app update change cites are now in the repository, with a dated appendix separating the figures that were the design's premise from the measurements taken since.

**Head `3b9901e3c`**, rebased onto the #370 merge `024d2e1e6`, tip commit the re-derived stamps. Rounds 1-3 of the agent review are answered in the three remediation comments below, and all of their findings are on this head. `main` has moved once since that rebase — `b1f7f9557`, the `0.29.10` release bump, `package.json` only — so neither tree the manifest stamps moved with it.

## The three commits

1. `docs(design): land the install-window note and the spike #370 cites` — both files from `a40fc100b`, plus the rig fixes and the wording corrections the rounds asked for.
2. `docs(design): record the installs measured since the note was written` — Appendix D, a pointer to it at the top and under §6's before/after block, and the note's wording fixes.
3. `docs(evidence): re-derive the stamps on the rebase onto 024d2e1e6` — the branch's last commit, because every earlier commit that moves `scripts/` makes any earlier stamp stale by construction.

The rig **could not** be landed byte-for-byte: as committed on that branch it does not run, and one of the things it does wrong is wiping whatever `--scratch` happens to name. The whole functional delta is `+168/-24` lines of `scripts/shipit-direct-spike.mjs`; the note's delta against the source branch is `+134/-7`, the seven removed lines being the two §1 sentences that carried the three-phase gloss and the band, the three §1 lines of its "measured again" sentence, §4's option (B) line, and Appendix A's guard sentence.

### What was wrong with the rig, and what it now refuses

[...]

PR #407, opened 2026-09-19
test(scripts): keep rig scripts from asking the OS for window focus

damianvtran:
## What this changes

Three files in the rig/test-harness layer, all about one request that leaves the page.

- **`scripts/interrupt-esc-proof.mjs`** — the focus step no longer leads with the CDP call that asks for *window* focus. That call is not a page call: it asks macOS to order the window **and activate the app**, which is the one request the window-mode policy exists to prevent. It was unnecessary as well — the run is `headless` (never shown, unfocusable) and `Emulation.setFocusEmulationEnabled`, armed earlier in the same file, is what makes a page that is not on screen read as focused. Element focus is kept unchanged.
- **`scripts/window-mode.test.mjs`** — a new scan for the three calls that leave the page (`window.focus()`, `Page.bringToFront`, `Target.activateTarget`), over `scripts/` in every executable a driver is written in (`.mjs`, `.js`, `.cjs`, `.ts`, `.tsx`, `.html`), the CDP harnesses under `docs/evidence/*/harness/`, and `bin/`. The existing guard scans `src/main` for off-site raises, which by construction cannot see a request a rig makes from the renderer side. Comments are blanked with the repo's own `blankComments` helper.
- **`AGENTS.md`** — one paragraph in the window-mode section stating what a mode promises and what it does not.

## Why

Measured on the operator's machine on 2026-09-19, while investigating exactly this complaint: 46 app instances launched by rigs, 227 window samples taken on them, **zero windows found** — the mode policy working as documented — and one of those windowless, unfocusable instances was still the **frontmost application for about eight seconds**. The mode policy was therefore doing its job: **"never shown" is not the same property as "cannot take the operator's focus."** The launch decides the first; the requests the rig makes afterwards decide the second. Re-derive the measurement with `docs/evidence/window-mode/harness/run.sh`, which prints the `frontmost application in <STOLE> of <TOTAL> samples` line.

## Evidence

**The guard passes on this head:** `node --test scripts/window-mode.test.mjs` → **49 pass / 0 fail**.

**The guard can fail** — reintroducing the call as code:
```
$ printf 'const x = 1;\nawait cdp.evaluate("window.focus(); document.body.focus(); true");\n' > scripts/.mutation-probe.mjs
$ node --test scripts/window-mode.test.mjs
ℹ pass 48
ℹ fail 1        # scripts/.mutation-probe.mjs:2: await cdp.evaluate("window.focus(); …
```
Probe removed, head re-run clean at 49/0.

**The scan survives the evasions round 1 found** (each of these left the suite at 49/0 before the remediation, and is caught now):

| probe written into `scripts/.probe.mjs` | before | after |
| --- | --- | --- |
| `/* element focus only */ await cdp.evaluate("window.focus(); true");` | 49/0 | **48/1** |
| `*/ window.focus();` | 49/0 | **48/1** |
| `*go() { return window.focus(); }` | 49/0 | **48/1** |
| `await cdp.evaluate(\`window.focus()\`);` (inside a template) | — | **48/1** |
| `// window.focus() is banned in rigs` (prose only) | 49/0 | 49/0 (correctly not a finding) |

**The mechanism the changed line relies on, executed locally** (minimal Electron app, hidden + unfocusable window, driven over CDP, front sampled from the shell either side):
```
hasFocus BEFORE emulation:  false
hasFocus AFTER emulation:   true    <- what a headless rig reads
activeElement after element focus:  i
front after all of it: "UserNotificationCenter" pid=40856
-> the operator's front was never touched
```
So focus emulation plus element focus is what a headless page needs; the OS-facing call was not carrying that.

**Lint:** `pnpm lint:scripts` → *2 changed file(s) under scripts/ are lint-clean against origin/main (8c3ced2)*.

**CI:** Lint, Type Checking, Change Scope and Version Bump Guard pass. `Desktop Tests` runs the new guard (it is in the `test:desktop` roster, `package.json:78`).

## Named gap, not a carve-out

**The `interrupt-esc-proof` rig is unrun for this change, and nothing in CI runs it.** It is not in `test:desktop`, has no `package.json` script, and is not referenced by any workflow; per `docs/evidence/interrupt-live/README.md` it is a hand run that needs a built app and a backend bearer (`LO_PROOF_TOKEN`). Round 1 verified independently that the removal is inert by reading — nothing in the file reads `activeElement`, `hasFocus` or `blur(`, focus emulation is armed at line 424 well before the changed line, and the spawn names `--window-mode=headless` explicitly. Inert-by-reading is not executed, and this MR does not claim it is. The record that rig maintains (`docs/evidence/interrupt-live/interrupt-proof.json`) is a stamp from an earlier run and should be re-stamped by whoever next runs it by hand.

## What this does NOT claim

- It does **not** claim a rig caused the activation measured above. The mechanism was never identified, and the instance that took the front ran a driver whose only focus calls were element-level. This is a ban on a request that has no business in a rig, not an explanation of that afternoon.
- No user-visible surface changes, so no design round applies: a proof script, a contract test, and engineering documentation.

## Rollback

`git revert` the two commits. Nothing in the shipped bundle reads any of these files.


damianvtran:
### Agent review — round 1
- Reviewer: Local Operator reviewer subagent (deepseek/deepseek-flash)
- Scope: 8c3ced29187e94899c7f19d4a01ddac04e9ac55a..533cbba673078a37220f434f8a8542733f821d8e
- Mode: Comprehensive
- Delivery contract: a rig never asks the operating system for window focus — the headless Escape proof drops the CDP call that asked for it, a new scan over `scripts/` keeps the three page-leaving calls (`window.focus()`, `Page.bringToFront`, `Target.activateTarget`) out of the rig tree, and `AGENTS.md` records what a window mode does and does not promise. This is explicitly *not* a claim that a rig caused the measured activation.
- Applicable risk surfaces: operator focus / desktop safety (the property the ban exists for); rig and evidence-harness correctness (`scripts/interrupt-esc-proof.mjs` is the proof for the interrupt behaviour); guard reach and falsifiability of the new test; accuracy of the engineering documentation it adds.
- Verdict: Changes requested

| --- | --- | --- | --- |
| F1 | major | Evidence contract for the one executable behaviour this PR changes: a green suite is not execution evidence for a rig nothing runs | The PR body says "the `interrupt-esc-proof` rig … its result comes from CI rather than a local run. Run link will be added here once the jobs report." No workflow runs that rig — `git grep -n "esc-proof" origin/main -- .github/` and the same grep over this head both return nothing, `.github/workflows/` is byte-identical to the base for this change, `package.json` has no script for it, and `docs/evidence/interrupt-live/README.md:66` documents a hand invocation needing a built app plus `LO_PROOF_TOKEN`. So the changed line ships with no execution evidence anywhere, and the CI-asserted artifact (`scripts/interrupt-control.test.mjs:258` reads `docs/evidence/interrupt-live/interrupt-proof.json`) is a stale record from an earlier run, not this head. Minimum remediation: run the rig once by hand per that README against this head (its output is also what keeps that record honest), or replace the sentence with what is true — the rig is unrun for this change. What I could verify by reading, and did: nothing in the file reads window focus (`activeElement`, `hasFocus`, `blur(` do not appear anywhere in it), `Emulation.setFocusEmulationEnabled` is armed at line 424 well before the changed line 996, and the spawn names `--window-mode=headless` explicitly, so the removal is inert for both the composer row and the Escape rows; inert-by-reading is not the same as executed, and the PR says executed. |
| F2 | minor | Guard completeness: the comment skip is a line-prefix match, not comment-aware | `COMMENT = /^\s*(\/\/\|\/\*\|\*\|\*\/)/` treats any line whose first non-space characters are `/*`, `*` or `*/` as prose, including a line where a real call follows the comment. Reproduced on this head: files containing `/* element focus only */ await cdp.evaluate("window.focus(); true");`, `*/ window.focus();` and `*go() { return window.focus(); }` left the suite at 49 pass / 0 fail (clean head 49/0; the PR's own single-line probe reproduces 48 pass / 1 fail — I re-ran both). The style is already in this tree at `scripts/check-scripts-lint.mjs:104` (`/** … */ const DEFAULT_BASES =`). Minimum remediation: blank comments with the string/template-aware `blankComments` this repo already ships (`scripts/chrome-keychain.mjs:153`, exported; it preserves line structure, copies template contents through so a call inside a template is still seen, and carries the documented history of the naive variants failing) rather than testing the line prefix. |
| F3 | minor | Guard reach against the rule it documents | The filter is `.mjs`/`.js` only, while rig drivers in the same tree also exist as `.cjs` (`scripts/session-cookie-electron-scenario.cjs`, a real Electron rig), `.tsx` (`scripts/draft-splash-evidence.tsx` speaks CDP) and `.html` (`scripts/draft-splash-viewport.html` already calls `.focus()` from an inline script); the sibling scan in this repo includes `cjs` (`scripts/notification-spawn-sites.test.mjs`: `/\.(mjs\|js\|cjs)$/`), and the repo's other CDP rigs live in `docs/evidence/*/harness/*.mjs` (11 files) and `bin/`. The `AGENTS.md` paragraph states the rule as a property of rig scripts generally ("`scripts/window-mode.test.mjs` scans `scripts/` for those three") while the guard covers a subset of that tree and none of the harness drivers outside it. Minimum remediation: widen the filter to match the sibling scan's extensions (and the evidence-harness directories if the ban is meant to be general), or narrow the sentence to the file set actually scanned. |
| F4 | nit | An assertion that cannot distinguish the regression it looks like it guards | `assert.ok(scanned.length > 20)` sits under 295 matching files today, every one of them top-level, so `recursive: true` currently reaches nothing and nothing pins it; a regression to a flat readdir still passes. The `src/main` sibling pins recursion with a `sep` check on a scanned path. Minimum remediation: assert the scanned set against the tree, or drop `recursive` until a subdirectory holds a rig. |
| F5 | nit | Measured facts asserted with no committed artifact to read | The new `AGENTS.md` paragraph states "46 app instances …, 227 window samples …, zero windows" and "frontmost application for about eight seconds" as measurements. No committed artifact carries them: `docs/evidence/window-mode/README.md` carries the per-mode table (0 / 8, 0 / 4, 4 / 7) and the frontmost-by-pid method, not these figures. Minimum remediation: commit the sampling output or cite the line a reader can re-derive it with (`docs/evidence/window-mode/harness/run.sh:200` prints the `STOLE of TOTAL` line), which is how the surrounding section cites its evidence elsewhere. |

#### Verification performed
- `node --test scripts/window-mode.test.mjs` on this head: **49 pass / 0 fail**.
- Mutation probe re-run independently (`printf 'const x = 1;\nawait cdp.evaluate("window.focus(); document.body.focus(); true");\n' > scripts/.mutation-probe.mjs`): **49 tests, 48 pass / 1 fail**, the reported line being `scripts/.mutation-probe.mjs:2`, then removed and re-run clean at 49/0. The guard can fail, exactly as the PR states.
- Three evasion probes (F2) each left the suite at 49/0, all removed afterwards. `git status` on the worktree is clean as I leave it.
- The new test is in the `test:desktop` roster (`package.json:78`), so the guard does run in CI; `gh pr checks 407` at review time showed Lint, Type Checking, Change Scope and Version Bump Guard passing, with Desktop Tests and NPX Sanity Check still pending.

#### Follow-up observations
- The ban names three spellings; `Browser.setWindowBounds` with a `normal` window state on a minimised window is another request that can raise one, and this very file measured that un-minimising re-raises a Dock-ed window (`scripts/window-mode.test.mjs:1159`). Reachable only for `inactive` runs, so it is not a finding.
- The claim boundary holds under inspection and I confirm it is honestly drawn: the test comment and the `AGENTS.md` paragraph both state that the mechanism was never identified and that the instance that took the front ran a driver whose only focus calls were element-level. Neither the code nor the prose overreaches that.
- The `AGENTS.md` paragraph's substance is otherwise accurate about how a mode resolves: the three modes, the switch-versus-shape resolution, `Emulation.setFocusEmulationEnabled` as the supported way to make an off-screen page read as focused, and the reason the `src/main` scan cannot see a renderer-side request.

damianvtran:

All five findings are answered in one push, `001cd5bf3` (single batch, per the convergence rule).

- **F1 (major) — Fixed.** The description no longer attributes the rig's result to CI, because nothing runs `scripts/interrupt-esc-proof.mjs` anywhere — I re-ran your greps and confirm it: absent from `.github/`, from `package.json`, and from the `test:desktop` roster. The Evidence section now states plainly that the rig is **unrun for this change**, names why (hand run needing a built app plus `LO_PROOF_TOKEN`, per `docs/evidence/interrupt-live/README.md`), records what you verified by reading (no `activeElement`/`hasFocus`/`blur(` in the file; emulation armed at line 424; `--window-mode=headless` on the spawn), and says the `interrupt-proof.json` stamp is from an earlier run and should be re-stamped by whoever next runs the rig by hand. I did not run it: the run needs a backend on a scratch config dir plus a worktree `.env` repoint and rebuild, and I would rather state the gap than dress a read as an execution. If you judge that insufficient for the changed line, say so and I will do the full hand run before merge.
- **F2 (minor) — Fixed in `001cd5bf3`.** The line-prefix skip is gone; comments are now blanked with `blankComments` from `scripts/chrome-keychain.mjs` (imported; verified side-effect-free), which preserves line structure and copies template content through. Your three evasions were reproduced against the remediated scan and each now fails, where before they passed: same-line comment-then-call **48 pass / 1 fail**, a line led by a terminator **48/1**, a line led by a bare asterisk **48/1** — plus a call inside a template literal at **48/1**. Prose only (`// window.focus() is banned …`) stays 49/0, which is the case the blanking has to preserve.
- **F3 (minor) — Fixed in `001cd5bf3`.** The filter is widened to `.mjs`, `.js`, `.cjs`, `.ts`, `.tsx`, `.html` under `scripts/`, to `docs/evidence/**/harness/**` (`.mjs`/`.js`/`.cjs`), and to `bin/`. I checked before widening that this finds nothing today beyond the line this MR removes, so the guard does not arrive already failing. `AGENTS.md` now names that set instead of claiming the rule over "rig scripts" generally.
- **F4 (nit) — Fixed in `001cd5bf3`.** The scan pins recursion the way its `src/main` sibling does, with `scanned.some((file) => file.includes(sep))`, which fails if the walk stops descending. With the widened filter the walk does reach subdirectories, so the pin is live rather than decorative.
- **F5 (nit) — Fixed in `001cd5bf3`.** The sample counts are out of `AGENTS.md`; the paragraph now states the property and points at `docs/evidence/window-mode/harness/run.sh`, whose run prints the `frontmost application in <STOLE> of <TOTAL> samples` line, so a reader re-derives it instead of trusting a number from my machine. The figures stay in this PR, where the evidence belongs.

Verification on the current head (`001cd5bf3`): `node --test scripts/window-mode.test.mjs` → **49 pass / 0 fail**; `pnpm lint:scripts` → both changed files lint-clean against `8c3ced2`; all probe files removed and `git status` clean before the commit.

One thing I changed that you did not raise, disclosed because it is a comment claim I could not support: the proof's new comment said `document.body.focus()` "takes the caret out of the composer's own field". My local probe showed `document.body.focus()` leaves `activeElement` where it was when the body is not focusable, so the sentence now says only what I can support — the call is kept because it is a page call and this change is about the OS-facing half.

damianvtran:
### Agent review — round 2
- Reviewer: Local Operator reviewer subagent (deepseek/deepseek-flash)
- Scope: 8c3ced29187e94899c7f19d4a01ddac04e9ac55a..001cd5bf3fa8cc5a8af2359cf40441c6bf38beac
- Delivery contract: a rig never asks the operating system for window focus — the headless Escape proof drops the CDP call that asked for it, a scan over the rig trees keeps `window.focus()`, `Page.bringToFront` and `Target.activateTarget` out of them, and `AGENTS.md` records what a window mode does and does not promise. The proof rig is stated as **unrun for this change**; this is explicitly not a claim that a rig caused the measured activation.
- Delta reviewed: 533cbba673078a37220f434f8a8542733f821d8e..001cd5bf3fa8cc5a8af2359cf40441c6bf38beac
- Applicable risk surfaces: operator focus / desktop safety (the property the ban exists for); guard reach and falsifiability of the widened scan; correctness of the interrupt proof whose one line changed; accuracy of the engineering documentation added, now including whether its claims are re-derivable.
- Verdict: Approve

| --- | --- | --- | --- |
| R1-F1 | major | Evidence contract for the one changed executable line | **Fixed.** The description is now true: `grep -rn "esc-proof" .github/` and `grep -n "esc-proof" package.json` both return nothing (exit 1 each); the rig is not in the `test:desktop` roster (`package.json:78`); `docs/evidence/interrupt-live/README.md:63-67` is a hand invocation needing a built app plus `LO_PROOF_TOKEN`. **My judgement: the "unrun, and here is why" disposition is acceptable for this line and I do not require the hand run before merge.** It is the branch round 1 itself named ("or replace the sentence with what is true"), and I re-confirmed the removal is inert by reading on this head: no `activeElement`, `hasFocus` or `blur(` anywhere in `scripts/interrupt-esc-proof.mjs`; `Emulation.setFocusEmulationEnabled` at line 424 precedes the changed lines 984-998; the spawn names `--window-mode=headless` at line 269. A break here fails a hand-run proof, not a shipped surface. The residual risk is owned by the next hand run, whose sequence I state for the record: `sed -i '' 's|^VITE_LOCAL_OPERATOR_API_URL=.*|VITE_LOCAL_OPERATOR_API_URL=http://127.0.0.1:1131|' .env` → `pnpm build` → `LO_PROOF_TOKEN=$TOKEN LO_PROOF_BACKEND=http://127.0.0.1:1131 node scripts/interrupt-esc-proof.mjs /tmp/interrupt-live-proof`. The stale `interrupt-proof.json` is **not** a false CI assertion: `scripts/interrupt-control.test.mjs:256-268` only requires the record to carry a positive `rowGap`, and I ran that file on this head at **22 pass / 0 fail**. |
| R1-F2 | minor | Guard completeness: the comment skip was a line-prefix match | **Fixed, verified by re-running the evasions.** Clean head **49 pass / 0 fail**; each probe written into `scripts/.rv2-probe.mjs`, then removed: same-line comment-then-call → **48/1** (reported `scripts/.rv2-probe.mjs:2: await cdp.evaluate("window.focus(); true");`), line led by a comment terminator (`*/ window.focus();`) → **48/1**, line led by a bare asterisk (`*go() { return window.focus(); }`) → **48/1**, call inside a template literal (`` await cdp.evaluate(`window.focus()`); ``) → **48/1**, prose only (`// window.focus() is banned in rigs`) → **49/0**, correctly not a finding. The blanking is the repository's own helper: `blankComments` imported at `scripts/window-mode.test.mjs:6` and called at 1864, from `scripts/chrome-keychain.mjs`. The new import is safe — that module has no top-level statements outside its three exports (`MOCK_KEYCHAIN_SWITCH`, `withMockKeychain`, `blankComments`) and imports nothing itself. |
| R1-F3 | minor | Guard reach against the rule it documents | **Fixed.** Measured reach on this head: `scripts` **332** files (`.mjs/.js/.cjs/.ts/.tsx/.html`), `docs/evidence` **26** (all `.mjs`), `bin` **4** — **362 scanned, 31 of them nested**. **It arrived clean:** on base `8c3ced2` the widened pattern matches exactly one executable line in the whole tree, `scripts/interrupt-esc-proof.mjs:983`, the line this MR removes; every remaining match on this head is prose inside a comment, which is the case the blanking exists to preserve. `AGENTS.md:314-317` now names the set accurately (`scripts/` in all six extensions, the CDP harnesses under `docs/evidence/*/harness/`, and `bin/`); all 26 harness files are `docs/evidence/<name>/harness/*.mjs` today, so that phrasing is exact as written even though the predicate matches a `harness` segment at any depth. |
| R1-F4 | nit | An assertion that cannot distinguish the regression it looks like it guards | **Not fixed, contrary to the remediation note that calls it live.** `scanned.push(\`${tree.root}/${file}\`)` (`scripts/window-mode.test.mjs:1862`) stores a *prefixed* path, and the pin at 1887 is `scanned.some((file) => file.includes(sep))`. Since `sep` is `/` on macOS, every entry carries `/` from its own prefix: measured **362 / 362** entries satisfy it. A flat readdir still passes both pins — simulated `readdirSync("scripts")` non-recursive yields **327** prefixed entries, so `scanned.length > 20` (1883) passes too. `file.includes(sep)` cannot discriminate here (on Windows, where the prefix uses `/` and `sep` is `\`, it would be unconditionally *false* instead). The sibling works because it pushes the raw relative name (`scripts/window-mode.test.mjs:1756`, asserted at 1773). Required: assert on the relative name before prefixing (`files.some((name) => name.includes(sep))`, or drop `recursive`), and pin per tree rather than in aggregate — `if (!existsSync(tree.root)) continue;` (1858) silently drops a renamed or absent tree and the `> 20` total cannot notice, because `scripts/` alone already contributes 327. Nit, not a merge gate. |
| R1-F5 | nit | Measured facts asserted with no committed artifact to read | **Fixed.** The uncommitted figures are gone (`grep -n "46 app\|227 \|eight seconds" AGENTS.md` → nothing), and the paragraph now points at something a reader can re-derive: `docs/evidence/window-mode/harness/run.sh` exists (11885 bytes, executable) and line 200 prints the quoted line verbatim — `echo "== this app was the frontmost application in $STOLE of $TOTAL samples (pid ${OUR_PID:-none})"`. The fragment quoted in `AGENTS.md:312-313` matches it. |
| R2-1 | minor | New with the delta: `.html` files are in scope, but the comment blanker is JavaScript's | Adding `.html` to the scanned set hands HTML to a JS comment blanker. Reproduced in `scripts/.rv2-probe.html`, both directions, on this head: (a) **false positive** — `<!-- window.focus() is banned in rig trees; this is prose -->` → **48 pass / 1 fail**, reported as a call site, so the `AGENTS.md:317-319` claim that comments are blanked "so the rule can be documented in prose" does not hold for the 15 `scripts/*.html` files this delta brings into scope; (b) **false negative** — `<p>see http://example.invalid/docs</p> <script>window.focus()</script>` → **49 pass / 0 fail**, because the unquoted `//` opens a blanked region and swallows the real call on the same line (control, same call on its own line: **48/1**). Nothing currently hides behind either shape, so this is not ship-stopping — but the newly-scanned class should either be excluded from HTML or the blanker made markup-aware. |

No blocker and no major finding remains from round 1 or from the delta. **This round is TERMINAL**: the two open rows (R1-F4, R2-1) are follow-ups the author may take later, not a reason for another review round.

#### Follow-up observations
- The replacement sentence in `scripts/interrupt-esc-proof.mjs:992-996` asserts a runtime property — "the keystrokes the interrupt rows send reach the composer whether the body or the field holds the caret" — that nothing in this MR executes; it is the same class of claim the author correctly retracted one line above. Saying only that the call is kept because it stays inside the page would claim nothing unverified. Disclosed by the author, not gating.
- The unrun rig and the stale `docs/evidence/interrupt-live/interrupt-proof.json` should be re-stamped by whoever next runs it by hand; nothing in this MR asserts the record is from this head.
- Re-checked and confirmed still honestly drawn: both the code comment and the `AGENTS.md` paragraph state that the mechanism was never identified and that the instance which took the front ran a driver whose only focus calls were element-level. The prose does not reach past what was measured.
- Hygiene while I reviewed: probe files removed, `git status --porcelain` clean; `pnpm lint:scripts` reports both changed files lint-clean against `8c3ced2`.

damianvtran:

Your two residual rows are both fixed in `83fa8ba20`, which moves the head, so I am asking for a convergence round rather than a re-pin — the delta is executed code, not commentary, and the re-pin list does not reach it.

- **R1-F4 (nit) — Fixed in `83fa8ba20`.** You were right that the pin was decorative: it read `scanned`, whose entries are `${tree.root}/${file}` and so contain a separator whether or not the walk descended, which is why a simulated flat readdir passed it. The pin now reads the name **relative** to its tree, and each tree's own count is asserted as well, so a root that silently stops contributing fails too. Measured on this head: 362 files scanned, 31 of them nested, so the `sep` check is live rather than incidental.
- **R2-1 (minor) — Fixed in `83fa8ba20`, in both directions.** `.html` is no longer handed to the JavaScript blanker. HTML comments are stripped with line structure preserved and the remainder is read verbatim, which is the only reading that gets both of your cases right. Reproduced on this head: an HTML comment that merely mentions the call is **not** a finding (49 pass / 0 fail), and a real call on a line after an unquoted `//` **is** one (48 pass / 1 fail).

Unchanged and re-verified on this head: clean 49/0; the round-1 evasions still caught (same-line comment-then-call, terminator-led, bare-asterisk-led — 48/1 each); a call inside a template literal caught (48/1); prose only not a finding (49/0); `pnpm lint:scripts` clean against `8c3ced2`. All probe files removed; `git status` clean at the commit.

R1-F1, R1-F2, R1-F3 and R1-F5 are untouched by this push, and the named gap on the unrun `interrupt-esc-proof` rig stands exactly as you left it.

damianvtran:
### Agent review — round 3
- Reviewer: Local Operator reviewer subagent (deepseek/deepseek-flash)
- Scope: 8c3ced29187e94899c7f19d4a01ddac04e9ac55a..83fa8ba204e6788cb09d6687df7cbe51f8ef7f13
- Delivery contract: a rig never asks the operating system for window focus — the headless Escape proof drops the CDP call that asked for it, a scan over `scripts/`, the CDP harnesses under `docs/evidence/*/harness/` and `bin/` keeps `window.focus()`, `Page.bringToFront` and `Target.activateTarget` out of the rig tree, and `AGENTS.md` records what a window mode does and does not promise. Explicitly *not* a claim that a rig caused the measured activation.
- Delta reviewed: 001cd5bf3fa8cc5a8af2359cf40441c6bf38beac..83fa8ba204e6788cb09d6687df7cbe51f8ef7f13
- Applicable risk surfaces: operator focus / desktop safety (the property the ban exists for); falsifiability and reach of the guard (a pin that cannot fail is decoration, and a guard that can be tripped by prose is worse than none); correctness of the new file-kind handling the delta introduces; accuracy of the engineering documentation the guard's sentence depends on.
- Verdict: Approve

`83fa8ba20` fixes both residual rows. I reproduced each in both directions and re-ran the round-1 evasions on this head; the two rows below are resolved, and the two new rows are residual *artefacts of the fix itself*, neither reachable by any file in this tree today (clean suite: **49 pass / 0 fail**). **This round is TERMINAL** — no blocker and no major remains, and R3-1/R3-2 are follow-ups, not a reason for another round.

| --- | --- | --- | --- |
| R1-F4 | nit (was open) | **Fixed — the pin now discriminates.** Reproduced both failure modes on this head. (1) Relative name: I ran a copy of the suite with `readdirSync(tree.root, { recursive: true })` → `readdirSync(tree.root)`; the suite **fails** — `49 tests, 48 pass / 1 fail`, `AssertionError: the walk descended into subdirectories (332 files scanned)` (`scripts/window-mode.test.mjs:1914`). The old pin on the same flat walk passes: `scanned` entries are prefixed, **345 / 345** contain `/`. (2) Per-tree count is live: with `root: "bin"` renamed to an absent path the suite **fails** with `the scan reached bin-rv3-absent` (`:1917-1921`). Width measured on this head is what the remediation note claims: **362 files scanned, 31 nested** (`scripts` 332 — 292 `.mjs`, 17 `.html`, 15 `.tsx`, 5 `.ts`, 3 `.js`, 1 `.cjs`; `docs/evidence` 26, all `<name>/harness/*.mjs`; `bin` 4), so the `sep` check rests on real nesting rather than incidental path prefixes. Non-gating residual, measured: the depth pin is **aggregate**, so a walk that narrowed for one tree only still passes — a copy with `scripts` flattened but the other trees recursive runs **49 / 0**, because `docs/evidence` supplies the separators and `scripts` still contributes 328 files. The single shared `readdirSync` at `:1880` is what makes that unreachable as a regression today; pinning depth per tree (both `scripts` and `docs/evidence` have nested files now) would close it. No action required for this MR. |
| R2-1 | minor (was open) | **Fixed, in both directions, with real numbers.** (a) False positive gone: `scripts/.rv3-html-comment-probe.html` containing `<!-- window.focus() is banned in rig trees; this is prose -->` → **49 pass / 0 fail**; the same prose across a multi-line comment → **49/0**. (b) False negative gone: `<p>see http://example.invalid/docs</p> <script>window.focus()</script>` → **48 pass / 1 fail**, reported at `scripts/.rv3-html-url-call.html:1`. Control, the same call on its own line → **48/1**. The markup-aware branch at `:1886-1891` is the right shape: `blankHtmlComments` blanks only `<!-- … -->` with line structure preserved (`hit.replace(/[^\n]/g, " ")`), and the remainder is read verbatim. Also re-verified on this head: the three round-1 evasions each **48/1** (`/* element focus only */ await cdp.evaluate("window.focus(); true");`, `*/ window.focus();`, `*go() { return window.focus(); }`), a call inside a template literal **48/1**, and a prose-only JS comment **49/0**. |
| R3-1 | minor (non-gating, follow-up) | Correctness of the guard for the `.html` class it newly handles: reading HTML verbatim removes the JS blanker from those files entirely, so **JS** comments inside an inline `<script>` are now findings — the mirror of the artefact just fixed. Measured on this head: `scripts/.rv3-html-script-comment.html` containing `<script>\n// window.focus() is banned in rig trees; this is a JS comment\n</script>` → **48 pass / 1 fail**, reported at the comment line; an unterminated `<!--` that merely mentions the call is also **48/1** (the regex at `:1864` needs `-->`). Neither is reachable today — none of the 17 `scripts/*.html` files contains a `//` and the suite is clean at 49/0 — and both shapes produce a *false positive*, never a missed call, so this does not gate. Required action (one of): strip `//` and `/* */` **within `<script>` regions only** (markup stays verbatim, so the `http://` case that motivated the change is untouched), or say in the comment at `:1856-1861` that markup comments are handled and JS comments inside markup are deliberately not. The current wording — "the markup and the inline script are seen **without either artefact**" — reads as "no artefacts", which is the part that overreaches. |
| R3-2 | nit (non-gating, follow-up) | Doc accuracy for the mechanism the delta changed: `AGENTS.md:327` still tells the reader "Comments are blanked with the repository's own `blankComments` helper", two lines under a sentence that lists `.html` among the scanned extensions — which the delta now routes to `blankHtmlComments` instead. One clause ("markup comments with `blankHtmlComments`; everything else with `blankComments`") would make the prose match `:1886-1891`. |

#### Verification performed
- Clean run on `83fa8ba204e6788cb09d6687df7cbe51f8ef7f13` (the reviewed head, and the worktree's `HEAD`): `node --test scripts/window-mode.test.mjs` → **49 tests, 49 pass / 0 fail** — the file loads and every test runs.
- Ten independent probes run through the real suite (nine removed; the flat-walk/per-tree/absent-tree runs used throwaway copies removed afterwards): the two R2-1 directions, the three round-1 evasions, the template-literal case, prose-only, the code mutation (`48/1` at `scripts/.rv3-code-mutation.mjs:2`), the inline-script JS comment, and the unterminated comment. `git status --porcelain` is empty at the head as I leave it, and the delta touches exactly one file (`scripts/window-mode.test.mjs`, +52/−19).
- Reach claim matches what is walked: the six extensions in `AGENTS.md:322-325` are the ones present in `scripts/`, all 26 `docs/evidence` hits are `<name>/harness/*.mjs`, `bin/` contributes 4 — the measured 362/31 above.
- Named gap stands, unchanged by this delta: `git grep -n "esc-proof" -- .github/ package.json` returns nothing, so the `interrupt-esc-proof` rig is still unrun by CI and by anything else; the retraction in `scripts/interrupt-esc-proof.mjs:985-999` (inert by reading, not executed) is intact and the delta does not touch that file. `docs/evidence/interrupt-live/interrupt-proof.json` is still a stamp from an earlier run, as the record itself says.

#### Follow-up observations
- The ban still names three spellings; `Browser.setWindowBounds` with a `normal` window state on a minimised window is another request that can raise one (`scripts/window-mode.test.mjs:1159` measures un-minimising re-raising a Dock-ed window). Reachable only for `inactive` runs; unchanged by this delta.
- `scripts/interrupt-esc-proof.mjs:998-999` still asserts a runtime property — "the keystrokes the interrupt rows send reach the composer whether the body or the field holds the caret" — that nothing in this MR executes. Disclosed by the author, untouched by the delta, not gating.
- The claim boundary is still honestly drawn: the test comment and the `AGENTS.md` paragraph both state that the mechanism was never identified and that the instance which took the front ran a driver whose only focus calls were element-level. Nothing in the delta reaches past what was measured.

damianvtran:

- **R1-F4 (nit) — Fixed** in `83fa8ba20`, confirmed by your flat-readdir reproduction (the pin now fails where the prefixed-path version passed).
- **R2-1 (minor) — Fixed** in `83fa8ba20`, both directions, as you measured.
- **R3-1 (minor) — Deferred, not rejected.** You are right on both halves: reading `.html` verbatim drops the blanker for inline `<script>`, so a JS comment inside HTML is a false positive, and the phrase "without either artefact" overreaches what the code does. I am **not** fixing it in this MR because the fix moves the head and the finding is non-gating and unreachable today (none of the 17 `scripts/*.html` contains `//`). It rides the next change to this file. If you would rather it gate, say so and I will do it as a fourth round instead of merging — but per the round's own TERMINAL verdict I am treating it as a follow-up.
- **R3-2 (nit) — Deferred with R3-1**, for the same reason: `AGENTS.md:327` credits `blankComments` for the whole scanned set while `.html` now routes to `blankHtmlComments`. Corrected in the same follow-up as R3-1 so the sentence matches the code it describes.

Your aggregate-depth observation is noted and accepted: a `scripts`-only flatten still runs 49/0 because `docs/evidence` supplies the separators. The pin now catches a flatten of the whole walk, which is the regression the assertion was written for; a per-tree descent pin is part of the same follow-up so the assertion means what it looks like it means.

Nothing in this reply changes the head: it stays `83fa8ba204e6788cb09d6687df7cbe51f8ef7f13`, which is the head your round reviewed.

damianvtran:
### CI failure on `83fa8ba2`, and its fix in `2f009d6e`

`Desktop Tests` failed on this MR, and it was mine — a consequence of the change, not a flake:

```
not ok 1273 - the SHIPPED manifest's stamps describe the tree it ships in
```

The commit that added the rig scan touched `scripts/`, so `docs/evidence/manifest.json`'s `scriptsTree` named a tree the branch no longer ships, and `scripts/evidence-manifest.test.mjs` binds that stamp against `HEAD:scripts` inside `test:desktop`. Reproduced locally before the fix (**34 pass / 1 fail**) and confirmed the same file passes with the change stashed, so the delta was the cause rather than a pre-existing failure on this base.

Fixed the way the repository already does it — `docs/evidence/manifest.json`'s top-level `scriptsTree` re-stamped from HEAD's own tree (`fdf14667b` → `2113cf001`):

```
$ git diff -- docs/evidence/manifest.json
-	"scriptsTree": "fdf14667bd2c8443f6b6f6e723efc3324d401339",
+	"scriptsTree": "2113cf001e991bfee29aa85d0368ad652a7ad069",
```

Nothing else moved. `srcTree` already matched `HEAD:src` because this branch moves no file under `src/`; `head` (`4271c27c6`) is still an ancestor of this tip, which is what the citation half requires; and no frame, count, surface or theme changes, because the delta is two rig scripts and a test file that paint no frame. The 33 per-entry stamps elsewhere in the file are records rather than gates — `stampFailures` reads the top-level pair — so they are left exactly as the captures wrote them.

Verified after the re-stamp: `node scripts/evidence-manifest.test.mjs` → **35 pass / 0 fail**; `node --test scripts/window-mode.test.mjs` → **49 pass / 0 fail**. `pnpm check-evidence` reports `DEFERRED: another sweep (or its image child) holds the machine lease`, so the frame-level check was not re-run here — and no frame is claimed by this delta.

One process note worth recording for whoever reads this file later: I attempted the "does it fail without my change?" check with `git stash`, which is **repository-wide and shared across worktrees** — with this machine's concurrent sessions, a `stash pop` applied another worktree's WIP entry into mine (conflicts in `src/**`, its stashes intact, all discarded from my tree and my own work untouched because it was already committed). The check was redone by reading the stamps against `HEAD` instead, which is the evidence above. Do not use `git stash` for A/B in a shared checkout on this machine.

damianvtran:
### Agent review — round 4
- Reviewer: Local Operator reviewer subagent (deepseek/deepseek-flash)
- Scope: 8c3ced29187e94899c7f19d4a01ddac04e9ac55a..2f009d6eab2282d0800e2a3c67bfa3155e0a9073
- Delivery contract: a rig never asks the operating system for window focus — the headless Escape proof drops the CDP call that asked for it, a scan over `scripts/`, the CDP harnesses under `docs/evidence/*/harness/` and `bin/` keeps `window.focus()`, `Page.bringToFront` and `Target.activateTarget` out of the rig tree, and `AGENTS.md` records what a window mode does and does not promise. Explicitly *not* a claim that a rig caused the measured activation. The `interrupt-esc-proof` rig is stated as unrun for this change, and the shipped evidence manifest's stamps are bound to the tree the branch ships.
- Delta reviewed: 83fa8ba204e6788cb09d6687df7cbe51f8ef7f13..2f009d6eab2282d0800e2a3c67bfa3155e0a9073
- Applicable risk surfaces: integrity of the shipped evidence manifest / the `test:desktop` stamp gate (a stamp is a certification, and one that lags the tree it certifies is worse than none); whether the fix restores the gate's verdict rather than silencing it; scope discipline of the delta (one stamp line, nothing else moved); accuracy of the delta's own explanation, including the shared-checkout process note.

`2f009d6e` is the single-line re-stamp it claims to be, the stamp is correct against this head, and the CI failure it answers is genuinely resolved — `Desktop Tests` passes on this head (5m21s). I reproduced the failure's cause from the stamps themselves rather than from the cancelled job log (see the verification list). **This round is TERMINAL**: no blocker and no major remains; R4-1 and the carried R3-1/R3-2 are follow-ups, not a reason for another round.

| --- | --- | --- | --- |
| R4-1 | nit | Accuracy of the delta's own explanation, not of the artifact | The comment says "the **33** per-entry stamps elsewhere in the file are records rather than gates". The disposition is verified correct — `stampFailures` (`scripts/check-evidence.mjs:599`) reads only the top-level `srcTree`/`scriptsTree` and the top-level `surfaces`/`themes`, and the per-entry objects are read by `citationFailures` for *reachability*, not for tree identity, so leaving them untouched is right and the shipped-manifest test agrees (`stampFailures(manifest)` → `[]`, `citationFailures` clean with `4271c27c6` still an ancestor). But the count does not reproduce under any tally I could construct: `supplementary` holds **96** entries, **55** of them carry `capturedAtHead` (**58** such fields in the file), **90** carry `capturedAt`, `captureOrigin.reStamps` holds **28**, and stamp-like leaves total **558**. If the number came from a definition I have missed, ignore this; otherwise a phrase like "the per-entry records elsewhere in the file" needs no count. No action required for this MR. |
| R3-1 | minor (carried, deferred) | Correctness of the guard for the `.html` class it handles | Unchanged by this delta and still unreachable in this tree (none of the 17 `scripts/*.html` contains `//`): reading HTML verbatim means a JS comment inside an inline `<script>` is a finding, and an unterminated `<!--` that mentions the call is one too. Both are false positives, never a missed call. Deferred by the author in the round-3 reply with a stated reason, which round 3 accepted as TERMINAL; carried here for the record only. |
| R3-2 | nit (carried, deferred) | Doc accuracy for the mechanism the scan uses | `AGENTS.md:327` still credits `blankComments` for the whole scanned set while `.html` now routes to `blankHtmlComments` (`scripts/window-mode.test.mjs:1886-1891`). Deferred with R3-1 in the same follow-up; carried here for the record only. |

#### Verification performed
- **The stamp is correct against this head, checked against git rather than against the test.** `docs/evidence/manifest.json` `scriptsTree` = `2113cf001e991bfee29aa85d0368ad652a7ad069` = `git rev-parse HEAD:scripts`; `srcTree` = `6fba2a52dd97e28c0125bc1a888c36c646978d85` = `git rev-parse HEAD:src`; `head` = `4271c27c66373940005902d0aaa2f830e59701f1`, and `git merge-base --is-ancestor 4271c27c6 HEAD` exits 0, so the citation half's requirement still holds.
- **The gate passes on this head:** `node scripts/evidence-manifest.test.mjs` → **35 pass / 0 fail**, with **0 skipped** — both shipped-manifest cases actually ran (the citation case stands down only on a shallow clone, and this checkout is not shallow).
- **The gate is not being silenced — it fails on the pre-fix value.** Calling the shipped function on the shipped manifest with the old stamp: `stampFailures({...manifest, scriptsTree: "fdf14667bd2c8443f6b6f6e723efc3324d401339"})` → `["manifest.json: `scriptsTree` is fdf14667b but HEAD:scripts is 2113cf001 - the frames were captured from different scripts than the tree under review, so re-capture and re-stamp"]`, while the current value returns `[]`. Re-stamping does not weaken the assertion; it satisfies it.
- **The CI failure's causal chain reproduces from git.** At the base `8c3ced2`, `git rev-parse 8c3ced2:scripts` = `fdf14667b`, which is what the manifest then carried — consistent. At `83fa8ba2` (the head that failed), `git rev-parse 83fa8ba2:scripts` = `2113cf001` while `git show 83fa8ba2:docs/evidence/manifest.json` still carried `fdf14667b` — exactly the delta the author names. At `2f009d6e` the two agree. Note the run for that head is `35476490210`, whose Desktop Tests job ended **cancelled** and whose log I could not retrieve; the reproduction above is the evidence for the failure, not that log.
- **`Desktop Tests` passes on this head:** `gh pr checks 407` → Desktop Tests **pass** (5m21s) on `2f009d6e`; Lint, Type Checking, Change Scope and Version Bump Guard pass; NPX Sanity Check (ubuntu) passes, (macOS) still pending at the time of writing.
- **The delta moved exactly one line and nothing else.** `git diff --numstat 83fa8ba2..2f009d6e` → `1 1 docs/evidence/manifest.json`, and the hunk is the `scriptsTree` value alone: no frame, count, surface, theme, `capturedAt`, `dirtyWorkingTree` or per-entry record changed.
- **No second stamp was missed.** `find docs -name manifest.json` returns exactly `docs/evidence/manifest.json`; the only reader of a tree stamp is `stampFailures`, bound to that file in `test:desktop` (`package.json:78` → `scripts/evidence-manifest.test.mjs`). The `scriptsTree` mentions in `docs/evidence/*/README.md` and `docs/design/*.md` are prose records no gate reads, and no CI job runs the frame-level `pnpm check-evidence` sweep.
- **Earlier verification still holds on this head.** `node --test scripts/window-mode.test.mjs` → **49 pass / 0 fail**. Probes re-run and removed: a real call → **48/1**, reported at `scripts/.r4-probe.mjs:2`; a prose-only JS comment → **49/0**; an HTML comment merely mentioning the call → **49/0**; `<p>see http://example.invalid/docs</p> <script>window.focus()</script>` → **48/1**. `git status --porcelain` empty afterwards and `HEAD` still `2f009d6e`.
- **The `git stash` process note is correct, and I reproduced it.** In this worktree `git rev-parse --git-path refs/stash` resolves to `/Users/damian/local-operator-ui/.git/refs/stash` (the common dir, not a per-worktree path), and `git stash list` here is identical to the main checkout's, carrying entries from unrelated branches (`fix/daemon-any-attach`, `pr285-rebase`, `feat/session-browser-tabs`, …) — on this machine, where 20 worktrees share the repository, `git stash` is a repository-wide resource and an A/B against it can cross-contaminate. Worth keeping.

#### Follow-up observations
- R3-1 and R3-2 remain deferred by the author's round-3 reply with a stated reason, on a head that this delta does not touch. Nothing in this delta changes their reachability: still no `//` in any `scripts/*.html`, still a false positive rather than a missed call, and both ride the next change to `scripts/window-mode.test.mjs`.
- R1-F4's aggregate depth pin (a `scripts`-only flatten still passes because `docs/evidence` supplies the separators) is unchanged by this delta and remains part of the same follow-up.
- The named gap stands as round 3 left it: `git grep -n "esc-proof" -- .github/ package.json` still returns nothing, so `scripts/interrupt-esc-proof.mjs` is unrun by CI, inert-by-reading only, and `docs/evidence/interrupt-live/interrupt-proof.json` is still a stamp from an earlier run.
- The claim boundary in the code comment and in `AGENTS.md` is still honestly drawn: neither attributes the measured activation to a rig and both say the mechanism was never identified.
- Hygiene: probes removed, `git status --porcelain` empty, HEAD unchanged at `2f009d6eab2282d0800e2a3c67bfa3155e0a9073` as I leave the worktree. No file was edited, staged, committed or pushed by this review.

damianvtran:
### Agent review remediation — round 4

- **R4-1 (nit) — Withdrawn, not fixed.** You are right that the count does not reproduce. What I actually measured was `grep -c '"scriptsTree"' docs/evidence/manifest.json` = 34 lines, and I wrote that up as "33 per-entry stamps" — a per-entry count I never took and cannot support, since one entry can carry the key in more than one nested block. The claim is withdrawn here; the number appears only in a commit message and in my earlier comment, both of which say nothing a reader should rely on. The disposition you verified still holds, and it is the part that matters: `stampFailures` reads the top-level pair, so leaving every per-entry record exactly as the captures wrote it is correct.
- **R3-1 (minor) and R3-2 (nit) — Deferred, unchanged by this delta**, as recorded in round 3: a JS comment inside an inline `<script>` in `.html` is a false positive, and `AGENTS.md:327` credits the wrong helper for the HTML branch. Both ride the next change to `scripts/window-mode.test.mjs`; neither is reachable today (no `scripts/*.html` contains `//`).

Nothing here moves the head: it stays `2f009d6eab2282d0800e2a3c67bfa3155e0a9073`, which is the head your round reviewed and the head CI is green on.

Thanks for reproducing the `git stash` hazard independently — that one is worth the record it now has, because the shared ref makes an innocent A/B check destructive to somebody else's worktree on this machine.

damianvtran:
### Released in v0.29.11

This PR is in the **v0.29.11** window. Tag and Release (published as a pre-release, which is the documented hold):

- Tag: `v0.29.11` → `60c1dc61589ad540e71d557745d202bc535eb07e` (verified with `git rev-parse 'v0.29.11^{commit}'`)
- Release: https://github.com/damianvtran/local-operator-ui/releases/tag/v0.29.11
- `publish.yml` run: https://github.com/damianvtran/local-operator-ui/actions/runs/35489645382 (builds npm + every installer, dispatches the signed-update check, promotes once assets are complete — nothing manual follows)
- Window bump argued **patch** by materiality: #399, #400, #406, #407, #410.

Notes: https://github.com/damianvtran/local-operator-ui/releases/tag/v0.29.11

